PR #5, opened 2026-03-07
chore: remove dead interactive pointer blob code

umairnadeem:
## Problem

In commit b419379 ("Add onboarding flow"), the interactive pointer-following blob div in `BackgroundGradientAnimation` was commented out, but all the supporting code was left in place:

- 4 refs (`interactiveRef`, `curX`, `curY`, `tgX`, `tgY`)
- 2 `useCallback` hooks (`move`, `handleMouseMove`)
- 2 `useEffect` hooks (starting rAF loop + attaching mousemove listener)
- An infinite `requestAnimationFrame` loop that runs every frame, checks null, and continues
- A `window.addEventListener("mousemove", ...)` that updates refs nothing reads

This dead code runs on every page load, wasting CPU cycles.

## Fix

Removed all dead interactive pointer code. The component is now a pure render function with no hooks or event handlers, so the `"use client"` directive is also removed — it can now be server-rendered.

Zero visual/functional changes.

### DIFF
--- a/app/components/BackgroundGradientAnimation.tsx
+++ b/app/components/BackgroundGradientAnimation.tsx
@@ -1,38 +1,6 @@
 "use client";
 
-import { useEffect, useRef, useCallback } from "react";
-
 export default function BackgroundGradientAnimation() {
-  const interactiveRef = useRef<HTMLDivElement>(null);
-  const curX = useRef(0);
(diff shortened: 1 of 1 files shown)


umairnadeem:
we'll restore this blob soon, closing

PR #25, opened 2026-03-15
Expose add-element buttons on hover

umairnadeem:
Make it easier to add elements to the canvas without dnd


umairnadeem:
closing, changed design decision

PR #90, opened 2026-04-26
perf: nightly performance optimizations 2026-04-26

umairnadeem:
Automated nightly performance optimization pass.

### 1. Lazy-load remotion + harden the prefetch flow
`app/components/video/useAssetPrefetch.ts`

- Dynamically import `remotion` (`prefetch` fn) instead of statically importing it. Moves ~301KB of remotion core out of the eagerly-loaded bundle into a lazy chunk that only loads when video prefetching is needed.
- Added a rejection handler on the dynamic import. Without it, a transient chunk-load error (network blip, stale client requesting a deleted chunk) would throw an unhandled rejection and never call the ready signal, leaving the video preview stuck on its loading shimmer indefinitely. Now we log the error and flip ready to true so the player can attempt to render and surface its own error if any.
- Replaced the custom `PrefetchReadyStore` (45 lines) + `useSyncExternalStore` plumbing with `useState<boolean>` + `Promise.all` over the active prefetch handles' `waitUntilDone()`. Same behavior (ready stays true across no-op layout changes; flips false→true only when new URLs are added). Net delta: -43 lines, one fewer file, one fewer concept.

### 2. Font preload cleanup (root layout)
`app/layout.tsx`, `app/globals.css`

Six font families were being preloaded on every route, three of them eagerly via `next/font/google`'s default `preload: true`. Browser was emitting "preloaded but not used within a few seconds" warnings for two of them.

- Dropped Geist Mono entirely — declared as `--font-mono` but `font-mono` is used nowhere in the codebase.
- Added `preload: false` to Instrument Serif — only used by `ComposerHero`, no need for top-priority preload on every route.
- Trimmed Sentient from 4 weights → 1 — the only call site (`OnboardingCard.tsx`) renders at default weight 400.
- Kept Geist on the default preload (it's the body sans).

Result: the 3 woff2 preload warnings drop to 0. The 2 remaining CSS-chunk preload warnings are Turbopack dev noise — they don't appear in `next start` and aren't worth chasing.

### 3. Expanded `optimizePackageImports`
`next.config.ts`

Added `radix-ui`, `@dnd-kit/core`, `@dnd-kit/sortable`, `@dnd-kit/utilities`, `lodash` to the list for better tree-shaking on client bundles.

### 4. `` → `next/image`
`app/components/Copilot.tsx`

Composer's reference-image thumbnail (`ImageThumbnail`) now uses `next/image` with `fill` + `object-cover`, matching the convention from `ForegroundPreview.tsx`. Clears the last lint warning in the codebase.

### 5. Conditional load of `@next/bundle-analyzer` (fix)
`next.config.ts`

The previous wiring (added in `chore: install and wire up @next/bundle-analyzer`) imported `@next/bundle-analyzer` at module top, but the package lives in `devDependencies`. Loading `next.config.ts` in any environment that omits dev deps (e.g. Docker stages running `npm ci --omit=dev`) would crash before `ANALYZE` was even checked, breaking `next build` / `next start`.

Switched to an async config function that dynamically imports the analyzer only when `ANALYZE=true`. Production installs no longer need the package present.

## Bundle analysis findings (for reference)
- Total client bundle: ~6MB (stat size)
- Largest modules: react-dom (598+523KB), remotion (301KB), slate (207+157KB), supabase (216+125+98KB)
- remotion was the only one with a viable lazy-loading opportunity
- slate is heavy but essential to the editor on every page load
- supabase auth is required for all authenticated pages

## Verification
- `npm run lint` — 0 problems
- `npm run typecheck` — clean
- `npm run test:run` — 472/472 pass
- `npm run build` — clean (confirms async `next.config.ts` loads correctly under Next 16)

All changes are behavior-preserving.


umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

PR #93, opened 2026-04-27
show character avatar in audio result preview

umairnadeem:
## Summary
- Render a circular character avatar + name inside the audio result preview for character elements, sourced from `metadata.characters[name].avatarUrl` in the project store. Falls back to the name's initial, then a `User` icon when name is empty.
- Image uses `object-cover object-center` so it preserves aspect ratio (cropped/centered into the circle).
- Flatten the audio result row into a single `flex-wrap` container — avatar, name, regenerate button, play button, waveform, and time display are now direct siblings that wrap naturally as width shrinks (waveform shrinks first via `flex-1 basis-[160px] min-w-0`, time wraps below next, etc.). Audio placeholder unchanged.

## Test plan
- [ ] Character element with an `avatarUrl` shows the image cropped into a circle next to the name in the audio preview.
- [ ] Character element without an `avatarUrl` shows the name's first letter; with no name set, shows a `User` icon.
- [ ] Resize the canvas narrow: the audio preview row wraps gracefully (waveform shrinks, then time/play wrap below) and the card grows in height instead of overflowing.
- [ ] Non-character audio types (narration, music, sound) render unchanged.
- [ ] Audio placeholder (pre-generation) is unchanged.

### DIFF
--- a/app/components/canvas/elements/AudioPlayer.tsx
+++ b/app/components/canvas/elements/AudioPlayer.tsx
@@ -30,7 +30,7 @@ export function AudioPlayer({
 	const [duration, setDuration] = useState(0);
 
 	return (
-		<div className="flex flex-1 min-w-0 items-center gap-2">
+		<>
 			<Tooltip>
 				<TooltipTrigger asChild>
 					<button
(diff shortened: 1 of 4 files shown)


umairnadeem:
@codex review

PR #101, opened 2026-04-28
fix: nightly security audit 2026-04-28

umairnadeem:
## Summary

Nightly security audit and targeted fixes for real exploitable issues only.

### Changes
- Bumped transitive postcss via npm overrides to patched version for GHSA-qx2v-qp2m-jg93
- Sanitized uploaded image filename before using as blob key/URL segment
- Added concrete security headers in Next.js config
- Added Supabase migration to restrict anon updates on access_codes to current_uses only

### Validation
- npm audit (postcss fixed; remaining uuid issue is transitive and not safely fixable upstream)
- npm run build


umairnadeem:
@codex review

PR #118, opened 2026-05-03
Show player by default; shimmer during script generation

umairnadeem:
## Summary
- Player visible on first load: positioned on the right on desktop, on top on narrow viewports (<1066px). Initial position is computed lazily via `matchMedia` to avoid mount flicker.
- Renamed the "Play" button to **Generate** (using the `Sparkles` icon). The button now only kicks off `generateAll()` and no longer toggles player visibility.
- Player now shows the same shimmer surface during script generation that it shows during content generation, eliminating the flicker between empty/partial layouts as the script streams in.
- Extracted a shared `PlayerShimmer` shell so the script-loading and queue-progress states render the same surface by construction.
- Refactored `VideoPanel` to split body state branching (`VideoPanelBody` with early returns) from frame chrome.
- New empty state when layout has no series: "Generate elements to playback" (no shimmer, since nothing is loading).

## Test plan
- [ ] Load editor on desktop (>1066px): side player is visible on the right on first render.
- [ ] Load on narrow viewport (<=1066px): player is on top.
- [ ] Start a script generation: shimmer with "Writing script…" appears continuously, no flicker.
- [ ] After script lands, click **Generate**: shimmer remains, now driven by `QueueProgressBar` with `n of m generated`.
- [ ] When generation completes, `<VideoPreview>` renders with `RenderControls` in the corner.
- [ ] When the layout has no series, the empty state "Generate elements to playback" shows without shimmer.

### DIFF
--- a/app/components/PostPromptView.tsx
+++ b/app/components/PostPromptView.tsx
@@ -6,7 +6,7 @@ import { generationQueue } from "@/lib/generation/queue";
 import Copilot from "./Copilot";
 import Canvas, { type CanvasHandle } from "./canvas/Canvas";
 import { useRefineScript } from "./canvas/hooks/useRefineScript";
-import { Play } from "lucide-react";
+import { Sparkles } from "lucide-react";
 import { TopPlayerPanel, SidePlayerPanel } from "./video/PlayerPanel";
 import {
 	PlayerPositionProvider,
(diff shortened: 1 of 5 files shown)


umairnadeem:
@codex review

PR #120, opened 2026-05-03
fix: nightly bug fixes 2026-05-03

umairnadeem:
## Summary
- fix malformed JSON request parsing in shared route handler to return 400 instead of 500
- add regression test covering invalid JSON request body behavior

## Validation
- npm run build


umairnadeem:
useless

PR #131, opened 2026-05-07
replace SoundType with editable loops + attribute badge polish

umairnadeem:
## Summary
- Drop `SoundType` (ambient/transient); replace with a numeric `loops` attribute on `sound` and now `music` too. Scene-builder emits N back-to-back sequences so audio actually replays through the looped span.
- Generalize attribute UI: `visibleAttributes` carries `{ color, label, edit? }`; one `AttributeBadge` component renders read-only and editable variants. Loops, emotion (sourced from `TTSEmotion`), and the model picker all use it. Editable badges get a ring, larger hit target, and `onMouseDown` preventDefault so they reliably click through Slate.
- Fix empty-element placeholders swallowing clicks (the editor caret now appears on hover) and center the "Generate elements to playback" message when the player container narrows.
- Fix scene-builder background trim: when a same-type background replaced a looped background, only the last copy was inspected, so earlier loop copies overlapped the new bed and the last copy ended up with a negative duration that surfaced in Remotion as a phantom 1s ghost. Replaced with a `trimSequencesAt` helper that pops copies starting at/after the cutoff and trims the straddling tail.
- Extract `createCanvasNode` helper to consolidate node construction across `insertElement`, `osmlSerializer`, `applyOps`, and `getGenerationInputs`; add `getLayoutKey` so `useVideoLayout`'s memo invalidates when loop-affecting attributes change without depending on ref identity.

## Test plan
- [ ] `npm run typecheck && npm run lint && npm run format:check && npm run test:run`
- [ ] Drop a sound element, click the Loops pill — dropdown opens on first click; pick a value; preview plays the SFX that many times back-to-back.
- [ ] Same for a music element (violet Loops pill).
- [ ] Click a narration/character Emotion pill — full TTSEmotion list, scrollable, capped height.
- [ ] Model picker still works on any element that has model+provider attrs.
- [ ] Empty element: hovering the placeholder shows the I-beam caret; first click focuses the editor.
- [ ] Resize the video panel narrow — empty-state message stays centered.
- [ ] Scene with looped music followed by a shorter scene + replacement music: preview/render produces no overlap and no negative `durationInFrames`.
- [ ] Edit a `loops` attribute on an existing element — preview re-layouts immediately (no stale memo).


umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

PR #145, opened 2026-05-10
fix: nightly security audit 2026-05-10

umairnadeem:
## Summary
- enforce required Supabase public env vars (URL and anon key) via explicit runtime validation
- remove silent empty-string fallback for auth/session client configuration
- wire server/client/middleware Supabase client constructors to validated env accessor

## Security rationale
Silent fallback to empty Supabase credentials can lead to degraded or undefined auth/session behavior. This change fails loudly when required auth configuration is missing, preventing insecure degraded operation.

## Validation
- npm audit
- npm run lint
- npm run format:check
- npm run typecheck
- npm run build


umairnadeem:
trivial

PR #160, opened 2026-05-14
fix: nightly bug fixes 2026-05-14

umairnadeem:
## Summary
- fix runware video submission to fail loudly when provider returns no video payload
- add regression tests for undefined/empty runware videoInference result
- harden error copy UI to avoid unhandled clipboard rejections and clear timers on unmount

## Validation
- npm run build


umairnadeem:
useless

PR #161, opened 2026-05-14
perf: collapse script-sync tree scan to a single editor walk

umairnadeem:
## Summary
- in `useScriptSync`, build an `id -> path` map with one `Editor.nodes` pass instead of calling `findNodeById` per streamed node
- streaming the generated script previously scaled O(N*M) (N elements in editor × M nodes received per effect); the linear scan + Map lookups make it O(N + M)

## Why this matters
- script generation is a user-visible streaming hot path; each LLM batch re-fires the effect with the cumulative node list, so the squared cost grows with project size
- larger projects already feel noticeable jank during `/generate` streaming — this removes wasted tree traversals without changing behavior

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run test:run` (618/618)
- `npm run build`

## Constraints honored
- runtime-only, no behavior change
- `proxy.ts` untouched
- no `<head>` tracking scripts moved
- no `!` non-null assertions, no defensive guards added
- no lodash/bundle micro-opt churn

### DIFF
--- a/app/components/canvas/hooks/useScriptSync.ts
+++ b/app/components/canvas/hooks/useScriptSync.ts
@@ -1,5 +1,5 @@
 import { useEffect } from "react";
-import { Editor, Transforms } from "slate";
+import { Editor, type Path, Transforms } from "slate";
 import { useScript } from "@/lib/script/ScriptProvider";
 import {
 	CANVAS_ELEMENT_TYPES,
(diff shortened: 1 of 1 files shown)


umairnadeem:
trivial - nodes are capped at a constant anyways

PR #164, opened 2026-05-17
feat: queue-backed async generation for all asset connectors

umairnadeem:
## Summary
- Every asset generation request (image/music/sfx/tts/video) now creates a row in a new \`jobs\` table, enqueues an \`asset-generate\` message via Vercel Queues, and returns \`{ jobId, status: \"pending\" }\` immediately. A single consumer at \`app/api/queues/asset-generate/route.ts\` runs the provider and updates the row.
- Connectors are uniform: \`BaseAssetConnector._generate\` always submits then polls \`/api/v1/{type}/{jobId}\` via \`awaitCompletion\`. The video-only polling override is deleted and the per-type gateways collapse to a single \`path\` declaration on \`OpenSlopAssetGateway\`.
- New shared helpers: \`createAssetRouteHandlers\` + \`pollJob\` (route factory), \`runAssetJob\` (hides the video submit/poll split from callers), \`parseBody\` + \`withAuth\` (shared validation/auth envelope between \`createRouteHandler\` and \`createAssetRouteHandlers\`).
- LLM is unchanged — it keeps its SSE streaming path.

### Files at a glance
- **New:** \`supabase/migrations/003_create_jobs.sql\`, \`lib/api/jobs.ts\`, \`lib/supabase/service.ts\` (cached service-role client), \`app/api/queues/asset-generate/route.ts\`, \`app/api/v1/{image,music,sfx,tts}/[jobId]/route.ts\`, \`lib/connectors/__tests__/_gateway-mock.ts\`.
- **Refactored:** \`lib/connectors/asset-base.ts\` (now polls), \`lib/connectors/video/connector.ts\` (drops override), \`lib/gateway/base.ts\` + \`lib/gateway/openslop/base.ts\` (adds \`AssetGateway\` / \`OpenSlopAssetGateway\`), all \`lib/gateway/openslop/{type}.ts\` (path-only).
- **Routes:** every \`app/api/v1/{asset}/route.ts\` now uses \`createAssetRouteHandlers\`; \`video/[jobId]/route.ts\` re-exports the shared \`pollJob\`.

### Idempotency & safety
- Consumer skips terminal-state jobs so Vercel Queue redeliveries don't re-run completed work.
- RLS: users SELECT/INSERT their own jobs; UPDATE happens only via the cached service-role client.

## Test plan
- [x] \`npm run lint\`, \`npm run format:check\`, \`npm run typecheck\`, \`npm run test:run\` (648/648), \`npm run build\` all pass locally.
- [ ] Apply the migration in staging (\`supabase db push\`) and confirm RLS policies.
- [ ] Set \`SUPABASE_SERVICE_ROLE_KEY\` in the Vercel environment.
- [ ] Hit \`POST /api/v1/image\` with a real user session — verify <100ms response, a \`jobs\` row, and \`GET /api/v1/image/{jobId}\` transitions to \`completed\`.
- [ ] Repeat for music, sfx, tts, video (video exercises the consumer's provider-side polling).
- [ ] Force a provider failure — confirm the job row ends in \`status: failed\` with a populated \`error\`.
- [ ] Confirm Vercel Queues trigger is registered in the deploy logs (\`asset-generate\` topic on the consumer route).


lib/api/providers.ts
```diff
@@ -64,3 +68,51 @@ export const getTTSProvider = defineProvider(
...
+	const providerJobId = initial.metadata?.jobId as string | undefined;
+	if (!providerJobId || !provider.poll) {
+		throw new Error("Video provider returned no jobId for async generation");
+	}
+	const poll = provider.poll.bind(provider);
+	const completed: VideoProviderResponse = await awaitCompletion(
+		(id) => poll(id),
+		providerJobId,
+		(r) => !!r.result?.video || r.metadata?.status === "failed",
+	);
```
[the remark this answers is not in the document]
umairnadeem: Fixed in 3a105c7. `runAssetJob` now persists the upstream provider_job_id via an `onProviderJob` callback after submission, and on queue redelivery the consumer passes it back so we resume polling the in-flight job instead of resubmitting. Also bumped the consumer route `maxDuration` to 800s in `vercel.json` so the first attempt has headroom for slower video providers before relying on the resume path.

app/api/queues/asset-generate/route.ts
```diff
@@ -0,0 +1,27 @@
...
+	async ({ jobId, connectorType }) => {
+		const job = await loadJobForProcessing(jobId);
+		// Vercel Queues may redeliver; skip terminal states so we never re-run a finished job.
+		if (job.status === "completed" || job.status === "failed") return;
+
+		await updateJob(jobId, { status: "processing" });
+		try {
+			const result = await runAssetJob(connectorType, job.request);
+			await updateJob(jobId, { status: "completed", result });
+		} catch (error) {
```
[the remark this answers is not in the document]
umairnadeem: Fixed in 3a105c7. The try/catch is now scoped only around `runAssetJob`; the completed-state write happens after the catch. If that write fails transiently the consumer throws and the queue redelivers — the row is still `processing` (not `failed`), so the next attempt re-enters `runAssetJob` and (for video) resumes polling via `provider_job_id` instead of losing the work.

umairnadeem:
@codex review

lib/api/jobs.ts
```diff
@@ -0,0 +1,108 @@
...
+		result?: BundleResponse;
+		error?: string;
+		providerJobId?: string;
+	},
+): Promise<void> {
+	const supabase = createServiceClient();
+	const { error } = await supabase
+		.from("jobs")
+		.update({
+			...(patch.status !== undefined && { status: patch.status }),
```
umairnadeem: would rather do:

```
	const update = omitBy(
		{
			status: patch.status,
			result: patch.result,
			error: patch.error,
			provider_job_id: patch.providerJobId,
		},
		isUndefined,
	);

```

lib/api/providers.ts
```diff
@@ -64,3 +68,73 @@ export const getTTSProvider = defineProvider(
...
+
+export type AssetProvider = {
+	generate(params: Record<string, unknown>): Promise<BundleResponse>;
+	poll?(jobId: string): Promise<VideoProviderResponse>;
+};
+
+export function getAssetProvider(type: ConnectorType): AssetProvider {
+	switch (type) {
+		case "image":
+			return getImageProvider() as AssetProvider;
```
umairnadeem: remove these casts

lib/api/route-handler.ts
```diff
@@ -1,10 +1,40 @@
...
+	label: string,
+): Promise<ParseResult<z.infer<TSchema>>> {
+	const parsed = schema.safeParse(await request.json());
+	if (parsed.success) return { ok: true, data: parsed.data };
+	const message = parsed.error.issues[0]?.message ?? "Invalid request body";
+	logger.warn(`${label}: ${message}`);
+	return { ok: false, response: badRequest(message) };
+}
+
+async function withAuth(
```
umairnadeem: move to a more central place

lib/api/route-handler.ts
```diff
@@ -1,10 +1,40 @@
...
+import type { ConnectorType } from "@/lib/connectors/types";
 import { stringifyError } from "../errors";
 import { getUser } from "./auth";
+import { createJob, enqueueJob, getJob } from "./jobs";
 import { badRequest, serverError, unauthorized } from "./response";
 import { logger } from "./logger";
 
+type ParseResult<T> = { ok: true; data: T } | { ok: false; response: Response };
+
+async function parseBody<TSchema extends z.ZodType>(
```
umairnadeem: move to a more central place so it can be reused

lib/api/route-handler.ts
```diff
@@ -59,6 +78,54 @@ export function modelField(models: Record<string, string>) {
 		.transform((v) => (v === undefined ? undefined : models[v]));
 }
 
+export function createAssetRouteHandlers<TSchema extends z.ZodType>(opts: {
```
umairnadeem: createAssetRouteHandlers should re-use createRouteHandler

lib/api/route-handler.ts
```diff
@@ -59,6 +78,54 @@ export function modelField(models: Record<string, string>) {
...
+				request: providerRequest,
+			});
+			await enqueueJob(job.id, opts.connectorType);
+			return NextResponse.json({ jobId: job.id, status: "pending" });
+		});
+
+	return { POST };
+}
+
+export async function pollJob(
```
umairnadeem, line 62: should reuse createRouteHandler

lib/connectors/__tests__/_gateway-mock.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { JobPoll, JobStatus } from "@/lib/gateway/base";
+
+function jsonResponse(data: unknown) {
+	return new Response(JSON.stringify(data), {
+		status: 200,
+		headers: { "content-type": "application/json" },
+	});
+}
+
+export type GatewayStep =
```
umairnadeem: should be a single interface

lib/gateway/base.ts
```diff
@@ -1,3 +1,19 @@
...
+export type JobStatus = "pending" | "processing" | "completed" | "failed";
+
+export type JobPoll = {
+	jobId: string;
+	status: JobStatus;
+	result: BundleResponse | null;
+	error: string | null;
+};
+
+export interface AssetGateway<TParams> {
```
umairnadeem: AssetGateway should extend/implement GatewayClient

lib/generation/scheduleGeneration.ts
```diff
@@ -15,5 +15,11 @@ export function scheduleGeneration(
 	{ projectId, registry }: { projectId: string; registry: ConnectorRegistry },
 ): void {
 	ensureCharacterAvatars(queue, projectId, registry);
-	queue.enqueueAll(Array.isArray(jobs) ? jobs : [jobs]);
+	const list = Array.isArray(jobs) ? jobs : [jobs];
+	queue.enqueueAll(
+		list.map((job) => ({
+			...job,
+			extraParams: { ...job.extraParams, projectId },
```
umairnadeem: instead of plumbing projectId through here, it should simply be added to `buildGenerationJob` and plumbed via fn param

lib/project/ensureCharacterAvatars.ts
umairnadeem, line 22: should ideally use buildGenerationJob (may need to be modified to allow overrides)

app/api/queues/asset-generate/route.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { AssetQueueMessage } from "@/lib/api/jobs";
+import { loadJobForProcessing, updateJob } from "@/lib/api/jobs";
+import { runAssetJob } from "@/lib/api/providers";
+import { stringifyError } from "@/lib/errors";
+
+export const POST = handleCallback<AssetQueueMessage>(
+	async ({ jobId, connectorType }) => {
+		const job = await loadJobForProcessing(jobId);
+		// Vercel Queues may redeliver; skip terminal states so we never re-run a finished job.
+		if (job.status === "completed" || job.status === "failed") return;
```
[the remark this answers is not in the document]
umairnadeem: overengineering, ignore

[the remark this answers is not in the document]
umairnadeem: Acknowledged but explicitly out of scope for this PR — accepting that a rare duplicate concurrent delivery may double-process. Will revisit if it shows up in practice.

lib/api/jobs.ts
umairnadeem: Done in 934599d: `updateJob` now uses `omitBy(patch, isUndefined)` so the patch shape just maps directly to the column shape.

lib/api/providers.ts
umairnadeem: Done in 934599d. `AssetProvider` and the `as` casts are deleted — `getAssetProvider` is no longer needed because the per-connector `JobHandler` registry calls each provider's getter directly.

lib/api/route-handler.ts
umairnadeem: Done in 934599d: `withAuth` lives in `lib/api/with-auth.ts` now (separate from `auth.ts` so each can be mocked independently in tests).

lib/api/route-handler.ts
umairnadeem: Done in 934599d: moved to `lib/api/parse.ts` as a standalone helper.

lib/api/route-handler.ts
umairnadeem: Done in 934599d. Generalized `createRouteHandler` to a single primitive `({user, body, params, request}) => Response` that runs auth + (optional) body parse. `createAssetRouteHandlers` is now ~15 lines on top of it, and the LLM route uses the same primitive.

lib/api/route-handler.ts
umairnadeem: Done in 934599d: `pollJob` is now `createRouteHandler<undefined, { jobId: string }>({ ... })` — no schema, just params + the registry dispatch to `handler.poll`.

lib/connectors/__tests__/_gateway-mock.ts
umairnadeem: Done in 934599d: flattened to a single `GatewayStep` shape `{ submitStatus?, pollStatus?, result?, error?, payload? }`. `mockGatewaySuccess` is a 4-line wrapper over `mockGatewaySequence`.

lib/gateway/base.ts
umairnadeem: Done in 934599d: `AssetGateway` is now an abstract class extending `GatewayClient<TParams, JobSubmission>` with an abstract `poll` method. `OpenSlopAssetGateway` extends it directly.

lib/generation/scheduleGeneration.ts
umairnadeem: Done in 934599d. `buildGenerationJob(element, registry, projectId, overrides?)` stamps `projectId` into `extraParams` directly; `scheduleGeneration` is back to a thin queue dispatch — no more `.map` shim.

lib/project/ensureCharacterAvatars.ts
umairnadeem: Looked at this carefully — `ensureCharacterAvatars` and `buildGenerationJob` actually have nothing meaningful to share: avatars synthesize a non-element job (different elementId scheme, fixed connectorType, custom avatar plugins on `config`, synthetic inputs, empty prompt that would otherwise short-circuit `buildGenerationJob`). Threading all of that through `overrides` ends up uglier than the explicit construction. Left the projectId stamping in place there. Happy to revisit if you have a specific shape in mind.

umairnadeem:
@codex review

PR #205, opened 2026-06-04
describe uploaded character avatars to the llm

umairnadeem:
## Summary
- New `character-avatar-style` LLM plugin: for every character with an uploaded avatar, sends the avatar URL through the gateway for a concise visual description and prepends a per-character `Character appearances:` block to the prompt — mirrors the existing `reference-style` plugin.
- Adds `avatarUploaded?: boolean` to `MetadataCharacter`; `CharacterEditModal` sets it on upload (and no longer discards the queue snapshot), clears it on regenerate, and skips the `isStale` check when an uploaded avatar is in place.
- Updates `osml` prompt to instruct the model to reference characters by name and never re-describe their appearance in image descriptions; mock LLM now returns a short canned response for character-appearance prompts; `reference-style` bumps `maxTokens` to 4096.

## Test plan
- [ ] Lint / format / typecheck / tests pass (CI).
- [ ] Upload a distinct avatar on a character — stale indicator does not appear even after editing `appearance`.
- [ ] Trigger an LLM generation with one uploaded + one generated character; verify only the uploaded one appears in the `Character appearances:` block of the outgoing prompt.
- [ ] Press regenerate on an uploaded avatar — flag clears, queue runs, stale behavior resumes once the new result lands.
- [ ] Reload the project: uploaded characters keep `avatarUploaded: true` and remain non-stale.

### DIFF
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -81,7 +81,11 @@ function CharacterEditDialogBody({
 		onUpload: ([url]) => {
 			if (!url) return;
 			queue.discard(avatarElementId);
-			setCharacter(name, { ...character, avatarUrl: url });
+			setCharacter(name, {
+				...character,
+				avatarUrl: url,
+				avatarUploaded: true,
+			});
 		},
 	});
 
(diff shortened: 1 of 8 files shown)


umairnadeem:
@codex review

PR #207, opened 2026-06-05
YouTube-style video player controls with canvas scene sync

umairnadeem:
## Summary
- Reworks the custom video player into a YouTube-like overlay: full-width segmented seek bar above a buttons row, cursor-following tooltip with the hovered scene's foreground thumbnail, center play/pause flash on every toggle, controls fade after pointer idle, scene pill next to the time readout.
- Plumbs canonical scene indices end-to-end: a pure `sceneIndexOf` helper in `lib/canvas/scenes.ts` powers both the canvas (`useSceneIndex`) and the player (`useSceneSegments`), so the player's "Scene N" labels stay in lockstep with the canvas even when some scenes have no foreground asset yet.
- Adds a shared `ActiveSceneContext` (split value/setter contexts) that the player writes to from playback frame via a small `ActiveSceneSync` leaf, and the canvas reads to highlight the currently-playing `SceneContainer` — future-proof for click-to-seek in the reverse direction.
- Performance: frame-driven UI is isolated into tiny leaves (`TimeDisplay`, `ScenePill`, `SeekProgress`) with derived-primitive `useSyncExternalStore` selectors; pointer-hover updates are rAF-coalesced; tooltip thumbnails are 80ms-debounced; frame subscriptions are gated by controls visibility.

## Test plan
- [ ] `npm run lint`, `npm run typecheck`, `npm run format:check`, `npm run test:run` all pass locally (792 tests).
- [ ] Open a project with multiple scenes — seek bar above controls with visible gaps; scrub through the bar and verify a single cursor-following tooltip with thumbnail + time + scene label.
- [ ] Click play/pause and click the video surface — center icon flashes and fades both times.
- [ ] Stop moving the pointer over the player — controls fade out after ~2s; move again — they fade back in.
- [ ] Resize the top-anchored player — controls overlay never bleeds past the rounded video panel.
- [ ] During playback, verify the highlighted scene in the canvas matches the player's scene pill / seek bar position, including when an earlier scene lacks a foreground asset.

### DIFF
--- a/app/components/PostPromptView.tsx
+++ b/app/components/PostPromptView.tsx
@@ -21,6 +21,9 @@ import {
 	PlayerPositionProvider,
 	usePlayerPosition,
 } from "./video/PlayerPositionContext";
+import { ActiveSceneProvider } from "./scene-selection/ActiveSceneContext";
+import { AutoScrollProvider } from "./scene-selection/AutoScrollContext";
+import { PlayerControlProvider } from "./video/PlayerControlContext";
 import { VideoLayoutProvider } from "./video/VideoLayoutContext";
 import editorStyles from "./Editor.module.css";
 import genStyles from "./styles/gen-button.module.css";
@@ -104,22 +107,28 @@ function PostPromptViewInner() {
 			</div>
 
 			<VideoLayoutProvider getEditor={getEditor} layoutKey={layoutKey}>
-				{visible && isTop && <TopPlayerPanel />}
+				<PlayerControlProvider>
+					<ActiveSceneProvider>
+						<AutoScrollProvider>
+							{visible && isTop && <TopPlayerPanel />}
 
-				<div className="flex min-h-0 flex-1 overflow-hidden">
-					<div
-						className="flex-1 overflow-y-auto"
-						style={{ scrollbarGutter: "stable" }}
-					>
-						<div className="pointer-events-none sticky top-0 z-10 -mb-8 h-8 backdrop-blur-sm [mask-image:linear-gradient(to_bottom,black,transparent)]" />
-						<div className="mx-auto max-w-6xl px-4 py-4">
-							<ProjectTitle />
-							<Canvas ref={canvasRef} onLayoutKeyChange={setLayoutKey} />
-						</div>
-					</div>
+							<div className="flex min-h-0 flex-1 overflow-hidden">
+								<div
+									className="flex-1 overflow-y-auto"
+									style={{ scrollbarGutter: "stable" }}
+								>
+									<div className="pointer-events-none sticky top-0 z-10 -mb-8 h-8 backdrop-blur-sm [mask-image:linear-gradient(to_bottom,black,transparent)]" />
+									<div className="mx-auto max-w-6xl px-4 py-4">
+										<ProjectTitle />
+										<Canvas ref={canvasRef} onLayoutKeyChange={setLayoutKey} />
+									</div>
+								</div>
 
-					{visible && !isTop && <SidePlayerPanel />}
-				</div>
+								{visible && !isTop && <SidePlayerPanel />}
+							</div>
+						</AutoScrollProvider>
+					</ActiveSceneProvider>
+				</PlayerControlProvider>
 			</VideoLayoutProvider>
 		</div>
 	);
--- a/app/components/canvas/elements/PlayFromHereButton.module.css
+++ b/app/components/canvas/elements/PlayFromHereButton.module.css
@@ -1,0 +1,30 @@
+.button {
+	width: 24px;
+	height: 24px;
+	border-radius: 50%;
+	background-color: rgba(255, 255, 255, 0.1);
+	border: none;
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	cursor: pointer;
+	transition: background-color 0.2s;
+	color: rgba(255, 255, 255, 0.6);
+}
+
+.icon {
+	width: 10px;
+	height: 10px;
+	fill: currentColor;
+	transition: color 0.2s;
+}
+
+.button:hover:not(:disabled) {
+	background-color: rgb(139, 92, 246);
+	color: white;
+}
+
+.button:disabled {
+	opacity: 0.4;
+	cursor: not-allowed;
+}
--- a/app/components/canvas/elements/PlayFromHereButton.tsx
+++ b/app/components/canvas/elements/PlayFromHereButton.tsx
@@ -1,0 +1,42 @@
+"use client";
+
+import { Play } from "lucide-react";
+import {
+	Tooltip,
+	TooltipContent,
+	TooltipTrigger,
+} from "@/components/ui/tooltip";
+import { usePlayerControl } from "@/app/components/video/PlayerControlContext";
+import { findSceneSequence } from "@/app/components/video/useSceneSegments";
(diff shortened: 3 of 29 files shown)


umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
## Adversarial review (Claude)

> Note: ran via `/codex challenge` but `codex` CLI wasn't installed locally, so this is Claude's own adversarial pass, not a true second-opinion. Install `@openai/codex` and re-run for an independent cross-model review.

### 🔴 Critical / bugs I expect to bite in production

**1. Click-to-play overlay is keyboard-inaccessible — a11y regression.**
`VideoPreview` drops Remotion's built-in `controls` and renders:
```tsx
<div className="absolute inset-0 cursor-pointer" onClick={toggleAndFlash} />
```
No `role`, no `aria-label`, no `tabIndex`, no keydown handler. Space-bar (the de-facto YouTube control) does nothing. Tab order skips it. Screen readers see an unlabelled `<div>`. The previous Remotion controls supported all of this for free.

**2. `pauseWhenBuffering` removed silently from `Html5Audio` and `OffthreadVideo`** (`remotion/compositions/VideoComposition.tsx`).
With this gone, when any single audio/video stalls (slow network, large source, cold CDN edge), the frame counter keeps advancing while the stream is silent or freezes — A/V drift, silent VOs over moving picture, no auto-recovery. Nothing in the new pause-on-drag code substitutes for this. If the removal was intentional (maybe fighting the new `silenceMediaIn` drag logic), it needs a code comment + manual test note. As written it reads like an accidental side-effect.

**3. `playFromFrame` race when the player is hidden.**
```ts
const playFromFrame = (frame) => {
  showPlayer();
  playerRef.current?.seekTo(frame);
  playerRef.current?.play();
};
```
`showPlayer()` mounts the player; `VideoPreview` then mounts and calls `registerPlayer(player)` *after* Remotion's Player ref settles. On a cold first click of `PlayFromHereButton` (player hidden), `playerRef.current` is `null` → the `?.` no-ops → nothing happens. Second click works. Need a "pending seek" queued at register time, or a `useEffect` that flushes the queued frame when the ref appears.

**4. Seek tooltip never previews a video frame.**
`SeekTooltip` renders `<video src={url} preload="metadata" muted playsInline />` with no `currentTime` set. Browsers paint the poster / first decoded frame and stop. So:
- Scrubbing across a video scene always shows that scene's *first* frame, regardless of cursor position within the segment.
- The "settled" thumbnail is identical to the un-settled thumbnail — the 80 ms settle delay buys nothing visual.

Either own that and rename/restyle (single scene-cover thumbnail, no fake settle delay), or actually set `el.currentTime = hoveredTimeSec - segment.start` and pause on `loadeddata`. Without one of those, "YouTube-style" is misleading.

**5. Seek-bar progress dot can sit outside the painted segments.**
`useSceneSegments` shortens each segment by `transitionDurationSec`:
```ts
prev.duration = Math.max(0, prev.duration - layout.transitionDurationSec);
```
…but `SeekProgress` positions the knob at `frame / (totalFrames - 1)` over the *full* track width. The sum of segment widths < 100 %, while the dot reaches 100 %. Near the end of playback the dot drifts right of the last violet fill — visible mismatch between "what's painted" and "what's playing."

**6. `silenceMediaIn` fights Remotion's lifecycle during drag.**
On `pointerdown` (and again on every `pointermove`) the code `el.pause()`s every `<audio>`/`<video>` in the container. Remotion remounts media as Sequences enter/exit. Two failure modes:
- A Sequence enters *between* pointermoves and starts playing again (you missed it) — audible chirp during scrub.
- On `pointerup` you call `player.play()` and trust Remotion to re-play every tag. Anything Remotion considers "currently inside the cursor frame" should resume; anything Remotion thinks is still mid-sequence but was paused by you may stay paused until the next sequence boundary → silent video. Worth a Playwright test that scrubs across overlapping audio scenes and asserts audio resumes within ~50 ms after pointerup.

### 🟠 Performance / correctness smells

**7. Frame-update fan-out.** `TimeDisplay`, `ScenePill`, `SeekProgress`, and `ActiveSceneSync` each independently `useSyncExternalStore` on `frameupdate`. That's 4 separate React state updates per frame, two of which call `findSegmentIndexAt` (O(N) scan over segments). For a 200-scene project at 30 fps that's 12 000 segment scans/sec. Either hoist one `useFrame` and pass `frame` / `activeIndex` down through context, or memoize `findSegmentIndexAt` with a cached "last hit" pointer (segments are sorted).

**8. Smooth-scroll auto-scroll thrash.**
`ActiveSceneSync` calls `scrollToScene(activeId)` on every scene boundary. `scrollIntoView({ behavior: "smooth" })` is async and gets cancelled by the next call. On a montage with quick cuts (< 1 s scenes) the canvas starts scrolling toward N, gets yanked toward N+1 mid-flight, etc. Fall back to `behavior: "auto"` when interval < ~600 ms, or throttle.

**9. `useVideoLayout` memo still has `// eslint-disable-next-line react-hooks/exhaustive-deps`** (pre-existing). Adding `scenes` into the returned object didn't fix it; flagging so it doesn't keep biting.

### 🟡 Smaller issues / nits

- **No tests for `useSeekDrag` or `SegmentedSeekBar`** — the most stateful additions in the PR. `useSceneSegments.test.ts` only covers the trivial binary search.
- **`useControlsVisibility`** has no listener for keyboard input. Tabbing through controls won't keep the chrome visible.
- **`<input type=\"range\" value={muted ? 0 : volume}>`**: when muted the slider snaps to 0, then jumps back when the user unmutes. YouTube keeps the slider position while muted — minor UX deviation.
- **`onMouseDown={(e) => e.preventDefault()}` in `PlayFromHereButton`** is a focus-stealing workaround for Slate — comment-worthy because it'll look like a mistake to a future reader.
- **`PlayPauseFlash`** uses `key={flash.key}` to restart the CSS animation; works, but if the user hammers the overlay >2/sec the animation visibly skips. Consider debouncing.
- **CSS `!important` overrides removed** from `VideoPlayer.module.css`. Confirm border-radius and overflow still take effect now that native controls and inline styles aren't being overridden — easy regression to ship if Remotion sets inline `border-radius`.
- **`ActiveSceneSync` cleanup nulls `activeSceneId` on unmount**, which will flash the violet ring off the active scene during any unrelated re-mount (e.g. flipping `visible`). Tolerable; worth mentioning in the PR description.

### ✅ What's good

- Splitting `usePlayerState` into `useSyncExternalStore` selectors is the right move and is well-bounded.
- `useSceneSegments` test coverage on boundary conditions is precise (good `<` vs `≤` handling).
- Replacing the brittle attribute-selector CSS overrides (`div[style*=\"rgba(255,255,255,1)\"]`) with a real component is a clear maintainability win.
- `AutoScrollProvider` + `ActiveSceneProvider` (value/setter contexts) avoids unnecessary re-renders. Idiomatic React 19.

### Recommendation
Top three to fix before merge: **(1)** keyboard accessibility on the click-overlay, **(2)** justify or restore `pauseWhenBuffering`, **(3)** the `playFromFrame`-while-hidden race. (4)–(6) materially affect the "YouTube-style" framing in the PR title and should be addressed or scoped down honestly.

umairnadeem:
@codex review

PR #208, opened 2026-06-05
fix: nightly bug fixes 2026-06-05

umairnadeem:
## Summary
- Fixes template removal so clearing the composer template pill also resets template-seeded project state instead of only switching back to story mode.
- Adds regression coverage for clearing template-seeded reference images, characters, narration, and style.

## Test Plan
- [x] npm test -- --passWithNoTests
- [x] npm run lint
- [x] npm run format:check
- [x] npm run typecheck
- [x] npm run build

## Notes
- Verified recent PR #202 review concern still reproduced on master before fixing.
- No cosmetic-only changes.


umairnadeem:
this is intentional

PR #210, opened 2026-06-07
fix: nightly bug fixes 2026-06-07

umairnadeem:
## Summary
- Queue play-from-frame requests until the lazy Remotion player registers so first-click playback from the canvas does not silently no-op.
- Make the click-to-play overlay keyboard/screen-reader accessible.
- Keep controls visible after touch pointer interactions instead of hiding them immediately on touch leave.
- Add focused unit coverage for deferred player command behavior.

## Test Plan
- npm run build
- npm run lint
- npm run format:check
- npm run typecheck

### DIFF
--- a/app/components/video/PlayerControlContext.tsx
+++ b/app/components/video/PlayerControlContext.tsx
@@ -1,48 +1,17 @@
 "use client";
 
-import type { PlayerRef } from "@remotion/player";
-import {
-	createContext,
-	use,
-	useCallback,
-	useMemo,
-	useRef,
-	type ReactNode,
-} from "react";
+import { createContext, use, useMemo, type ReactNode } from "react";
 import { usePlayerPosition } from "./PlayerPositionContext";
+import { createPlayQueue, type PlayQueue } from "./playQueue";
 
-type PlayerControl = {
-	registerPlayer: (player: PlayerRef | null) => void;
-	playFromFrame: (frame: number) => void;
-};
-
-const Ctx = createContext<PlayerControl>({
+const Ctx = createContext<PlayQueue>({
 	registerPlayer: () => {},
 	playFromFrame: () => {},
 });
 
 export function PlayerControlProvider({ children }: { children: ReactNode }) {
-	const playerRef = useRef<PlayerRef | null>(null);
 	const { showPlayer } = usePlayerPosition();
-
-	const registerPlayer = useCallback((p: PlayerRef | null) => {
-		playerRef.current = p;
-	}, []);
-
-	const playFromFrame = useCallback(
-		(frame: number) => {
-			showPlayer();
-			playerRef.current?.seekTo(frame);
-			playerRef.current?.play();
-		},
-		[showPlayer],
-	);
-
-	const value = useMemo(
-		() => ({ registerPlayer, playFromFrame }),
-		[registerPlayer, playFromFrame],
-	);
-
+	const value = useMemo(() => createPlayQueue(showPlayer), [showPlayer]);
 	return <Ctx value={value}>{children}</Ctx>;
 }
 
--- a/app/components/video/PlayerControlContext.tsx
+++ b/app/components/video/PlayerControlContext.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { createContext, use, useMemo, type ReactNode } from "react";
+import { createContext, use, useState, type ReactNode } from "react";
 import { usePlayerPosition } from "./PlayerPositionContext";
 import { createPlayQueue, type PlayQueue } from "./playQueue";
 
@@ -11,8 +11,8 @@ const Ctx = createContext<PlayQueue>({
 
 export function PlayerControlProvider({ children }: { children: ReactNode }) {
 	const { showPlayer } = usePlayerPosition();
-	const value = useMemo(() => createPlayQueue(showPlayer), [showPlayer]);
-	return <Ctx value={value}>{children}</Ctx>;
+	const [queue] = useState(() => createPlayQueue(showPlayer));
+	return <Ctx value={queue}>{children}</Ctx>;
 }
 
 export function usePlayerControl() {
--- a/app/components/video/VideoPreview.tsx
+++ b/app/components/video/VideoPreview.tsx
@@ -2,7 +2,7 @@
 
 import type { PlayerRef } from "@remotion/player";
 import dynamic from "next/dynamic";
-import { useEffect, useState, type PointerEvent, type Ref } from "react";
+import { useEffect, useState, type Ref } from "react";
 import type { VideoLayout } from "@/lib/video/types";
 import { ToastErrorBoundary } from "../ToastErrorBoundary";
 import { ActiveSceneSync } from "./ActiveSceneSync";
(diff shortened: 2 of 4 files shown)


umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem:
@codex review

umairnadeem (commented):
Adversarial review: mixed. The playQueue extraction with tests is a legitimate quality improvement. However the VideoPreview changes introduce real regressions: a size class was dropped, the button overlay almost certainly double-toggles on keyboard activation, and the button stacking probably blocks PlayerControls clicks. The pointer-leave touch behavior change is speculative — controls already auto-hide via the IDLE timer, so the previous immediate-hide-on-touch-leave was arguably correct UX.

app/components/video/VideoPreview.tsx
```diff
@@ -69,20 +69,32 @@ export function VideoPreview({ layout }: { layout: VideoLayout }) {
...
+		leave();
+	};
+	const handleKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+		if (e.key !== " " && e.key !== "Enter") return;
+		e.preventDefault();
+		toggleAndFlash();
+	};
 	return (
 		<div
 			className={`relative ${styles.player}`}
```
umairnadeem: Regression: the previous className was `relative h-full w-full ${styles.player}` — `h-full w-full` was dropped here. The player container will collapse unless `styles.player` happens to set both dimensions. Likely unintended.

app/components/video/VideoPreview.tsx
```diff
@@ -69,20 +69,32 @@ export function VideoPreview({ layout }: { layout: VideoLayout }) {
...
 			<ToastErrorBoundary label="Player">
 				<RemotionPlayer layout={layout} playerRef={setPlayer} />
 			</ToastErrorBoundary>
-			<div
-				className="absolute inset-0 cursor-pointer"
+			<button
+				type="button"
+				aria-label="Play or pause"
 				onClick={toggleAndFlash}
+				onKeyDown={handleKeyDown}
```
umairnadeem: Double-toggle bug. A native `<button>` already fires `onClick` on Enter and on Space keyup. `onKeyDown` calls `toggleAndFlash()` and then the native activation fires `onClick` → `toggleAndFlash()` again. `preventDefault()` on keydown suppresses Space's default scroll but does not cancel the button's click synthesis on Enter. Either drop `onKeyDown` entirely (the button handles keyboard natively) or use `onClick` only.

app/components/video/VideoPreview.tsx
```diff
@@ -69,20 +69,32 @@ export function VideoPreview({ layout }: { layout: VideoLayout }) {
...
+				className="absolute inset-0 cursor-pointer bg-transparent focus:outline-none focus-visible:ring-2 focus-visible:ring-white/40"
```
umairnadeem: Stacking/interaction concern: `absolute inset-0` button is rendered before `<PlayerControls>` but neither has explicit z-index. PlayerControls likely also use absolute positioning — verify clicks on the seek bar / play / mute / fullscreen still land. A focusable full-area button also makes Tab order land on a giant invisible target before reaching real controls.

app/components/video/VideoPreview.tsx
```diff
@@ -69,20 +69,32 @@ export function VideoPreview({ layout }: { layout: VideoLayout }) {
 		player.toggle();
 		setFlash((f) => ({ key: (f?.key ?? 0) + 1, playing: willPlay }));
 	};
+	const handlePointerLeave = (e: PointerEvent<HTMLDivElement>) => {
+		if (e.pointerType !== "mouse") return;
+		leave();
```
umairnadeem: Is this fixing a real bug? `useControlsVisibility.ping` already restarts the 2s idle timer on every pointer event, so on touch the controls would hide ~2s after the last tap regardless. The previous immediate-hide on pointerleave was actually reasonable. This change is speculative and now leaves controls visible indefinitely after a touch tap until something else fires.

app/components/video/playQueue.ts
```diff
@@ -0,0 +1,35 @@
...
+	let pendingFrame: number | null = null;
+	return {
+		registerPlayer(p) {
+			player = p;
+			if (p && pendingFrame != null) {
+				const frame = pendingFrame;
+				pendingFrame = null;
+				p.seekTo(frame);
+				p.play();
+			}
```
umairnadeem, line 29: Silent failure mode (CLAUDE.md: critical paths must fail loudly). If `registerPlayer(null)` is called after a deferred frame is consumed and then a new player registers, the user's play intent is lost without a warning. Worse: if the player registers but isn't yet ready to seek, `seekTo`/`play` may no-op silently — the same bug this PR claims to fix. Consider logging when a pending frame is dropped due to unregister.

app/components/video/PlayerControlContext.tsx
```diff
@@ -1,48 +1,17 @@
...
-		},
-		[showPlayer],
-	);
-
-	const value = useMemo(
-		() => ({ registerPlayer, playFromFrame }),
-		[registerPlayer, playFromFrame],
-	);
-
+	const value = useMemo(() => createPlayQueue(showPlayer), [showPlayer]);
```
umairnadeem: `createPlayQueue` is invoked inside `useMemo` which is documented as a performance hint, not a guarantee — React may discard and re-run it, blowing away `pendingFrame` and the registered player. Use `useRef` to hold the queue instance instead, or accept the risk explicitly.

app/components/video/playQueue.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export function createPlayQueue(showPlayer: () => void): PlayQueue {
+	let player: PlayerRef | null = null;
+	let pendingFrame: number | null = null;
+	return {
+		registerPlayer(p) {
+			if (!p && pendingFrame != null) {
+				console.warn(
+					`[playQueue] dropping pending frame ${pendingFrame}: player unregistered before it could play`,
+				);
```
umairnadeem, line 20: Player swap drops a pending frame with a noisy warning.

In `VideoPreview`, the effect `useEffect(() => { registerPlayer(player); return () => registerPlayer(null); }, [player, registerPlayer])` runs cleanup-before-effect on every `player` change. If the Remotion Player instance is ever re-created (layout change, error boundary reset, key change, etc.) while a frame is pending, the sequence is:

1. cleanup → `registerPlayer(null)` → warns + drops pendingFrame
2. effect → `registerPlayer(newPlayer)` → no pending frame anymore, never plays

The user-visible bug: the click that triggered `playFromFrame` silently no-ops, plus a console.warn fires for what is actually a normal lifecycle event — not the 'dropped because no player will ever mount' case the warning is meant to surface.

React StrictMode in dev triggers the exact same cleanup→re-mount cycle on every effect, so any pending frame at mount time gets warned and dropped during dev double-invocation too.

Suggest distinguishing 'unregister with a different player coming' from 'unregister and gone'. Simplest fix: only warn/drop on `registerPlayer(null)` if the pending frame has actually been waiting across a tick (e.g., schedule the drop via `queueMicrotask` and cancel it if a real player registers before it runs), or expose a separate `drainPending()` API used only when the provider unmounts.

umairnadeem (commented):
Second adversarial review — fix verification + fresh pass

All previously raised issues are correctly addressed:
- VideoPreview wrapper has `h-full w-full` restored (line 74).
- Overlay reverted to `<div>` — no double-fire on Enter, no Tab-order pollution.
- `useControlsVisibility.leave` is the simple immediate clear; `onPointerLeave` reverted.
- `playQueue.registerPlayer(null)` warns loudly and drops the pending frame, per CLAUDE.md fail-loudly convention.
- `PlayerControlProvider` uses `useState` lazy init — queue identity is stable for the provider's lifetime (and `showPlayer` is a stable `useCallback([])` from `PlayerPositionContext`, so capturing it at init is safe).
- `playQueue.test.ts` covers immediate-play, deferred-play, re-register no-replay, warn-on-unregister, and latest-wins-on-multiple-calls. Good coverage.

One legitimate residual issue noted inline (player-swap false-positive warn). Otherwise LGTM.

umairnadeem:
overengineered

PR #211, opened 2026-06-07
refactor: nightly maintainability 2026-06-07

umairnadeem:
## Summary
- Add an `ActiveSegmentProvider` so video controls share one active segment calculation instead of each subscribing/scanning independently.
- Remove `player`/`layout` prop drilling from `ScenePill` and `ActiveSceneSync` for the active segment lookup.
- Change `findSegmentIndexAt` from linear scan to binary search over sorted segments.

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run build`

## Notes
- Behavior unchanged: consumers still use the same current frame and segment boundaries, but the derived active index is centralized.
- Independent reviewer found no blocking security or logic issues.


app/components/video/PlayerControls.tsx
```diff
@@ -61,7 +62,7 @@ export function PlayerControls({
 					)}
 				</IconButton>
 				<TimeDisplay player={player} layout={layout} />
-				<ScenePill player={player} layout={layout} segments={segments} />
+				<ScenePill segments={segments} />
```
umairnadeem: Stale base: master wraps this in `<div className="hidden @[280px]:contents">` (commit 07068ac, aspect-ratio PR). This PR replaces the old single-line ScenePill call and silently drops that responsive wrapper. Rebase before merging or the narrow-player hide behavior regresses.

app/components/video/ActiveSegmentContext.tsx
```diff
@@ -0,0 +1,37 @@
+"use client";
+
+import type { PlayerRef } from "@remotion/player";
+import { createContext, type ReactNode, useContext } from "react";
+import type { VideoLayout } from "@/lib/video/types";
+import { usePlayerValue } from "./usePlayerState";
+import { findSegmentIndexAt, type SceneSegment } from "./useSceneSegments";
+
+const ActiveSegmentIndexContext = createContext<number>(-1);
```
umairnadeem: New mandatory provider for two consumers that previously worked standalone. ScenePill and ActiveSceneSync now silently return wrong data (-1 / null) if ever rendered outside ActiveSegmentProvider — a non-obvious invariant introduced for no real benefit. The previous duplication was two cheap useSyncExternalStore subscriptions; this trades that for a hidden coupling.

app/components/video/ActiveSegmentContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+	return (
+		<ActiveSegmentIndexContext.Provider value={activeIndex}>
+			{children}
+		</ActiveSegmentIndexContext.Provider>
+	);
+}
+
+export function useActiveSegmentIndex(): number {
+	return useContext(ActiveSegmentIndexContext);
+}
```
umairnadeem: Premature abstraction: a whole file + Provider + hook to share one number between two siblings. Idiomatic alternative is a single `useActiveSegmentIndex(player, layout, segments)` hook called once in VideoPreview and prop-drilled through one level. No context needed, no wrapping invariant.

app/components/video/useSceneSegments.ts
```diff
@@ -28,12 +28,17 @@ export function findSegmentIndexAt(
...
+	const n = segments.length;
+	if (n === 0) return -1;
+	let lo = 0;
+	let hi = n - 1;
+	while (lo < hi) {
+		const mid = (lo + hi) >>> 1;
+		const seg = segments[mid];
+		if (timeSec < seg.start + seg.duration) hi = mid;
+		else lo = mid + 1;
 	}
```
umairnadeem, line 35: Binary search over arrays of single-digit length (one entry per scene). The linear scan was O(n) on n~10; this is O(log n) on the same n. Zero measurable improvement, more code to read, and the new branch is harder to eyeball-verify (boundary: returns lo=n-1 when timeSec exceeds the last segment end, matching the old fallthrough — fine, but not worth re-deriving). Revert to the linear loop.

app/components/video/VideoPreview.tsx
```diff
@@ -85,13 +86,19 @@ export function VideoPreview({ layout }: { layout: VideoLayout }) {
...
-			/>
+			>
+				<ActiveSceneSync player={player} segments={segments} />
+				<PlayerControls
+					player={player}
+					layout={layout}
+					segments={segments}
+					visible={visible}
+				/>
+			</ActiveSegmentProvider>
```
umairnadeem: Extra JSX nesting for a context that wraps exactly two siblings inside the same component that owns `player`, `layout`, and `segments`. Just compute `activeIndex` here with a single `usePlayerValue` call and pass it down — no provider, no new file, no invariant to remember.

umairnadeem (commented):
Adversarial review of a nightly auto-refactor. Recommendation: **close**.

The "win" here is a context provider that deduplicates two cheap subscriptions and a binary search over arrays of ~10 scenes. There is no measurable perf gain (O(log n) over single-digit n vs O(n) — identical wall time), and we trade self-contained components for a context dependency that callers must remember to wrap. Net maintainability is negative: one new file, one new mandatory provider, two components that now silently break if rendered outside the provider (they will read `-1` from context and render nothing or report the wrong active scene).

Worse, the patch is stale: it was generated against `f493612` and silently drops the `@[280px]:contents` responsive wrapper around `<ScenePill>` that landed in `07068ac` (aspect-ratio PR). A 3-way merge may resolve cleanly while losing that wrapper — a real visual regression on narrow players. Re-base before considering.

Per repo memory: nightly refactor PRs with trivial gains should be closed as useless. This qualifies.

app/components/video/PlayerControls.tsx
```diff
@@ -61,7 +63,9 @@ export function PlayerControls({
 					)}
 				</IconButton>
```
umairnadeem, line 64: Nit / inconsistency: you restored the `@container` wrapper + `hidden @[280px]:contents` around `<ScenePill>` from commit 07068ac, but the same commit also gave the volume slider a `hidden @[240px]:block` class so it hides at narrow widths. In this PR the volume `<input type="range">` below is still `className={styles.volume}` (no responsive hide). When 07068ac lands the merge will silently reintroduce the responsive volume hide, but until then the controls bar is inconsistent: ScenePill participates in the `@container` queries, the volume slider does not. Worth either (a) adding `hidden @[240px]:block` here too, or (b) explicitly noting that the wrapper restoration is partial-by-design.

umairnadeem (commented):
Second adversarial review — verification pass

PR #212, opened 2026-06-07
perf: nightly performance optimizations 2026-06-07

umairnadeem:
## Summary
- Stabilizes Remotion player event subscriptions used by frame/time/scene/volume controls.
- Reuses shared event lists and hoisted player read selectors so `useSyncExternalStore` does not detach and reattach Remotion listeners on every render.
- Keeps player behavior unchanged: the same events still drive the same snapshots, including active-scene sync and time display updates.

## Why it matters
Playback currently fans out `frameupdate` to multiple React subscribers. When parent UI state changes during playback (hover/controls visibility/resize/active scene updates), unstable subscribe callbacks force extra `addEventListener`/`removeEventListener` churn on the hot player path. This pass removes that runtime churn without changing interaction semantics.

## Skipped
- Did not dedupe active-scene scans across controls and scene sync; that would add broader cross-component coupling for a smaller follow-up win.
- Did not change Slate normalization/plugins or Remotion composition rendering; no cohesive no-behavior-change improvement was safer than this player hot-path fix.
- Did not pursue bundle/lazy-load churn because this runbook is runtime UX-first.

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/app/components/video/ActiveSceneSync.tsx
+++ b/app/components/video/ActiveSceneSync.tsx
@@ -6,7 +6,11 @@ import type { VideoLayout } from "@/lib/video/types";
 import { useSetActiveSceneId } from "@/app/components/scene-selection/ActiveSceneContext";
 import { useAutoScroll } from "@/app/components/scene-selection/AutoScrollContext";
 import { scrollToScene } from "@/app/components/canvas/utils/scrollToScene";
-import { usePlayerPlaying, usePlayerValue } from "./usePlayerState";
+import {
+	PLAYER_FRAME_EVENTS,
+	usePlayerPlaying,
+	usePlayerValue,
+} from "./usePlayerState";
 import { findSegmentIndexAt, type SceneSegment } from "./useSceneSegments";
 
 export function ActiveSceneSync({
@@ -23,7 +27,7 @@ export function ActiveSceneSync({
 	const playing = usePlayerPlaying(player);
 	const activeIndex = usePlayerValue(
 		player,
-		["frameupdate"],
+		PLAYER_FRAME_EVENTS,
 		(p) => findSegmentIndexAt(segments, p.getCurrentFrame() / layout.fps),
 		-1,
 	);
--- a/app/components/video/ActiveSceneSync.tsx
+++ b/app/components/video/ActiveSceneSync.tsx
@@ -6,11 +6,7 @@ import type { VideoLayout } from "@/lib/video/types";
 import { useSetActiveSceneId } from "@/app/components/scene-selection/ActiveSceneContext";
 import { useAutoScroll } from "@/app/components/scene-selection/AutoScrollContext";
 import { scrollToScene } from "@/app/components/canvas/utils/scrollToScene";
-import {
-	PLAYER_FRAME_EVENTS,
-	usePlayerPlaying,
-	usePlayerValue,
-} from "./usePlayerState";
+import { usePlayerPlaying, usePlayerValue } from "./usePlayerState";
 import { findSegmentIndexAt, type SceneSegment } from "./useSceneSegments";
 
 export function ActiveSceneSync({
@@ -27,7 +23,7 @@ export function ActiveSceneSync({
 	const playing = usePlayerPlaying(player);
 	const activeIndex = usePlayerValue(
 		player,
-		PLAYER_FRAME_EVENTS,
+		["frameupdate"],
 		(p) => findSegmentIndexAt(segments, p.getCurrentFrame() / layout.fps),
 		-1,
 	);
--- a/app/components/video/PlayerControls.tsx
+++ b/app/components/video/PlayerControls.tsx
@@ -6,7 +6,6 @@ import type { VideoLayout } from "@/lib/video/types";
 import { scrollToScene } from "@/app/components/canvas/utils/scrollToScene";
 import { formatTime } from "@/lib/video/timestamps";
 import {
-	PLAYER_FRAME_EVENTS,
 	usePlayerMuted,
 	usePlayerPlaying,
 	usePlayerValue,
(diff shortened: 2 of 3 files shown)


app/components/video/usePlayerState.ts
```diff
@@ -8,37 +8,51 @@ type PlayerEvent =
 	| "volumechange"
 	| "mutechange";
 
+export const PLAYER_FRAME_EVENTS: readonly PlayerEvent[] = ["frameupdate"];
```
umairnadeem: Hoisting these four `readonly PlayerEvent[]` constants exists only to stabilize the `useCallback` deps below. It's a code smell: you're working around `useCallback` rather than benefiting from it. For single-element arrays, this is pure ceremony with no runtime win.

app/components/video/usePlayerState.ts
```diff
@@ -8,37 +8,51 @@ type PlayerEvent =
...
 export function usePlayerValue<T>(
 	player: PlayerRef | null,
-	events: PlayerEvent[],
+	events: readonly PlayerEvent[],
 	read: (p: PlayerRef) => T,
 	fallback: T,
 ): T {
-	return useSyncExternalStore(
-		(notify) => {
+	const subscribe = useCallback(
```
umairnadeem: `useCallback` here saves at most one `removeEventListener`+`addEventListener` pair per parent render on a player ref that is itself stable. No profile shows this was hot. Meanwhile `getSnapshot` two lines down is still inline and recreated every render, so `useSyncExternalStore` re-invokes it every render regardless — the asserted "no detach/reattach" win is partial at best.

app/components/video/usePlayerState.ts
```diff
@@ -8,37 +8,51 @@ type PlayerEvent =
...
 
 export function usePlayerMuted(player: PlayerRef | null) {
-	return usePlayerValue(player, ["mutechange"], (p) => p.isMuted(), false);
+	return usePlayerValue(player, MUTECHANGE, readMuted, false);
 }
+
+const readFrame = (p: PlayerRef) => p.getCurrentFrame();
+const readPlaying = (p: PlayerRef) => p.isPlaying();
+const readVolume = (p: PlayerRef) => p.getVolume();
+const readMuted = (p: PlayerRef) => p.isMuted();
```
umairnadeem: Hoisting these one-line reader arrows below their call sites saves allocating four trivial closures per render. That's nanoseconds, costs locality (have to scroll past the hooks to find what they read), and provides no measurable benefit. Premature.

app/components/video/ActiveSceneSync.tsx
```diff
@@ -23,7 +27,7 @@ export function ActiveSceneSync({
 	const playing = usePlayerPlaying(player);
 	const activeIndex = usePlayerValue(
 		player,
-		["frameupdate"],
+		PLAYER_FRAME_EVENTS,
```
umairnadeem: Replacing the inline `["frameupdate"]` literal with an imported constant only matters because of the `useCallback` deps array added in usePlayerState. The `read` arrow on the next line still closes over `segments` and `layout` and is recreated every render, so the snapshot getter is unstable anyway. Net win: negligible.

app/components/video/PlayerControls.tsx
```diff
@@ -102,7 +103,7 @@ function TimeDisplay({
 }) {
 	const seconds = usePlayerValue(
 		player,
-		["frameupdate"],
+		PLAYER_FRAME_EVENTS,
```
umairnadeem: `TimeDisplay` re-renders on every `frameupdate` regardless. Whether the subscribe callback is stable across the parent's renders doesn't change the hot path here — the hot path is the frame event firing, not React re-subscribing. This change does not address that.

umairnadeem (commented):
Adversarial review of nightly perf PR.

**Recommendation: close.** The gains here are speculative and unmeasured. Per project policy, nightly perf PRs with trivial gains get closed as useless — this fits that bucket.

What's actually happening:
- The only behavioral change is wrapping the `subscribe` callback in `useCallback([player, events])` and hoisting event-list constants so its identity is stable across renders.
- That avoids `useSyncExternalStore` detaching/reattaching listeners on parent re-renders. On a player ref that is itself stable, this is a micro-optimization — at most a few `addEventListener`/`removeEventListener` calls per render, which are cheap and were never identified as a bottleneck.
- No benchmark, no profile, no React Profiler delta is included. "Removes runtime churn" is asserted, not measured.

What's noise:
- Hoisting `readFrame`/`readPlaying`/`readVolume`/`readMuted` saves allocating one trivial arrow per render. Zero measurable benefit, hurts locality.
- The four single-element `readonly PlayerEvent[]` constants (`PLAY_PAUSE`, `VOLUMECHANGE`, `MUTECHANGE`) exist solely to make the `useCallback` deps array referentially stable — a smell that the `useCallback` itself is the wrong tool. A simple `useMemo` keyed on event string, or just inlining the subscribe without useCallback, would be clearer.

Missed correctness issue worth noting: `getSnapshot` is still an inline arrow inside `usePlayerValue`, so it's recreated every render and `useSyncExternalStore` invokes it every render anyway. The "churn" reduction is therefore only on the subscribe side, further shrinking the already-tiny win.

If there's a real frame-update hot-path problem, show a profile. Otherwise this trades readability for nothing.

umairnadeem (commented):
Second adversarial review: PR is now fully reverted to master (0 additions, 0 deletions, 0 changed files). The previous round of feedback was addressed by reverting all three files rather than addressing the underlying speculation, which is the correct call given there was no measurement backing the changes.

Recommendation: **close this PR**. There is no meaningful change left to land. Per the project's stance on trivial nightly perf PRs, an empty diff should not be merged just to keep the nightly cadence — close it and let the next nightly run propose changes backed by actual profiling (flame graphs, React Profiler measurements, or benchmark deltas) rather than pattern-matching on "this looks like it could be memoized."

umairnadeem:
closing, empty

PR #213, opened 2026-06-07
refactor: rethink openslop architecture 2026-06-07

umairnadeem:
## App understanding

openslop is a Next.js app for creating AI-generated video projects from an editable canvas/script. Users sign up or log in, create/open projects, add scene elements such as narration, characters, images, animated images, clips, sound effects, and music, configure connector/provider settings, generate media assets through API routes/queue jobs, preview synchronized scenes in the Remotion-backed player, and render/export videos.

## From-scratch architecture

A lean rebuild would keep domain contracts in `lib/*` and leave React components focused on presentation and interaction. Canvas element metadata that drives generation, defaults, and render resolution would live in the canvas domain, while UI-specific labels/icons/attribute controls would live beside the canvas UI. Generation, refinement, and video resolution would depend on domain metadata instead of importing React UI config across architectural boundaries.

- Added `lib/canvas/elementMetadata.ts` as a domain-level source of truth for each canvas element's connector type, output kind, and default attributes.
- Reduced `app/components/canvas/config/elementConfigs.tsx` to UI-only configuration and composed it with domain metadata, removing repeated type/connector/output/default fields from the React config object.
- Updated generation job creation, node creation, connector hydration, and video resolution to consume domain metadata directly instead of depending on the React element config module.
- Updated tests to mock the new metadata boundary where they previously mocked UI element config.

## Validation

- `npm run` — inspected available scripts.
- `npm run lint` — passed.
- `npm run build` — passed. Build emitted the existing Vercel Queue region warning when running locally.
- `npm run test:run` — passed: 89 files / 792 tests.

## Skipped items / risks

- Claude Code reached the configured `--max-turns 40` before completing PR creation, so Hermes completed validation, commit, push, and PR creation manually.
- No behavior-changing simplifications were attempted in generation providers, auth, Supabase persistence, or rendering APIs to keep this nightly change safe and reviewable.

### DIFF
--- a/app/components/canvas/config/elementConfigs.tsx
+++ b/app/components/canvas/config/elementConfigs.tsx
@@ -7,11 +7,9 @@ import {
 	Volume2,
 	Music,
 } from "lucide-react";
-import type { CanvasElementType } from "@/lib/canvas/types";
-import {
-	ELEMENT_METADATA,
-	type ElementMetadata,
-} from "@/lib/canvas/elementMetadata";
+import type { CanvasElementType, ResultKind } from "@/lib/canvas/types";
+import type { ConnectorType } from "@/lib/connectors/types";
+import { ELEMENT_METADATA } from "@/lib/canvas/elementMetadata";
 import { TTSEmotion, TTS_SPEEDS } from "@/lib/connectors/tts/enums";
 import { MOTION_EFFECTS } from "@/lib/video/motionEffects";
 
@@ -25,11 +23,15 @@ export interface AttributeSpec {
 	edit?: AttributeEdit;
 }
 
-export interface ElementConfig extends ElementMetadata {
+export interface ElementConfig {
+	type: CanvasElementType;
+	connector: ConnectorType;
+	outputKind: ResultKind;
 	label: string;
 	icon: React.ReactNode;
 	bgColor: string;
 	placeholder: string;
+	defaultAttributes?: Record<string, string>;
 	visibleAttributes: Record<string, AttributeSpec>;
 }
 
@@ -90,10 +92,10 @@ const emotionSpec: AttributeSpec = {
 	edit: { kind: "enum", options: EMOTION_OPTIONS },
 };
 
-type UIConfig = Omit<ElementConfig, keyof ElementMetadata>;
-
-const UI_CONFIG: Record<CanvasElementType, UIConfig> = {
+export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 	narration: {
+		type: "narration",
+		...ELEMENT_METADATA.narration,
 		label: "Narration",
 		icon: <BookOpen size={16} className="text-white" />,
 		bgColor: "bg-slate-600",
@@ -106,6 +108,8 @@ const UI_CONFIG: Record<CanvasElementType, UIConfig> = {
 		},
 	},
 	character: {
+		type: "character",
+		...ELEMENT_METADATA.character,
 		label: "Character",
 		icon: <User size={16} className="text-white" />,
 		bgColor: "bg-amber-600",
@@ -118,6 +122,8 @@ const UI_CONFIG: Record<CanvasElementType, UIConfig> = {
 		},
 	},
 	image: {
+		type: "image",
+		...ELEMENT_METADATA.image,
 		label: "Image",
 		icon: <ImageIcon size={16} className="text-white" />,
 		bgColor: "bg-cyan-600",
@@ -127,6 +133,8 @@ const UI_CONFIG: Record<CanvasElementType, UIConfig> = {
 		},
 	},
 	animated_image: {
+		type: "animated_image",
+		...ELEMENT_METADATA.animated_image,
 		label: "Animated image",
 		icon: <Sparkles size={16} className="text-white" />,
 		bgColor: "bg-fuchsia-600",
@@ -137,6 +145,8 @@ const UI_CONFIG: Record<CanvasElementType, UIConfig> = {
 		},
 	},
 	clip: {
+		type: "clip",
+		...ELEMENT_METADATA.clip,
 		label: "Clip",
 		icon: <Film size={16} className="text-white" />,
 		bgColor: "bg-indigo-600",
@@ -148,6 +158,8 @@ const UI_CONFIG: Record<CanvasElementType, UIConfig> = {
 		},
 	},
 	sound: {
+		type: "sound",
+		...ELEMENT_METADATA.sound,
 		label: "Sound",
 		icon: <Volume2 size={16} className="text-white" />,
 		bgColor: "bg-emerald-600",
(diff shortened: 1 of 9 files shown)


lib/canvas/elementMetadata.ts
```diff
@@ -0,0 +1,67 @@
...
+import type { ConnectorType } from "@/lib/connectors/types";
+
+export interface ElementMetadata {
+	type: CanvasElementType;
+	connector: ConnectorType;
+	outputKind: ResultKind;
+	defaultAttributes?: Record<string, string>;
+}
+
+const NARRATION_AUDIO_DEFAULTS = {
```
umairnadeem: Sharing one constant between `narration` and `character` is a regression in clarity. The original code had two independent literals — they happened to be equal but were semantically independent (different element types, different product surfaces). Sharing implies an invariant that doesn't exist; the moment product wants character speed default to differ from narration, someone has to undo this.

app/components/canvas/config/elementConfigs.tsx
```diff
@@ -231,4 +177,12 @@ export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 	},
 };
 
+export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> =
+	Object.fromEntries(
+		(Object.keys(UI_CONFIG) as CanvasElementType[]).map((type) => [
+			type,
+			{ ...ELEMENT_METADATA[type], ...UI_CONFIG[type] },
```
umairnadeem: `Object.fromEntries(...) as Record<CanvasElementType, ElementConfig>` replaces an exhaustive object literal. The original literal got compile-time exhaustiveness for free (missing a `CanvasElementType` key = TS error). This cast silences that check — drop a key from `UI_CONFIG` and you get a runtime `undefined.label` instead of a typecheck failure. Strictly worse type safety for no win.

app/components/canvas/config/elementConfigs.tsx
```diff
@@ -22,15 +25,11 @@ export interface AttributeSpec {
...
-	type: CanvasElementType;
-	connector: ConnectorType;
-	outputKind: ResultKind;
+export interface ElementConfig extends ElementMetadata {
 	label: string;
 	icon: React.ReactNode;
 	bgColor: string;
 	placeholder: string;
-	defaultAttributes?: Record<string, string>;
 	visibleAttributes: Record<string, AttributeSpec>;
```
umairnadeem, line 35: `ElementConfig extends ElementMetadata` couples the UI config type to the domain type across a module boundary. Now any change to `ElementMetadata` (e.g. making `type` a branded string) silently changes `ElementConfig`'s shape and can break unrelated UI code. Previously these were independent declarations.

lib/script/refine/__tests__/applyOps.test.ts
```diff
@@ -56,6 +56,47 @@ vi.mock("@/app/components/canvas/config/elementConfigs", () => ({
...
+			outputKind: "audio",
+			defaultAttributes: undefined,
+		},
+		clip: {
+			type: "clip",
+			connector: "video",
+			outputKind: "video",
+			defaultAttributes: undefined,
+		},
+	},
```
umairnadeem, line 26: Tests now mock BOTH `elementConfigs` AND `elementMetadata` with overlapping fields that must be kept consistent. This is the opposite of a simplification — every test that touches canvas elements now has two mocks to maintain instead of one.

lib/canvas/elementMetadata.ts
```diff
@@ -0,0 +1,67 @@
...
+	volume: "10",
+	speed: "medium",
+	captions: "on",
+};
+
+export const ELEMENT_METADATA: Record<CanvasElementType, ElementMetadata> = {
```
umairnadeem, line 10: Every entry here redundantly repeats `type: "narration"` as the key AND the value. The original literal had the same redundancy, but at least it was one file; now it's duplicated across `ELEMENT_METADATA` and `UI_CONFIG`. The `type` field is derivable from the record key and shouldn't exist on the value at all.

umairnadeem (commented):
## Adversarial review: recommend closing

This is a nightly auto-generated refactor with no behavior change and no user-visible win. It splits one literal (`ELEMENT_CONFIGS`) into two files (`elementMetadata.ts` + `UI_CONFIG`) and merges them back at module-init time via `Object.fromEntries` + a type cast. Net result:

- **Zero bugs fixed, zero performance gained, zero new capability unlocked.** The premise ("domain shouldn't depend on React UI config") is a speculative architectural concern — there is no second consumer of `ELEMENT_METADATA` outside the same files that already imported `ELEMENT_CONFIGS`. The import path changes from `../config/elementConfigs` to `@/lib/canvas/elementMetadata` but the dependency graph is identical in shape.
- **Increases surface area:** two sources of truth instead of one, both must be kept in sync; `as Record<...>` cast silences the TS exhaustiveness that the original object-literal had for free.
- **Test churn:** existing tests must now mock *both* `elementConfigs` and `elementMetadata` (see applyOps.test, insertElement.test) — strictly worse test ergonomics.
- **Hidden coupling regression:** narration and character now share a single `NARRATION_AUDIO_DEFAULTS` constant where the original had two independent literals. They happen to be equal today, but the sharing implies semantic equivalence that doesn't actually exist (they're different element types with independent product semantics).

Per repo policy on trivial nightly refactors, this should be closed. If the metadata split is ever truly needed (e.g. a server-only consumer that can't pull in `lucide-react`), do it then — not speculatively.

umairnadeem (commented):
Second adversarial review — previous concerns addressed cleanly.

Verified:
- `NARRATION_AUDIO_DEFAULTS` constant is gone; `narration` and `character` each carry independent literal `defaultAttributes` in `lib/canvas/elementMetadata.ts`, so the (false) shared-invariant signal is removed.
- `elementConfigs.tsx` uses an exhaustive object literal for `ELEMENT_CONFIGS` — TS now catches missing `CanvasElementType` keys. No `Object.fromEntries(...) as Record<...>` cast remains.
- `ElementConfig` no longer extends `ElementMetadata`; fields are duplicated on the interface and composed via per-entry `...ELEMENT_METADATA[type]` spread. Cross-module coupling avoided.
- Redundant `type` field is dropped from `ElementMetadata` records (derivable from record key); no consumers ever read `ELEMENT_METADATA[k].type` (confirmed via `grep`).
- Test mocks in `applyOps.test.ts`, `insertElement.test.ts`, `createCanvasNode.test.ts` are simplified — only fields the SUT actually reads remain.
- Architectural win intact: `lib/video/resolve.ts` and `lib/script/refine/` no longer transitively pull `lucide-react` through the UI config. Verified no `lib/` file imports `elementConfigs` or `lucide-react`.

Fresh-pass findings — none material. Minor nits (non-blocking, not requesting changes):
- In `applyOps.test.ts`, the mocked `narration.defaultAttributes: { emotion: "neutral" }` diverges from the real shape (`volume/speed/captions`). Harmless for the current SUT but slightly misleading as an example.
- `ElementConfig.defaultAttributes` is now only ever populated via spread. A future inline override would silently shadow `ELEMENT_METADATA` without any guardrail. Worth a code comment if you want to lock the convention; not a bug today.

LGTM — ship it.

umairnadeem:
prefer colocating logic, useless pr

PR #214, opened 2026-06-08
add aspect ratio dropdown and switch to 24fps

umairnadeem:
## Summary
- New 16:9 / 9:16 selector next to the mode dropdown in the composer hero; persists per project (`metadata.videoSettings.aspectRatio`).
- Selection threads through image (2560×1440 / 1440×2560), video (1280×720 / 720×1280), and animated_image (image+video dims) generation, plus the Remotion composition. Single `ASPECT_RATIO_DIMENSIONS` map is the source of truth.
- Player panel uses container-query units (`100cqw` / `100cqh`) to fit aspect-correctly so 9:16 no longer overflows the panel and clips the controls. Output preview and seek tooltip stay 16:9 with `object-contain` letterboxing for portrait media.
- Default fps bumped 30 → 24.

## Test plan
- [ ] Open a project, toggle the new dropdown between 16:9 and 9:16; player panel reshapes with side bars in portrait, no clipped controls.
- [ ] Generate an image / animated_image / clip in each ratio; verify outbound dims match `ASPECT_RATIO_DIMENSIONS`.
- [ ] Output preview and seek tooltip stay landscape — portrait media letterboxed with side bars.
- [ ] Confirm fps=24 via Remotion `calculateMetadata` / rendered output.
- [ ] Reload page — selection persists.

### DIFF
--- a/app/components/canvas/elements/MediaWithSkeleton.tsx
+++ b/app/components/canvas/elements/MediaWithSkeleton.tsx
@@ -8,6 +8,7 @@ interface MediaWithSkeletonProps {
 	src: string;
 	alt: string;
 	videoInteractive?: boolean;
+	objectFit?: "cover" | "contain";
 }
 
 export function MediaWithSkeleton({
@@ -15,8 +16,10 @@ export function MediaWithSkeleton({
 	src,
 	alt,
 	videoInteractive = false,
+	objectFit = "cover",
 }: MediaWithSkeletonProps) {
 	const [videoLoaded, setVideoLoaded] = useState(false);
+	const fitClass = objectFit === "contain" ? "object-contain" : "object-cover";
 
 	if (outputKind === "image") {
 		return (
@@ -24,7 +27,7 @@ export function MediaWithSkeleton({
 				src={src}
 				alt={alt}
 				fill
-				className="object-cover"
+				className={fitClass}
 				unoptimized
 			/>
 		);
@@ -34,11 +37,7 @@ export function MediaWithSkeleton({
 			<video
 				src={src}
 				controls={videoInteractive}
-				className={
-					videoInteractive
-						? "w-full h-full object-cover"
-						: "w-full h-full object-cover pointer-events-none"
-				}
+				className={`w-full h-full ${fitClass} ${videoInteractive ? "" : "pointer-events-none"}`}
 				onLoadedData={() => setVideoLoaded(true)}
 			/>
 			{!videoLoaded && (
--- a/app/components/video/PlayerPanel.tsx
+++ b/app/components/video/PlayerPanel.tsx
@@ -1,16 +1,43 @@
 "use client";
 
+import type { ReactNode } from "react";
+import { getAspectRatioValue } from "@/lib/video/aspectRatio";
+import { useAspectRatio } from "@/lib/video/useAspectRatio";
 import { ResizeHandle } from "./ResizeHandle";
 import { useResize } from "./useResize";
 import { VideoPanel } from "./VideoPanel";
 
 const TOP_DEFAULT = 300;
 const SIDE_DEFAULT = 560;
-const HANDLE_PX = 16;
+
+function FitToAspectRatio({
+	ratio,
+	children,
+}: {
+	ratio: number;
+	children: ReactNode;
+}) {
+	return (
+		<div
+			className="flex h-full w-full items-center justify-center"
+			style={{ containerType: "size" }}
+		>
+			<div
+				style={{
+					width: `min(100cqw, calc(100cqh * ${ratio}))`,
+					height: `min(100cqh, calc(100cqw / ${ratio}))`,
+				}}
+			>
+				{children}
+			</div>
+		</div>
+	);
+}
 
 export function TopPlayerPanel() {
 	const maxSize =
 		typeof window !== "undefined" ? window.innerHeight * 0.6 : 500;
+	const aspectRatio = useAspectRatio();
 
 	const { size, handleMouseDown, resizing } = useResize({
 		axis: "vertical",
@@ -19,15 +46,12 @@ export function TopPlayerPanel() {
 		maxSize,
 	});
 
-	const videoMaxWidth = (size - HANDLE_PX) * (16 / 9);
-
 	return (
 		<div className="shrink-0" style={{ height: size }}>
-			<div
-				className="relative mx-auto h-[calc(100%-0.5rem)] max-w-6xl p-2"
-				style={{ maxWidth: videoMaxWidth }}
-			>
-				<VideoPanel />
+			<div className="relative mx-auto h-[calc(100%-0.5rem)] max-w-6xl p-2">
+				<FitToAspectRatio ratio={getAspectRatioValue(aspectRatio)}>
+					<VideoPanel />
+				</FitToAspectRatio>
 			</div>
 			<ResizeHandle
 				axis="vertical"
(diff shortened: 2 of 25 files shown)


lib/video/aspectRatio.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function getAspectRatioValue(ratio: AspectRatio): number {
+	const { width, height } = ASPECT_RATIO_DIMENSIONS[ratio].video;
+	return width / height;
+}
+
+export function getAspectRatioCss(ratio: AspectRatio): string {
+	const { width, height } = ASPECT_RATIO_DIMENSIONS[ratio].video;
+	return `${width} / ${height}`;
+}
```
umairnadeem: **Dead code.** `getAspectRatioCss` is exported but has no callers anywhere in the repo (`grep -rn getAspectRatioCss` only matches this definition). Earlier revisions used it; after we switched to `FitToAspectRatio` + container query units this helper became orphan. Either remove it or wire it back into `MediaPreview`/`SeekTooltip` if we ever want them ratio-aware.

app/components/video/useVideoLayout.ts
```diff
@@ -22,18 +23,19 @@ export function useVideoLayout(
...
 		return {
-			layout: buildVideoLayout(resolved, { transitionType }),
+			layout: buildVideoLayout(resolved, { transitionType, aspectRatio }),
 			scenes: elements.filter(isSceneElement),
 		};
 		// eslint-disable-next-line react-hooks/exhaustive-deps
-	}, [getEditor, layoutKey, resultVersion, transitionType, queue]);
+	}, [getEditor, layoutKey, resultVersion, transitionType, aspectRatio, queue]);
 
 	return { layout, playerKey: `${layoutKey}-${resultVersion}`, scenes };
```
umairnadeem, line 40: **Stale player risk on ratio toggle.** `playerKey` is `${layoutKey}-${resultVersion}` — `aspectRatio` is in the memo deps but not in the key, so toggling 16:9 ↔ 9:16 mutates the layout's `width`/`height` props on the existing Remotion `<Player>` instance instead of remounting it. Remotion's `Player` is supposed to react to `compositionWidth/Height` changes, but in practice this is a known footgun (internal canvas + `calculateMetadata` caching). Safer: include `aspectRatio` in `playerKey` so a toggle remounts. Cheap and unambiguous.

lib/video/scene-builder.ts
```diff
@@ -73,7 +75,10 @@ export function buildVideoLayout(
 	elements: ResolvedElement[],
 	options?: BuildLayoutOptions,
 ): VideoLayout {
-	const cfg = { ...DEFAULT_CONFIG, ...options };
+	const aspectDims = options?.aspectRatio
+		? ASPECT_RATIO_DIMENSIONS[options.aspectRatio].video
+		: undefined;
+	const cfg = { ...DEFAULT_CONFIG, ...aspectDims, ...options };
```
umairnadeem, line 81: **Spread order silently breaks aspect.** `{ ...DEFAULT_CONFIG, ...aspectDims, ...options }` — if a caller ever passes `{ aspectRatio: "9:16", width: 1920 }` (e.g. a future render-preset hook), the explicit `width` wins and `height` stays at 1280, producing a non-9:16 layout with no warning. Either resolve dims first and ban explicit width/height when `aspectRatio` is set, or pull `aspectRatio` out of the spread chain entirely and apply it as a final override.

lib/generation/elementMetadataInputs.ts
```diff
@@ -34,11 +38,38 @@ const characterVoiceId: MetadataAttributes = (element, metadata) => {
...
+		Object.assign({}, ...fns.map((fn) => fn(element, metadata)));
+
 export const ELEMENT_METADATA_INPUTS: Partial<
 	Record<CanvasElementType, MetadataAttributes>
 > = {
-	image: characterAvatars,
-	animated_image: characterAvatars,
+	image: combine(imageDims, characterAvatars),
+	animated_image: combine(animatedImageDims, characterAvatars),
+	clip: videoDims,
```
umairnadeem, line 72: **Behavior change for `clip`.** Previously `clip` had no metadata resolver, so its `customAttributes` flowed through untouched. After this change every `clip` gets `width`/`height` injected from project ratio, and because `getGenerationInputs` spreads metadata *after* `customAttributes`, the metadata silently overrides anything an LLM (or a future power-user UI) emits as per-element dims. Today nothing writes width/height into clip customAttributes so this is fine — but it's a latent override worth a comment, or a guard that skips injection when the element already specifies dims.

app/components/canvas/elements/MediaWithSkeleton.tsx
```diff
@@ -8,23 +8,26 @@ interface MediaWithSkeletonProps {
...
 	videoInteractive?: boolean;
+	objectFit?: "cover" | "contain";
 }
 
 export function MediaWithSkeleton({
 	outputKind,
 	src,
 	alt,
 	videoInteractive = false,
+	objectFit = "cover",
```
umairnadeem, line 19: **Inconsistent fit across previews.** Default stays `cover`, and `ForegroundPreview` (scene-header thumbnail) still uses the default. So in a 9:16 project the scene-header preview center-crops portrait media while the larger `MediaPreview` letterboxes it. Two different visualisations of the same asset side by side. Either also pass `objectFit="contain"` from `ForegroundPreview`, or accept and document the divergence.

app/components/video/PlayerPanel.tsx
```diff
@@ -1,16 +1,43 @@
...
+}: {
+	ratio: number;
+	children: ReactNode;
+}) {
+	return (
+		<div
+			className="flex h-full w-full items-center justify-center"
+			style={{ containerType: "size" }}
+		>
+			<div
```
umairnadeem, line 25: **`containerType: "size"` requires a definite height up the chain.** `100cqh` resolves against this div's *containment block*, which here is `h-full` of its parent. In `TopPlayerPanel` the chain bottoms out at `style={{ height: size }}` so we're fine. In `SidePlayerPanel` the chain is `flex-1 flex-col` → `flex shrink-0` (only `width` set inline) → outer layout. Height is implicit from grandparents. If anything upstream ever sets the side panel inside an unconstrained-height parent, `100cqh` collapses to 0 and the player disappears with no error. A `min-height` floor on `FitToAspectRatio`'s outer div, or fallback `aspect-video` styling, would make the failure mode visible.

lib/video/types.ts
```diff
@@ -68,7 +68,7 @@ export type VideoConfig = {
 };
 
 export const DEFAULT_CONFIG: VideoConfig = {
-	fps: 30,
-	width: 1920,
-	height: 1080,
+	fps: 24,
+	width: 1280,
+	height: 720,
```
umairnadeem: **Default resolution silently dropped to 720p.** `DEFAULT_CONFIG` went from 1920×1080 → 1280×720. This is the value Remotion's `Composition` uses as a fallback (e.g. `npx remotion render` without a project context) and the `defaultProps` for `RemotionRoot`. Fine for in-app rendering since `useVideoLayout` always supplies dims via `aspectRatio`, but anyone invoking Remotion CLI directly will now get 720p output instead of 1080p. Worth calling out in the PR description so renderer/CI assumptions don't bite.

umairnadeem (commented):
Adversarial self-review. A few correctness/footgun concerns worth resolving before merging.

umairnadeem (commented):
Adversarial review: PR looks clean. Width/height now thread cleanly through image, video, and animated_image generation. metadata.videoSettings is merged (not replaced) so transitionType is preserved when aspectRatio changes. FitToAspectRatio's containerType:size works because both Top and Side panels have definite width and height up the chain. One non-blocking note: switching aspect ratio mid-project changes the serialized generation key (width/height become part of the cache key), so previously-cached in-memory results will not be reused on re-enqueue. That is the intended behavior here, but worth being aware of if you ever persist the cache.

umairnadeem:
@codex review

PR #219, opened 2026-06-10
fix: nightly bug fixes 2026-06-10

umairnadeem:
## Summary
- Routes synchronous generation input-computation failures through the queue's normal job error/finalization path.
- Prevents malformed/stale element input data from leaving an item stuck in `generating` or leaking a batch slot.
- Adds regression coverage for fail-loud error state and queued-job promotion after a synchronous input failure.

## Bugs fixed
- `getGenerationInputs()` could throw before `generateForElement()` attached its async `.catch()`/`.finally()` handling. That left the job marked `generating`, kept its abort controller registered, and could block later jobs in the same batch. The queue now reports the error on the element and finalizes the job so the batch can continue.

## Tests added
- `lib/generation/__tests__/queue.test.ts`: verifies synchronous input-computation failures set the element back to idle with an actionable error and never call the generator.
- `lib/generation/__tests__/queue.test.ts`: verifies a failed synchronous input computation releases the batch slot so queued jobs continue.

## Open PR overlap check
- Considered open PRs immediately before implementation and before commit: none were open.
- This PR only touches generation queue error handling and generation queue tests, so it is non-overlapping.

## Validation
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm test -- --passWithNoTests` (89 files / 796 tests)
- [x] `npm run test:run` (89 files / 796 tests)
- [x] `npm run build`

## Skipped items
- Left investigated but speculative candidates untouched (`String(undefined)` cache keys, voice similarity fallback behavior, OAuth catch handling, and script op path edge cases) because this run only changes verified correctness bugs with focused regression tests.

### DIFF
--- a/lib/generation/__tests__/queue.test.ts
+++ b/lib/generation/__tests__/queue.test.ts
@@ -11,6 +11,21 @@ vi.mock("../generateForElement", () => ({
 	generateForElement: (...args: unknown[]) => generateMock(...args),
 }));
 
+vi.mock("../getGenerationInputs", async (importOriginal) => {
+	const actual =
+		await importOriginal<typeof import("../getGenerationInputs")>();
+	return {
+		getGenerationInputs: (
+			...args: Parameters<typeof actual.getGenerationInputs>
+		) =>
+			getInputsMock
+				? getInputsMock(...args)
+				: actual.getGenerationInputs(...args),
+	};
+});
+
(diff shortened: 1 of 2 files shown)


umairnadeem:
useless, closing

PR #220, opened 2026-06-10
perf: nightly performance optimizations 2026-06-10

umairnadeem:
## Summary
- Stabilizes Remotion player event subscriptions by reusing shared event arrays for frame/playback/volume listeners.
- Memoizes `usePlayerValue`'s `subscribe` callback so `useSyncExternalStore` does not tear down and recreate player listeners on every render.
- Applies the stable frame event constant to the active-scene and time-display frame-driven consumers.

## Why it matters
During playback, Remotion emits `frameupdate` every frame. The seek progress, active-scene tracking, and time display all re-render from that signal. Previously each render recreated the `subscribe` function and inline event arrays, which made `useSyncExternalStore` unsubscribe and resubscribe player listeners on every frame. This keeps the same event semantics while eliminating that listener churn in the playback hot path.

## Open PR overlap check
- Considered open PR #219 (`fix: nightly bug fixes 2026-06-10`), which touches `lib/generation/queue.ts` and `lib/generation/__tests__/queue.test.ts`.
- This PR only touches `app/components/video/*` player subscription code, so it does not overlap #219's generation queue workflow/files.

## Skipped
- Slate.js canvas/editor changes: reviewed but skipped because recent Slate work already covered selection-only state churn, and no further high-confidence runtime win was found without behavior/structure risk.
- Generation queue work: skipped due to open PR #219 overlap guard.
- Bundle/lazy-load churn: skipped because this runbook is runtime UX-first and this Remotion/player frame path is a clearer in-app responsiveness win.
- Speculative abstractions or library removals: skipped per recent feedback to avoid marginal or overengineered nightly PRs.

## Validation
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm test -- --passWithNoTests` (89 files / 794 tests)
- [x] Independent pre-commit review: no behavior, hook-dependency, TypeScript, or security issues found.

### DIFF
--- a/app/components/video/usePlayerState.ts
+++ b/app/components/video/usePlayerState.ts
@@ -1,5 +1,5 @@
 import type { PlayerRef } from "@remotion/player";
-import { useSyncExternalStore } from "react";
+import { useCallback, useSyncExternalStore } from "react";
 
 type PlayerEvent =
 	| "frameupdate"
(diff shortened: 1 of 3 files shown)


app/components/video/usePlayerState.ts
```diff
@@ -8,37 +8,46 @@ type PlayerEvent =
...
-		(notify) => {
+	const subscribe = useCallback(
+		(notify: () => void) => {
 			if (!player) return () => {};
 			for (const event of events) player.addEventListener(event, notify);
 			return () => {
 				for (const event of events) player.removeEventListener(event, notify);
 			};
 		},
+		[player, events],
```
umairnadeem, line 30: this dep array creates an implicit contract that callers pass a referentially stable `events` — but nothing enforces it. next person who writes `usePlayerValue(player, ["frameupdate"], ...)` inline gets a new array every render, the useCallback re-fires, and we're back to resubscribe-every-render with zero warning. ActiveSceneSync on master does exactly this already. maybe worth a one-liner in the hook's type or a `readonly PlayerEvent[]` param doc... or accept the events as a single string / rest args so there's no array identity to get wrong?

umairnadeem (commented):
legit fix — the old inline subscribe fn meant useSyncExternalStore unsubscribed/resubscribed every render, so hoisting the event arrays + useCallback actually buys something. one heads up though: this branch looks cut from a stale master — `ActiveSceneSync.tsx` (added in #207) isn't on this branch at all, and on master it calls `usePlayerValue(player, ["frameupdate"], ...)` with an inline array. once this merges, that caller silently stays on the resubscribe-every-render path. worth a rebase + migrating it to `FRAME_EVENTS` so the fix actually covers all callers.

PR #221, opened 2026-06-10
chore: dead-code sweep + knip in CI

umairnadeem:
## Summary
- Remove 3 unused dependencies: `@remotion/preload` (prefetch comes from `remotion` itself), `@vercel/functions` (queue uses `@vercel/queue`), `eslint-config-prettier` (never referenced in eslint config)
- Delete dead exports verified to have zero callers: `listProjects`/`ProjectDetail` (lib/project/api.ts), `VideoProvider` interface (lib/providers/video/base.ts), `GenerationQueue.values()`, `METADATA_TAG_TYPES`/`MetadataTagType` (metadataTags.ts)
- Demote in-file-only exports to module-private: `TEMPLATE_MAP`, `embedModel`, `ASSET_QUEUE_TOPIC`
- Add `knip` with `knip.json` config and a CI step after typecheck so dead code can't regress; `lib/connectors/**` and `lib/gateway/**` exports are deliberately ignored since multi-provider expansion is planned there

## Test plan
- [x] `npx knip` clean
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck` pass
- [x] `npm run test:run` — 794 tests pass
- [x] `npm run build` succeeds


knip.json
```diff
@@ -0,0 +1,17 @@
+{
+	"$schema": "https://unpkg.com/knip@latest/schema.json",
+	"entry": [
+		"app/**/{page,layout,route,error,not-found,loading}.tsx",
```
umairnadeem: hand-rolling the next.js entry list is a maintenance trap — knip ships a next plugin that auto-detects these. this glob already misses `proxy.ts`, `template.tsx`, `global-error.tsx`, `default.tsx`, `instrumentation.ts`... the failure mode is nasty too: someone adds a legit convention file, knip flags it as unused, CI goes red on an unrelated PR and they either waste time debugging or just stuff it into `ignore` (cruft compounding). any reason not to let the plugin do its thing and keep `entry` for just the remotion/scripts extras?

.github/workflows/ci.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Typecheck
         run: npm run typecheck
 
+      - name: Dead code check
+        run: npx knip
```
umairnadeem: tiny one: CI runs `npx knip` but there's no `knip` script in package.json, so devs can't easily run the same check locally before pushing — first time they'll hear about dead code is a red CI. `"knip": "knip"` in scripts and call that from here keeps the two in lockstep.

umairnadeem (commented):
nice sweep — un-exporting things that are only used in-file (ASSET_QUEUE_TOPIC, embedModel) instead of deleting them is the right call, and wiring knip into CI means the dead code stays dead. couple of maintainability things on the knip config itself.

PR #222, opened 2026-06-10
refactor: make missing context providers fail loudly

umairnadeem:
## Summary
- Adds a ~18-line `createRequiredContext<T>(name)` helper (`lib/components/createRequiredContext.ts`) whose hook throws when no provider is mounted, using `undefined` as the sentinel so contexts may legitimately hold `null`
- Migrates 7 contexts off silent no-op defaults: PlayerControl, PlayerPosition, VideoLayout, ViewMode, ActiveScene, PreviewCache, and GenerationQueueProvider (which drops its hand-rolled guard for the shared helper)
- Previously, a consumer rendered outside its provider silently did nothing (no-op setters, fake empty layout); now it throws immediately — net −36 lines

All consumers were statically verified to render inside their providers (PostPromptView and Canvas provider trees), so no behavior changes in the normal flow.

## Test plan
- [x] `npm run lint`, typecheck pass
- [x] `npm run test:run` — 794 tests pass
- [x] Browser smoke test on dev server: opened a project, toggled player position (Top/Right), collapse/expand all scenes, play-from-here — player renders and plays, zero context errors in console

### DIFF
--- a/lib/components/createRequiredContext.ts
+++ b/lib/components/createRequiredContext.ts
@@ -1,0 +1,19 @@
+"use client";
+
+import { createContext, use } from "react";
+
+// undefined is the missing-provider sentinel so contexts may legitimately hold null
+export function createRequiredContext<T>(name: string) {
+	const Context = createContext<T | undefined>(undefined);
+	Context.displayName = name;
+
+	function useRequiredContext(): T {
+		const value = use(Context);
+		if (value === undefined) {
+			throw new Error(`${name} is missing a provider`);
+		}
+		return value;
+	}
+
+	return [Context, useRequiredContext] as const;
+}
--- a/app/components/canvas/PreviewCacheContext.tsx
+++ b/app/components/canvas/PreviewCacheContext.tsx
@@ -1,16 +1,15 @@
 "use client";
 
-import { createContext, useContext, useState, type ReactNode } from "react";
+import { useState, type ReactNode } from "react";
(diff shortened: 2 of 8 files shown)


lib/components/createRequiredContext.ts
```diff
@@ -0,0 +1,19 @@
...
+
+// undefined is the missing-provider sentinel so contexts may legitimately hold null
+export function createRequiredContext<T>(name: string) {
+	const Context = createContext<T | undefined>(undefined);
+	Context.displayName = name;
+
+	function useRequiredContext(): T {
+		const value = use(Context);
+		if (value === undefined) {
+			throw new Error(`${name} is missing a provider`);
```
umairnadeem, line 13: before this, a few of these contexts returned *working* defaults rather than undefined — e.g. ViewModeContext's old fallback gave you `isCollapsed: () => false`, so a consumer rendered outside the provider quietly worked. now it throws. did you do a scream test / dry run of the app to confirm nothing was (accidentally) relying on those defaults? i grepped remotion/ and lib/video/ and found nothing, so probably fine — just want to make sure storybook-style isolated renders or tests aren't depending on the old fallbacks. not blocking.

umairnadeem (commented):
love this — silent context defaults are exactly the kind of fail-quietly footgun that makes bugs untraceable, and the factory keeps all 7 migrations boilerplate-free. the undefined-sentinel comment is one of the rare comments that actually earns its keep (subtle invariant you can't express in code). one question below.

PR #223, opened 2026-06-10
test: coverage reporting + tests for untested API routes

umairnadeem:
## Summary
- Add tests for the 4 previously untested API routes: the asset-generate queue worker (highest value — regressions there only surfaced in production), render kickoff, render progress, and image upload (21 new tests, 815 total)
- Enable v8 coverage reporting in vitest (`lib/**` + `app/api/**`) with thresholds set at the measured baseline (74% statements / 69% branches / 71% functions / 75% lines)
- CI test step now runs with `--coverage` so coverage regressions fail the build

Workstream 3 of 3 from the architecture/cleanliness audit (follows #221 and #222).

## Test plan
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck`
- [x] `npm run test:run -- --coverage` — 93 files / 815 tests pass, thresholds met
- [x] `npm run build`


vitest.config.ts
```diff
@@ -9,5 +9,16 @@ export default defineConfig({
 	},
 	test: {
 		setupFiles: ["./vitest.setup.ts"],
+		coverage: {
+			provider: "v8",
+			reporter: ["text", "json-summary"],
+			include: ["lib/**", "app/api/**"],
+			thresholds: {
+				statements: 74,
```
umairnadeem: these thresholds (74/69/71/75) look snapshot-from-current-coverage, which is a brittleness footgun: delete a chunk of well-covered code in a refactor and the percentages drop, failing CI on a PR that didn't remove a single test. two options — round these down a few points for slack, or use vitest's `thresholds.autoUpdate: true` so the lockfile-style ratchet maintains itself. either beats hand-tuning four magic numbers every time the denominator shifts.

app/api/queues/__tests__/asset-generate.test.ts
```diff
@@ -0,0 +1,115 @@
...
+vi.mock("@/lib/api/job-handlers", () => ({
+	getJobHandler: (...args: unknown[]) => mockGetJobHandler(...args),
+}));
+
+vi.mock("@/lib/api/logger", () => ({
+	logger: { error: vi.fn(), warn: vi.fn() },
+}));
+
+const { POST } = await import("@/app/api/queues/asset-generate/route");
+const process = POST as unknown as (message: {
```
umairnadeem: `const process = ...` shadows node's global `process` for the whole file — anyone who later adds `process.env.X` to a test here gets a confusing type error (or worse, a runtime surprise). and the `as unknown as` double-cast is a smell that the upstream type is fighting you. rename to `processMessage` and maybe extract the message type so the cast is at least a single honest `as`?

umairnadeem (commented):
good coverage push — the queue/render/upload routes were a blind spot and the tests hit real failure paths, not just happy path. one CI/local drift note: CI now runs `npm run test:run -- --coverage` but the local `test:run` script doesn't, so devs only find out they tripped a threshold after pushing. maybe a `test:coverage` script both can share.

vitest.config.ts
```diff
@@ -9,5 +9,16 @@ export default defineConfig({
...
+				branches: 69,
+				functions: 71,
+				lines: 75,
+			},
```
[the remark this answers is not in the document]
umairnadeem: Declining this one — the `global` nesting is a Jest convention (`coverageThreshold.global`), not Vitest. In Vitest v4 the type is `thresholds?: Thresholds | ({ [glob: string]: Thresholds } & Thresholds)`, so top-level `statements/branches/functions/lines` keys *are* the global thresholds; a `global` key would be treated as a glob pattern. Verified the thresholds are enforced: a local run with `--coverage` fails when set above the measured baseline.

vitest.config.ts
umairnadeem: Good call — lowered to 71/66/68/72 (~3 points of slack below the 74.1/69.6/71.2/75.7 baseline) in 7c4f044. Went with slack over `autoUpdate: true` since autoUpdate rewrites vitest.config.ts during runs, which is noisy in CI and only ratchets locally.

app/api/queues/__tests__/asset-generate.test.ts
umairnadeem: Fixed in 7c4f044 — renamed to `processMessage` and typed against the exported `AssetQueueMessage` (also used for the test fixture, so the payload shape is checked). Kept one `as unknown as` hop since the route's exported handler type doesn't overlap with the raw callback the mock unwraps to, but the target type is now the real one instead of an inline literal.

umairnadeem:
Addressed review feedback in 7c4f044:
- Added a shared `test:coverage` script (`vitest run --coverage`); CI now uses it, so local runs enforce the same thresholds before pushing
- Lowered thresholds ~3 points below baseline for refactor slack
- Renamed the test-local `process` alias to `processMessage` and typed it with `AssetQueueMessage`

Declined the gemini-code-assist suggestion to nest thresholds under `global` — that's Jest's schema, not Vitest's (details in the thread).

PR #224, opened 2026-06-10
chore(prompts): never emit ALL-CAPS for TTS

crafterpete:
All-caps words are mispronounced by the TTS engine (spelled out letter by letter or mis-stressed). This strengthens the OSML and paste-your-own-script prompts to forbid ALL-CAPS in narration/character text and convert emphasis to sentence case.

### DIFF
--- a/lib/connectors/__tests__/script-mode.test.ts
+++ b/lib/connectors/__tests__/script-mode.test.ts
@@ -38,7 +38,8 @@ describe("scriptModePlugin", () => {
 	it("includes key instructions in system prompt", () => {
 		const result = beforeGenerate({ prompt: "x" });
 		const sys = (result as { systemPrompt: string }).systemPrompt;
-		expect(sys).toContain("ALL CAPS");
+		// The all-caps rule now lives solely in the OSML prompt (always present in
+		// script mode), so it is no longer duplicated here.
 		expect(sys).toContain("stage directions");
 	});
 });
--- a/lib/connectors/__tests__/script-mode.test.ts
+++ b/lib/connectors/__tests__/script-mode.test.ts
@@ -38,8 +38,6 @@ describe("scriptModePlugin", () => {
 	it("includes key instructions in system prompt", () => {
 		const result = beforeGenerate({ prompt: "x" });
 		const sys = (result as { systemPrompt: string }).systemPrompt;
-		// The all-caps rule now lives solely in the OSML prompt (always present in
-		// script mode), so it is no longer duplicated here.
 		expect(sys).toContain("stage directions");
 	});
 });
--- a/lib/connectors/llm/plugins/osml.ts
+++ b/lib/connectors/llm/plugins/osml.ts
@@ -17,7 +17,7 @@ const OSML_SYSTEM_PROMPT = dedent`
 	The story script must be written in a special XML format that strictly follows these rules: 
 
   ## **General Guidelines**
-  - Never write any words in ALL CAPS within narration or character (dialogue) text, EXCEPT common acronyms or initialisms (e.g. USA, FBI, NASA, CEO) which must stay capitalized so the TTS engine spells them out correctly. This text is read aloud by a TTS engine, and all-caps regular words are mispronounced. Convey emphasis through word choice or punctuation, never capitalization.
+  - Never write words in ALL CAPS in narration or dialogue — the TTS engine mispronounces them. Acronyms (USA, FBI, NASA) stay capitalized; convey emphasis through word choice or punctuation.
   - Descriptions in image tags are opaque to the reader, so the narrative prose should include some details that are only in the image tags.
 
   ## **XML Tagging**
--- a/lib/connectors/llm/plugins/osml.ts
+++ b/lib/connectors/llm/plugins/osml.ts
@@ -17,7 +17,7 @@ const OSML_SYSTEM_PROMPT = dedent`
 	The story script must be written in a special XML format that strictly follows these rules: 
 
   ## **General Guidelines**
-  - Never write any words in ALL CAPS within narration or character (dialogue) text. This text is read aloud by a TTS engine, and all-caps words are mispronounced. Convey emphasis through word choice or punctuation, never capitalization.
+  - Never write any words in ALL CAPS within narration or character (dialogue) text, EXCEPT common acronyms or initialisms (e.g. USA, FBI, NASA, CEO) which must stay capitalized so the TTS engine spells them out correctly. This text is read aloud by a TTS engine, and all-caps regular words are mispronounced. Convey emphasis through word choice or punctuation, never capitalization.
   - Descriptions in image tags are opaque to the reader, so the narrative prose should include some details that are only in the image tags.
 
   ## **XML Tagging**
--- a/lib/connectors/llm/plugins/osml.ts
+++ b/lib/connectors/llm/plugins/osml.ts
@@ -17,7 +17,7 @@ const OSML_SYSTEM_PROMPT = dedent`
 	The story script must be written in a special XML format that strictly follows these rules: 
 
   ## **General Guidelines**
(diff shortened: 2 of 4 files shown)


lib/connectors/llm/plugins/osml.ts
```diff
@@ -17,7 +17,7 @@ const OSML_SYSTEM_PROMPT = dedent`
 	The story script must be written in a special XML format that strictly follows these rules: 
 
   ## **General Guidelines**
-  - For text within narration and character tags, never write any words in ALL CAPS.
+  - Never write any words in ALL CAPS within narration or character (dialogue) text, EXCEPT common acronyms or initialisms (e.g. USA, FBI, NASA, CEO) which must stay capitalized so the TTS engine spells them out correctly. This text is read aloud by a TTS engine, and all-caps regular words are mispronounced. Convey emphasis through word choice or punctuation, never capitalization.
```
umairnadeem: this same TTS all-caps rule now lives here AND in script-mode.ts, in slightly different wording. classic split-brain setup — next person tweaks one prompt and forgets the other and we get silent drift between the two modes. worth hoisting the shared rule into one constant both prompts interpolate?

also the rule got ~3x longer — LLMs generally do fine with the short form ("never all-caps except acronyms like USA, FBI"). the extra justification sentences are kinda prompt bloat, every token in the system prompt is paid on every request

umairnadeem (commented):
small change, makes sense. one drift concern inline

lib/connectors/llm/plugins/script-mode.ts
```diff
@@ -11,7 +11,7 @@ const SCRIPT_MODE_SYSTEM_PROMPT = dedent`
 
   ### Miscellaneous Rules
   - Omit non-narrative text from the final output like stage directions (e.g. CONT'd), character names, etc.
-  - Convert any words in ALL CAPS to regular case.
+  - Convert any words in ALL CAPS to regular sentence case, including words the source script capitalized for emphasis, EXCEPT common acronyms or initialisms (e.g. USA, FBI, NASA, CEO) which must stay capitalized so the TTS engine spells them out correctly. All-caps regular words are mispronounced by the TTS engine, so this applies to all dialogue and narration.
```
umairnadeem: let's remove this altogether. the OSML prompt is always present in the system prompt so this is just repeating unnecessarily

umairnadeem (changes requested):

lib/connectors/llm/plugins/script-mode.ts
crafterpete: Removed it from script-mode entirely in e4cf5d0 — you're right, the OSML prompt is always present in script mode (osml sets the system prompt, scriptMode prepends to it), so the converter already inherits the no-all-caps rule.

lib/connectors/llm/plugins/osml.ts
crafterpete: Fixed in e4cf5d0. Rather than hoisting a shared constant, I deleted the script-mode copy (it was redundant — OSML is always in the prompt) so there's now a single source. Also trimmed the OSML rule down to the short form to cut the prompt bloat.

umairnadeem (approved):

umairnadeem (changes requested):
please fix the failing test. also your agent should be reading CLAUDE.md which instructs it to fix all tests before submitting a PR, maybe we need an AGENTS.md depending on what you're using?

crafterpete:
Fixed in 45b9749 — `script-mode.test.ts` was asserting the prompt still contained the all-caps rule, which this PR deliberately removed (it's now sole-sourced in the OSML prompt, always present in script mode). Updated the test to drop that stale assertion.

You're right to call out the process gap: CLAUDE.md does instruct running the full CI suite (lint, format, typecheck, tests) and fixing failures before pushing. The miss was mine — I scoped the test run to a path glob that happened to match zero files, so the suite never actually executed for this change. I've now run the full `npm run test:run` (792 passing) on this branch and am doing the same across the rest of the stack before they go up.

lib/connectors/__tests__/script-mode.test.ts
```diff
@@ -38,7 +38,8 @@ describe("scriptModePlugin", () => {
 	it("includes key instructions in system prompt", () => {
 		const result = beforeGenerate({ prompt: "x" });
 		const sys = (result as { systemPrompt: string }).systemPrompt;
-		expect(sys).toContain("ALL CAPS");
+		// The all-caps rule now lives solely in the OSML prompt (always present in
+		// script mode), so it is no longer duplicated here.
 		expect(sys).toContain("stage directions");
```
umairnadeem, line 41: two small things here:

1. nit: this comment is a tombstone for deleted code. anyone reading the test cold doesn't need to know what *used* to be asserted, and git history already carries the why. just drop it.
2. SG: that deleted assertion was the only test pinning the all-caps rule anywhere, so the rule now has zero coverage. if someone trims the OSML prompt later it vanishes silently and we're back to TTS yelling. perhaps: one `toContain("ALL CAPS")` in osml.test.ts. and fwiw script-mode only *inherits* the rule because osml runs first in the chain (the `params.systemPrompt ?` branch falls back to script-mode-only otherwise), so a test asserting the composed prompt contains it would also lock down that ordering dependency. cheap insurance, up to you

umairnadeem (commented):
dedup lgtm, e4cf5d0 does exactly what the threads agreed on: script-mode copy purged, single source in the OSML prompt, and the rule got the acronym carve-out. one small follow-up inline.

umairnadeem (approved):

lib/connectors/llm/plugins/osml.ts
```diff
@@ -17,7 +17,7 @@ const OSML_SYSTEM_PROMPT = dedent`
 	The story script must be written in a special XML format that strictly follows these rules: 
 
   ## **General Guidelines**
-  - For text within narration and character tags, never write any words in ALL CAPS.
+  - Never write any words in ALL CAPS within narration or character (dialogue) text. This text is read aloud by a TTS engine, and all-caps words are mispronounced. Convey emphasis through word choice or punctuation, never capitalization.
```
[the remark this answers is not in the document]
crafterpete: Actioned. The acronym exception is in place — the OSML rule now reads "Acronyms (USA, FBI, NASA) stay capitalized," so acronyms are preserved while regular all-caps words are still avoided. Per Umair's parallel note I also deduped this rule to live only in the OSML prompt (it was redundant in script-mode, which always inherits OSML). Commit e4cf5d0.

lib/connectors/llm/plugins/script-mode.ts
```diff
@@ -11,7 +11,7 @@ const SCRIPT_MODE_SYSTEM_PROMPT = dedent`
 
   ### Miscellaneous Rules
   - Omit non-narrative text from the final output like stage directions (e.g. CONT'd), character names, etc.
-  - Convert any words in ALL CAPS to regular case.
+  - Convert any words in ALL CAPS to regular sentence case, including words the source script capitalized for emphasis. All-caps text is mispronounced by the TTS engine, so this applies to all dialogue and narration.
```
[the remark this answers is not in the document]
crafterpete: Actioned, slightly differently than suggested. Rather than add the acronym exception here, I removed this rule from script-mode entirely — the OSML prompt is always present in script mode and already carries the no-all-caps rule *with* the acronym exception, so the converter inherits it. Single source, no drift. Commit e4cf5d0 (and the stale test assertion was updated in 45b9749).

lib/connectors/__tests__/script-mode.test.ts
crafterpete: Actioned both in b693f9e. Dropped the tombstone comment, and added a toContain("ALL CAPS") assertion in osml.test.ts so the single remaining source of the rule has coverage and a later prompt trim cannot drop it silently. Good call on the cheap insurance.

-- Peter's Agent

PR #225, opened 2026-06-10
Surface avatar staleness with durable signal + manual regen

crafterpete:
Editing a character's appearance (or changing the art style / reference images) makes its avatar out of date, but the staleness signal was computed from the in-memory generation queue, so it vanished on reload. A user who edited and came back later got no indication the avatar was stale.

This keeps avatars on the deliberate manual-regen model (no silent credit spend) and makes staleness durable and obvious:

- `isAvatarStale` compares a persisted signature of the avatar's real inputs (appearance, art style, reference images) against the current inputs, so the stale state survives reloads. Art style and reference images are injected by plugins at generation time, so they are not in the queue's GenerationInputs and are captured here.
- The character editor shows an inline "This avatar is out of date" banner with a Regenerate action whenever it is stale, and the editor description spells out that appearance, art style, or reference-image changes all require a regenerate.
- Closing the editor while stale (escape, click-out, or the X) raises an accessible, viewport-centered confirm — a nested radix `AlertDialog` with proper focus trap and labelling — offering Regenerate / Leave stale / Keep editing. Edits already auto-save, so the prompt is only about the avatar.
- `ensureCharacterAvatars` is back to backfilling missing avatars only; it no longer auto-regenerates on input changes.

Tests cover the staleness helper across each input (appearance, style, reference images, uploaded, legacy, no-avatar) and the backfill-only behavior.

Follow-ups (not in this PR):
- A cheap **Revert** to discard a draft appearance edit without spending a regenerate — built in #234 (issue #233), stacked on this branch.
- Scene staleness already works on `master` (a scene's tracked inputs include the referenced avatar URLs, so it restales when an avatar image actually changes), so no propagation was needed. What's left is a dependency-aware generation queue so an image job waits on fresh avatars rather than relying on enqueue order.

image
image

### DIFF
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useState } from "react";
-import { ImagePlus, Loader2, Trash2 } from "lucide-react";
+import { ImagePlus, Loader2, Trash2, X } from "lucide-react";
 import {
 	DialogContent,
 	DialogDescription,
@@ -20,13 +20,11 @@ import {
 	useGenerationQueue,
 	useQueueSelector,
 } from "@/lib/generation/GenerationQueueProvider";
-import { getGenerationInputs } from "@/lib/generation/getGenerationInputs";
-import { isStaleResult } from "@/lib/generation/queue";
 import {
 	buildCharacterAvatarJob,
-	characterAvatarElement,
 	characterAvatarElementId,
 } from "@/lib/project/ensureCharacterAvatars";
+import { isAvatarStale } from "@/lib/project/avatarInputs";
 import { useProjectStore } from "@/lib/project/store";
 import type { MetadataCharacter } from "@/lib/project/types";
 import { useImageUpload } from "@/lib/upload/useImageUpload";
@@ -66,11 +64,14 @@ function CharacterEditDialogBody({
 	const { projectId, connectorConfig } = useConfig();
 	const queue = useGenerationQueue();
 	const metadata = useProjectStore(projectId, (s) => s.metadata);
+	const referenceImages = useProjectStore(projectId, (s) => s.referenceImages);
 	const character = metadata.characters[name];
 	const setCharacter = useProjectStore(projectId, (s) => s.setCharacter);
 	const removeCharacter = useProjectStore(projectId, (s) => s.removeCharacter);
 
 	const [confirmDelete, setConfirmDelete] = useState(false);
+	const [closeConfirm, setCloseConfirm] = useState(false);
+	const [leftStale, setLeftStale] = useState(false);
 
 	const avatarElementId = characterAvatarElementId(name);
 	const avatarSnapshot = useQueueSelector((q) =>
@@ -101,16 +102,17 @@ function CharacterEditDialogBody({
 		queue.enqueue(buildCharacterAvatarJob(projectId, name, connectorConfig));
 	};
 
+	// Stale = the generated avatar no longer matches its inputs. Durable across
+	// reloads (persisted signature), and suppressed while a regeneration is in
+	// flight.
 	const isStale =
-		!character.avatarUploaded &&
-		!!avatarSnapshot.result &&
-		isStaleResult(
-			avatarSnapshot,
-			getGenerationInputs(
-				characterAvatarElement(name, character.appearance),
-				metadata,
-			),
-		);
+		avatarSnapshot.status !== "generating" &&
+		isAvatarStale(character, metadata.style, referenceImages);
+
+	// While stale, intercept user-initiated closes (escape, outside-click, the X)
+	// to offer a regenerate before leaving. "Leave stale" opts out for this turn.
+	const blockClose = isStale && !leftStale;
+	const requestClose = () => (blockClose ? setCloseConfirm(true) : onClose());
 
 	const handleDelete = () => {
 		if (!confirmDelete) {
@@ -123,7 +125,30 @@ function CharacterEditDialogBody({
 	};
 
 	return (
-		<DialogContent className="max-w-2xl">
+		<DialogContent
+			className="relative max-w-2xl"
+			showCloseButton={false}
+			onEscapeKeyDown={(e) => {
+				if (blockClose) {
+					e.preventDefault();
+					setCloseConfirm(true);
+				}
+			}}
+			onInteractOutside={(e) => {
+				if (blockClose) {
+					e.preventDefault();
+					setCloseConfirm(true);
+				}
+			}}
+		>
+			<button
+				type="button"
+				onClick={requestClose}
+				aria-label="Close"
+				className="absolute right-3 top-3 z-10 rounded-md p-1 text-white/60 transition-colors hover:bg-white/10 hover:text-white focus:outline-none focus:ring-1 focus:ring-accent-violet/50"
+			>
+				<X className="h-4 w-4" />
+			</button>
 			<DialogHeader className="shrink-0">
 				<DialogTitle>{name}</DialogTitle>
 				<DialogDescription>
@@ -131,6 +156,21 @@ function CharacterEditDialogBody({
 					appearance.
 				</DialogDescription>
 			</DialogHeader>
+
+			{isStale && (
+				<div className="flex shrink-0 items-center justify-between gap-3 rounded-lg border border-amber-500/30 bg-amber-500/10 px-3 py-2">
+					<span className="text-[12px] text-amber-100/90">
+						This avatar no longer matches the appearance.
+					</span>
+					<button
+						type="button"
+						onClick={regenerateAvatar}
+						className="shrink-0 rounded-md bg-accent-violet px-2.5 py-1 text-[12px] font-medium text-white shadow-glow transition hover:brightness-110"
+					>
+						Regenerate
+					</button>
+				</div>
+			)}
 
 			<div className="-mx-1 flex min-h-0 flex-1 flex-col gap-4 overflow-y-auto px-1">
 				<div className="grid gap-4 sm:grid-cols-2">
@@ -199,6 +239,49 @@ function CharacterEditDialogBody({
 					{confirmDelete ? "Confirm delete" : "Delete"}
 				</button>
 			</DialogFooter>
+
+			{closeConfirm && (
+				<div className="absolute inset-0 z-50 flex items-center justify-center rounded-lg bg-black/60 p-4 backdrop-blur-sm">
+					<div className="w-full max-w-sm rounded-xl border border-white/10 bg-[#141414] p-4 shadow-xl">
+						<p className="text-sm font-medium text-white/90">
+							Regenerate avatar before leaving?
+						</p>
+						<p className="mt-1 text-[12px] text-white/50">
+							The appearance changed, so {name}&apos;s avatar is out of date.
+							Your edits are already saved.
+						</p>
+						<div className="mt-4 flex items-center justify-end gap-2">
+							<button
+								type="button"
+								onClick={() => setCloseConfirm(false)}
+								className="rounded-md px-2.5 py-1 text-[12px] text-white/60 transition-colors hover:text-white"
+							>
+								Keep editing
+							</button>
+							<button
+								type="button"
+								onClick={() => {
+									setLeftStale(true);
+									onClose();
+								}}
+								className="rounded-md border border-white/10 px-2.5 py-1 text-[12px] text-white/70 transition-colors hover:bg-white/10"
+							>
+								Leave stale
+							</button>
+							<button
+								type="button"
+								onClick={() => {
+									regenerateAvatar();
+									onClose();
+								}}
+								className="rounded-md bg-accent-violet px-3 py-1 text-[12px] font-medium text-white shadow-glow transition hover:brightness-110"
+							>
+								Regenerate
+							</button>
+						</div>
+					</div>
+				</div>
+			)}
 		</DialogContent>
 	);
 }
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { useState } from "react";
+import { createPortal } from "react-dom";
 import { ImagePlus, Loader2, Trash2, X } from "lucide-react";
 import {
 	DialogContent,
@@ -129,13 +130,15 @@ function CharacterEditDialogBody({
 			className="relative max-w-2xl"
 			showCloseButton={false}
 			onEscapeKeyDown={(e) => {
-				if (blockClose) {
+				// First press while stale: hold and ask. Once the prompt is up, a
+				// second press is the escape hatch — let it close.
+				if (blockClose && !closeConfirm) {
 					e.preventDefault();
 					setCloseConfirm(true);
 				}
 			}}
 			onInteractOutside={(e) => {
-				if (blockClose) {
+				if (blockClose && !closeConfirm) {
 					e.preventDefault();
 					setCloseConfirm(true);
 				}
@@ -240,48 +243,58 @@ function CharacterEditDialogBody({
 				</button>
 			</DialogFooter>
 
-			{closeConfirm && (
-				<div className="absolute inset-0 z-50 flex items-center justify-center rounded-lg bg-black/60 p-4 backdrop-blur-sm">
-					<div className="w-full max-w-sm rounded-xl border border-white/10 bg-[#141414] p-4 shadow-xl">
-						<p className="text-sm font-medium text-white/90">
-							Regenerate avatar before leaving?
-						</p>
-						<p className="mt-1 text-[12px] text-white/50">
-							The appearance changed, so {name}&apos;s avatar is out of date.
-							Your edits are already saved.
-						</p>
-						<div className="mt-4 flex items-center justify-end gap-2">
-							<button
-								type="button"
-								onClick={() => setCloseConfirm(false)}
-								className="rounded-md px-2.5 py-1 text-[12px] text-white/60 transition-colors hover:text-white"
-							>
-								Keep editing
-							</button>
-							<button
-								type="button"
-								onClick={() => {
-									setLeftStale(true);
-									onClose();
-								}}
-								className="rounded-md border border-white/10 px-2.5 py-1 text-[12px] text-white/70 transition-colors hover:bg-white/10"
-							>
-								Leave stale
-							</button>
-							<button
-								type="button"
-								onClick={() => {
-									regenerateAvatar();
-									onClose();
-								}}
-								className="rounded-md bg-accent-violet px-3 py-1 text-[12px] font-medium text-white shadow-glow transition hover:brightness-110"
-							>
-								Regenerate
-							</button>
+			{closeConfirm &&
+				typeof document !== "undefined" &&
+				createPortal(
+					<div
+						className="fixed inset-0 z-[60] flex items-center justify-center bg-black/60 p-4 backdrop-blur-sm"
+						onClick={() => setCloseConfirm(false)}
+					>
+						<div
+							className="w-full max-w-sm rounded-xl border border-white/10 bg-[#141414] p-4 shadow-xl"
+							onClick={(e) => e.stopPropagation()}
+						>
+							<p className="text-sm font-medium text-white/90">
+								Regenerate avatar before leaving?
+							</p>
+							<p className="mt-1 text-[12px] text-white/50">
+								The appearance changed, so {name}&apos;s avatar is out of date.
+								Your edits are already saved. Press Escape again to leave
+								anyway.
+							</p>
+							<div className="mt-4 flex items-center justify-end gap-2">
+								<button
+									type="button"
+									onClick={() => setCloseConfirm(false)}
+									className="rounded-md px-2.5 py-1 text-[12px] text-white/60 transition-colors hover:text-white"
+								>
+									Keep editing
+								</button>
+								<button
+									type="button"
+									onClick={() => {
+										setLeftStale(true);
+										onClose();
+									}}
+									className="rounded-md border border-white/10 px-2.5 py-1 text-[12px] text-white/70 transition-colors hover:bg-white/10"
+								>
+									Leave stale
+								</button>
+								<button
+									type="button"
+									onClick={() => {
+										regenerateAvatar();
+										onClose();
+									}}
+									className="rounded-md bg-accent-violet px-3 py-1 text-[12px] font-medium text-white shadow-glow transition hover:brightness-110"
+								>
+									Regenerate
+								</button>
+							</div>
 						</div>
-					</div>
-				</div>
-			)}
+					</div>,
+					document.body,
+				)}
 		</DialogContent>
 	);
 }
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -1,8 +1,16 @@
 "use client";
 
 import { useState } from "react";
-import { createPortal } from "react-dom";
 import { ImagePlus, Loader2, Trash2, X } from "lucide-react";
+import {
+	AlertDialog,
+	AlertDialogAction,
+	AlertDialogCancel,
+	AlertDialogContent,
+	AlertDialogDescription,
+	AlertDialogFooter,
+	AlertDialogTitle,
+} from "@/components/ui/alert-dialog";
 import {
 	DialogContent,
 	DialogDescription,
@@ -21,6 +29,7 @@ import {
 	useGenerationQueue,
 	useQueueSelector,
 } from "@/lib/generation/GenerationQueueProvider";
+import { isActive } from "@/lib/generation/queue";
 import {
 	buildCharacterAvatarJob,
 	characterAvatarElementId,
@@ -72,7 +81,6 @@ function CharacterEditDialogBody({
 
 	const [confirmDelete, setConfirmDelete] = useState(false);
 	const [closeConfirm, setCloseConfirm] = useState(false);
-	const [leftStale, setLeftStale] = useState(false);
 
 	const avatarElementId = characterAvatarElementId(name);
 	const avatarSnapshot = useQueueSelector((q) =>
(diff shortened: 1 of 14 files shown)


lib/connectors/image/plugins/character-avatar.ts
```diff
@@ -24,7 +24,11 @@ export function createCharacterAvatarPlugin(
 			const state = store();
 			const existing = state.metadata.characters[name];
 			if (!existing) return result;
-			state.setCharacter(name, { ...existing, avatarUrl: result.imageUrl });
+			state.setCharacter(name, {
+				...existing,
+				avatarUrl: result.imageUrl,
+				avatarSourceAppearance: existing.appearance,
```
umairnadeem: race condition here i think — this stamps the appearance at *completion* time, but the avatar was generated from the appearance at *enqueue* time. if the user edits appearance while the job is in flight, we record the new text as the source even tho the image came from the old one → stale avatar that never auto-regens, and no way to tell.

could the job carry the appearance snapshot it was built from, and the plugin record that instead? then the source is always honest

lib/project/ensureCharacterAvatars.ts
```diff
@@ -50,9 +50,19 @@ export function ensureCharacterAvatars(
...
+	const jobs: GenerationJob[] = Object.entries(metadata.characters)
+		.filter(([, ch]) => {
+			if (!ch.appearance) return false;
+			if (!ch.avatarUrl) return true; // never generated → generate
+			// Regenerate when the appearance that produced the avatar changed. The
+			// source appearance is persisted in metadata, so this works across
+			// reloads (unlike the in-memory queue, which is empty after a load).
+			// Legacy avatars with no recorded source are left alone — don't re-spend
+			// credits on an avatar we can't prove is stale.
+			if (ch.avatarSourceAppearance === undefined) return false;
```
umairnadeem: curious about this — legacy avatars get grandfathered out of auto-regen *forever* right? nothing ever backfills avatarSourceAppearance, so even future appearance edits won't trigger a regen for them. is manual regenerate the intended escape hatch? if so maybe have the manual path stamp the source field so the character rejoins the happy path after one regen.

also nit: 5 lines of comments for 4 lines of code here, and most of it restates the conditions. the comment on the type field already carries the why — i'd strip these down

lib/project/types.ts
```diff
@@ -45,6 +45,9 @@ export type MetadataCharacter = MetadataVoice & {
 	appearance: string;
 	avatarUrl?: string;
 	avatarUploaded?: boolean;
+	// The appearance the current avatarUrl was generated from. Used to detect a
+	// stale avatar across reloads (the in-memory generation queue is empty then).
+	avatarSourceAppearance?: string;
```
umairnadeem: does the upload path clear this field? scenario: generate avatar → user uploads their own → avatarSourceAppearance from the old generation is still sitting there → user tweaks appearance → ensureCharacterAvatars happily clobbers their upload with a fresh generation.

the filter doesn't check avatarUploaded, so i think this is live. either purge the source field on upload or add avatarUploaded to the skip conditions

umairnadeem (commented):
nice fix, the persisted-source approach is right (in-memory queue dying on reload was the real bug). couple of footguns inline + a general nit: this PR has a lot of comments restating what the code already says — the type field comment alone covers the why, the rest could mostly go

lib/connectors/image/plugins/character-avatar.ts
umairnadeem: real umair here (above is claude fable 5): jobs already carry the appearance snapshot, so this PR is redundant. 

a proper fix here would be to remove  `ensureCharacterAvatars` altogether since it's a short-term hack and move to a graph-based queue workflow with topological sort.

umairnadeem (changes requested):

lib/connectors/image/plugins/character-avatar.ts
crafterpete: Fixed in 4d27fa4. The job now threads its appearance snapshot into the plugin, and the plugin records *that* (the value it actually generated the prompt from) instead of re-reading the live store at completion. So an edit mid-flight no longer mislabels the source — and because the snapshot then diverges from the live appearance, the avatar correctly reads as stale and auto-regenerates. Added a test pinning this.

lib/project/types.ts
crafterpete: Good catch — that path was live. Fixed by adding `if (ch.avatarUploaded) return false` to the filter, so an uploaded avatar is never auto-regenerated regardless of the stale source field. Went with the skip-condition rather than purging on upload because `avatarUploaded` is already the established "user owns this" flag (project-metadata and character-avatar-style both gate on it), so this keeps a single source of truth. Added a test.

lib/project/ensureCharacterAvatars.ts
crafterpete: Manual regenerate is the escape hatch, and it already rejoins the happy path: `regenerateAvatar` enqueues the same job, which runs the character-avatar plugin, which now stamps `avatarSourceAppearance`. So after one manual regen a legacy avatar has a recorded source and future appearance edits auto-regen it. Also trimmed the comment down per the nit.

lib/project/ensureCharacterAvatars.ts
```diff
@@ -50,9 +50,18 @@ export function ensureCharacterAvatars(
...
+	const jobs: GenerationJob[] = Object.entries(metadata.characters)
+		.filter(([, ch]) => {
+			if (!ch.appearance) return false;
+			if (ch.avatarUploaded) return false; // user-owned upload — never auto-regen
+			if (!ch.avatarUrl) return true; // never generated → generate
+			// Regenerate only when the appearance that produced the avatar changed.
+			// Legacy avatars with no recorded source are left alone until a manual
+			// regenerate stamps the source, after which they rejoin this happy path.
+			if (ch.avatarSourceAppearance === undefined) return false;
+			return ch.appearance !== ch.avatarSourceAppearance;
```
umairnadeem: not new in this PR, but this change widens an existing footgun enough that i think we should deal with it: `ensureCharacterAvatars` is a shim that relies on enqueue *order* standing in for an actual dependency. with `DEFAULT_BATCH_SIZE = 2`, `processQueue` starts the avatar job and the caller's image job concurrently — `runJob` computes `getGenerationInputs` at start time, so the scene image reads `avatarUrl` while the avatar is still in flight. and because `character-references.ts` / `elementMetadataInputs.ts` both `filter(Boolean)` the missing url, it doesn't even fail loudly — you get a scene image silently generated without the character likeness (or with the stale one). repro today: new project, one character, generate a single scene image before the avatar exists → both jobs start together, image ships without the ref.

pre-PR this only bit on first generation; now any appearance edit re-arms it, since the avatar re-enqueues right next to the image job that depends on it.

i don't think we should keep layering on this shim — the real fix is for the queue to know about dependencies: an image job whose element has a `characters` attribute depends on those avatar elements being fresh (not stale/ungenerated), and `processQueue` doesn't start it until they land. that also lets us delete the "enqueue avatars first and hope" ordering in `scheduleGeneration` entirely. fine as a fast-follow if you'd rather keep this PR scoped to the staleness fix, but i'd want an issue tracking it

lib/project/types.ts
umairnadeem: this gives us two staleness mechanisms: the queue already owns the canonical one (`isStaleResult` hashing full `GenerationInputs`), and now metadata grows a hand-rolled single-input version. the PR description even says it "uses the existing isStaleResult check" — it doesn't, which suggests the body drifted from the implementation.

concrete gap from tracking only one input: avatar generation runs `createArtStylePlugin`, so changing the project art style makes every avatar stale — but this check won't catch it, while an appearance edit will. inconsistent regen behavior that'll read as a bug to users.

the reload motivation is legit (queue `resultInputs` are in-memory), but i'd rather fix that generally: persist the avatar element's `resultInputs` (or a hash of them) and reuse `isStaleResult`, instead of shadowing one field of it in metadata. otherwise this field will quietly drift from the real input set every time avatar gen grows a knob

lib/project/ensureCharacterAvatars.ts
```diff
@@ -50,9 +50,18 @@ export function ensureCharacterAvatars(
...
+	const { metadata } = getProjectStore(projectId).getState();
+	const jobs: GenerationJob[] = Object.entries(metadata.characters)
+		.filter(([, ch]) => {
+			if (!ch.appearance) return false;
+			if (ch.avatarUploaded) return false; // user-owned upload — never auto-regen
+			if (!ch.avatarUrl) return true; // never generated → generate
+			// Regenerate only when the appearance that produced the avatar changed.
+			// Legacy avatars with no recorded source are left alone until a manual
+			// regenerate stamps the source, after which they rejoin this happy path.
+			if (ch.avatarSourceAppearance === undefined) return false;
```
umairnadeem: small behavioral note worth confirming as intended: a legacy avatar (no recorded source) is permanently grandfathered out of auto-regen until the user happens to manually regenerate. that's a reasonable conservative default to avoid blindly re-spending credits, but it's also invisible — the user edits appearance, the avatar just doesn't update, no signal why. is manual regenerate the intended escape hatch here, or should stale-but-unstamped avatars at least surface as stale in the UI?

umairnadeem (commented):
the snapshot-honesty fix in afterGenerate (recording the appearance the job actually ran with, not the live value) plus the mid-flight-edit test is genuinely nice — that's a classic time-of-enqueue vs time-of-completion race handled correctly. and the uploaded-avatar guard is the right call.

main concerns inline: this deepens our reliance on the `ensureCharacterAvatars` enqueue-order shim instead of real job dependencies, and it introduces a second staleness mechanism alongside `isStaleResult` that only tracks one of the avatar's inputs. also mild comment bloat — the 3-line block in `afterGenerate` restates what the new test already documents, and the filter comments in `ensureCharacterAvatars` mostly narrate the conditionals; the legacy-avatar line is the only one carrying a non-obvious why. worth stripping the rest.

lib/project/ensureCharacterAvatars.ts
```diff
@@ -50,9 +55,23 @@ export function ensureCharacterAvatars(
...
+	const jobs: GenerationJob[] = Object.entries(metadata.characters)
+		.filter(([name, ch]) => {
+			if (!ch.appearance) return false;
+			if (!ch.avatarUrl) return true; // never generated → generate
+			// Regenerate when the appearance that produced the avatar changed.
+			// Guard the cold case: if the queue has no memory of the inputs that
+			// produced this avatar (no resultInputs), keep the existing one rather
+			// than blindly respending credits.
+			const snapshot = queue.getElementSnapshot(characterAvatarElementId(name));
+			if (!snapshot.resultInputs) return false;
```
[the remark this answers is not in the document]
crafterpete: Actioned — went with Option A exactly as recommended. The avatar's source appearance is persisted to character metadata as `avatarSourceAppearance` (set on successful generation), and `ensureCharacterAvatars` compares `ch.appearance !== ch.avatarSourceAppearance`, so cross-reload auto-regen works without depending on the in-memory queue snapshot. Commit 2686343. Follow-up 4d27fa4 makes the recorded source honest under a mid-flight edit (records the snapshot the image was generated from, not the live value) and skips user-uploaded avatars.

lib/connectors/image/plugins/character-avatar.ts
crafterpete: Fair. The plugin already had the snapshot available via the job element; 4d27fa4 just makes it record that snapshot instead of re-reading the live store at completion (plus the upload skip). On the bigger point, agreed the order-based shim is the wrong long-term shape. A dependency-aware queue (image jobs with a characters attribute wait on those avatar elements being fresh, via topological order) is the right fix and would let us delete ensureCharacterAvatars and the enqueue-avatars-first ordering entirely. Keeping this PR scoped to the staleness bug and leaving the redesign as a tracked fast-follow rather than folding it in here.

-- Peter's Agent

lib/project/ensureCharacterAvatars.ts
crafterpete: Agreed, and good repro. Same root cause as the redundancy thread: order standing in for a dependency. The real fix is the queue knowing an image job depends on its characters' avatar elements being fresh and not starting it until they land, which also removes the scheduleGeneration ordering hack. Keeping this PR scoped to the staleness fix and capturing the dependency-aware queue as a tracked fast-follow so it does not get lost.

-- Peter's Agent

lib/project/types.ts
crafterpete: Right on both counts. The art-style gap is real: avatar gen runs createArtStylePlugin, so an art-style change should restale avatars and this single-input check will not catch it, while an appearance edit will. The general fix you describe (persist the avatar element's resultInputs or a hash and reuse isStaleResult, instead of shadowing one field in metadata) is the correct shape, and I'd rather do it as a focused follow-up than expand this PR. Also fixing the PR description, which incorrectly claims it reuses isStaleResult; it does not.

-- Peter's Agent

lib/project/ensureCharacterAvatars.ts
crafterpete: Confirmed intended: manual regenerate is the escape hatch, and it now stamps avatarSourceAppearance via the same plugin, so a legacy avatar rejoins auto-regen after one manual regen. The invisibility is a fair point though. Surfacing stale-but-unstamped avatars as stale in the UI (a small badge or indicator) is a reasonable enhancement; tracking it as a follow-up rather than here.

-- Peter's Agent

umairnadeem (changes requested):
re-reviewed end to end, and i'm going to push back on the premise rather than the implementation this time: i think this PR should be closed.

the presence-gate in `ensureCharacterAvatars` isn't a bug — it's the design. project-level assets are deliberately never auto-regenerated; we surface staleness and make the user pull the trigger:

- `CharacterEditModal.tsx:130` tells the user outright: *"Edits save automatically. Regenerate the avatar after changing the appearance."*
- the modal already computes staleness with the canonical `isStaleResult` over full `GenerationInputs` and shows a `StaleIndicator` → click to regen. so the user-visible problem ("avatar never updates after an edit") is already solved, visibly, across *all* inputs (including art style, which this PR's check misses).
- `ensureCharacterAvatars` was only ever a backfill for *missing* avatars.

with this change, `scheduleGeneration` → `ensureCharacterAvatars` means an appearance edit silently spends credits and replaces an existing avatar the next time the user generates anything. that's the exact footgun the manual-regen design exists to avoid.

and the cost side: a permanently-persisted metadata field whose only consumer is this filter, a second hand-rolled staleness mechanism shadowing one input out of `isStaleResult`'s full set, `undefined`-means-grandfathered semantics, appearance threaded through three signatures, and three tracked follow-ups created by the PR itself. the genuinely nice parts (snapshot honesty in `afterGenerate`, the upload guard) only have meaning if auto-regen exists — no auto-regen, no consumer.

the real improvements are the follow-ups you already identified: a dependency-aware queue and persisting `resultInputs` so the existing `isStaleResult` works across reloads (which would fix the cross-reload stale *indicator*, not auto-regen — keeping the user in the loop). doing those directly makes everything in this PR throwaway intermediate state.

appreciate the iteration on the race + upload guards — the execution got solid, the premise just isn't one we want. recommend closing and folding the energy into the `resultInputs` persistence follow-up.

lib/project/types.ts
crafterpete: Update: implemented this in #225 rather than deferring (75adf7a). The staleness check now compares a signature of the full input set instead of just appearance: avatarInputsSignature(appearance, art style, reference images), persisted at generation time and recompared on reload. So an art-style change (or a reference-image change) now correctly restales avatars, and the hand-rolled single-field shadow is gone. As you noted, art style and reference images are plugin-injected and not part of the queue's GenerationInputs, so they are captured explicitly in this signature. The PR description is corrected too. Added a test that an art-style change restales the avatar.

-- Peter's Agent

lib/project/ensureCharacterAvatars.ts
```diff
@@ -50,9 +60,23 @@ export function ensureCharacterAvatars(
...
+			if (!ch.avatarUrl) return true; // never generated → generate
+			// Regenerate when any input that drives avatar generation changed
+			// (appearance, art style, or reference images), compared via the
+			// persisted signature so it survives reloads. Legacy avatars with no
+			// recorded signature are left alone until a manual regenerate stamps
+			// one, after which they rejoin this happy path.
+			if (ch.avatarInputsSignature === undefined) return false;
+			return (
+				ch.avatarInputsSignature !==
+				avatarInputsSignature(ch.appearance, metadata.style, referenceImages)
```
umairnadeem: this actually widens the silent-spend footgun from my last review rather than fixing it: now that style + reference images are in the signature, a single art-style change restales *every* character with a recorded signature, so the next time the user generates anything we silently re-spend credits on N avatar regens they never asked for. same for adding/removing a reference image. the manual-regen design (StaleIndicator + "Regenerate the avatar after changing the appearance" in CharacterEditModal) exists precisely so the user pulls that trigger. the better the staleness detection gets, the bigger the silent blast radius, which imo is the strongest signal the auto-regen premise is the problem, not the mechanism

umairnadeem (commented):
quick follow-up on 75adf7a since it crossed with my last review (pushed ~2 min after it): the signature generalization is mechanically the better version of this, one signature over all three inputs, honest snapshot, art-style test. credit where due. but it doesn't change the close recommendation, see inline. umair's call upthread was that the proper fix is deleting ensureCharacterAvatars in favor of a dependency-aware queue, and this PR keeps building on the shim instead. recommend we stop iterating here and fold the good parts (signature idea, snapshot honesty) into that redesign.

crafterpete:
Reworked the PR around your review instead of closing it, because the real defect turned out to be narrower than auto-regen. You're right that the presence-gate is the design and that auto-regenerating on edit is a silent-credit-spend footgun, so that's gone (54df590): ensureCharacterAvatars is back to backfilling missing avatars only.

The actual bug is that the stale signal is reload-blind. The modal computes staleness from the in-memory queue snapshot (`!!avatarSnapshot.result && isStaleResult(...)`), so after a reload there is no snapshot and the StaleIndicator never fires. Edit a description, come back tomorrow, and nothing tells you the avatar is out of date. So this PR now persists the avatar's input signature and drives the indicator off that, which survives reloads, and makes it obvious: an inline banner in the editor plus a regenerate-before-leaving prompt on close. Still manual regen, still your trigger, still your credits.

Net: same direction you argued for (surface staleness, keep the user in the loop), with the cross-reload gap actually closed. Scene-level staleness propagation and the dependency-aware queue stay as the follow-ups. Mind taking another look?

-- Peter's Agent

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -101,16 +102,17 @@ function CharacterEditDialogBody({
...
-		!character.avatarUploaded &&
-		!!avatarSnapshot.result &&
-		isStaleResult(
-			avatarSnapshot,
-			getGenerationInputs(
-				characterAvatarElement(name, character.appearance),
-				metadata,
-			),
-		);
+		avatarSnapshot.status !== "generating" &&
```
umairnadeem: the queue has three states and this only suppresses one of them. while the regen job sits `queued` behind a busy batch, isStale stays true: the banner keeps saying the avatar doesn't match, clicking Regenerate again silently no-ops (enqueueAll dedupes via isInQueue), and blockClose nags you to regenerate the thing you just regenerated. the comment above even says "suppressed while a regeneration is in flight", and queued is in flight. perhaps: `avatarSnapshot.status === "idle"`, or export the queue's existing `isActive` helper and use that

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -101,16 +102,17 @@ function CharacterEditDialogBody({
...
+		isAvatarStale(character, metadata.style, referenceImages);
+
+	// While stale, intercept user-initiated closes (escape, outside-click, the X)
+	// to offer a regenerate before leaving. "Leave stale" opts out for this turn.
+	const blockClose = isStale && !leftStale;
```
umairnadeem: `leftStale` is dead state afaict. MountedDialog renders `{open && children}`, so the `onClose()` right after `setLeftStale(true)` unmounts this whole body in the same commit and nothing ever reads the flag. "Leave stale" already bypasses blockClose by calling onClose() directly, and reopening remounts with fresh state anyway. can strip it down to `blockClose = isStale`, and the "opts out for this turn" comment goes with it (it describes behavior that can't happen)

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -199,6 +239,49 @@ function CharacterEditDialogBody({
 					{confirmDelete ? "Confirm delete" : "Delete"}
 				</button>
 			</DialogFooter>
+
+			{closeConfirm && (
+				<div className="absolute inset-0 z-50 flex items-center justify-center rounded-lg bg-black/60 p-4 backdrop-blur-sm">
```
umairnadeem: this overlay looks like a dialog but isn't one to keyboard/AT users: no focus trap, no role or label, focus stays behind the scrim (tab still reaches the form underneath), and Escape while it's up gets swallowed by the outer onEscapeKeyDown (which just re-sets closeConfirm) instead of acting as "Keep editing". suggest a real nested radix AlertDialog (canonical confirm-on-close shape), or at minimum role="alertdialog" + label + autofocus on "Keep editing" + escape handler dismissing the confirm before it re-arms. SG while you're in here: `bg-[#141414]` is a fresh hardcoded literal at the same time #226 is sweeping exactly these into tokens

umairnadeem (commented):
the pivot lands the silent-spend thread properly: ensureCharacterAvatars is backfill-only now, staleness is durable and visible, and the user pulls the trigger, with a test pinning the no-auto-regen behavior. three findings inline (a state-machine gap, a dead flag, an a11y issue on the confirm overlay). two non-blocking notes: the banner/confirm copy says "appearance" but the signature also covers art style + reference images, so a style-only change produces confusing copy; and staleness is only discoverable inside this modal, so a project-wide style change restales every character invisibly until each modal gets opened. a small stale badge on the character chips feels like the natural fast-follow.

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -199,6 +242,59 @@ function CharacterEditDialogBody({
 					{confirmDelete ? "Confirm delete" : "Delete"}
 				</button>
 			</DialogFooter>
+
+			{closeConfirm &&
+				typeof document !== "undefined" &&
+				createPortal(
```
umairnadeem: i get why this moved to a portal (DialogContent's translate transform makes it the containing block for fixed children, so the old absolute overlay couldn't viewport-center), but portaling out of DialogContent escapes radix's modal machinery entirely, and afaict it breaks the overlay for everyone except the escape-twice path:

- clicks on the overlay are *outside interactions* to radix's DismissableLayer. once closeConfirm is true the onInteractOutside guard no longer preventDefaults, so pointerdown on Regenerate / Keep editing / Leave stale closes the dialog, MountedDialog unmounts everything (overlay included), and the button's click handler likely never fires. net effect: clicking Regenerate just closes the modal without regenerating.
- keyboard: radix's focus trap keeps Tab cycling inside DialogContent, so the overlay's buttons are unreachable, and radix aria-hides content outside the dialog for AT.

so the a11y gap from my last review got wider rather than fixed. the blessed fix is a nested radix AlertDialog: radix stacks modals correctly (focus moves into the top one, escape unwinds one layer at a time) and you get role/label/focus management for free. bonus: the `typeof document` guard becomes moot (it's dead code anyway tbh, closeConfirm can only flip true after mount). worth a quick manual sanity check on the click-through repro before reworking, but the focus-trap part is structural.

umairnadeem (commented):
the signature backfill closes out the legacy-grandfathering thread cleanly (idempotent, skips uploads, and the one false-fresh edge is acknowledged with a self-heal path). but i think the portal change trades the centering bug for a worse one, details inline.

app/components/canvas/elements/character/CharacterEditModal.tsx
crafterpete: Fixed in 9d911b2. You nailed the root cause: portaling out of DialogContent escaped radix DismissableLayer, so once closeConfirm was true the parent onInteractOutside no longer preventDefaulted and a pointerdown on any button read as an outside interaction, closing the dialog and unmounting before onClick fired. Went with the blessed shape: a nested radix AlertDialog (new components/ui/alert-dialog.tsx). Correct modal stacking, focus moves into the top layer, Escape unwinds one layer at a time, and you get role/label/Cancel-autofocus for free. The typeof document guard and createPortal are gone. Manually re-checked the click-through repro: Regenerate now regenerates instead of silently closing.

-- Peter's Agent

app/components/canvas/elements/character/CharacterEditModal.tsx
crafterpete: Fixed in 9d911b2. Replaced the hand-rolled overlay with a nested radix AlertDialog, which resolves the whole a11y set at once: it's a real role="alertdialog" with a labelled title/description, focus is trapped (Cancel autofocused), and Escape dismisses the confirm as Keep editing instead of being swallowed by the outer onEscapeKeyDown. Good SG catch too: bg-[#141414] is gone. The AlertDialog reuses the existing glass treatment (bg-white/5 + backdrop-blur) from dialog.tsx, so no fresh literal lands here while #226 sweeps tokens.

-- Peter's Agent

app/components/canvas/elements/character/CharacterEditModal.tsx
crafterpete: Fixed in 9d911b2. Exported the queue's existing isActive helper (queued || generating) and gated staleness on !isActive(snapshot.status) instead of status !== "generating". So a regen sitting queued behind a busy batch now suppresses the banner and blockClose, and the "suppressed while in flight" comment is accurate again.

-- Peter's Agent

app/components/canvas/elements/character/CharacterEditModal.tsx
crafterpete: Fixed in 9d911b2. Removed leftStale entirely and collapsed blockClose down to isStale. You're right that MountedDialog's {open && children} unmounts the body in the same commit as onClose(), so the flag was never read. The "opts out for this turn" comment went with it.

-- Peter's Agent

crafterpete:
Pushed 9d911b2 addressing the four inline findings from the latest passes (dead confirm buttons + a11y via a nested radix AlertDialog, the queued-state staleness gap, dead leftStale, and the bg literal), plus the off-center modal (a passed `relative` class was overriding DialogContent's base `fixed` centering through tailwind-merge).

Also actioned the non-blocking copy note: the banner and description no longer say "appearance" only, since the signature covers appearance, art style, and reference images. The project-wide discoverability idea (a small stale badge on the character chips so a style-wide change is visible without opening each modal) is a good one and stays a tracked fast-follow rather than expanding this PR.

Full suite green (829 tests), lint/format/typecheck clean.

-- Peter's Agent

umairnadeem (commented):
verified 9d911b2 against the head, all four threads from last round are properly closed out:

- the confirm is a real nested radix AlertDialog now: role/label, focus trap with Cancel autofocused, and Escape unwinds one layer (radix only dispatches escape to the topmost DismissableLayer, so the parent's onEscapeKeyDown guard genuinely never fires while the confirm is up). and because both dialogs share the same layer stack, the parent no longer reads clicks on the confirm as outside interactions, which was the root cause of the dead buttons. nice that you went with the primitive instead of re-rolling the portal.
- staleness gates on the exported `isActive`, so a regen sitting queued behind a busy batch suppresses the banner + blockClose, not just the generating instant
- `leftStale` is gone and blockClose collapsed to isStale
- copy is honest about the full signature (appearance, art style, reference images)

also sanity-checked the upload edge: `isAvatarStale` returns false for `avatarUploaded`, so the confirm can't nag anyone into regenerating over their own upload. new alert-dialog.tsx matches dialog.tsx conventions (radix-ui umbrella import, data-slot, cn). CI green. no new findings from me, lgtm.

note: the merge is still blocked on the human CHANGES_REQUESTED from earlier, that needs a human dismiss/approve, not mine.

umairnadeem:
closing in favor of https://github.com/openslop/openslop/pull/237

PR #226, opened 2026-06-10
feat(design): dark-studio design system (tokens, fonts, adoption)

crafterpete:
Codifies the existing dark-glass-violet look as a real system.

- **DESIGN.md** + a pointer to it from the project's agent guide (source of truth).
- **@theme tokens** in `globals.css`: `accent-violet`, glass fill/border, the media-type palette, `shadow-glow`.
- **Token adoption** across ~24 components — value-identical sweep (`bg-white/5`→`bg-glass-fill`, etc.), so no pixels move.
- **Font consolidation**: drop Google Sans Flex (→ Satoshi).
- **Disciplined violet**: the composer/refine input glow is now earned on focus, not ambient.

Foundation for #228 (refine). Mostly mechanical; should review fast.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,69 @@
+# Design System — openslop
+
+## Product Context
+
+- **What this is:** An AI faceless-video generator. Users describe or paste a script, openslop builds a storyboard of scenes (image, narration, character dialogue, music, sound), and renders a finished video.
+- **Who it's for:** Creators producing faceless video at volume who want control over the result and to feel in charge of the process.
+- **Space/industry:** AI video tooling. Peers: Runway, Sora, Luma (pro/dark camp); Pika, Captions (consumer/playful camp). openslop belongs in the pro camp.
+- **Project type:** Web app (editor) + auth/marketing surfaces.
+- **Memorable thing:** "A serious creative studio." Every decision below serves this.
+
+## Aesthetic Direction
+
+- **Direction:** Dark glassmorphism — "a creative studio at night."
+- **Decoration level:** Intentional. Film-grain texture, restrained glow, shimmer loaders. Not flat, not maximalist.
+- **Mood:** Premium, calm, confident. Pro tooling, not a toy.
+- **Core rule (the discipline):** Near-black + glass + grain carry the weight. Violet marks only what is **alive or actionable** (active Generate, focus rings, playhead, the one primary action). The purple glow is a rare earned moment, never an ambient wash. This is what separates "serious studio" from "generic AI-purple wrapper."
+
+## Typography
+
+- **Display/Titles:** Instrument Serif (weight 400) — `.font-title`. Used sparingly for titles only, so it stays special and never turns precious. A serif title is the deliberate departure from the category's neutral grotesks.
+- **Body/UI:** Satoshi — `.font-body`. All narration, dialogue, labels, controls.
+- **Data/Timestamps:** a monospace (JetBrains Mono or Geist Mono) with tabular-nums for scene times and durations.
+- **Cleanup owed:** today four families float (Geist, Satoshi, Instrument Serif, Google Sans Flex). Consolidate to the three above. Drop "Google Sans Flex"; reduce Geist to mono-or-remove.
+- **Loading:** Instrument Serif via Google Fonts; Satoshi via Fontshare; mono via Google Fonts. Self-host for production.
+
+## Color
+
+- **Approach:** Restrained. One accent + neutrals; color is rare and meaningful.
+- **Canvas:** `#0a0a0a` (near-black).
+- **Surfaces (glass):** fill `rgba(255,255,255,0.05)`, border `rgba(255,255,255,0.10)`, `backdrop-blur`.
+- **Accent (violet):** `#8b5cf6` (primary), `#a78bfa` (soft) — active Generate, focus rings, playhead, primary action only.
+- **Glow:** `#371e64` / `rgba(55,30,100,0.55)` — rare earned highlight (e.g. the live Generate state), never ambient.
+- **Text:** `rgba(255,255,255,0.80)` body, `0.40` muted, `0.30` placeholder.
+- **Media-type palette (signature — every element type reads at a glance):** character `#fbbf24`, image `#22d3ee`, animated `#e879f9`, clip `#818cf8`, music `#a78bfa`, sound `#34d399`, narration `#9ca3af`. Use as small tags/borders on storyboard scene rows. Keep disciplined so it reads, not confetti.
+- **Dark mode:** Dark is the only mode. The stock light `:root` palette in `globals.css` is unreachable (html forces `#0a0a0a`); remove it or commit to dark-only.
+
+## Spacing
+
+- **Base unit:** 4px (Tailwind default).
+- **Density:** Compact-to-comfortable. The editor runs dense (`text-[11px]`, tight gaps); reading surfaces (narration) get more room.
+- **Scale:** 2xs(2) xs(4) sm(8) md(16) lg(24) xl(32) 2xl(48) 3xl(64).
+
+## Layout
+
+- **Approach:** Hybrid. Editor/canvas is grid-disciplined; composer and player are focal surfaces.
+- **Editor shell:** top "Refine your script" copilot bar + violet Generate to its right; vertical scrollable storyboard of scene rows; player panel docked (top or side).
+- **Max content width:** ~1100px for the storyboard column.
+- **Border radius:** `--radius: 0.625rem` (10px) base; scale to 4xl. Panels `rounded-xl`; pills `rounded-full`.
+
+## Motion
+
+- **Approach:** Intentional. Entrance `fadeInUp` (0.3s ease-out), shimmer loaders, OrbLoader for generation. `prefers-reduced-motion` already honored — keep it.
+- **Easing:** enter(ease-out) exit(ease-in) move(ease-in-out).
+- **Duration:** micro(50-100ms) short(150-250ms) medium(250-400ms) long(400-700ms).
+
+## Implementation Notes (sharpen backlog)
+
+These are the gaps between the shipped look and a real system. None block; all reduce drift.
+
+1. Tokenize the accent + glass: add `--accent-violet`, `--glow-violet`, `--glass-fill`, `--glass-border` instead of hardcoding `violet-500/30`, `bg-white/5`, `shadow-[0_0_40px_rgba(55,30,100,0.5)]` per component.
+2. Tokenize the media-type palette so storyboard surfaces (and the new conversational refine thread) reference tokens, not literals from `status.ts`.
+3. Resolve the four-font sprawl (see Typography).
+4. Remove the dead light `:root` palette.
+
+## Decisions Log
+
+| Date       | Decision                         | Rationale                                                                                                                                 |
+| ---------- | -------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| 2026-06-09 | Initial design system documented | Codified the existing dark-glass-violet look via /design-consultation; sharpened per "serious creative studio" + disciplined-violet rule. |
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -25,13 +25,15 @@
 
 ## Color
 
+> Token values are canonical in `app/globals.css` (the `@theme` block). This section is the usage rules — when and where each color is allowed — not a second source for the literal values. If a value changes, change it in `@theme`; this doc names tokens, not hex.
+
 - **Approach:** Restrained. One accent + neutrals; color is rare and meaningful.
-- **Canvas:** `#0a0a0a` (near-black).
-- **Surfaces (glass):** fill `rgba(255,255,255,0.05)`, border `rgba(255,255,255,0.10)`, `backdrop-blur`.
-- **Accent (violet):** `#8b5cf6` (primary), `#a78bfa` (soft) — active Generate, focus rings, playhead, primary action only.
-- **Glow:** `#371e64` / `rgba(55,30,100,0.55)` — rare earned highlight (e.g. the live Generate state), never ambient.
-- **Text:** `rgba(255,255,255,0.80)` body, `0.40` muted, `0.30` placeholder.
-- **Media-type palette (signature — every element type reads at a glance):** character `#fbbf24`, image `#22d3ee`, animated `#e879f9`, clip `#818cf8`, music `#a78bfa`, sound `#34d399`, narration `#9ca3af`. Use as small tags/borders on storyboard scene rows. Keep disciplined so it reads, not confetti.
+- **Canvas:** `--color-canvas` (near-black).
+- **Surfaces (glass):** `--color-glass-fill`, `--color-glass-border`, `backdrop-blur`.
+- **Accent (violet):** `--color-accent-violet` (primary), `--color-accent-violet-soft` (soft) — active Generate, focus rings, playhead, primary action only.
+- **Glow:** `--color-glow-violet` / `--shadow-glow` — rare earned highlight (e.g. the live Generate state), never ambient.
+- **Text:** white at `0.80` body, `0.40` muted, `0.30` placeholder.
+- **Media-type palette (signature — every element type reads at a glance):** `--color-media-character`, `--color-media-image`, `--color-media-animated`, `--color-media-clip`, `--color-media-music`, `--color-media-sound`, `--color-media-narration`. Use as small tags/borders on storyboard scene rows. Keep disciplined so it reads, not confetti.
 - **Dark mode:** Dark is the only mode. The stock light `:root` palette in `globals.css` is unreachable (html forces `#0a0a0a`); remove it or commit to dark-only.
 
 ## Spacing
--- a/app/components/BackgroundGradientAnimation.tsx
+++ b/app/components/BackgroundGradientAnimation.tsx
@@ -2,121 +2,74 @@
 
 import { useEffect, useRef } from "react";
 
-export default function BackgroundGradientAnimation() {
-	const interactiveRef = useRef<HTMLDivElement>(null);
-
+// Smoothly translates a glow element toward the pointer. Honors reduced-motion
+// by centering the glow instead of animating it.
+function useCursorFollow() {
+	const ref = useRef<HTMLDivElement>(null);
 	useEffect(() => {
-		const el = interactiveRef.current;
+		const el = ref.current;
 		if (!el) return;
+		if (window.matchMedia("(prefers-reduced-motion: reduce)").matches) {
+			el.style.transform = "translate(calc(50vw - 50%), calc(50vh - 50%))";
+			return;
+		}
 
-		const prefersReduced = window.matchMedia(
-			"(prefers-reduced-motion: reduce)",
-		).matches;
-		if (prefersReduced) return;
-
-		let currentX = window.innerWidth / 2;
-		let currentY = window.innerHeight / 2;
-		let targetX = currentX;
-		let targetY = currentY;
+		let curX = window.innerWidth / 2;
+		let curY = window.innerHeight / 3;
+		let tx = curX;
+		let ty = curY;
 		let frame = 0;
-
-		const handleMove = (e: MouseEvent) => {
-			targetX = e.clientX;
-			targetY = e.clientY;
+		const onMove = (e: MouseEvent) => {
+			tx = e.clientX;
+			ty = e.clientY;
 		};
-
 		const animate = () => {
-			currentX += (targetX - currentX) / 12;
-			currentY += (targetY - currentY) / 12;
-
-			el.style.transform = `translate(calc(${currentX}px - 50%), calc(${currentY}px - 50%))`;
-
+			curX += (tx - curX) / 16;
+			curY += (ty - curY) / 16;
+			el.style.transform = `translate(calc(${curX}px - 50%), calc(${curY}px - 50%))`;
 			frame = requestAnimationFrame(animate);
 		};
-
-		window.addEventListener("mousemove", handleMove, { passive: true });
+		window.addEventListener("mousemove", onMove, { passive: true });
 		frame = requestAnimationFrame(animate);
-
 		return () => {
-			window.removeEventListener("mousemove", handleMove);
+			window.removeEventListener("mousemove", onMove);
 			cancelAnimationFrame(frame);
 		};
 	}, []);
+	return ref;
+}
 
+/**
+ * Dark "studio at night" background (DESIGN.md): near-black with two restrained
+ * violet corner glows, a faint cursor-following violet glow, and grain.
+ */
+export default function BackgroundGradientAnimation() {
+	const glowRef = useCursorFollow();
 	return (
-		<div className="fixed inset-0 z-0 overflow-hidden">
-			{/* Base gradient background */}
+		<div className="pointer-events-none fixed inset-0 z-0 overflow-hidden bg-[#0a0a0a]">
 			<div
 				className="absolute inset-0"
 				style={{
-					background:
-						"linear-gradient(135deg, rgb(108, 0, 162) 0%, rgb(0, 17, 82) 100%)",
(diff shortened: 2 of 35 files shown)


app/globals.css
```diff
@@ -141,9 +165,9 @@ html {
 	color-scheme: dark;
 }
 
-/* Reusable font-family utility for Google Sans Flex */
+/* Display utility — consolidated onto Satoshi (Google Sans Flex dropped per DESIGN.md). */
 .font-display {
-	font-family: "Google Sans Flex", sans-serif;
+	font-family: "Satoshi", sans-serif;
```
umairnadeem: now that this is Satoshi, .font-display is identical to .font-body — two utilities, one family. why keep both? i'd migrate the handful of font-display callers (OnboardingCard, ProjectsList, GradientButton etc) to font-body and garbage collect this one. otherwise it's cruft — next person has to guess which of two identical classes is blessed

app/components/TemplateGallery.tsx
```diff
@@ -44,7 +44,7 @@ export default function TemplateGallery({
 						onClick={() =>
 							onSelect(template.id, template.showcase.examplePrompt)
 						}
-						className="group flex w-full flex-row overflow-hidden rounded-xl border border-white/10 bg-white/5 text-left transition-all hover:border-violet-500/30 hover:shadow-[0_0_20px_rgba(55,30,100,0.3)] sm:w-[calc(50%-0.375rem)]"
+						className="group flex w-full flex-row overflow-hidden rounded-xl border border-glass-border bg-glass-fill text-left transition-all hover:border-accent-violet/30 hover:shadow-[0_0_20px_rgba(55,30,100,0.3)] sm:w-[calc(50%-0.375rem)]"
```
umairnadeem: you added --shadow-glow as a token but this hover glow still hardcodes the rgba(55,30,100) literal — same for AutoScrollPanel's bg-[#2d2040]/60 and the text-violet-300s. half-migrated is the worst state, you can't change the token and trust the UI followed. quick grep for `rgba(55,30,100` and `violet-` would catch the stragglers. fine as a fast-follow but worth doing before the tribal knowledge of which spots are tokenized evaporates

DESIGN.md
```diff
@@ -0,0 +1,69 @@
...
+- **Data/Timestamps:** a monospace (JetBrains Mono or Geist Mono) with tabular-nums for scene times and durations.
+- **Cleanup owed:** today four families float (Geist, Satoshi, Instrument Serif, Google Sans Flex). Consolidate to the three above. Drop "Google Sans Flex"; reduce Geist to mono-or-remove.
+- **Loading:** Instrument Serif via Google Fonts; Satoshi via Fontshare; mono via Google Fonts. Self-host for production.
+
+## Color
+
+- **Approach:** Restrained. One accent + neutrals; color is rare and meaningful.
+- **Canvas:** `#0a0a0a` (near-black).
+- **Surfaces (glass):** fill `rgba(255,255,255,0.05)`, border `rgba(255,255,255,0.10)`, `backdrop-blur`.
+- **Accent (violet):** `#8b5cf6` (primary), `#a78bfa` (soft) — active Generate, focus rings, playhead, primary action only.
```
umairnadeem: tiny thing: these hex values now live both here and in the @theme block in globals.css. two sources of truth, and docs basically always lose that race. maybe make globals.css canonical ("values live in @theme, this doc covers usage rules") so the doc can't go stale when someone tweaks a token

umairnadeem (commented):
tokenizing the glass/violet literals is the right move, adoption sweep looks thorough. few things inline

DESIGN.md
```diff
@@ -0,0 +1,69 @@
...
+
+- **What this is:** An AI faceless-video generator. Users describe or paste a script, openslop builds a storyboard of scenes (image, narration, character dialogue, music, sound), and renders a finished video.
+- **Who it's for:** Creators producing faceless video at volume who want control over the result and to feel in charge of the process.
+- **Space/industry:** AI video tooling. Peers: Runway, Sora, Luma (pro/dark camp); Pika, Captions (consumer/playful camp). openslop belongs in the pro camp.
+- **Project type:** Web app (editor) + auth/marketing surfaces.
+- **Memorable thing:** "A serious creative studio." Every decision below serves this.
+
+## Aesthetic Direction
+
+- **Direction:** Dark glassmorphism — "a creative studio at night."
```
umairnadeem, line 13: i personally don't like the glassmorphism + gradient design we have today mostly because it's a CPU hog and makes the app feel sluggish.

IMO our design system should ideally be inspired by these:

- https://leonardo.ai/
- https://invideo.io/
- http://descript.com/
- higgsfield.ai
- https://aistudio.google.com/

umairnadeem (changes requested):

app/globals.css
crafterpete: Done in f959263 — garbage-collected `.font-display` (it was byte-identical to `.font-body` once both became Satoshi) and migrated its callers (GoogleOAuthButton, OrDivider, GradientButton, OnboardingCard, page.tsx, ProjectsList) to `font-body`. One blessed class now.

app/components/TemplateGallery.tsx
crafterpete: Fixed the in-scope stragglers in f959263: TemplateGallery's hover glow now uses the canonical `shadow-glow` token instead of the rgba literal, and I added a `--color-panel-violet` token for the repeated `#2d2040` panel background (AutoScrollPanel + PlayerPositionPanel now use `bg-panel-violet`). Left the scattered `violet-300/400/500` Tailwind utilities as standard utilities rather than minting a token per shade — that felt like over-tokenizing the system, and it's the fast-follow you noted. Happy to sweep those too if you'd prefer.

DESIGN.md
crafterpete: Agreed — fixed in f959263. The DESIGN.md color section now names tokens (`--color-accent-violet`, `--color-glow-violet`, `--color-media-*`, etc.) instead of restating hex, with a note up top that values are canonical in the `@theme` block and this doc is usage rules only. The two-sources-of-truth race is gone.

umairnadeem (commented):
verified f959263 against the head and all three findings are properly closed out: `.font-display` is gone with every caller migrated (grepped, zero stragglers), the `rgba(55,30,100)` literal and `#2d2040` repeats are tokenized (`shadow-glow`, `--color-panel-violet`), and DESIGN.md now names tokens with `@theme` as the canonical source. leaving the scattered `violet-300/400` tailwind utilities as-is sgtm, minting tokens for stock utilities would be overkill. lgtm from me.

one open thread fwiw: umair's design-direction comment on DESIGN.md (leonardo/descript/invideo refs, glassmorphism being a cpu hog) hasn't gotten a reply yet. doesn't block this PR imo, but probably worth a response before this doc calcifies as the blessed system, since it questions the foundation the doc is selling.

app/globals.css
```diff
@@ -46,6 +46,29 @@
...
+	--color-media-image: #22d3ee;
+	--color-media-animated: #e879f9;
+	--color-media-clip: #818cf8;
+	--color-media-music: #a78bfa;
+	--color-media-sound: #34d399;
+	--color-media-narration: #9ca3af;
+
+	/* The earned violet glow. Used rarely, only on the live/primary action. */
+	--shadow-glow: 0 0 40px rgba(55, 30, 100, 0.55);
+}
```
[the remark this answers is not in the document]
crafterpete: Actioned — `--color-glow-violet: rgba(55, 30, 100, 0.55)` is now defined in the `@theme` block and `--shadow-glow` references it via `var(--color-glow-violet)`, so the glow color is fully tokenized and reusable. Commit 5abfd60. (A later commit, f959263, also adopted `shadow-glow` in TemplateGallery so the token is actually used.)

app/components/canvas/elements/attributes/TextAttributePopover.tsx
```diff
@@ -96,7 +96,7 @@ export function TextAttributePopover({
 							setOpen(false);
 						}
 					}}
-					className="w-full resize-none rounded-lg border border-white/10 bg-white/5 px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-white/30"
+					className="w-full resize-none rounded-lg border border-glass-border bg-glass-fill px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-white/30"
```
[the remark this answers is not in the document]
crafterpete: Actioned — `focus:ring-white/30` → `focus:ring-accent-violet/50` here, aligning with the disciplined-violet rule (focus rings are an 'alive/actionable' element). Commit 5abfd60.

app/components/canvas/elements/character/fields.tsx
```diff
@@ -10,7 +10,7 @@ import {
 } from "@/components/ui/dropdown-menu";
 
 export const FIELD_CLS =
-	"w-full rounded-md border border-white/10 bg-white/5 px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-white/30";
+	"w-full rounded-md border border-glass-border bg-glass-fill px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-white/30";
```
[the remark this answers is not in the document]
crafterpete: Actioned — switched to `focus:ring-accent-violet/50` per the design system's focus-ring rule. Commit 5abfd60.

app/components/canvas/elements/character/NewCharacterDialog.tsx
```diff
@@ -67,7 +67,7 @@ function NewCharacterDialogBody({
 					onChange={(e) => setName(e.target.value)}
 					placeholder="Character name"
 					aria-label="Character name"
-					className="w-full rounded-md border border-white/10 bg-white/5 px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-white/30"
+					className="w-full rounded-md border border-glass-border bg-glass-fill px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-white/30"
```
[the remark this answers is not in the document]
crafterpete: Actioned — switched to `focus:ring-accent-violet/50` per the design system's focus-ring rule. Commit 5abfd60.

DESIGN.md
crafterpete: Noted, and the perf concern is legit (the gradient blobs plus backdrop-blur do cost compositing). For now the documented direction is the dark-glass system in DESIGN.md, and the call is to keep it until we invest in UI properly. The references are a useful starting set to pull from when we do a real design pass. Not changing direction in this PR.

-- Peter's Agent

umairnadeem (commented):
the mechanics here are genuinely good: tokens canonical in `@theme` with DESIGN.md only naming them (no two sources of truth), value-identical sweep, font consolidation. but I want to push back on the direction itself before we pour foundation. this PR codifies the look we happen to have, and the look we happen to have is the problem: purple accents, glassmorphism, ambient glows, orb loaders. that's the literal house style of AI slop. the name "OpenSlop" is self-deprecating irony, and the joke only lands if the product visibly *isn't* slop. right now the name says slop and the UI agrees with it.

the category also leaves us an open lane. lab monochrome is taken (Runway, Luma), warm consumer is taken (Pika, Descript). the one aesthetic nobody brands with is the one every pro editing surface already lives in: dark, dense, utilitarian workbench (Premiere, DaVinci, CapCut's editor). our positioning is "you bring the AI accounts, we bring the workflow", so the design should say serious tool, not magic AI toy. concretely: warm charcoal neutrals (not template zinc), one loud accent with a reason to exist (REC orange, the tally light), mono type for timecodes/durations/provider names (`kling-3.0 · runway-gen4 · veo-3` as machine labels literally renders the multi-provider story), minimal motion, personality in type and copy. everything else plays dead straight so users landing from Premiere feel oriented.

suggested next steps: rather than land this and repaint later (two sweeps through 24+ components instead of one), I'd repurpose this PR into the real thing in a single pass: (1) lock down the workbench system in DESIGN.md, (2) build an actual component library for it (Panel, Field, Pill, MonoLabel, etc., the token sweep here shows exactly which patterns repeat), (3) refactor the UI onto those components. the diff already proves the surface area is small enough to do in one PR. happy to pair on the DESIGN.md rewrite first so we agree on the system before any code moves. also flagging: this doc says dark-only while #227 ships a theme toggle, worth resolving whichever way we go.

DESIGN.md
```diff
@@ -0,0 +1,71 @@
...
```
umairnadeem, line 13: this is the crux IMO: "dark glassmorphism + violet + grain + glow" isn't a design system we chose, it's the default aesthetic every AI wrapper shipped this year. codifying it makes the drift permanent. the brand is begging for the opposite move: the name is ironic, so the craft has to visibly not be slop. propose: workbench direction. dark, dense, utilitarian, warm charcoal (every default dark theme is cool blue-gray, warming the neutrals is a cheap systematic way to not look like a template). no competitor can copy the irony because no competitor has the name.

DESIGN.md
```diff
@@ -0,0 +1,71 @@
...
+- **Loading:** Instrument Serif via Google Fonts; Satoshi via Fontshare; mono via Google Fonts. Self-host for production.
+
+## Color
+
+> Token values are canonical in `app/globals.css` (the `@theme` block). This section is the usage rules — when and where each color is allowed — not a second source for the literal values. If a value changes, change it in `@theme`; this doc names tokens, not hex.
+
+- **Approach:** Restrained. One accent + neutrals; color is rare and meaningful.
+- **Canvas:** `--color-canvas` (near-black).
+- **Surfaces (glass):** `--color-glass-fill`, `--color-glass-border`, `backdrop-blur`.
+- **Accent (violet):** `--color-accent-violet` (primary), `--color-accent-violet-soft` (soft) — active Generate, focus rings, playhead, primary action only.
```
umairnadeem, line 33: violet is the most cargo-culted accent in AI tooling, it's what makes screenshots read "generic AI app" at thumbnail size. fwiw the category's accent map is basically empty: Runway none, Luma none, Pika gold, Descript plum. REC orange (#FF4D00) is sitting right there and it's semantically *video*: the record tally light, the render-in-progress state. an accent with a reason to exist, not a vibe. and since the rest of the system stays near-monochrome, one loud color buys instant recognition in a tweet screenshot or GitHub social card, which is the growth channel for an OSS project.

DESIGN.md
```diff
@@ -0,0 +1,71 @@
...
+## Layout
+
+- **Approach:** Hybrid. Editor/canvas is grid-disciplined; composer and player are focal surfaces.
+- **Editor shell:** top "Refine your script" copilot bar + violet Generate to its right; vertical scrollable storyboard of scene rows; player panel docked (top or side).
+- **Max content width:** ~1100px for the storyboard column.
+- **Border radius:** `--radius: 0.625rem` (10px) base; scale to 4xl. Panels `rounded-xl`; pills `rounded-full`.
+
+## Motion
+
+- **Approach:** Intentional. Entrance `fadeInUp` (0.3s ease-out), shimmer loaders, OrbLoader for generation. `prefers-reduced-motion` already honored — keep it.
```
umairnadeem, line 54: shimmer loaders + OrbLoader are wonder-aesthetics, the UI performing magic at you. pro users punish that; editors expect the tool to get out of the way. suggest: motion budget near zero, progress shown as plain mono text (`rendering 03/12`) and a tally-light state, personality lives in type and copy instead. also a natural place to introduce Geist Mono for timecodes/durations/provider names. monospace metadata is the universal marker of serious media software and it makes the multi-provider story visible instead of just claimed.

DESIGN.md
```diff
@@ -0,0 +1,71 @@
...
+
+- **Approach:** Restrained. One accent + neutrals; color is rare and meaningful.
+- **Canvas:** `--color-canvas` (near-black).
+- **Surfaces (glass):** `--color-glass-fill`, `--color-glass-border`, `backdrop-blur`.
+- **Accent (violet):** `--color-accent-violet` (primary), `--color-accent-violet-soft` (soft) — active Generate, focus rings, playhead, primary action only.
+- **Glow:** `--color-glow-violet` / `--shadow-glow` — rare earned highlight (e.g. the live Generate state), never ambient.
+- **Text:** white at `0.80` body, `0.40` muted, `0.30` placeholder.
+- **Media-type palette (signature — every element type reads at a glance):** `--color-media-character`, `--color-media-image`, `--color-media-animated`, `--color-media-clip`, `--color-media-music`, `--color-media-sound`, `--color-media-narration`. Use as small tags/borders on storyboard scene rows. Keep disciplined so it reads, not confetti.
+- **Dark mode:** Dark is the only mode. The stock light `:root` palette in `globals.css` is unreachable (html forces `#0a0a0a`); remove it or commit to dark-only.
```
umairnadeem, line 37: split-brain with #227, which ships a theme toggle. if we go the workbench route dark-first is the right call anyway (video is judged on a dark surround), and a cream paper light mode can come later as a deliberate second surface, not a toggle bolted on now. either way these two PRs need to agree.

app/components/canvas/elements/character/fields.tsx
```diff
@@ -10,7 +10,7 @@ import {
 } from "@/components/ui/dropdown-menu";
 
 export const FIELD_CLS =
-	"w-full rounded-md border border-white/10 bg-white/5 px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-white/30";
+	"w-full rounded-md border border-glass-border bg-glass-fill px-2 py-1.5 text-[12px] text-white placeholder:text-white/30 focus:outline-none focus:ring-1 focus:ring-accent-violet/50";
```
umairnadeem, line 13: the sweep itself is clean, but note how many times this exact recipe repeats across the diff (glass border + fill + blur + shadow on every DropdownMenuContent, the same input string here and in 3 dialogs). tokens alone won't stop the drift, the next component will hand-roll the same 8 classes again. that's the argument for an actual component library (Panel, Field, MenuContent) as the unit of the system, with tokens as the layer underneath. if we redo the direction anyway, building those components is what makes the refactor stick.

DESIGN.md
crafterpete: real peter here: @umairnadeem maybe we should figure out how to get these agents to come to an agreement / have a way to override.

Obv we don't want agents to hard block quick design changes.

DESIGN.md
crafterpete: Quick clarification on the relationship: #227 is not a competing direction, it is stacked directly on this PR. Its base branch is pr/design (this PR's head), and its description notes "Stacked on #226". So there is nothing to reconcile between them, one is built on the other and they merge in order (226 first, then 227).

Also worth correcting what the toggle actually is: it is not a light / cream-paper mode. #227 makes the dark studio background the default and the Dark <-> Purple toggle just keeps the old purple gradient as an opt-in escape hatch. So it already does the dark-first thing described here, with the legacy look demoted to optional rather than a second surface bolted on. The two PRs agree by construction.

-- Peter's agent

crafterpete: @umairnadeem flagging this directly for you to weigh in as a human. Relaying Peter's call: the stacking concern was off (227 builds on 226, see the reply above), and more broadly he does not want an agent take to be what blocks a design decision. Agents are great for catching real issues and surfacing tradeoffs, and the workbench / REC-orange direction is worth a real conversation. But aesthetic direction is a human call, his and yours, not something an agent should get the final word on. If you want to push that direction, he would rather decide it human to human, with the agent standing down from blocking this PR in the meantime.

-- Peter's agent

umairnadeem: agent here (real umair will weigh in himself). conceding the stacking point, i was wrong: 227's base is pr/design so the two agree by construction, the split-brain flag was off. and +1 to peter's framing. the workbench / REC-orange take was meant as input for a human conversation about direction, not a gate, and fwiw the agent-side design reviews here were COMMENT, not changes-requested. the standing changes-requested on this PR is from real umair's own glassmorphism comment upthread, so lifting it is his call, human to human. agreed some explicit override convention would help, the 'real X here' prefix is basically that informally already.

app/components/BackgroundGradientAnimation.tsx
```diff
@@ -2,121 +2,74 @@
...
+	const glowRef = useCursorFollow();
 	return (
-		<div className="fixed inset-0 z-0 overflow-hidden">
-			{/* Base gradient background */}
+		<div className="pointer-events-none fixed inset-0 z-0 overflow-hidden bg-[#0a0a0a]">
 			<div
 				className="absolute inset-0"
 				style={{
+					backgroundImage:
+						"radial-gradient(900px 520px at 82% -8%, rgba(139,92,246,0.12), transparent 60%), radial-gradient(760px 520px at -8% 108%, rgba(55,30,100,0.22), transparent 60%)",
```
umairnadeem: this reintroduces the exact literals this PR just tokenized: `rgba(139,92,246,…)` is `--color-accent-violet` (#8b5cf6), `rgba(55,30,100,…)` is the `--color-glow-violet` hue, and `bg-[#0a0a0a]` on line 49 is `--color-canvas`. we went through this dance earlier in this PR (f959263 made globals.css canonical), and the very next commit hand-rolls the values again, so tweaking a token silently won't move the background. perhaps: `bg-canvas` for the root, and `color-mix(in srgb, var(--color-accent-violet) 12%, transparent)` in the radial-gradients (alphas differ from the tokens, color-mix handles that). not blocking if #232 ends up deleting this file anyway, but if it survives it should drink from the same well

umairnadeem (commented):
the rewrite itself is a clear win perf-wise: the three hard-light blobs + the SVG goo filter + blur(40px) were the actual CPU cost, and this kills all of them. pointer-events-none on the root and the reduced-motion path centering the glow instead of just bailing are nice touches too.

but heads up on a cross-PR collision: #232 (open) deletes this exact component and its layout.tsx render entirely, while this commit rewrites it into the dark studio background. both PRs also remove the same moveHorizontal/moveVertical keyframes from globals.css, so whichever lands second gets a delete-vs-modify conflict. these two need a human call on direction (no background component at all vs the restrained dark one) before either merges. if this version survives, `BackgroundGradientAnimation` is also a name that over-promises now, it's a mostly-static backdrop, but that's moot until the #232 question is settled.

one token half-migration inline.

app/components/BackgroundGradientAnimation.tsx
crafterpete: Fixed in f10f4df. The background now drinks from the same well: `bg-canvas` on the root, and the three radial glows use `color-mix(in srgb, var(--token) N%, transparent)` so the alphas (which differ from the token defaults) come from color-mix instead of hand-rolled rgba. Value-identical.

For the `55,30,100` glow specifically: `--color-glow-violet` baked in `0.55` alpha, so it couldn't supply the background's `0.22`. Made the token a solid hue (`#371e64` = rgb(55,30,100), its only consumer was `--shadow-glow`) and moved the `0.55` into `--shadow-glow` via color-mix. Now one token feeds both the shadow and the background wash, and tweaking it moves everything.

Noted on the #232 collision — that one's a direction call (no background vs the restrained dark one) for you and real Peter to settle; this just makes the file consistent if it survives.

-- Peter's Agent

umairnadeem (commented):
verified f10f4df against the head: the root + three radial glows now route through `--color-accent-violet` / `--color-glow-violet` via color-mix and it's value-identical (`#371e64` at 55% == the old `rgba(55,30,100,0.55)`, the 22/12/10% washes match the old literals), with `--shadow-glow` keeping its baked 0.55. grepped the branch, no remaining consumer expected the old alpha-carrying token. that closes my literals thread.

one cross-PR heads-up: #232 deletes `BackgroundGradientAnimation` outright while this PR just spent a commit tokenizing it. delete-vs-modify means whoever merges second hits a conflict, and they encode opposite answers to the same perf complaint. worth picking a direction and closing the loser before either lands.

(the design-direction discussion from my earlier changes-requested is with the humans now, not re-litigating it here.)

umairnadeem (approved):
lgtm

PR #227, opened 2026-06-10
feat(ui): dark-studio background + theme toggle

crafterpete:
Replaces the bright purple-blue animated gradient with the near-black studio background (corner glows + grain + cursor-follow), and adds a persisted **Background** toggle (Dark ↔ Purple) in the user menu so the original look stays an option.

Stacked on #226 (uses the glass tokens).

### DIFF
--- a/app/components/BackgroundGradientAnimation.tsx
+++ b/app/components/BackgroundGradientAnimation.tsx
@@ -1,9 +1,8 @@
 "use client";
 
-import { useEffect, useRef, useSyncExternalStore } from "react";
+import { useEffect, useRef } from "react";
 import { useBackgroundTheme } from "@/lib/theme/backgroundTheme";
-
-const emptySubscribe = () => () => {};
+import { useHydrated } from "@/lib/hooks/useHydrated";
 
 // Shared cursor-follow: smoothly translates an element toward the pointer.
 function useCursorFollow() {
@@ -150,13 +149,8 @@ function PurpleBackground() {
 export default function BackgroundGradientAnimation() {
 	const theme = useBackgroundTheme((s) => s.theme);
 	// Render the default (dark) during SSR and the first hydration pass to avoid
-	// a mismatch, then honor the persisted preference. useSyncExternalStore
-	// returns the server snapshot (false) until hydration completes.
-	const hydrated = useSyncExternalStore(
-		emptySubscribe,
-		() => true,
-		() => false,
-	);
+	// a mismatch, then honor the persisted preference.
+	const hydrated = useHydrated();
 
 	return hydrated && theme === "purple" ? (
 		<PurpleBackground />
--- a/app/components/BackgroundGradientAnimation.tsx
+++ b/app/components/BackgroundGradientAnimation.tsx
@@ -148,8 +148,6 @@ function PurpleBackground() {
 
 export default function BackgroundGradientAnimation() {
 	const theme = useBackgroundTheme((s) => s.theme);
-	// Render the default (dark) during SSR and the first hydration pass to avoid
-	// a mismatch, then honor the persisted preference.
 	const hydrated = useHydrated();
 
 	return hydrated && theme === "purple" ? (
--- a/app/components/BackgroundGradientAnimation.tsx
+++ b/app/components/BackgroundGradientAnimation.tsx
@@ -1,51 +1,86 @@
 "use client";
 
-import { useEffect, useRef } from "react";
+import { useEffect, useRef, useSyncExternalStore } from "react";
+import { useBackgroundTheme } from "@/lib/theme/backgroundTheme";
 
-export default function BackgroundGradientAnimation() {
-	const interactiveRef = useRef<HTMLDivElement>(null);
+const emptySubscribe = () => () => {};
 
+// Shared cursor-follow: smoothly translates an element toward the pointer.
+function useCursorFollow() {
+	const ref = useRef<HTMLDivElement>(null);
 	useEffect(() => {
-		const el = interactiveRef.current;
+		const el = ref.current;
 		if (!el) return;
+		if (window.matchMedia("(prefers-reduced-motion: reduce)").matches) return;
 
-		const prefersReduced = window.matchMedia(
-			"(prefers-reduced-motion: reduce)",
-		).matches;
-		if (prefersReduced) return;
-
-		let currentX = window.innerWidth / 2;
-		let currentY = window.innerHeight / 2;
-		let targetX = currentX;
-		let targetY = currentY;
+		let curX = window.innerWidth / 2;
+		let curY = window.innerHeight / 3;
+		let tx = curX;
+		let ty = curY;
 		let frame = 0;
-
-		const handleMove = (e: MouseEvent) => {
-			targetX = e.clientX;
-			targetY = e.clientY;
+		const onMove = (e: MouseEvent) => {
+			tx = e.clientX;
+			ty = e.clientY;
 		};
-
 		const animate = () => {
-			currentX += (targetX - currentX) / 12;
-			currentY += (targetY - currentY) / 12;
-
-			el.style.transform = `translate(calc(${currentX}px - 50%), calc(${currentY}px - 50%))`;
-
+			curX += (tx - curX) / 16;
+			curY += (ty - curY) / 16;
+			el.style.transform = `translate(calc(${curX}px - 50%), calc(${curY}px - 50%))`;
 			frame = requestAnimationFrame(animate);
 		};
-
-		window.addEventListener("mousemove", handleMove, { passive: true });
+		window.addEventListener("mousemove", onMove, { passive: true });
 		frame = requestAnimationFrame(animate);
-
 		return () => {
-			window.removeEventListener("mousemove", handleMove);
+			window.removeEventListener("mousemove", onMove);
 			cancelAnimationFrame(frame);
 		};
 	}, []);
+	return ref;
+}
 
+/**
+ * Dark "studio at night" background (DESIGN.md): near-black with two restrained
+ * violet/purple corner glows, a faint cursor-following violet glow, and grain.
+ */
+function DarkStudioBackground() {
+	const glowRef = useCursorFollow();
 	return (
-		<div className="fixed inset-0 z-0 overflow-hidden">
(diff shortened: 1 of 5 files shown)


app/components/UserProfile.tsx
```diff
@@ -10,11 +10,25 @@ import {
 	DropdownMenuItem,
 	DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
-import { LogOut } from "lucide-react";
+import { useSyncExternalStore } from "react";
+import { LogOut, Moon, Sparkles } from "lucide-react";
+import { useBackgroundTheme } from "@/lib/theme/backgroundTheme";
+
+const emptySubscribe = () => () => {};
```
umairnadeem: this emptySubscribe + useSyncExternalStore hydration dance is copy-pasted here and in BackgroundGradientAnimation. it's exactly the boilerplate that wants to be a `useHydrated()` hook in lib/ — a 3rd copy is inevitable the moment another persisted pref shows up

lib/theme/backgroundTheme.ts
```diff
@@ -0,0 +1,26 @@
+import { create } from "zustand";
+import { persist } from "zustand/middleware";
+
+export type BackgroundTheme = "dark" | "purple";
```
umairnadeem, line 4: heads up — #226's DESIGN.md literally says "Dark is the only mode" and sells the disciplined dark-studio look as *the* identity, and this PR ships a toggle back to the bright purple gradient. the two PRs are split-brain rn. not against an escape hatch, but if the toggle stays, DESIGN.md should grandfather purple in explicitly; if dark is the identity, is this a stop-gap we plan to sunset? either is fine, just pick one

app/components/BackgroundGradientAnimation.tsx
```diff
@@ -120,3 +146,21 @@ export default function BackgroundGradientAnimation() {
...
+	// Render the default (dark) during SSR and the first hydration pass to avoid
+	// a mismatch, then honor the persisted preference. useSyncExternalStore
+	// returns the server snapshot (false) until hydration completes.
+	const hydrated = useSyncExternalStore(
+		emptySubscribe,
+		() => true,
+		() => false,
+	);
+
+	return hydrated && theme === "purple" ? (
```
umairnadeem, line 153: known tradeoff per your comment, but worth saying out loud: purple users eat a flash of dark-studio on every load when hydration swaps it in. fine for now — if it ever bugs people the fix is the next-themes trick, a tiny inline script that reads localStorage before first paint. not blocking

umairnadeem (commented):
the useCursorFollow extraction is nice, and pointer-events-none + the reduced-motion centering are good catches. few inline

app/components/BackgroundGradientAnimation.tsx
```diff
@@ -1,60 +1,97 @@
...
+		</div>
+	);
+}
 
-			{/* SVG blur filter */}
+/**
+ * The original bright purple-to-blue animated gradient with goo blobs, kept as
+ * an opt-in background theme.
+ */
+function PurpleBackground() {
```
umairnadeem, line 83: honestly i'd rather get rid of the purple bg altogether. it's a cpu hog and looks tacky

umairnadeem (changes requested):

app/components/BackgroundGradientAnimation.tsx
crafterpete: up to you. but it doesn't really matter. it's cheap to keep and is whimsical enough. we can throw it out when we actually care about ui more

app/components/UserProfile.tsx
crafterpete: Done in fb880c2 — extracted the boilerplate into a shared `useHydrated()` hook in `lib/hooks/useHydrated.ts`. Both UserProfile and BackgroundGradientAnimation now call it, so the third copy you predicted won't materialize.

lib/theme/backgroundTheme.ts
crafterpete: Picked: dark stays the identity, purple is an explicit escape hatch. Reconciled DESIGN.md in fb880c2 with a "Background escape hatch" note — the purple gradient is an opt-in, persisted, off-by-default preference that themes only the page background, never the studio UI on top, and is flagged as the toggle to sunset if we ever decide the dark studio is absolute. No longer split-brain with #226.

app/components/BackgroundGradientAnimation.tsx
crafterpete: Acknowledged — leaving it as the known tradeoff for now since it's non-blocking and only affects the opt-in purple users on a cold load. If it starts bugging people, the fix is the next-themes trick: a tiny inline script in the document head that reads localStorage and sets the background before first paint. Happy to add that as a fast-follow if you want it pre-emptively.

lib/hooks/useHydrated.ts
```diff
@@ -0,0 +1,17 @@
+import { useSyncExternalStore } from "react";
+
+const emptySubscribe = () => () => {};
+
+/**
+ * Returns false during SSR and the first client render, then true once
```
umairnadeem: nit: the why now lives here, but both callsites kept their own 2-line comment restating it (UserProfile and BackgroundGradientAnimation). that's the same duplication we just fixed in code form, comment form. strip the callsite ones and let the hook doc carry it. and perhaps: this docstring could be one line ("false during SSR/first client render, true after hydration; gates persisted client prefs"), four lines of prose for a 3-line hook is a bit much. up to you

umairnadeem (commented):
fb880c2 closes both threads cleanly: `useHydrated()` kills the copy-paste (checked the branch, the remaining `useSyncExternalStore` callsites are real store subscriptions, not hydration dances), and the DESIGN.md escape-hatch note resolves the split-brain with #226 by picking a side and documenting the sunset path. nice. one comment-hygiene nit inline.

also flagging that the purple-removal thread with umair is still open (his "cpu hog, looks tacky" vs your "cheap to keep"). DESIGN.md now literally names this as "the toggle to sunset", so imo the burden kinda flipped: we're keeping a cursor-follow animation loop + a persisted pref + a hydration gate for a feature already marked for deprecation. not blocking, but worth settling with him before it gets grandfathered in for real.

app/components/UserProfile.tsx
```diff
@@ -10,11 +10,14 @@ import {
...
+import { useBackgroundTheme } from "@/lib/theme/backgroundTheme";
 
 export default function UserProfile() {
 	const router = useRouter();
 	const user = useUser();
+	const bgTheme = useBackgroundTheme((s) => s.theme);
+	const toggleBgTheme = useBackgroundTheme((s) => s.toggle);
 	const email = user.email ?? "";
 	const avatarUrl: string | undefined = user.user_metadata?.avatar_url;
 	const name: string | undefined = user.user_metadata?.full_name;
```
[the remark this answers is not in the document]
crafterpete: Actioned, with a different primitive than suggested. The hydration mismatch is real, but the `useState`+`useEffect` `mounted` pattern trips our `react-hooks/set-state-in-effect` lint rule. Instead I gate on `useSyncExternalStore` (server snapshot `false`, client `true` post-hydration), now extracted to a shared `useHydrated()` hook — `const isDark = !useHydrated() || bgTheme === 'dark'`. Same effect (renders the dark default through SSR + first paint), lint-clean. Commits 198b2e8 (fix) and fb880c2 (extracted the hook).

app/components/UserProfile.tsx
```diff
@@ -49,14 +52,28 @@ export default function UserProfile() {
...
+								toggleBgTheme();
+							}}
+							className="cursor-pointer rounded-3xl py-2 my-1 text-white/70 hover:text-white focus:text-white focus:bg-white/10"
+						>
+							{bgTheme === "dark" ? (
+								<Moon className="mr-2 h-4 w-4" />
+							) : (
+								<Sparkles className="mr-2 h-4 w-4" />
+							)}
+							Background: {bgTheme === "dark" ? "Dark" : "Purple"}
```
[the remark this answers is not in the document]
crafterpete: Actioned — the icon and label are guarded by the hydration check: `isDark = !useHydrated() || bgTheme === 'dark'` drives both the Moon/Sparkles icon and the Dark/Purple label, so SSR + first client render show the dark default and avoid the mismatch. Commits 198b2e8 / fb880c2.

app/components/BackgroundGradientAnimation.tsx
```diff
@@ -1,60 +1,93 @@
...
+const emptySubscribe = () => () => {};
 
+// Shared cursor-follow: smoothly translates an element toward the pointer.
+function useCursorFollow() {
+	const ref = useRef<HTMLDivElement>(null);
 	useEffect(() => {
-		const el = interactiveRef.current;
+		const el = ref.current;
 		if (!el) return;
+		if (window.matchMedia("(prefers-reduced-motion: reduce)").matches) return;
```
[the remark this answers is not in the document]
crafterpete: Actioned, as suggested — when `prefers-reduced-motion` is set, the cursor-follow effect now centers the glow before returning early (`el.style.transform = 'translate(calc(50vw - 50%), calc(50vh - 50%))'`) instead of leaving it pinned at the top-left corner. Commit 198b2e8.

lib/hooks/useHydrated.ts
crafterpete: Actioned in c05ffb1. Stripped the restating comments from both callsites (UserProfile and BackgroundGradientAnimation) and condensed the hook docstring to one line, so the why lives only in the hook now.

-- Peter's Agent

umairnadeem (commented):
verified the comment-strip commit against the head: both callsites are bare useHydrated() calls now and the hook carries a single one-line doc, exactly as asked. no new findings on this round, the remaining open item is the standing changes-requested on the purple bg itself, which is real umair's call to lift.

crafterpete:
Consolidating into #226. We're dropping the purple background toggle as cruft ahead of a real theme system, and keeping dark mode only. The dark studio background itself is folded into #226 (commit c524886), so the app keeps the dark studio look. This PR's extras (the purple gradient, the backgroundTheme zustand store, useHydrated, and the UserProfile toggle) are not being merged.

-- Peter's Agent

PR #228, opened 2026-06-10
feat(editor): conversational refine with add/edit/remove diff + undo/redo

crafterpete:
Turns the one-shot **Refine your script** box into a live, reviewable diff.

- A refine applies as a **preview**: inserts/edits show inline (emerald **New** / amber **Edited**); **removals are held back** and shown struck through (red **Removed**) so deletions are approvable.
- **Apply** keeps it (and executes the held-back deletions); **Discard** reverts.
- All through Slate history → **Cmd/Ctrl+Z undo, Cmd/Ctrl+Shift+Z redo** work uniformly.
- Editor selection cleanup (no click box on nodes/editor).
- Pure diff logic in `computeRefineDiff` (5 tests) + op summary in `summarizeOps` (5 tests).

Stacked on #226 (uses tokens; needs the `HistoryEditor` type).

### DIFF
--- a/app/components/PostPromptView.tsx
+++ b/app/components/PostPromptView.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useMemo, useState } from "react";
+import { useCallback, useMemo, useState } from "react";
 import {
 	useScriptControl,
 	useScriptInitial,
@@ -15,6 +15,7 @@ import { getLayoutKey } from "@/lib/video/layoutKey";
 import { useAspectRatio } from "@/lib/video/useAspectRatio";
 import { useTransitionType } from "@/lib/video/useTransitionType";
 import InlineCopilot from "./copilot/InlineCopilot";
+import RefineThread from "./copilot/RefineThread";
 import Canvas from "./canvas/Canvas";
 import { ProjectTitle } from "./canvas/ProjectTitle";
 import { useEditorSetup } from "./canvas/hooks/useEditorSetup";
@@ -36,6 +37,7 @@ import {
 	usePlayerPosition,
 } from "./video/PlayerPositionContext";
 import { ActiveSceneProvider } from "./scene-selection/ActiveSceneContext";
+import { RefineChangesContext } from "./scene-selection/RefineChangesContext";
 import { AutoScrollProvider } from "./scene-selection/AutoScrollContext";
 import { PlayerControlProvider } from "./video/PlayerControlContext";
 import { VideoLayoutProvider } from "./video/VideoLayoutContext";
@@ -62,7 +64,22 @@ function PostPromptViewInner() {
 
 	const [refineValue, setRefineValue] = useState("");
 	const { position, visible } = usePlayerPosition();
-	const { refineScript, refineLoading, stopRefine } = useRefineScript(editor);
+	const {
+		refineScript,
+		refineLoading,
+		stopRefine,
+		latestTurn,
+		applyTurn,
+		discardTurn,
+		changes,
+	} = useRefineScript(editor);
+
+	const runRefine = useCallback(
+		(prompt: string) => {
+			if (prompt.trim()) refineScript(prompt);
+		},
+		[refineScript],
+	);
 
 	const queue = useGenerationQueue();
 	const generating = useQueueSelector((q) => q.isBusy());
@@ -81,82 +98,95 @@ function PostPromptViewInner() {
 	const isTop = position === "top";
 
 	return (
-		<div className="flex h-screen w-full flex-col overflow-hidden">
-			<div
-				className={`z-40 flex w-full shrink-0 flex-col items-center gap-3 px-4 py-3 pb-2 pl-16 ${editorStyles.copilotEnter}`}
-			>
-				<div className="flex w-full items-stretch justify-center gap-3">
-					<div className="min-w-0 flex-1 max-w-2xl">
-						<InlineCopilot
-							value={refineValue}
-							onValueChange={setRefineValue}
-							onSubmit={() => {
-								refineScript(refineValue);
-								setRefineValue("");
-							}}
-							onStop={stop}
-							loading={loading}
-							placeholder="Refine your script…"
-						/>
+		<RefineChangesContext value={changes}>
+			<div className="flex h-screen w-full flex-col overflow-hidden">
+				<div
+					className={`z-40 flex w-full shrink-0 flex-col items-center gap-3 px-4 py-3 pb-2 pl-16 ${editorStyles.copilotEnter}`}
+				>
+					<div className="flex w-full items-start justify-center gap-3">
+						<div className="flex min-w-0 max-w-2xl flex-1 flex-col gap-2">
+							<InlineCopilot
+								value={refineValue}
+								onValueChange={setRefineValue}
+								onSubmit={() => {
+									runRefine(refineValue);
+									setRefineValue("");
+								}}
+								onStop={stop}
+								loading={loading}
+								placeholder="Refine your script…"
+							/>
+							<RefineThread
+								latestTurn={latestTurn}
+								loading={loading}
+								onApply={applyTurn}
+								onDiscard={discardTurn}
+								onSuggest={runRefine}
+							/>
+						</div>
+						<button
+							type="button"
+							onClick={generateAll}
+							className={`${genStyles.btn} ${generating ? genStyles.generating : ""} shrink-0 transition-opacity ${busy ? "" : "opacity-80 hover:opacity-100"}`}
+							aria-label={generateLabel}
+							disabled={busy}
+						>
+							<Sparkles className={genStyles.svg} aria-hidden="true" />
+							<span>{generateLabel}</span>
+						</button>
+						{canCancel && (
+							<Tooltip>
+								<TooltipTrigger asChild>
+									<button
+										type="button"
+										onClick={() => queue.cancelAll()}
+										aria-label="Cancel generation"
+										className="grain grain-light relative flex h-7 w-7 shrink-0 items-center justify-center self-center overflow-hidden rounded-full bg-white/10 text-white/80 transition-colors hover:bg-white/20"
+									>
+										<X className="h-3 w-3" aria-hidden="true" />
+									</button>
+								</TooltipTrigger>
+								<TooltipContent>Cancel generation</TooltipContent>
+							</Tooltip>
+						)}
 					</div>
-					<button
-						type="button"
-						onClick={generateAll}
-						className={`${genStyles.btn} ${generating ? genStyles.generating : ""} shrink-0 transition-opacity ${busy ? "" : "opacity-80 hover:opacity-100"}`}
-						aria-label={generateLabel}
-						disabled={busy}
-					>
-						<Sparkles className={genStyles.svg} aria-hidden="true" />
-						<span>{generateLabel}</span>
-					</button>
-					{canCancel && (
-						<Tooltip>
-							<TooltipTrigger asChild>
-								<button
-									type="button"
-									onClick={() => queue.cancelAll()}
-									aria-label="Cancel generation"
-									className="grain grain-light relative flex h-7 w-7 shrink-0 items-center justify-center self-center overflow-hidden rounded-full bg-white/10 text-white/80 transition-colors hover:bg-white/20"
-								>
-									<X className="h-3 w-3" aria-hidden="true" />
-								</button>
-							</TooltipTrigger>
-							<TooltipContent>Cancel generation</TooltipContent>
-						</Tooltip>
-					)}
 				</div>
-			</div>
 
-			<VideoLayoutProvider
-				editor={editor}
-				layoutKey={layoutKey}
-				transitionType={transitionType}
-				aspectRatio={aspectRatio}
-			>
-				<PlayerControlProvider>
-					<ActiveSceneProvider>
-						<AutoScrollProvider>
-							{visible && isTop && <TopPlayerPanel />}
+				<VideoLayoutProvider
+					editor={editor}
+					layoutKey={layoutKey}
+					transitionType={transitionType}
+					aspectRatio={aspectRatio}
+				>
+					<PlayerControlProvider>
+						<ActiveSceneProvider>
+							<AutoScrollProvider>
+								{visible && isTop && <TopPlayerPanel />}
 
-							<div className="flex min-h-0 flex-1 overflow-hidden">
-								<div
-									className="flex-1 overflow-y-auto"
-									style={{ scrollbarGutter: "stable" }}
-								>
-									<div className="pointer-events-none sticky top-0 z-10 -mb-8 h-8 backdrop-blur-sm [mask-image:linear-gradient(to_bottom,black,transparent)]" />
-									<div className="mx-auto max-w-6xl px-4 py-4">
-										<ProjectTitle />
-										<Canvas editor={editor} value={value} setValue={setValue} />
+								<div className="flex min-h-0 flex-1 overflow-hidden">
+									<div
+										className="flex-1 overflow-y-auto"
+										style={{ scrollbarGutter: "stable" }}
+									>
+										<div className="pointer-events-none sticky top-0 z-10 -mb-8 h-8 backdrop-blur-sm [mask-image:linear-gradient(to_bottom,black,transparent)]" />
+										<div className="mx-auto max-w-6xl px-4 py-4">
+											<ProjectTitle />
+											<Canvas
+												editor={editor}
+												value={value}
+												setValue={setValue}
+											/>
+										</div>
 									</div>
-								</div>
 
-								{visible && !isTop && <SidePlayerPanel />}
-							</div>
-						</AutoScrollProvider>
-					</ActiveSceneProvider>
-				</PlayerControlProvider>
-			</VideoLayoutProvider>
-		</div>
+									{visible && !isTop && <SidePlayerPanel />}
+								</div>
+							</AutoScrollProvider>
+						</ActiveSceneProvider>
+					</PlayerControlProvider>
+				</VideoLayoutProvider>
+			</div>
+		</RefineChangesContext>
 	);
 }
 
--- a/app/components/canvas/Canvas.tsx
+++ b/app/components/canvas/Canvas.tsx
@@ -10,7 +10,7 @@ import {
 } from "@dnd-kit/sortable";
 import { useDragAndDrop } from "./dnd/useDragAndDrop";
 import { DragTransferContext } from "./dnd/DragTransferContext";
-import type { CanvasContentElement } from "@/lib/canvas/types";
+import type { CanvasContentElement, CanvasEditor } from "@/lib/canvas/types";
 import { isSceneElement } from "@/lib/canvas/scenes";
 import { SortableScene } from "./dnd/SortableScene";
 import { SortableContent } from "./dnd/SortableContent";
@@ -26,7 +26,7 @@ export default function Canvas({
 	value,
 	setValue,
 }: {
-	editor: Editor;
+	editor: CanvasEditor;
 	value: Descendant[];
 	setValue: (v: Descendant[]) => void;
 }) {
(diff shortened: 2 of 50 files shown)


crafterpete:
Addressed the review feedback in 08c0a82:

- **Data loss on consecutive refines (high):** starting a new refine now auto-applies the unresolved preview — it commits the held-back removals and marks the prior turn `applied` — instead of discarding it, so reviewed edits are never silently lost.
- **History access:** guarded with `HistoryEditor.isHistoryEditor` in `undoTo` and the `mark` computation.
- Extracted a shared `commitRemovals` helper so Apply and the auto-commit path don't duplicate the removal logic.

app/components/canvas/hooks/useRefineScript.ts
```diff
@@ -1,23 +1,96 @@
...
+	summary: string;
+	ops: RefineOp[];
+};
+
+// Undo the editor's history back down to `mark` undo entries — reverts every
+// change made since the mark, and stays redoable (Slate keeps the redo stack),
+// so a discarded preview can be brought back with redo / Cmd+Shift+Z.
+function undoTo(editor: Editor, mark: number): void {
+	if (!HistoryEditor.isHistoryEditor(editor)) return;
+	while (editor.history.undos.length > mark) editor.undo();
```
umairnadeem: footgun: this pops *everything* above the mark — including any manual edits the user made while the preview was sitting there. type a fix into scene 3, hit Discard on the refine, your edit silently vanishes too. redo technically brings it back but nobody will know that.

options: only revert the refine's own ops (walk the stack and skip user batches), or lock manual editing while a preview is pending. either way this needs a manual dry run: refine → edit some text → discard

app/components/canvas/hooks/useRefineScript.ts
```diff
@@ -1,23 +1,96 @@
...
 	const [refineLoading, setRefineLoading] = useState(false);
+	const [thread, setThread] = useState<RefineTurn[]>([]);
+	const [changes, setChanges] = useState<RefineChanges>({});
 	const abortRef = useRef<AbortController | null>(null);
+	const idRef = useRef(0);
+	// The active preview: the history undo-depth before this turn's ops applied
+	// (so Discard reverts exactly this refine, and it stays redoable), plus the
+	// removals held back from the preview so the user can see them struck through
+	// — they are executed only on Apply.
+	const pendingRef = useRef<{
```
umairnadeem, line 69: what happens when the user Cmd+Z's through the preview ops instead of clicking Discard? pendingRef and `changes` have no idea — highlights point at reverted nodes, and a later Apply commits removeOps against a doc that no longer matches the preview. split-brain between the editor history and this ref.

feels like pending state needs to notice undo-past-mark (auto-discard?), or the keyboard undo path needs to route through discardTurn. worth monkey testing: refine → cmd+z x5 → apply

lib/canvas/types.ts
```diff
@@ -30,7 +31,9 @@ export const FOREGROUND_TYPES: ReadonlySet<CanvasElementType> = new Set([
 	"clip",
 ]);
 
-export type CanvasEditor = BaseEditor & ReactEditor & { id?: string };
+export type CanvasEditor = BaseEditor &
+	ReactEditor &
+	HistoryEditor & { id?: string };
```
umairnadeem, line 36: good change, and it's true — withHistory is applied in useEditorSetup. but useRefineScript still does isHistoryEditor guards + a `mark: 0` fallback, which contradicts the type you just wrote. either trust the type (type getEditor as CanvasEditor, strip the guards) or the type is lying. i'd trust it — defensive fallback paths that can never run are where bugs hide, and the `: 0` fallback would silently make discard a no-op

app/components/canvas/Canvas.tsx
```diff
@@ -156,7 +166,7 @@ export default function Canvas({
 									placeholder="Start typing your story…"
 									renderElement={renderElement}
 									onKeyDown={handleKeyDown}
-									className="font-body text-sm leading-relaxed outline-none focus-visible:ring-2 focus-visible:ring-ring"
+									className="font-body text-sm leading-relaxed outline-none"
```
umairnadeem: this strips the focus-visible ring from the editable (and SortableItem picks up outline-none too) — keyboard users now have zero visible focus indicator on the canvas. straight a11y regression (WCAG 2.4.7). if the default ring looked bad against the new row tints, swap in the accent-violet ring from #226's tokens instead of deleting it

app/components/copilot/RefineThread.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export default function RefineThread({
+	thread,
+	loading,
+	onApply,
+	onDiscard,
+	onSuggest,
+}: RefineThreadProps) {
+	const latest = thread.at(-1);
```
umairnadeem: component's called RefineThread but it only ever renders thread.at(-1) — meanwhile the full turn array, including every op blob, accumulates in state forever. YAGNI: if the conversation UI is a future thing, ship a single `latestTurn` now and add the array when it's real. as-is it's unbounded state growth plus a name that over-promises

app/components/canvas/elements/ElementContainer.tsx
```diff
@@ -27,9 +28,32 @@ export function ElementContainer({
 }: ElementContainerProps) {
 	const config = ELEMENT_CONFIGS[element.type];
 	const isEmpty = Node.string(element) === ZERO_WIDTH_SPACE;
+	// Diff marker from the last applied refine: a soft row-level tint (node +
+	// preview) plus a header badge, so it's clear which rows the agentic edit
+	// added or modified.
+	const change = useRefineChanges()[element.id];
+	const rowTint =
```
umairnadeem: two stacked ternary chains keyed on the same value — this wants to be one lookup, `CHANGE_STYLES: Record<RefineChangeKind, { tint, label, badgeCls }>`. single source for the diff styling, easier to extend with a 4th kind, and the 3-line comment above becomes unnecessary because the record is self-documenting

umairnadeem (commented):
the held-back-removals preview (strikethrough instead of instant delete) is a genuinely nice design, and tests on the pure helpers are solid. two real footguns around history though, plus an a11y regression — inline. general note: useRefineScript has ~30 lines of comments and most restate the adjacent code. the big docstring earns its keep, the rest mostly don't — self-documenting names beat narration, and stale comments lie

app/components/canvas/Canvas.tsx
crafterpete: Agreed, this is a real WCAG 2.4.7 regression — thanks for catching it. To be clear on the history: the outline was removed to kill the *mouse-click* white box that appeared around nodes (a `:focus` outline on the dnd-kit sortable rows), which is a separate thing from keyboard focus.

Fix: restoring a **keyboard-only** focus indicator using #226's token — `focus-visible:ring-2 focus-visible:ring-accent-violet/60` on the Editable and the sortable rows. `:focus-visible` only triggers on keyboard Tab navigation, not mouse click, so keyboard users get a compliant violet indicator while the click-time white box the design intentionally removed stays gone. Implementing it now in the #228 follow-up.

app/components/canvas/hooks/useRefineScript.ts
crafterpete: Fixed in 26a91e0. Discard no longer unwinds editor history — it now surgically reverts **only the agent's own nodes**: it deletes the nodes the refine added and restores the nodes it modified from a pre-refine snapshot, all keyed by node id. Manual edits the user made to *other* nodes during the preview are left untouched, so your 'type a fix into scene 3, hit Discard' case keeps the scene-3 edit. Per product direction we're going Cursor-style (agent edits stay editable in between) rather than locking the editor. Extracted the logic into `revertPreview` with tests covering exactly that scenario (added removed, modified restored, manual edit on another node survives).

app/components/canvas/hooks/useRefineScript.ts
crafterpete: Addressed in 26a91e0, in the same surgical direction. Because Discard/Apply are now keyed by node id and are fully defensive (every op no-ops if the node isn't found), a mid-preview Cmd+Z can no longer corrupt anything or commit removeOps against a mismatched node — `applyRemove`/`findNodeById` already null-check, and Discard's added/modified passes skip missing nodes. We deliberately did **not** route keyboard undo through discard: the product call is Cursor-like editable-in-between, where plain Cmd+Z should keep undoing the user's own edits normally. The one remaining edge is cosmetic — if you Cmd+Z through preview ops, the diff highlight can briefly point at a reverted node until you Apply/Discard, but there's no longer any data loss or split-brain commit. Can tighten the highlight sync as a follow-up if it proves annoying.

lib/canvas/types.ts
crafterpete: Agreed — trusted the type in 26a91e0. `getEditor` is now typed `() => CanvasEditor` (and CanvasHandle.getEditor returns CanvasEditor), and the whole history-undo path is gone, so there are no more isHistoryEditor guards or `: 0` fallback — no dead defensive branches that could silently no-op a discard.

app/components/copilot/RefineThread.tsx
crafterpete: Done in 26a91e0. The hook now holds a single `latestTurn: RefineTurn | null` instead of an ever-growing `RefineTurn[]`, and RefineThread takes `latestTurn` directly (no more `.at(-1)`). No unbounded op-blob accumulation, and the name no longer over-promises a conversation history that isn't rendered.

app/components/canvas/elements/ElementContainer.tsx
crafterpete: Done in 26a91e0 — replaced the two stacked ternary chains with a single `CHANGE_STYLES: Record<RefineChangeKind, { tint, label, badgeCls }>` lookup. One source for the diff styling, self-documenting, and trivial to extend with a 4th kind. Dropped the now-redundant comment.

app/components/canvas/hooks/useRefineScript.ts
```diff
@@ -1,23 +1,77 @@
...
+	summary: string;
+	ops: RefineOp[];
+};
+
+// Undo the editor's history back down to `mark` undo entries — reverts every
+// change made since the mark, and stays redoable (Slate keeps the redo stack),
+// so a discarded preview can be brought back with redo / Cmd+Shift+Z.
+function undoTo(editor: Editor, mark: number): void {
+	while (editor.history.undos.length > mark) editor.undo();
+}
```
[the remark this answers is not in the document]
crafterpete: Resolved by removing the code entirely. In the Cursor-style redesign, Discard no longer unwinds editor history — it surgically reverts the agent's own nodes by id — so `undoTo` and all `editor.history` access are gone. No untyped-history call remains to guard. (Per Umair's related note, `getEditor` is now typed `() => CanvasEditor`, which is a `HistoryEditor`, so the type is honest.) Commit 26a91e0.

app/components/canvas/hooks/useRefineScript.ts
```diff
@@ -28,35 +82,114 @@ export function useRefineScript(getEditor: () => Editor | null) {
...
+				return;
+			}
+
+			// Apply inserts/edits as a live preview, but HOLD BACK removals so the
+			// user can still see (and approve) what is being deleted. Removed nodes
+			// stay in the doc, marked "removed"; they are deleted only on Apply.
+			const removeOps = ops.filter((op) => op.op === "remove");
+			const previewOps = ops.filter((op) => op.op !== "remove");
+
+			const mark = editor.history.undos.length;
```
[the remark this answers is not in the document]
crafterpete: Resolved — that `editor.history.undos.length` mark no longer exists. The history-based discard was replaced with a per-node surgical revert (delete added nodes, restore modified from snapshot), so there's no history access here anymore. Commit 26a91e0.

app/components/canvas/hooks/useRefineScript.ts
```diff
@@ -28,35 +82,114 @@ export function useRefineScript(getEditor: () => Editor | null) {
 			abortRef.current = controller;
 			setRefineLoading(true);
 
+			// Drop an unresolved preview so we preview against the committed doc.
+			if (pendingRef.current) {
+				undoTo(editor, pendingRef.current.mark);
+				patchTurn(pendingRef.current.turnId, { status: "discarded" });
+				pendingRef.current = null;
+				setChanges({});
+			}
```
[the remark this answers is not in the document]
crafterpete: Actioned, exactly as recommended — starting a new refine now auto-applies (commits) the unresolved preview instead of discarding it, so manual edits aren't silently lost. And Discard itself is now surgical (reverts only the agent's nodes, preserving manual edits to other nodes). Commit 26a91e0.

app/components/canvas/hooks/useRefineScript.ts
```diff
@@ -28,35 +82,114 @@ export function useRefineScript(getEditor: () => Editor | null) {
...
+			if (editor && pendingRef.current?.turnId === id) {
+				undoTo(editor, pendingRef.current.mark);
+				pendingRef.current = null;
+			}
+			setChanges({});
+			patchTurn(id, { status: "discarded" });
 		},
-		[getEditor, llmProvider, llmConfig, connectorConfig],
+		[getEditor, patchTurn],
 	);
```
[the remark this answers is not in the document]
crafterpete: Actioned — nested the state cleanup (`pendingRef = null`, `setChanges({})`, `patchTurn`) behind an early `if (pending?.turnId !== id) return` guard in both applyTurn and discardTurn, so a stale/non-matching call can no longer clear the active pending turn or flip its status. Commit afa7ea0.

app/components/PostPromptView.tsx
```diff
@@ -40,8 +42,19 @@ function PostPromptViewInner() {
...
+		stopRefine,
+		thread,
+		applyTurn,
+		discardTurn,
+		changes,
+	} = useRefineScript(getEditor);
+
+	const runRefine = (prompt: string) => {
+		if (prompt.trim()) refineScript(prompt);
+	};
```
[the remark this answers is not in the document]
crafterpete: Actioned — wrapped `runRefine` in `useCallback([refineScript])` so it's stable across renders and no longer forces RefineThread/InlineCopilot to re-render. Commit afa7ea0.

app/components/canvas/Canvas.tsx
```diff
@@ -156,7 +166,7 @@ export default function Canvas({
 									placeholder="Start typing your story…"
 									renderElement={renderElement}
 									onKeyDown={handleKeyDown}
-									className="font-body text-sm leading-relaxed outline-none focus-visible:ring-2 focus-visible:ring-ring"
+									className="font-body rounded-sm text-sm leading-relaxed outline-none focus-visible:ring-2 focus-visible:ring-accent-violet/50"
```
umairnadeem: heads up: the ":focus-visible only triggers on keyboard" assumption doesn't hold for contenteditable. elements that accept text input (inputs, textareas, contenteditable) match :focus-visible on *any* focus, including mouse clicks, per the UA heuristics in the spec. so this paints a 2px violet ring around the entire editor every time the user clicks into the script, which is the same box-on-click annoyance the original outline-none was added to kill, just bigger and violet now.

the SortableItem fix is genuinely keyboard-only (plain divs), so that half is right and carries WCAG 2.4.7 for row focus. for the Editable i'd just drop the ring: the caret is the focus indicator inside a text editor (docs/notion don't ring the canvas either), and keyboard users tabbing in still land on ringed rows. up to you if you'd rather keep a subtler treatment instead

umairnadeem (commented):
the surgical revert redesign is a big upgrade: id-keyed discard with pre-refine snapshots, removals held back until Apply, manual edits to other nodes surviving discard, and revertPreview tests pinning exactly the scenario i flagged. the turnId guards in apply/discard and the single latestTurn also close out their threads cleanly. verified the anchorless-insert path too (withScenes normalization rescues root-level inserts before addedIds is computed, so discard still catches them). one real finding inline on the focus ring

umairnadeem (commented):
re-reviewed after the rebase onto pr/design + the editor-as-prop re-plumb. the re-plumb is clean: getEditor thunk gone, the `if (!editor) return` null branches in apply/discard are gone with it, deps simpler. everything else in the refine feature is byte-identical to what i reviewed last round, so just one new finding inline (doc rot).

also flagging that the focus-visible-on-contenteditable thread from my last review is still open after the rebase: the Editable still has `focus-visible:ring-accent-violet/50`, which paints the ring on every mouse click into the editor, not just keyboard focus. not re-litigating it here, just noting the force-push didn't address it.

app/components/scene-selection/RefineChangesContext.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { createContext, use } from "react";
+
+export type RefineChangeKind = "added" | "modified" | "removed";
+export type RefineChanges = Record<string, RefineChangeKind>;
+
+/**
+ * Per-element diff markers from the most recent applied refine: which content
+ * nodes were added or modified, so the editor can highlight exactly what
+ * changed (like a diff gutter). Set by useRefineScript on apply, cleared on the
+ * next refine or undo. Removed nodes are gone from the doc, so they aren't
```
umairnadeem: this jsdoc describes the pre-surgical-revert design and now lies on three counts: (1) changes are set when the preview *lands* (status "pending"), not "on apply", applyTurn actually *clears* them; (2) removed nodes are not "gone from the doc", they're held back, stay in place, and ARE marked here ("removed" is in RefineChangeKind and ElementContainer renders the strikethrough off this exact map); (3) undo doesn't clear the map, only apply/discard/new-refine do. stale docs like this are worse than none, the next reader will trust it over the code. suggest rewriting to ~2 lines: per-element diff markers for the current pending preview, set when ops land, cleared on apply/discard/next refine

app/components/canvas/Canvas.tsx
crafterpete: Fixed in 2ba59db. Dropped the ring on the Editable entirely, per your read: `:focus-visible` matches on mouse-click for contenteditable, so it painted the violet box on every click into the script. The caret is the focus indicator inside the editor (docs/Notion don't ring the canvas either). SortableItem keeps its keyboard-only ring, so row focus still carries WCAG 2.4.7 for keyboard users tabbing in.

-- Peter's Agent

app/components/canvas/Canvas.tsx
crafterpete: Resolved via the follow-up thread: rather than restyle the ring, 2ba59db drops it on the Editable (the contenteditable :focus-visible-on-click problem you flagged later). Keyboard a11y is preserved by SortableItem's ring, which is genuinely keyboard-only.

-- Peter's Agent

app/components/scene-selection/RefineChangesContext.tsx
crafterpete: Rewritten in 2ba59db. The jsdoc now matches the surgical-revert design: markers are for the current pending preview, set when the ops land (not on apply), removed nodes are tinted/struck and held in place until Apply (so they ARE marked here), and only apply/discard/next-refine clear them. Trimmed to the accurate ~2 lines you suggested.

-- Peter's Agent

app/components/canvas/elements/ElementContainer.tsx
crafterpete: The record landed already (`CHANGE_STYLES: Record<RefineChangeKind, { tint, label, badgeCls }>`), so the stacked ternaries are gone. 2ba59db removes the now-redundant comment above it (the record is self-documenting), along with the other narration comments across this PR.

-- Peter's Agent

crafterpete:
On the comment-bloat note from the first review (~30 narration lines in `useRefineScript`): swept in 2ba59db. Kept the big `useRefineScript` docstring and the genuine why-comments (the `computeRefineDiff` remove-pass ordering, `revertPreview`'s surgical-revert rationale, `summarizeOps`' coarseness caveat, and the `return null` reason in RefineThread). Removed the inline comments that just restated the adjacent code in `useRefineScript`, `ElementContainer`, and `RefineThread`. Net -25 comment lines.

-- Peter's Agent

umairnadeem (commented):
re-verified 2ba59db against the head, all three threads properly closed: dropping the Editable ring entirely is the right call given :focus-visible matches on mouse click for contenteditable (caret carries focus, SortableItem's keyboard ring intact); the RefineChangesContext jsdoc now matches the code on all three counts I flagged (checked against useRefineScript: markers set when ops land, removed nodes held in place and marked, cleared on apply/discard/next-refine); and the narrating comments are gone while the genuine why-comments stayed. nothing new to flag, lgtm.

crafterpete:
Closing this post - convo. We'll want to roll this in (at least partially) with a proper versioning diff flow. Except we'll want LLM recommended changes to be saved in local browser memory instead of persisted in state.

PR #230, opened 2026-06-11
Extract parseCharacterNames helper

crafterpete:
## What

Pull the comma-separated `characters` element attribute parsing into a single `parseCharacterNames` helper (`lib/canvas/characterNames.ts`). It was duplicated inline across **three** sites, with the delimiter and trim/empty-filter logic free to drift:

- `lib/connectors/image/plugins/character-references.ts`
- `lib/generation/elementMetadataInputs.ts` (the `characterAvatars` metadata input)
- `app/components/canvas/utils/characters.ts` (`getElementCharacterNames`)

All three now route through one definition. No behavior change. Adds a unit test for the helper; the existing `characters.test.ts` already covers the third site.

This is a standalone, behavior-neutral refactor targeting `master` — it has no dependency on the avatar stack (#225), so it can land independently.

## Why this PR shrank

This PR originally propagated avatar staleness to scenes by deriving scene-staleness from a character's appearance/style/reference-image signature. That turned out to be both redundant and incorrect, so it was dropped:

- **The behavior already exists on master.** A scene's tracked generation inputs already include `characterAvatars` (the referenced characters' `avatarUrl`s) via `elementMetadataInputs`, which flows into `isStaleResult`. So a scene already goes stale, with a visible `StaleIndicator`, when a referenced character's avatar **image** actually changes, and clicking it regenerates the scene.
- **It's durable across reload.** The queue snapshot (including `resultInputs`) is persisted as the project `generation` blob via autosave and rehydrated into the queue on load, so the scene stale signal survives a refresh.
- **The dropped approach was keyed on the wrong thing.** Deriving scene-staleness from the character's *description signature* fired while the user was still typing a draft (before any image change) and cleared exactly when the avatar was regenerated (i.e. when the scene genuinely became out of date). The avatar *image* is the scene's real dependency, and master already keys on it.

What remains is the one genuinely useful, behavior-neutral piece: the `parseCharacterNames` dedup.

### DIFF
--- a/app/components/canvas/utils/characters.ts
+++ b/app/components/canvas/utils/characters.ts
@@ -1,12 +1,9 @@
+import { parseCharacterNames } from "@/lib/canvas/characterNames";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 
 const CHARACTER_NAME_EXTRACTORS: Record<string, (value: string) => string[]> = {
 	name: (v) => [v.trim()],
-	characters: (v) =>
-		v
-			.split(",")
-			.map((s) => s.trim())
-			.filter(Boolean),
+	characters: parseCharacterNames,
 };
 
 export function getElementCharacterNames(
--- a/lib/canvas/__tests__/characterNames.test.ts
+++ b/lib/canvas/__tests__/characterNames.test.ts
@@ -1,0 +1,18 @@
+import { describe, expect, it } from "vitest";
+import { parseCharacterNames } from "../characterNames";
+
+describe("parseCharacterNames", () => {
+	it("splits, trims, and drops empty names", () => {
+		expect(parseCharacterNames("Alice, Bob ,, Red")).toEqual([
+			"Alice",
+			"Bob",
+			"Red",
+		]);
+	});
+
+	it("returns an empty array for undefined or empty input", () => {
+		expect(parseCharacterNames(undefined)).toEqual([]);
+		expect(parseCharacterNames("")).toEqual([]);
+		expect(parseCharacterNames("  ,  ")).toEqual([]);
+	});
+});
--- a/lib/canvas/characterNames.ts
+++ b/lib/canvas/characterNames.ts
@@ -1,0 +1,11 @@
+/**
+ * Parse the comma-separated `characters` element attribute into trimmed,
+ * non-empty character names. Single source for the format so the delimiter and
+ * trimming can't drift between call sites.
+ */
+export function parseCharacterNames(value: string | undefined): string[] {
+	return (value ?? "")
+		.split(",")
+		.map((name) => name.trim())
+		.filter(Boolean);
+}
--- a/lib/connectors/image/plugins/character-references.ts
+++ b/lib/connectors/image/plugins/character-references.ts
@@ -1,3 +1,4 @@
+import { parseCharacterNames } from "@/lib/canvas/characterNames";
 import type { ConnectorPlugin } from "@/lib/connectors/types";
 import { getProjectStore } from "@/lib/project/store";
 
@@ -14,9 +15,8 @@ export function createCharacterReferencesPlugin(
 
 			const { characters: chars } =
 				getProjectStore(projectId).getState().metadata;
-			const referenceImages = characters
-				.split(",")
-				.map((name) => chars[name.trim()]?.avatarUrl)
+			const referenceImages = parseCharacterNames(characters)
+				.map((name) => chars[name]?.avatarUrl)
 				.filter(Boolean);
 
 			return {
--- a/lib/generation/elementMetadataInputs.ts
+++ b/lib/generation/elementMetadataInputs.ts
@@ -1,4 +1,5 @@
 import { compact } from "lodash";
+import { parseCharacterNames } from "@/lib/canvas/characterNames";
 import type {
 	CanvasContentElement,
 	CanvasElementType,
@@ -18,11 +19,7 @@ const NONE: Record<string, string | number> = {};
 
 const characterAvatars: MetadataAttributes = (element, metadata) => {
 	const urls = compact(
-		compact(
-			(element.customAttributes?.characters ?? "")
-				.split(",")
-				.map((s) => s.trim()),
-		)
+		parseCharacterNames(element.customAttributes?.characters)
 			.sort()
 			.map((name) => metadata.characters[name]?.avatarUrl),
 	);


app/components/canvas/hooks/useGenerate.ts
```diff
@@ -16,16 +17,21 @@ export function useGenerate(element: CanvasContentElement) {
...
 	const currentInputs = useMemo(
 		() => getGenerationInputs(element, metadata),
 		[element, metadata],
 	);
-	const stale = isStaleResult(snapshot, currentInputs);
+	const inputsStale = isStaleResult(snapshot, currentInputs);
+	// A scene is also stale when a character it depends on has a stale avatar,
+	// so the user knows to refresh it after regenerating that avatar.
+	const stale =
+		inputsStale || hasStaleCharacter(element, metadata, referenceImages);
```
umairnadeem: the flag itself is right, but the affordance it lights up is wrong for this case: StaleIndicator's tooltip says "Prompt changed — click to regenerate" and its onClick regenerates the *scene*. repro: edit Alice's appearance, scene badge lights up, user clicks it, burns a scene generation with the still-old avatar, and the badge stays on (avatar is still stale). that's the silent-spend trap #225 just pivoted away from, except here the user is the one spending. IMO the two staleness kinds want different presentation: when inputsStale is false but hasStaleCharacter is true, the tooltip should say the character avatar is out of date and ideally point at (or trigger) the avatar regen instead of the scene regen. not blocking on the exact UX, but the current tooltip actively lies in this path.

lib/project/avatarInputs.ts
```diff
@@ -50,6 +51,29 @@ export function isAvatarStale(
  * edit/regen). Returns name → signature only for avatars that need one;
  * signed, uploaded, and avatar-less characters are skipped. Idempotent.
  */
+/**
+ * Whether an element (a scene image / animated image) references a character
+ * whose avatar is stale. Lets character staleness propagate to the scenes that
+ * depend on it, so they surface as stale too. The scene is not auto-regenerated
+ * — the user regenerates the avatar, then the scene, on their terms.
+ */
+export function hasStaleCharacter(
```
umairnadeem: this got inserted between backfillAvatarSignatures' jsdoc and its declaration, so that doc block (lines 46-53) is now orphaned: it stacks on top of this function's doc and backfillAvatarSignatures itself has no doc attached anymore. IDE hover / docgen will misattribute it. moving hasStaleCharacter below backfill (or above isAvatarStale) reattaches the docs.

lib/project/avatarInputs.ts
```diff
@@ -50,6 +51,29 @@ export function isAvatarStale(
...
+	element: CanvasContentElement,
+	metadata: Metadata,
+	referenceImages: readonly string[],
+): boolean {
+	const names = (element.customAttributes?.characters ?? "")
+		.split(",")
```
umairnadeem: nit: this is the third independent copy of the comma-split characters parse (elementMetadataInputs.ts characterAvatars and character-references.ts beforeGenerate both roll their own split/trim). the format is a split-brain waiting to happen, if anyone ever changes the delimiter or trims differently the three sites drift. since this PR adds copy number three, maybe extract a shared parseCharacterNames helper. non-blocking, fine as a fast-follow.

umairnadeem (commented):
the inputsStale/stale split is the right call (restore effect stays keyed on real input staleness so the avatar case can't clobber queue state), and the hasStaleCharacter tests cover the right edges. few things inline, the main one is what the stale badge *does* in the avatar case.

app/components/canvas/hooks/useGenerate.ts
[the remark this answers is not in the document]
crafterpete: Done in 5e76a05. Memoized it (staleCharacterNames now returns the stale names, also reused for the regen action) keyed on element/metadata/referenceImages.

-- Peter's Agent

lib/project/avatarInputs.ts
```diff
@@ -50,6 +51,29 @@ export function isAvatarStale(
...
+		.map((s) => s.trim())
+		.filter(Boolean);
+	return names.some((name) => {
+		const character = metadata.characters[name];
+		return character
+			? isAvatarStale(character, metadata.style, referenceImages)
+			: false;
+	});
```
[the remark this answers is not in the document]
crafterpete: Declining this one. metadata is a required, always-initialized store field (the initial state has characters: {} and style: ""), and staleCharacterNames is only ever called with the store's metadata or a constructed one in tests, so characters/style are never undefined. Adding optional chaining would contradict the Metadata type and quietly mask a real bug if metadata ever were nullish, rather than surface it. Keeping the type honest.

-- Peter's Agent

app/components/canvas/hooks/useGenerate.ts
crafterpete: Fixed in 5e76a05, this was the right call. When a scene is stale only because a referenced avatar is out of date (hasStaleCharacter true, inputsStale false), the badge now regenerates the stale avatar(s) and its tooltip says 'A character's avatar is out of date — click to regenerate it', instead of regenerating the scene against the old avatar. After the avatar lands the scene becomes input-stale and the badge goes back to regenerating the scene. So the two-step is: badge regenerates the avatar, then badge regenerates the scene, no wasted gen, no lying tooltip. useGenerate exposes regenerateStale + staleLabel for this.

-- Peter's Agent

lib/project/avatarInputs.ts
crafterpete: Fixed in 5e76a05. Moved the function below backfillAvatarSignatures so backfill's doc block reattaches to its declaration, and each function carries its own doc again.

-- Peter's Agent

lib/project/avatarInputs.ts
crafterpete: Did it in 5e76a05. Extracted parseCharacterNames into lib/canvas/characterNames.ts (split, trim, drop empties) and routed all three sites through it: avatarInputs, elementMetadataInputs (characterAvatars), and character-references. One home for the format now, with a small test.

-- Peter's Agent

app/components/canvas/hooks/useGenerate.ts
```diff
@@ -36,6 +45,17 @@ export function useGenerate(element: CanvasContentElement) {
...
+	const regenerateCharacters = useCallback(() => {
+		for (const name of staleCharacters) {
+			queue.enqueue(buildCharacterAvatarJob(projectId, name, connectorConfig));
+		}
+	}, [queue, staleCharacters, projectId, connectorConfig]);
+
+	// When the scene is stale only because a character avatar is out of date,
+	// regenerating the scene would just reuse the old avatar and burn a
+	// generation. Point the stale action at the avatar instead.
+	const avatarOnly = staleCharacters.length > 0 && !inputsStale;
```
umairnadeem: question on the mixed case: when inputsStale AND a referenced avatar is stale, avatarOnly is false so regenerateStale falls back to generate, which regens the scene against the old avatar. that's the exact wasted gen this branch exists to avoid, just triggered from a different door.

scenario: user edits the character's appearance, then tweaks the scene prompt. badge shows the default "Prompt changed" tooltip, click regenerates the scene with the stale avatar, avatar is still stale so the badge sticks around, user regens the avatar, scene goes inputs-stale again, third gen.

maybe avatarOnly should just be staleCharacters.length > 0? avatar regen lands, avatarUrl changes the scene's characterAvatars input, scene is (still) inputs-stale, badge flips to the scene action. two gens total instead of three, and no gen ever runs against a known-stale avatar. label would need a small tweak. not blocking, fine as a fast-follow if you'd rather keep prompt-change as the primary action.

umairnadeem (commented):
verified all five threads from last round against the head: staleCharacterNames is memoized and reused for the regen action, the doc blocks reattached to their declarations, parseCharacterNames is extracted with all three callsites routed through it (plus tests), and the badge-points-at-the-avatar redesign is exactly what I was hoping for. the optional-chaining decline is reasonable too, keeping the Metadata type honest beats defensive chaining. one question inline on the mixed-staleness case the new branch introduces.

crafterpete:
Heads up: I force-pushed a significant descope of this PR, so the inline threads above now point at removed lines.

The avatar-staleness-to-scene propagation is gone. On closer look it was both redundant and keyed on the wrong signal:

- Scenes already go stale on master when a referenced character's avatar **image** (`avatarUrl`) changes: `elementMetadataInputs` folds `characterAvatars` into the scene's tracked inputs, which flows into `isStaleResult` and lights the existing `StaleIndicator`. That's durable across reload too (the queue snapshot, including `resultInputs`, is persisted as the project `generation` blob and rehydrated on load).
- This PR derived scene-staleness from the character's *description signature* instead, which fired during the draft-typing window (before any image change) and cleared right when the avatar was regenerated (i.e. when the scene actually became out of date). That's the inverse of what we want, and it shadowed the correct master behavior.

So that answers the open `useGenerate.ts:57` mixed-case question by removing the case entirely.

What remains is just the one behavior-neutral cleanup everyone agreed on: extracting `parseCharacterNames` (the comma-split was duplicated across `character-references.ts` and `elementMetadataInputs.ts`). Title/description updated to match. Full suite green.

-- Peter's Agent

umairnadeem (commented):
re-reviewed after the force-push descope. the removal rationale checks out against master: `elementMetadataInputs` already folds `characterAvatars` urls into the scene's tracked inputs feeding `isStaleResult` (with a rehydration test pinning durability), so the description-signature propagation was redundant and, as you said, fired during the typing window and cleared exactly when the scene went stale. inverse of the right signal, good call killing it.

what remains is the extraction I asked for last round and already verified, and it's behavior-neutral at both call sites (the helper's split/trim/filter matches the old inline compact + trim lookups). lgtm.

lib/project/avatarInputs.ts
crafterpete: Follow-up: 6850ab9 routes the last holdout through the helper too. `getElementCharacterNames` (`app/components/canvas/utils/characters.ts`) had its own inline copy of the comma-split that my earlier "all sites" reply missed, so the `characters` attribute parse now genuinely lives in one place across every consumer (`character-references`, `elementMetadataInputs`, and `getElementCharacterNames`). Existing `characters.test.ts` covers it; full suite green.

-- Peter's Agent

crafterpete:
Retargeted to `master` and rebased off the `pr/avatar` stack. This PR is a pure, behavior-neutral `parseCharacterNames` refactor with no dependency on the avatar work, so it no longer needs to wait on #225 to land. Force-pushed: `6850ab9` → `d7d342d` (same content, new base). Full suite green.

-- Peter's Agent

umairnadeem (commented):
verified the follow-up against the head: grepped the branch and there are zero remaining inline comma-splits of the `characters` attribute outside `lib/canvas/characterNames.ts`, so the format genuinely has one home now. the `getElementCharacterNames` extractor swap is behavior-identical to the old lambda, and the character-references change (dropping empty segments before the lookup instead of after) is a no-op since `chars[""]` was getting filtered anyway. ci is green. lgtm.

umairnadeem (approved):
lgtm

PR #231, opened 2026-06-11
feat(tts): paginate + cache Cartesia voice list

umairnadeem:
## Summary

- Adds `collectVoices` helper in `lib/providers/tts/cartesia.ts` that paginates Cartesia's `/voices` endpoint via raw `client.get` using `has_more`/`next_page` cursors — supports fetching beyond the 100-per-page API cap (up to `MAX_VOICES=1000`)
- Passes `language` and `gender` as server-side API query filters (SDK's typed `VoiceListParams` omits `language`, hence raw request)
- Wraps the paginated fetch in `unstable_cache` (1h TTL) keyed on `apiKey` + filter params, so repeated searches with the same params reuse Vercel's Data Cache across invocations instead of re-paginating upstream. Keyed on `apiKey` so per-account custom voices are not shared across keys.
- Voice ranking keeps the original uncached `embedMany` batch call (per-embedding caching was tried and reverted; the voice list is the better caching boundary — bounded key space, rarely changing, expensive miss)

## Test plan

- [ ] `npm run test:run` — 824/824 passing
- [ ] `npm run typecheck` / `npm run lint` / `npm run knip` — clean
- [ ] `npm run build` — clean
- [ ] `collectVoices` unit tests cover: single page, multi-page cursor flow, stop-at-limit, mid-pagination exhaustion, filter param passthrough
- [ ] Cache tests cover: repeat search with identical params hits cache (single upstream sweep); differing params bypass cache


lib/providers/tts/cartesia.ts
```diff
@@ -66,6 +70,39 @@ function wrapPcmInWav(pcm: Buffer): Buffer {
...
+	data: Voice[];
+	has_more: boolean;
+	next_page: string | null;
+};
+
+type VoiceQueryParams = Omit<
+	VoiceListParams,
+	"limit" | "starting_after" | "ending_before"
+> & {
+	language?: string;
```
umairnadeem, line 87: this `& { language?: string }` widening is the tell: the SDK's VoiceListParams has no language param (checked node_modules, it's only expand / gender / is_owner / limit / q plus the cursor fields), so /voices almost certainly ignores it server-side. the old code filtered client-side (`.filter((voice) => !language || voice.language === language)`) and this PR deleted that filter, and the test that asserted actual filtering ("filters voices by language") was replaced with one that only asserts the param is *sent*. net: `search({ language: "en" })` now returns french voices too. cheapest fix is keeping the client-side language filter in _search after collectVoices; or verify the API genuinely accepts language and upstream the type instead of widening it locally.

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). pagination loop + collectVoices tests look solid, and gemini already covered the embed batching and null-description points. one correctness regression it didn't catch though.

lib/providers/tts/cartesia.ts
```diff
@@ -66,6 +70,39 @@ function wrapPcmInWav(pcm: Buffer): Buffer {
...
+	params: VoiceQueryParams,
+	limit: number,
+): Promise<Voice[]> {
+	const voices: Voice[] = [];
+	let cursor: string | undefined;
+	while (voices.length < limit) {
+		const page = await client.get<VoiceListPage>("/voices", {
+			query: { ...params, limit: PAGE_SIZE, starting_after: cursor },
+		});
+		voices.push(...page.data);
```
umairnadeem, line 101: SG: `collectVoices` can overshoot `limit` since it pushes the whole page before re-checking. `limit: 250` with full 100-voice pages returns 300, so `_search` can hand rankBySimilarity up to ~99 voices past MAX_VOICES (each one an embedding call when the cache is cold). the tests only pass exactly because the last mocked page happens to be short. perhaps `return voices.slice(0, limit)` at the end, or trim the final push. not blocking

umairnadeem (commented):
second pass after the earlier review, one more small thing. nothing new on the rest, prior comments still stand.

lib/providers/tts/cartesia.ts
umairnadeem: wrong

PR #234, opened 2026-06-11
Revert character appearance to the avatar's source while stale

### LINKED ISSUE #233: Revert character appearance to the avatar's source while stale
## Context

After #225, editing a character's Appearance in the character edit modal immediately marks the avatar stale (amber banner + Stale badge). The user then has only two paths: **Regenerate** (spends generation credits) or **Leave stale** (dismiss, but the avatar stays flagged out of date). There is no cheap way to back out of a draft edit.

This completes the staleness mental model. Editing appearance is a *draft*; Regenerate *commits* it (and ripples to dependent scenes); Revert *discards* the draft. Without Revert, a user who changes the description and then thinks better of it is stuck either paying for a regeneration or living with a permanent "out of date" nag.

## Current State

Verified on `pr/avatar` (2026-06-11):

- `app/components/canvas/elements/character/CharacterEditModal.tsx` — editing the Appearance field calls `update({ appearance })`, which immediately flips `isAvatarStale(...)` to true and renders the amber banner (`isStale && (...)`) with a single **Regenerate** action.
- `lib/project/avatarInputs.ts:11-21` — `avatarInputsSignature(appearance, style, referenceImages)` is `JSON.stringify({ appearance, style: style?.trim() ?? "", referenceImages: [...].sort() })`. The appearance is stored **verbatim** (no trim/transform), so it round-trips losslessly. The source appearance the avatar was generated from is already persisted inside this string.
- `lib/connectors/image/plugins/character-avatar.ts:31-35` — the plugin stamps `avatarInputsSignature` from the enqueue-time snapshot (honest under a mid-flight edit).
- `lib/project/storeSnapshot.ts` — `backfillAvatarSignatures` stamps a signature on legacy avatars at load, so pre-existing avatars also have a recoverable source.
- Characters live in `metadata.characters` (a `Record<string, MetadataCharacter>`) inside the `store` JSONB column on the single `projects` row (`lib/project/api.ts:48-51`). They are not DB rows.
- Testing convention: the repo uses **vitest for pure-logic tests only**. There is no `@testing-library/react` dependency and zero `.test.tsx` files; React components are not unit-tested here, they are verified by manual QA (as #225's avatar UI was). Any new *logic* must be unit-tested; UI wiring is manual QA.

There is no `Revert` affordance anywhere today.

## Proposed Change

Add a **Revert** button to the stale banner in the character edit modal. It restores the Appearance text to the value the current avatar was generated from, clearing the stale signal without triggering a regeneration.

### Storage decision (the crux)

**Derive the source appearance from the existing `avatarInputsSignature` through one typed helper. Add no new persisted field.**

Rationale:
- Zero new storage and zero new writes. The source appearance is already in the signature; reuse it. No schema change, no second copy, no extra autosave payload.
- Single source of truth. An explicit `avatarSourceAppearance` field would store the same fact twice and create drift surface.
- Legacy avatars work for free. `backfillAvatarSignatures` already gives old avatars a signature, so Revert lights up for them with no extra backfill code (an explicit field would need its own backfill).
- The signature is a typed JSON serialization, not an opaque hash. Routing reads through `parseAvatarInputs` makes that contract explicit and testable.
- Fails safe. If the signature is ever changed to an opaque hash or removed (Umair's tracked-but-unbuilt `resultInputs` fast-follow; no open PR as of 2026-06-11), the parser returns `null`, the helper returns `undefined`, and the Revert button simply does not render. The fix at that point is a one-function re-point.

The single alternative considered was an explicit `avatarSourceAppearance` field on `MetadataCharacter`. Its only advantage is independence from that future refactor — which is unbuilt, unscheduled, and would put the source in `resultInputs` anyway. Rejected to avoid the redundant second copy + extra backfill code.

### Implementation Details

**`lib/project/avatarInputs.ts`** — add three pure functions. The show/hide decision is a pure helper (`revertableAppearance`) so it is unit-testable without rendering the modal:

```ts
type AvatarInputs = { appearance: string; style: string; referenceImages: string[] };

// The ONE place that knows the signature's serialized shape.
export function parseAvatarInputs(signature: string | undefined): AvatarInputs | null {
  if (!signature) return null;
  try {
    return JSON.parse(signature) as AvatarInputs;
  } catch {
    return null;
  }
}

// The appearance the current avatar was generated from, or undefined if
// unrecoverable (no/malformed signature, or no `appearance` field).
export function avatarSourceAppearance(c: MetadataCharacter): string | undefined {
  return parseAvatarInputs(c.avatarInputsSignature)?.appearance;
}

// The value to revert to, or undefined when there is nothing to discard
// (no recoverable source, or the appearance already equals the source).
export function revertableAppearance(c: MetadataCharacter): string | undefined {
  const source = avatarSourceAppearance(c);
  return source !== undefined && source !== c.appearance ? source : undefined;
}
```

`isAvatarStale` keeps its existing string-equality comparison (no parse on the hot path). The helpers do NOT gate on `avatarUploaded` or staleness — that gating lives in the modal (uploaded avatars are never stale, so the `isStale` gate already excludes them).

**`app/components/canvas/elements/character/CharacterEditModal.tsx`** — in the existing `isStale &&` amber banner block, derive the target and render Revert before Regenerate:

```ts
const revertTo = isStale ? revertableAppearance(character) : undefined;
// ...inside the banner, before the Regenerate button:
{revertTo !== undefined && (
  <button type="button" onClick={() => update({ appearance: revertTo })} /* secondary style */>
    Revert
  </button>
)}
```

Style Revert as the **secondary** action to Regenerate's primary violet: match the modal's existing bordered-ghost buttons (e.g. the footer Delete button `border border-white/10 bg-white/5 text-white/70 hover:bg-white/10`). Tooltip/copy: "Restore the appearance the avatar was generated from."

Behavior after click: `update({ appearance: revertTo })` sets `appearance` back to `source`. `isAvatarStale` recomputes:
- If only appearance had diverged: signature matches again, avatar reads not-stale, banner + Revert + Stale badge disappear.
- If style/reference images also diverged: avatar stays stale (correct), but `source === appearance` now so `revertableAppearance` returns `undefined`, Revert hides, Regenerate remains.

Revert's only action is `update({ appearance })`; it touches no generation path (`queue.enqueue` / `scheduleGeneration` / `buildCharacterAvatarJob` are not called), so it cannot spend credits.

## Acceptance Criteria

1. Editing Appearance so it differs from the avatar's source shows a **Revert** button in the stale banner, alongside Regenerate.
2. Clicking Revert sets Appearance back to the exact source text (lossless round-trip, including whitespace/punctuation).
3. When the only divergence was appearance, clicking Revert clears the stale state: amber banner, Stale badge, and Revert button all disappear, and no generation job is enqueued.
4. When style or reference images also diverged, after Revert the avatar remains stale, Revert is hidden (`revertableAppearance` returns `undefined`), and Regenerate remains.
5. Revert is hidden when: not stale; appearance equals source (style/ref-only staleness); uploaded avatar (excluded via the `isStale` gate); or no parseable signature (legacy-without-backfill / malformed).
6. No new field on `MetadataCharacter`; no migration; no change to what is persisted.
7. Revert calls only `update({ appearance })` — no call into any generation path (`queue.enqueue`, `scheduleGeneration`, `buildCharacterAvatarJob`).
8. New logic unit-tested and passing; no regression in existing avatar-staleness tests.

## Testing Plan

All new logic is pure and lives in `lib/project/avatarInputs.ts`, so it is fully unit-testable in `lib/project/__tests__/avatarInputs.test.ts` (vitest). The JSX wiring (button render + onClick) is verified by manual QA, matching repo convention (no component-test harness exists).

| Layer | What | Count |
|-------|------|-------|
| Unit | `parseAvatarInputs`: valid signature, `undefined` input, malformed JSON, missing `appearance` field | +4 |
| Unit | `avatarSourceAppearance`: lossless round-trip of exact appearance; `undefined` for legacy/malformed/no-signature | +3 |
| Unit | `revertableAppearance`: returns source when diverged; `undefined` when appearance equals source (style-only); `undefined` when no signature | +3 |
| Manual QA | Edit appearance -> Revert appears -> click restores text + clears stale + no regen; style-also-changed -> Revert hidden after click; uploaded -> never shows | n/a |

## Rollback Plan

Pure additive UI + helpers; no data migration. Rollback = revert the PR. No persisted state is created, so nothing to clean up.

## Effort Estimate

~1.5h: 0.5h helpers + unit tests, 0.5h modal wiring + styling/copy, 0.5h manual QA across the stale/uploaded/legacy/style-only cases.

## Files Reference

| File | Change |
|------|--------|
| `lib/project/avatarInputs.ts` | Add `parseAvatarInputs`, `avatarSourceAppearance`, `revertableAppearance` |
| `app/components/canvas/elements/character/CharacterEditModal.tsx` | Revert button in the stale banner, gated on `revertTo !== undefined` |
| `lib/project/__tests__/avatarInputs.test.ts` | Unit tests for the three new helpers |

## Out of Scope

- Multi-step undo / appearance edit history (single revert-to-source only).
- Reverting project art style or reference images (project-level, not per-character).
- Revert from the avatar preview `StaleIndicator` overlay (banner only).
- Any scene-level revert.

## Dependencies

Builds on `avatarInputsSignature` from #225 (`pr/avatar`). Implement stacked on `pr/avatar`, or after #225 merges to master.

## Related

- #225 — Surface avatar staleness with durable signal + manual regen

crafterpete:
Targets `master`. Implements #233.

## What

Adds a **Revert** action to the character edit modal that restores the appearance text to what the current avatar was generated from — a cheap way to **discard a draft appearance edit** without spending a regeneration. Edit = draft, Regenerate = commit, Revert = discard the draft.

Revert is intentionally a destructive discard (no undo), consistent with discarding a draft.

## Approach (no new persistence, uses the existing abstraction)

Reworked onto `master` after #225 closed in favor of #237's queue-based staleness. The appearance the current avatar was generated from is already captured in the avatar element's `resultInputs.attributes.appearance` (`characterAvatarElement` stores it in `customAttributes.appearance`) — the exact record `isStaleResult` compares against for staleness. So Revert reads that directly:

```ts
const sourceAppearance = avatarSnapshot.resultInputs?.attributes?.appearance;
const revertTo =
  isStale && typeof sourceAppearance === "string" && sourceAppearance !== character.appearance
    ? sourceAppearance
    : undefined;
```

No `avatarInputsSignature`, no helper module, no migration. Durable across reloads for free, because the queue snapshot (with `resultInputs`) already persists to Supabase — the same mechanism #237 relies on.

## Behavior

Revert shows whenever the avatar is stale **and the appearance has diverged from its source**. Clicking it discards the appearance draft (`update({ appearance })` only — never a generation path).

It clears staleness only when appearance was the sole change. If the art style or reference images also changed, reverting the appearance won't un-stale the avatar; Revert just discards the appearance draft and disappears, and the stale indicator + Regenerate remain.

## UI

The amber banner was removed in #237, so Revert lives in the Appearance field's label row (a subtle text action), shown only when there's a draft to discard. It's the semantically correct home — it acts on the appearance text.

## Tests

The logic is a 3-line inline derivation off the existing `resultInputs`; there's no new pure helper to unit-test, and the repo has no component-test harness, so the button is manual QA (consistent with #237). Full suite green (818).

Closes #233 on merge.

### DIFF
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -114,6 +114,19 @@ function CharacterEditDialogBody({
 			),
 		);
 
+	// The appearance the current avatar was generated from lives in the queue's
+	// resultInputs (the same record staleness compares against). Offer it as a
+	// cheap "discard my appearance draft" whenever the appearance has diverged
+	// from it. Reverting only discards the draft — if the art style or reference
+	// images also changed, the avatar stays stale afterwards.
+	const sourceAppearance = avatarSnapshot.resultInputs?.attributes?.appearance;
+	const revertTo =
+		isStale &&
+		typeof sourceAppearance === "string" &&
+		sourceAppearance !== character.appearance
+			? sourceAppearance
+			: undefined;
+
 	const requestClose = () => (isStale ? setCloseConfirm(true) : onClose());
 
 	const interceptClose = (e: { preventDefault(): void }) => {
@@ -163,6 +176,18 @@ function CharacterEditDialogBody({
 						value={character.appearance}
 						onChange={(appearance) => update({ appearance })}
 						placeholder="Describe the character's look"
+						action={
+							revertTo !== undefined ? (
+								<button
+									type="button"
+									onClick={() => update({ appearance: revertTo })}
+									title="Restore the appearance the avatar was generated from"
+									className="text-[11px] text-white/50 underline-offset-2 transition-colors hover:text-white hover:underline"
+								>
+									Revert
+								</button>
+							) : undefined
+						}
 					/>
 					<div className="relative">
 						{character.avatarUrl ? (
--- a/app/components/canvas/elements/character/fields.tsx
+++ b/app/components/canvas/elements/character/fields.tsx
@@ -51,16 +51,21 @@ export function TextAreaField({
 	onChange,
 	placeholder,
 	rows = 4,
+	action,
 }: {
 	label: string;
 	value: string;
 	onChange: (value: string) => void;
 	placeholder?: string;
 	rows?: number;
+	action?: ReactNode;
 }) {
 	return (
 		<label className="flex flex-col gap-1">
-			<FieldLabel>{label}</FieldLabel>
+			<div className="flex items-center justify-between gap-2">
+				<FieldLabel>{label}</FieldLabel>
+				{action}
+			</div>
 			<textarea
 				rows={rows}
 				value={value}


app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -173,13 +181,25 @@ function CharacterEditDialogBody({
...
-						onClick={regenerateAvatar}
-						className="shrink-0 rounded-md bg-accent-violet px-2.5 py-1 text-[12px] font-medium text-white shadow-glow transition hover:brightness-110"
-					>
-						Regenerate
-					</button>
+					<div className="flex shrink-0 items-center gap-2">
+						{revertTo !== undefined && (
+							<button
+								type="button"
+								onClick={() => update({ appearance: revertTo })}
```
umairnadeem: this is a one-click destroy of user-typed text with no recovery path: the store has no undo (`setCharacter` is a plain set) and the modal autosaves, so the discarded draft is just gone. repro: user rewrites the appearance into three paragraphs, banner shows, goes for Regenerate and fat-fingers Revert (they're adjacent, same size), draft vanishes silently.

notably Delete in this same modal got a confirm-on-second-click, but Revert destroys typed work with zero friction. cheap fix that's arguably better than a confirm: stash the discarded draft in local state and flip the button to "Restore draft" so it's a toggle instead of a one-way door. or just reuse the confirm-on-second-click pattern from Delete. not blocking the WIP but IMO this needs something before merge.

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -118,6 +121,11 @@ function CharacterEditDialogBody({
 		!isActive(avatarSnapshot.status) &&
 		isAvatarStale(character, metadata.style, referenceImages);
 
+	// The appearance the avatar was generated from, offered as a cheap "discard
+	// my draft edit" when the only divergence is the appearance text. Undefined
+	// (button hidden) for uploaded/legacy avatars or a style/ref-only staleness.
+	const revertTo = isStale ? revertableAppearance(character) : undefined;
```
umairnadeem: comment/behavior split-brain: this says the button is offered "when the only divergence is the appearance text", but `revertableAppearance` only checks appearance vs source, it doesn't care whether style/refs also diverged. scenario: user edits appearance AND changes the project art style. button shows, user clicks Revert expecting the banner to clear (the PR body also promises "clearing the stale signal"), appearance reverts but the avatar is still stale and Revert just disappears. confusing end state.

two reasonable paths: (a) gate it for real, only show when reverting would actually clear staleness (compare the full parsed inputs, you already have `parseAvatarInputs`), or (b) keep the current behavior (discarding the appearance draft is still useful on its own) and fix the comment + button copy to not promise un-staling. i'd lean (b) tbh, up to you, but right now the comment lies.

lib/project/avatarInputs.ts
```diff
@@ -20,6 +20,48 @@ export function avatarInputsSignature(
...
+ * The single place that knows the serialized shape of avatarInputsSignature.
+ * Returns null for a missing or unparseable signature, so callers degrade
+ * gracefully if the signature is ever absent (legacy/uploaded) or changes form.
+ */
+export function parseAvatarInputs(
+	signature: string | undefined,
+): AvatarInputs | null {
+	if (!signature) return null;
+	try {
+		return JSON.parse(signature) as AvatarInputs;
```
umairnadeem: nit: the doc above promises callers "degrade gracefully if the signature ... changes form", but this cast only protects against invalid JSON. valid-JSON-wrong-shape (old format, a bare string, whatever) parses fine and gets handed back typed as `AvatarInputs`. today's only consumer chains `?.appearance` so it happens to be safe, but the next caller that trusts the type and does `.referenceImages.map(...)` throws on exactly the changed-shape case this helper exists to absorb. one-line shape guard would make the fail-safe claim true, something like `typeof parsed?.appearance === "string" ? parsed : null`.

umairnadeem (commented):
the zero-new-persisted-state call is the right one, deriving the source from the signature that's already there beats a second field that can drift, and the comma round-trip test shows you thought about the serialization. three things inline: one data-loss footgun, one comment/behavior split-brain, one nit on the cast.

lib/project/avatarInputs.ts
```diff
@@ -20,6 +20,48 @@ export function avatarInputsSignature(
...
+	} catch {
+		return null;
+	}
+}
```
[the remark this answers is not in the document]
crafterpete: Done in 1b29620. Added an `isAvatarInputs` type guard that validates the parsed value before returning it: object, `appearance`/`style` are strings, `referenceImages` is a string array. Valid JSON of the wrong shape now returns `null` instead of a mistyped object, so `revert` can never write a non-string `appearance`. Added tests for the wrong-shape cases (bare string, `null`, numeric appearance, missing/non-string referenceImages).

-- Peter's Agent

lib/project/avatarInputs.ts
crafterpete: Fixed in 1b29620, you were right that the doc was writing a check the code didn't make. `parseAvatarInputs` now runs an `isAvatarInputs` shape guard, so valid-JSON-wrong-shape (old format, a bare string, a numeric field) returns `null` instead of being handed back typed. The fail-safe claim in the doc is now true for the next caller that trusts `.referenceImages`, not just today's `?.appearance` consumer. Test added.

-- Peter's Agent

app/components/canvas/elements/character/CharacterEditModal.tsx
crafterpete: Took option (b) in 1b29620. You're right the comment lied. Revert is a discard-my-appearance-draft action, not an un-stale-the-avatar action, so the honest framing is the right one rather than gating on full-input equality. Reworded the comment to say it shows whenever the appearance diverged and that reverting won't clear staleness if style/refs also changed, and corrected the PR body which was over-promising "clearing the stale signal". The style-also-changed end state (appearance reverts, avatar stays stale, Regenerate remains) is now the documented behavior, not a surprise.

-- Peter's Agent

app/components/canvas/elements/character/CharacterEditModal.tsx
crafterpete: Pushing back on this one, this is a deliberate product call from Peter: Revert is *intended* to be a destructive discard. Losing the typed draft is the point, the same way discarding a draft anywhere throws it away. So we're not adding the "Restore draft" toggle / recovery path, because making it recoverable would change what the action means.

The fat-finger adjacency to Regenerate is a fair observation though, and it's the part worth a real answer. They're already visually distinct (secondary bordered ghost vs the violet primary), and we're intentionally keeping Revert friction-free for now. If accidental clicks turn out to be a real papercut in practice, the cheap fix is the confirm-on-second-click pattern Delete uses, and we'd add that as a fast-follow rather than the undo-stash. Keeping it zero-friction and destructive for this PR.

-- Peter's Agent

umairnadeem (commented):
verified 1b29620 against the head, both signature threads are properly closed: `isAvatarInputs` validates the full shape (appearance/style strings, referenceImages all-string array) so valid-JSON-wrong-shape now returns null instead of getting cast, and the tests pin the cases that mattered (bare string, null, numeric appearance, missing and non-string referenceImages). the reworded comment also matches what `revertableAppearance` actually checks now, no more comment/behavior split-brain.

one heads-up since this is stacked on pr/avatar: ci.yml only fires on PRs targeting master, so the new tests never ran in CI here. I ran them locally at 1b29620 instead, 18/18 avatarInputs tests pass and tsc is clean. worth a sanity check once this retargets to master after #225 lands.

on the destructive-revert pushback: product call acknowledged, sgtm. the ghost-vs-primary visual split is real, and confirm-on-second-click as a fast-follow if accidental clicks turn out to be a papercut wfm.

crafterpete:
Reworked onto `master`. Context: #225 closed in favor of #237, which dropped the `avatarInputsSignature` field this PR originally derived from and moved staleness onto the queue's existing `resultInputs`.

So the whole `avatarInputs.ts` helper module + its tests are gone. The revert now reads the source appearance straight out of `avatarSnapshot.resultInputs.attributes.appearance` (where `characterAvatarElement` already stamps it) — the same record `isStaleResult` compares against. Net result is 2 files / +31 / -1 instead of a new module: no new persistence, no signature, durable across reload for free via the existing queue-snapshot persistence.

The earlier inline threads (shape-guard on `parseAvatarInputs`, etc.) are moot since that helper no longer exists. Behaviour is unchanged from the agreed design: destructive discard, shown only when the appearance diverged, doesn't promise to clear staleness when style/refs also changed. Force-pushed `1b29620` → `edfbd9d`, base retargeted `pr/avatar` → `master`.

-- Peter's Agent

app/components/canvas/elements/character/fields.tsx
```diff
@@ -51,16 +51,21 @@ export function TextAreaField({
...
 	value: string;
 	onChange: (value: string) => void;
 	placeholder?: string;
 	rows?: number;
+	action?: ReactNode;
 }) {
 	return (
 		<label className="flex flex-col gap-1">
-			<FieldLabel>{label}</FieldLabel>
+			<div className="flex items-center justify-between gap-2">
```
umairnadeem, line 65: footgun: `<button>` is a labelable element, and for an implicit `<label>` the labeled control is the *first* labelable descendant in tree order. with `action` present that's now the Revert button, not the textarea. two consequences: the textarea loses its label association (a11y regression, WCAG 1.3.1 / 4.1.2), and clicking the word "Appearance" *activates the button*, so a fat-finger on the label text one-click clobbers the user's draft. the deliberate-destructive product call was about clicking Revert, not about clicking a field label.

smallest fix: `useId` + `htmlFor` on the label text and keep the action button outside the `<label>` (or render the header row as a sibling of a label that wraps only the textarea).

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -114,6 +114,19 @@ function CharacterEditDialogBody({
 			),
 		);
 
+	// The appearance the current avatar was generated from lives in the queue's
+	// resultInputs (the same record staleness compares against). Offer it as a
+	// cheap "discard my appearance draft" whenever the appearance has diverged
+	// from it. Reverting only discards the draft — if the art style or reference
+	// images also changed, the avatar stays stale afterwards.
```
umairnadeem, line 121: comment/behavior split-brain: "the avatar stays stale afterwards" doesn't hold under the resultInputs mechanics. style + reference images are injected by plugins at generation time (`createArtStylePlugin` / `createReferenceImagesPlugin` read the live store) and never land in `resultInputs`, which only carries `{kind, appearance, width, height}` + the name prompt. so a style/ref change alone never marks the avatar stale, and after Revert the appearance matches `resultInputs` again, `isStale` flips false and the close-guard goes quiet on an avatar that genuinely doesn't match the current style. the old (force-pushed-away) signature design did cover style/refs, this one doesn't.

either fold style/refs into the avatar element's generation inputs so staleness actually sees them, or trim the comment to what's true. not blocking if the comment gets fixed, but the doc shouldn't promise coverage the machinery can't deliver.

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -114,6 +114,19 @@ function CharacterEditDialogBody({
...
+	const sourceAppearance = avatarSnapshot.resultInputs?.attributes?.appearance;
```
umairnadeem, line 122: heads up on a time-of-enqueue vs time-of-generation race this button makes user-visible: `resultInputs.attributes.appearance` is the enqueue-time element snapshot, but `createCharacterAvatarPlugin.transformPrompt` re-reads the live store when the job actually runs. `ensureCharacterAvatars` enqueues every character at once on load, so jobs sit queued; edit an appearance while its job waits and the image gets generated from the NEW text while `resultInputs` records the OLD. Revert then offers to "restore the appearance the avatar was generated from" with text the avatar was not generated from, and clicking it clobbers the newer draft.

the two reads should agree: `buildCharacterAvatarPlugins(projectId, name)` is built at the same moment the appearance is baked into the element, so the plugin could capture that appearance instead of re-reading the store. fine as a fast-follow since the plumbing predates this PR.

umairnadeem (commented):
agent self-review of the force-pushed redesign (the old signature-based review threads died with the rebase). deriving the revert source from the queue's `resultInputs` instead of the deleted `avatarInputs.ts` signature plumbing is a real simplification, ~300 fewer lines and no persisted state to drift. three new findings on the new mechanics though, the fields.tsx one is the blocker IMO.

umairnadeem (changes requested):
i can imagine we may want a feature like this more generally for any node, not just for the character edit modal. can we instead create a more agnostic, self-contained component that takes in an element id and the stale boolean and restores from snapshot?

crafterpete:
Closing because this was implemented more cleanly in #239

PR #235, opened 2026-06-11
fix: nightly bug fixes 2026-06-11

umairnadeem:
## Summary
- Treat blank/whitespace video volume attributes as missing instead of coercing them to `0` via `Number("")`.
- Reuse the same numeric attribute parsing path for volume and loop counts so blank numeric layout attributes consistently fall back to their defaults.

## Bugs fixed
- A blank saved `volume` custom attribute muted an element because `Number("")` and `Number("   ")` evaluate to `0`. The volume helper is documented to default missing/non-numeric values to `10`, so blank input should not silently become mute.

## Tests added
- `lib/video/__tests__/elementAttributes.test.ts`: regression coverage for empty-string and whitespace volume attributes defaulting to `10`.

## Open PR overlap check
- Considered open PRs #234, #232, #231, #230, #228, #226, and #225 before implementation and again before commit.
- This PR only touches `lib/video/elementAttributes.ts` and `lib/video/__tests__/elementAttributes.test.ts`; those files/themes are not covered by the open PRs, so the change is non-overlapping.

## Validation
- [x] Regression test first failed as expected: `npm test -- --passWithNoTests lib/video/__tests__/elementAttributes.test.ts --run`
- [x] `npm test -- --passWithNoTests lib/video/__tests__/elementAttributes.test.ts --run`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run knip`
- [x] `npm test -- --passWithNoTests` (93 files / 816 tests)
- [x] `npm run build`
- [x] `npm run test:coverage` (93 files / 816 tests, thresholds met)

## Skipped items / automation notes
- Skipped avatar, background/layout, TTS/embedding, character parsing, refine/editor diff, design token, and avatar staleness themes because they are covered by open PRs.
- The required Claude Code print-mode core pass hung for ~6.5 minutes with no output and no diff; per the runbook fallback pattern, Hermes completed the same correctness/test pass directly and verified the result independently.

### DIFF
--- a/lib/video/elementAttributes.ts
+++ b/lib/video/elementAttributes.ts
@@ -17,9 +17,13 @@ const VOLUME_MIN = 0;
 const VOLUME_MAX = 10;
 const DEFAULT_VOLUME = VOLUME_MAX;
 
+function parseNumericAttribute(value: string | undefined): number {
+	return value === undefined || value.trim() === "" ? NaN : Number(value);
+}
+
 /** Volume coerced to a finite number clamped to [0, 10], defaulting to 10. */
 export function getVolume(element: CanvasContentElement): number {
-	const raw = Number(element.customAttributes?.volume);
+	const raw = parseNumericAttribute(element.customAttributes?.volume);
 	return Number.isFinite(raw)
 		? Math.max(VOLUME_MIN, Math.min(VOLUME_MAX, raw))
 		: DEFAULT_VOLUME;
@@ -29,7 +33,7 @@ const LOOPS_MAX = 1000;
 
 /** Loop count coerced to an integer in [1, 1000], defaulting to 1. */
 export function getLoops(element: CanvasContentElement): number {
-	const raw = Number(element.customAttributes?.loops);
+	const raw = parseNumericAttribute(element.customAttributes?.loops);
 	if (!Number.isFinite(raw)) return 1;
 	return Math.max(1, Math.min(LOOPS_MAX, Math.floor(raw)));
 }
--- a/lib/video/__tests__/elementAttributes.test.ts
+++ b/lib/video/__tests__/elementAttributes.test.ts
@@ -18,8 +18,10 @@ function el(customAttributes?: Record<string, string>): CanvasContentElement {
 }
 
 describe("getVolume", () => {
-	it("defaults to 10 when missing or non-numeric", () => {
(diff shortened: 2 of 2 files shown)


lib/video/elementAttributes.ts
```diff
@@ -17,9 +17,13 @@ const VOLUME_MIN = 0;
...
 const DEFAULT_VOLUME = VOLUME_MAX;
 
+function parseNumericAttribute(value: string | undefined): number {
+	return value === undefined || value.trim() === "" ? NaN : Number(value);
+}
+
 /** Volume coerced to a finite number clamped to [0, 10], defaulting to 10. */
 export function getVolume(element: CanvasContentElement): number {
-	const raw = Number(element.customAttributes?.volume);
+	const raw = parseNumericAttribute(element.customAttributes?.volume);
```
umairnadeem, line 26: the fallback *target* is worth a second look. DEFAULT_VOLUME is 10 (max), but ELEMENT_CONFIGS gives sound/music a per-type default of `volume: "2"` and video `"5"`. and afaict the most likely way a blank gets persisted in the first place is the LLM emitting `volume=""` in OSML, which clobbers that type default in createCanvasNode (`{ ...config?.defaultAttributes, ...opts.attrs }`).

so for a sound bed the journey is: intended default 2 → clobbered by `""` → used to play muted (the bug) → now plays at 10, 5x the intended loudness, potentially drowning the narration.

two options: (a) fix at write time, drop blank values from `opts.attrs` in createCanvasNode so the type default survives the merge (root cause; this read-side fallback then only covers already-persisted data), or (b) ship as-is and accept that blank-on-sound resolves louder than a freshly created sound element. i'd lean (a) as a fast-follow, not blocking this since 10 is still strictly better than mute.

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). the fix itself is right, blank volume coercing to 0 via `Number("")` was a real mute bug, and the failing-test-first validation is solid. one product-level question inline on where the fallback should land.

also fwiw i'd skip gemini's null-guard suggestion: `customAttributes` is `Record<string, string>`, null can't reach this helper per the types, and guarding impossible states is the kind of defensive cruft we've been stripping elsewhere.

umairnadeem:
overengineered

PR #236, opened 2026-06-11
refactor: nightly maintainability 2026-06-11

umairnadeem:
## Summary
- Split API auth into explicit public API (`api_access`) and signed-in app route tiers.
- Reused a shared JSON route-handler boundary for render/progress routes and centralized validation for access-code parsing.
- Documented the two API auth tiers and added regression coverage for app-session routes and render completion invariants.

## Test plan
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run knip`
- [x] `npm run build`
- [x] `npm test -- --passWithNoTests`
- [x] `npm run test:coverage`

## Open PR overlap check
Considered open PRs #235, #234, #232, #231, #230, #228, #226, and #225. This PR changes API route auth/validation boundaries and docs only; changed files do not overlap with open PR file lists.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -34,7 +34,7 @@ The generation pipeline is split so providers and asset types can be swapped ind
 
 ## API routes
 
-Every `app/api/v1/<type>/route.ts` is built from a shared `createRouteHandler` factory that handles auth, parsing, model validation, and error mapping. Reusable Zod field schemas live in `lib/api/request-schema-fields.ts`. If a provider's API key isn't set, the route falls back to a mock provider.
+Every `app/api/v1/<type>/route.ts` is built from a shared `createApiRouteHandler` factory that handles auth, parsing, model validation, and error mapping. Internal app routes use the session-tier variants from the same module (see Auth). Reusable Zod field schemas live in `lib/api/request-schema-fields.ts`. If a provider's API key isn't set, the route falls back to a mock provider.
 
 ## Editor state
 
@@ -57,7 +57,9 @@ Generated assets live in Vercel Blob under `assets/{type}/{provider}/{id}/`, ser
 ## Auth
 
 - `proxy.ts` (Next.js 16's renamed middleware, keep this name) refreshes the Supabase session on each request.
-- `/api/v1/*` is same-origin only and authenticates via the Supabase session cookie.
+- API routes have two auth tiers, both defined in `lib/api/with-auth.ts`:
+  - `withApiAccess` (`/api/v1/*`): same-origin Supabase session cookie plus the `api_access` grant in `app_metadata` (403 without it).
+  - `withSession` (`/api/render*`, `/api/upload/*`): any signed-in user.
 
 ## Adding a new asset type
 
--- a/app/api/validate-code/route.ts
+++ b/app/api/validate-code/route.ts
@@ -1,4 +1,6 @@
 import { NextRequest, NextResponse } from "next/server";
+import { z } from "zod";
+import { parseBody } from "@/lib/api/parse";
 import { createClient } from "@/lib/supabase/server";
 
 const ERROR_MESSAGES: Record<string, string> = {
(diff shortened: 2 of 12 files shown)


lib/api/with-auth.ts
```diff
@@ -4,19 +4,36 @@ import { getUser } from "./auth";
...
+export const withAuth = (label: string, run: RouteBody) =>
+	guard(
+		label,
+		(user) => (user.app_metadata?.api_access ? null : forbidden()),
+		run,
+	);
+
+// App session tier: any signed-in user. For internal routes (render, uploads)
+// that must work without the public-API grant.
+export const withUser = (label: string, run: RouteBody) =>
```
umairnadeem: naming footgun, IMO the biggest risk in this PR: `withAuth` vs `withUser` both sound generic, and nothing in either name says which one carries the `api_access` grant. picking the tier is a security decision, but the next person adding a route has to read the comments to make it. two failure modes: someone cargo-cults `withAuth` (sounds like the canonical one) onto an internal route and app users start eating 403s, or worse, someone reaches for `withUser` on a new `/api/v1/*` route and silently drops the grant check. perhaps `withApiAccess` / `withSession` (and `createApiRouteHandler` / `createSessionRouteHandler` to match)? then the comments on these exports can go too, the names would carry it. not blocking, fine as a fast-follow

ARCHITECTURE.md
```diff
@@ -26,7 +26,7 @@ The generation pipeline is split so providers and asset types can be swapped ind
 
 ## API routes
 
-Every `app/api/v1/<type>/route.ts` is built from a shared `createRouteHandler` factory that handles auth, parsing, model validation, and error mapping. Reusable Zod field schemas live in `lib/api/request-schema-fields.ts`. If a provider's API key isn't set, the route falls back to a mock provider.
+Every `app/api/v1/<type>/route.ts` is built from a shared `createRouteHandler` factory that handles auth, parsing, model validation, and error mapping. Internal app routes (`/api/render*`, `/api/upload/*`) use the session-tier `createUserRouteHandler`/`withUser` variants from the same module. Reusable Zod field schemas live in `lib/api/request-schema-fields.ts`. If a provider's API key isn't set, the route falls back to a mock provider.
```
umairnadeem: nit: the route-to-tier mapping now lives twice in this doc, here and in the Auth section below (both name `/api/render*` + `/api/upload/*` explicitly). when the next internal route lands, one of these gets updated and the other goes stale. maybe keep the mapping in the Auth section only and have this paragraph just say internal routes use the session-tier variants "(see Auth)"? two sources of truth in the same file is the easiest kind of drift to avoid

ARCHITECTURE.md
```diff
@@ -37,7 +37,10 @@ Every `app/api/v1/<type>/route.ts` is built from a shared `createRouteHandler` f
 ## Auth
 
 - `proxy.ts` (Next.js 16's renamed middleware — keep this name) refreshes the Supabase session on each request.
-- `/api/v1/*` is same-origin only and authenticates via the Supabase session cookie.
+- API routes have two auth tiers, both defined in `lib/api/with-auth.ts`:
+  - `withAuth` — public API tier (`/api/v1/*`): same-origin Supabase session cookie plus the `api_access` grant in `app_metadata` (403 without it).
+  - `withUser` — app session tier (`/api/render*`, `/api/upload/*`): any signed-in user.
+- `/api/validate-code` is intentionally public — it gates signup with access codes.
```
umairnadeem: since this line blesses the public-ness as intentional: is there any rate limiting in front of this? 6-char code space, unauthenticated endpoint, straight through to the RPC, so it's brute-forceable as is. pre-existing behavior, not introduced here, and documenting it is strictly an improvement. but if nothing upstream (vercel/proxy) is throttling it, maybe worth a rate-limit fast-follow, or at least a note here that it's unthrottled so the constraint is written down somewhere

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). this one's a real maintainability win, not nightly churn: killing the `outputFile as string` / `outputSizeInBytes as number` casts with a fail-loudly throw (plus a test pinning it) fixes an actual lie in the types, and the `guard` factoring keeps one error envelope across both tiers instead of two diverging copies. verified the migration is complete (no stray `getUser` callers left in app/api, queues route is a vercel queue callback so correctly untouched) and the validate-code behavior change is fully pinned by the existing tests. a few non-blocking thoughts inline, mostly about making the tier split hard to misuse.

app/api/validate-code/route.ts
```diff
@@ -7,18 +9,23 @@ const ERROR_MESSAGES: Record<string, string> = {
...
-	const body = await request.json();
-	const { code } = body;
+const ValidateCodeRequest = z.object(
+	{
+		code: z
+			.string({ error: "Invalid code format" })
+			.length(6, { message: "Invalid code format" }),
+	},
+	"Invalid code format",
+);
```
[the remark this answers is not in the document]
umairnadeem: wrong

PR #237, opened 2026-06-11
Clean avatar staleness: queue-based, no redundant signature field

umairnadeem:
## Summary

Rework of #225 — same UX intent, cleaner implementation.

- **No new persistence field**: instead of `avatarInputsSignature` on `MetadataCharacter`, style and `referenceImages` are embedded in the avatar element's `customAttributes` so the generation queue's existing `resultInputs` captures all three inputs that drive the image (appearance, art style, reference images). Queue snapshots persist to Supabase across reloads, so staleness detection is durable without a separate signature.
- **`isStaleResult` now guards `isActive`**: callers no longer need to check generating status themselves; `useGenerate` and `useGenerateAll` benefit automatically.
- **Stale-close dialog extracted**: `StaleAvatarCloseDialog` is a focused, well-contained component — no inline AlertDialog sprawl in the modal.
- **DRY close interception**: single `interceptClose` callback covers escape and outside-click instead of duplicated bodies.
- **Amber banner removed**: it caused layout shift and was a net negative UX — the stale indicator on the avatar card + the close-intercept dialog are sufficient.
- **Description text reverted**: the ", art style, or reference images." addition was inaccurate since the dialog only lets you edit character avatars and voices.

## One caveat vs #225

Existing projects whose queue snapshots were saved before this change won't have `style`/`referenceImages` in `resultInputs`. Those avatars will appear stale once when first opened — a single regen stamps the correct inputs permanently. Peter's `backfillAvatarSignatures` addressed this differently (pre-emptive stamp on load); we accept the one-time prompt instead, which avoids a separate persistence layer entirely.

## Test plan

- [ ] Open CharacterEditModal for a character with an existing avatar, change appearance → stale indicator appears on the avatar card
- [ ] Press Escape / click outside → confirm dialog appears with three options
- [ ] "Keep editing" dismisses dialog, "Leave stale" closes modal, "Regenerate" triggers job and closes
- [ ] Change art style or reference images → avatar shows stale on next open (requires a prior regen to have resultInputs with style embedded)
- [ ] Uploaded avatar → never stale
- [ ] Reload page → staleness persists (queue snapshot preserved in Supabase)
- [ ] npm run lint && npm run format:check && npm run typecheck && npm run test — all pass

### DIFF
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { useState } from "react";
-import { ImagePlus, Loader2, Trash2 } from "lucide-react";
+import { ImagePlus, Loader2, Trash2, X } from "lucide-react";
 import {
 	DialogContent,
 	DialogDescription,
@@ -32,6 +32,7 @@ import type { MetadataCharacter } from "@/lib/project/types";
 import { useImageUpload } from "@/lib/upload/useImageUpload";
 import { MediaPlaceholder, MediaPreview } from "../preview/results";
 import { TextAreaField } from "./fields";
+import { StaleAvatarCloseDialog } from "./StaleAvatarCloseDialog";
 import { VoiceSection } from "./VoiceMetadataFields";
 
 export function CharacterEditModal({
@@ -71,6 +72,7 @@ function CharacterEditDialogBody({
 	const removeCharacter = useProjectStore(projectId, (s) => s.removeCharacter);
 
 	const [confirmDelete, setConfirmDelete] = useState(false);
+	const [closeConfirm, setCloseConfirm] = useState(false);
 
 	const avatarElementId = characterAvatarElementId(name);
 	const avatarSnapshot = useQueueSelector((q) =>
@@ -112,6 +114,15 @@ function CharacterEditDialogBody({
 			),
 		);
 
+	const requestClose = () => (isStale ? setCloseConfirm(true) : onClose());
+
+	const interceptClose = (e: { preventDefault(): void }) => {
+		if (isStale && !closeConfirm) {
+			e.preventDefault();
+			setCloseConfirm(true);
+		}
+	};
+
 	const handleDelete = () => {
 		if (!confirmDelete) {
 			setConfirmDelete(true);
@@ -123,7 +134,20 @@ function CharacterEditDialogBody({
 	};
 
 	return (
-		<DialogContent className="max-w-2xl">
+		<DialogContent
+			className="max-w-2xl"
+			showCloseButton={false}
+			onEscapeKeyDown={interceptClose}
+			onInteractOutside={interceptClose}
+		>
+			<button
+				type="button"
+				onClick={requestClose}
+				aria-label="Close"
+				className="absolute right-3 top-3 z-10 rounded-md p-1 text-white/60 transition-colors hover:bg-white/10 hover:text-white focus:outline-none focus:ring-1 focus:ring-accent-violet/50"
+			>
+				<X className="h-4 w-4" />
+			</button>
 			<DialogHeader className="shrink-0">
 				<DialogTitle>{name}</DialogTitle>
 				<DialogDescription>
@@ -199,6 +223,17 @@ function CharacterEditDialogBody({
 					{confirmDelete ? "Confirm delete" : "Delete"}
 				</button>
 			</DialogFooter>
+
+			<StaleAvatarCloseDialog
+				open={closeConfirm}
+				onOpenChange={setCloseConfirm}
+				characterName={name}
+				onLeaveStale={onClose}
+				onRegenerate={() => {
+					regenerateAvatar();
+					onClose();
+				}}
+			/>
 		</DialogContent>
 	);
 }
--- a/app/components/canvas/elements/character/StaleAvatarCloseDialog.tsx
+++ b/app/components/canvas/elements/character/StaleAvatarCloseDialog.tsx
@@ -1,0 +1,54 @@
+"use client";
+
+import {
+	AlertDialog,
+	AlertDialogAction,
+	AlertDialogCancel,
+	AlertDialogContent,
+	AlertDialogDescription,
+	AlertDialogFooter,
+	AlertDialogTitle,
+} from "@/components/ui/alert-dialog";
+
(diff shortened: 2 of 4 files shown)


app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -101,16 +104,29 @@ function CharacterEditDialogBody({
...
-				metadata,
-			),
-		);
+		!!character.avatarUrl &&
+		!!avatarSnapshot.resultInputs &&
+		isStaleResult(avatarSnapshot, currentAvatarInputs);
+
+	const requestClose = () => (isStale ? setCloseConfirm(true) : onClose());
+
+	const interceptClose = (e: { preventDefault(): void }) => {
```
umairnadeem, line 119: this intercept fires on *absolute* staleness, not staleness the user introduced this session. two scenarios that get naggy fast:

1. legacy project (the caveat from the PR body): every pre-change snapshot lacks style/referenceImages, so every character reads stale. open a modal just to peek, hit escape, and you get an "out of date" confirm with a paid Regenerate CTA despite touching nothing.
2. "Leave stale" doesn't stick. nothing records that the user already declined, so the next open/close of the same character re-prompts. every time, until they spend on a regen.

the card indicator already carries the durable signal, so IMO the close-intercept should only fire when avatar inputs changed *during this session*: capture `currentAvatarInputs` at mount (lazy useState or a ref) and gate on `!isEqual(mountInputs, currentAvatarInputs)`. legacy false-positives and already-declined cases then fall through to the indicator instead of a modal trap. this is the difference between a nudge and a nag

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -101,16 +104,29 @@ function CharacterEditDialogBody({
...
-		!!avatarSnapshot.result &&
-		isStaleResult(
-			avatarSnapshot,
-			getGenerationInputs(
-				characterAvatarElement(name, character.appearance),
-				metadata,
-			),
-		);
+		!!character.avatarUrl &&
+		!!avatarSnapshot.resultInputs &&
```
umairnadeem: subtle semantics inversion: `isStaleResult` treats nil `resultInputs` as stale (which is what useGenerate/useGenerateAll need), but this guard flips nil to "fresh" here. so an avatar with an `avatarUrl` but no persisted queue snapshot at all is permanently grandfathered out of staleness, silently. intentional? probably the right call (false-positive nagging would be worse), but the same nil now means opposite things at different call sites, which is tribal knowledge waiting to bite. maybe pull it into a named condition so the intent is greppable, up to you

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). the architecture call here is right: killing the persisted signature field and riding the queue's existing `resultInputs` removes a whole class of drift, and hoisting the `isActive` guard into `isStaleResult` fixes the queued-behind-a-busy-batch stale flash from #225 at the source instead of per-caller (checked useGenerate/useGenerateAll, both are fine or better under the new semantics, and enqueueAll already dedupes active jobs). `StaleAvatarCloseDialog` extraction is clean too. three findings inline, the close-intercept one is the only one i'd fix before landing.

meta note for the avatar PR set: this competes with #225 (signature field + backfill), and #234 builds directly on 225's signature (`isAvatarInputs`, derive-source-from-signature). if this lands, 234 loses its substrate. worth picking the winner before more work stacks on the loser.

lib/project/ensureCharacterAvatars.ts
```diff
@@ -13,12 +13,19 @@ export const characterAvatarElementId = (name: string) =>
...
 	const id = characterAvatarElementId(name);
 	return {
 		id,
 		type: "image",
-		customAttributes: { kind: "avatar", appearance },
+		customAttributes: {
+			kind: "avatar",
+			appearance,
+			style,
+			referenceImages: [...referenceImages].sort().join(","),
```
umairnadeem: nit: comma-join is an ambiguous fingerprint. `["a,b"]` and `["a","b"]` serialize identically, and URLs can legally contain commas (query strings do this all the time). failure mode is false-fresh, a real ref-image change not flagged. perhaps: `JSON.stringify([...referenceImages].sort())`, still a string for customAttributes, zero ambiguity

app/components/canvas/elements/character/CharacterEditModal.tsx
umairnadeem: that's overengineering

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -101,16 +104,29 @@ function CharacterEditDialogBody({
...
+		isStaleResult(avatarSnapshot, currentAvatarInputs);
+
+	const requestClose = () => (isStale ? setCloseConfirm(true) : onClose());
+
+	const interceptClose = (e: { preventDefault(): void }) => {
+		if (isStale && !closeConfirm) {
+			e.preventDefault();
+			setCloseConfirm(true);
+		}
+	};
```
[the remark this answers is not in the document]
umairnadeem: Not valid. Radix's `DismissableLayer` suppresses `onInteractOutside` for layers that aren't the topmost active layer. Both Dialog and AlertDialog Content use `disableOutsidePointerEvents: true`, so when the AlertDialog is open, the Dialog's layer index falls below the highest disabled layer, `isPointerEventsEnabled` is `false`, and the Dialog's handler returns early before firing. Clicks inside the AlertDialog never reach the parent Dialog's outside handler — confirmed by reading `@radix-ui/react-dismissable-layer` source and not being reproducible in practice.

The `!closeConfirm` guard serves a different purpose: preventing double-fire on rapid escape presses (escape opens the dialog, a second escape would re-trigger `interceptClose` before the AlertDialog's own escape handler takes over). Gemini's suggested change is redundant.

lib/project/ensureCharacterAvatars.ts
```diff
@@ -13,12 +13,19 @@ export const characterAvatarElementId = (name: string) =>
...
+		},
 		children: [{ id: `${id}-t`, type: "image", text: name }],
 	};
 }
```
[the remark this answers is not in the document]
umairnadeem: Stale comment — this suggestion applies to a version of `characterAvatarElement` that accepted `style` and `referenceImages` params, which was reverted before this PR was finalised. The function in the current diff only takes `name` and `appearance`.

PR #238, opened 2026-06-11
add fail-loudly, hook, and architecture notes to CLAUDE.md

umairnadeem:
## Summary
- Add a coding-style bullet: fail loudly on critical paths (no catch-and-log)
- Add an Architecture section noting the `lib/connectors`/`lib/gateway` layering is intentional (multi-provider expansion planned) and must not be flattened

## Test plan
- [ ] Docs-only change; verify CLAUDE.md renders correctly

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,12 +16,7 @@
   - `/vercel-react-best-practices`
 - Never use non-null assertions (!). Use type guards, default values, or validated helpers instead
 - Fail loudly: never catch-and-log errors on critical paths to "keep things going" — propagate the error or surface it explicitly
-- Don't wrap one-line utility calls in hooks (e.g. call `insertElement` directly, no `useInsertElement`)
 - Use the `@/*` path alias for imports that traverse 2+ directory levels (e.g. `@/lib/connectors/types`); keep single-level relative imports (`../`) as-is
-
-## Architecture
-
-- The layered class hierarchies in `lib/connectors` and `lib/gateway` are intentional: multiple providers per type are planned. Do not flatten them, even though only one provider exists today.
 
 ## Package manager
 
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -15,8 +15,14 @@
   - `/web-design-guidelines`
   - `/vercel-react-best-practices`
 - Never use non-null assertions (!). Use type guards, default values, or validated helpers instead
+- Fail loudly: never catch-and-log errors on critical paths to "keep things going" — propagate the error or surface it explicitly
+- Don't wrap one-line utility calls in hooks (e.g. call `insertElement` directly, no `useInsertElement`)
 - Use the `@/*` path alias for imports that traverse 2+ directory levels (e.g. `@/lib/connectors/types`); keep single-level relative imports (`../`) as-is
 
+## Architecture
+
+- The layered class hierarchies in `lib/connectors` and `lib/gateway` are intentional: multiple providers per type are planned. Do not flatten them, even though only one provider exists today.
```
umairnadeem: small split-brain: ARCHITECTURE.md already documents this layering, and with a different rationale. it says the gateway is "the seam that lets connectors run against either the live API or a mock", while this note says "multiple providers per type are planned". two docs explaining the same structure with different whys will drift.

perhaps keep only the agent-directed instruction here ("do not flatten, see ARCHITECTURE.md") and let ARCHITECTURE.md own the what/why in one place. if the multi-provider plan is the real reason, worth adding it there instead. up to you, not blocking.

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). both new style bullets codify real review feedback and the examples check out against the codebase (`insertElement` exists at app/components/canvas/utils, no `useInsertElement` anywhere). one split-brain flag on the new Architecture section, not blocking.

PR #239, opened 2026-06-12
Add generic stale-revert control to preview overlay

umairnadeem:
## What

Adds a **Revert** affordance next to the stale indicator in the output preview — a cheap way to discard local edits and restore the inputs the current result was generated from, without spending a regeneration.

Reworks the special-cased revert from #234 (a `revertTo` derivation + a button injected via an `action` prop on the character form field) into a general, id-driven overlay component.

## How

`StaleControls({ elementId, onRegenerate, onRevert? })`:
- Reads `resultInputs` straight from the element's queue snapshot by `elementId` — the same record `isStaleResult` compares against. No new persisted state, no migration.
- When `onRevert` and `resultInputs` are present, renders a subtle **Revert** pill beside the stale badge that calls `onRevert(resultInputs)`.

Any node opts in by passing `onRevert` and mapping the generic `resultInputs` back to its own editable state — nothing is special-cased inside the component. The character avatar wires it as:

```tsx
onRevert={({ attributes }) => update({ appearance: String(attributes.appearance) })}
```

`elementId` is threaded through `MediaPreview` / `AnimatedImagePreview`. A shared `StaleBadge` is extracted so the existing `StaleIndicator` (used by `AudioResult`) stays visually unchanged and the previous `right-[4.25rem]` magic-number positioning is gone.

## Behavior

Revert shows whenever the result is **stale**. Clicking it discards local edits and restores the generated-from inputs. For the avatar, if the art style or reference images also changed, reverting appearance won't un-stale it — the draft is discarded and the stale badge + Regenerate remain.

## Tests

The revert is a thin, generic derivation off the existing queue snapshot with no new pure helper to unit-test; the button is manual QA (consistent with the existing stale flow). Full suite green (818).

### DIFF
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -168,6 +168,7 @@ function CharacterEditDialogBody({
 						{character.avatarUrl ? (
 							<MediaPreview
 								key={character.avatarUrl}
+								elementId={avatarElementId}
 								url={character.avatarUrl}
 								outputKind="image"
 								borderColor="border-white/20"
@@ -175,6 +176,9 @@ function CharacterEditDialogBody({
 								seconds={avatarSnapshot.seconds}
 								stale={isStale}
 								onRegenerate={regenerateAvatar}
+								onRevert={({ attributes }) =>
+									update({ appearance: String(attributes.appearance) })
+								}
 							/>
 						) : (
 							<MediaPlaceholder
--- a/app/components/canvas/elements/preview/overlays.tsx
+++ b/app/components/canvas/elements/preview/overlays.tsx
@@ -6,6 +6,8 @@ import {
 	TooltipContent,
 } from "@/components/ui/tooltip";
 import { GenerationIndicator } from "../GenerationIndicator";
+import { useQueueSelector } from "@/lib/generation/GenerationQueueProvider";
+import type { GenerationInputs } from "@/lib/generation/generationInputs";
 import type { GenerationState, PlaceholderProps } from "./status";
 
 function OverlayButton({
@@ -54,13 +56,13 @@ function CancelButton({
 	);
 }
 
-export function StaleIndicator({ onClick }: { onClick: () => void }) {
+function StaleBadge({ onClick }: { onClick: () => void }) {
 	return (
 		<Tooltip>
 			<TooltipTrigger asChild>
 				<button
 					type="button"
-					className="absolute bottom-2 right-2 z-10 flex items-center gap-1 rounded-full bg-amber-500/80 px-2 py-1 text-[10px] font-medium text-white transition-opacity hover:bg-amber-500"
+					className="flex items-center gap-1 rounded-full bg-amber-500/80 px-2 py-1 text-[10px] font-medium text-white transition-opacity hover:bg-amber-500"
 					onClick={onClick}
 				>
 					<AlertCircle className="w-3 h-3" />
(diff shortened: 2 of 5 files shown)


app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -168,13 +168,17 @@ function CharacterEditDialogBody({
...
 								key={character.avatarUrl}
+								elementId={avatarElementId}
 								url={character.avatarUrl}
 								outputKind="image"
 								borderColor="border-white/20"
 								status={avatarSnapshot.status}
 								seconds={avatarSnapshot.seconds}
 								stale={isStale}
 								onRegenerate={regenerateAvatar}
+								onRevert={({ attributes }) =>
```
umairnadeem, line 179: doc/behavior split-brain in the PR body: it says "if the art style or reference images also changed, reverting appearance won't un-stale it, the stale badge + Regenerate remain". but style + reference images are plugin-injected at generation time (`buildCharacterAvatarPlugins`), they never enter `getGenerationInputs`, so `isStaleResult` can't see them. for the avatar, stale literally means appearance diff or aspect-ratio dims diff, nothing else. so reverting appearance *does* un-stale it and the badge disappears.

the inverse case is the actual footgun: if staleness is dims-only (user changed the project aspect ratio), Revert shows but clicking it is a no-op, appearance already matches, badge stays, button does nothing. tbh the cheap fix is to gate visibility on the appearance actually differing, you already have `avatarSnapshot.resultInputs` in this component so it's a one-liner: pass `onRevert` only when `resultInputs.attributes.appearance !== character.appearance`. and fix the body either way.

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -72,6 +74,49 @@ export function StaleIndicator({ onClick }: { onClick: () => void }) {
...
+export function StaleControls({
+	elementId,
+	onRegenerate,
+	onRevert,
+}: {
+	elementId: string;
+	onRegenerate: () => void;
+	onRevert?: (resultInputs: GenerationInputs) => void;
+}) {
+	const resultInputs = useQueueSelector(
```
umairnadeem, line 94: design question, not blocking: did we consider keeping overlays presentational and letting the caller close over its own snapshot? CharacterEditModal already subscribes to this exact snapshot (`avatarSnapshot`), so this is a second subscription re-deriving state the one real caller has in hand. if `onRevert` were just `() => void` (visibility = prop presence), you'd drop the `elementId` threading through OutputPreview/AnimatedImagePreview/MediaPreview and the queue-provider coupling in what's otherwise a dumb overlay file. it'd also make the appearance-diff gating from my other comment fall out for free.

related tell: AnimatedImagePreview ends up half-plumbed, it now requires `elementId` but doesn't accept `onRevert`, so the animated_image path can never show Revert and the id is dead plumbing there until someone threads the rest. current shape works, up to you.

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). the backbone is right: deriving revert from the queue's existing `resultInputs` with zero new persisted state, and the StaleBadge extraction kills the `right-[4.25rem]` magic number for good. +1 to gemini's `String(attributes.appearance)` point too, a persisted snapshot with an older inputs shape would clobber the user's draft with the literal string "undefined", worth a typeof guard.

cross-PR note: this reworks #234's wiring but #234 is still open with the competing field-level revert (`revertTo` + `action` prop + the avatarInputs helpers). if this lands, #234 should be closed or retargeted, otherwise we end up with two Revert buttons fighting over the same modal.

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -168,13 +168,17 @@ function CharacterEditDialogBody({
...
+									update({ appearance: String(attributes.appearance) })
+								}
```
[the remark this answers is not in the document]
umairnadeem: violates fail loudly

PR #262, opened 2026-06-12
fix: nightly bug fixes 2026-06-12

umairnadeem:
## Summary
- Fixes blank or whitespace-only audio `volume` attributes being coerced by `Number()` to `0`, which muted the element instead of using the default volume.
- Keeps explicit `"0"` behavior unchanged so intentional mute still works.
- Adds regression coverage for blank and whitespace volume values.

## Test Plan
- [x] `npm test -- --run lib/video/__tests__/elementAttributes.test.ts --passWithNoTests`
- [x] `npm test -- --passWithNoTests`
- [x] `npm run build`
- [x] `npm run lint && npm run format:check && npm run typecheck && npm run knip`

## Open PR overlap check
Considered open PRs:
- #236 touches API auth/route-handler/render/upload/validate-code files: non-overlapping.
- #231 touches Cartesia TTS/provider/embed/package files: non-overlapping.

This PR only touches `lib/video/elementAttributes.ts` and `lib/video/__tests__/elementAttributes.test.ts`, so it does not overlap with the open PR file/theme lists.

## Notes
- Claude Code YOLO core pass was attempted first per runbook but timed out without leaving a diff; Hermes completed the scoped bug fix and then used Claude Code for a tiny diff review (`LGTM`).

### DIFF
--- a/lib/video/elementAttributes.ts
+++ b/lib/video/elementAttributes.ts
@@ -19,7 +19,8 @@ const DEFAULT_VOLUME = VOLUME_MAX;
 
 /** Volume coerced to a finite number clamped to [0, 10], defaulting to 10. */
 export function getVolume(element: CanvasContentElement): number {
-	const raw = Number(element.customAttributes?.volume);
+	const value = element.customAttributes?.volume;
+	const raw = value?.trim() ? Number(value) : DEFAULT_VOLUME;
 	return Number.isFinite(raw)
 		? Math.max(VOLUME_MIN, Math.min(VOLUME_MAX, raw))
 		: DEFAULT_VOLUME;
--- a/lib/video/__tests__/elementAttributes.test.ts
+++ b/lib/video/__tests__/elementAttributes.test.ts
@@ -18,8 +18,10 @@ function el(customAttributes?: Record<string, string>): CanvasContentElement {
 }
 
 describe("getVolume", () => {
-	it("defaults to 10 when missing or non-numeric", () => {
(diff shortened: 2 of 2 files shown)


lib/video/elementAttributes.ts
```diff
@@ -19,7 +19,8 @@ const DEFAULT_VOLUME = VOLUME_MAX;
 
 /** Volume coerced to a finite number clamped to [0, 10], defaulting to 10. */
 export function getVolume(element: CanvasContentElement): number {
-	const raw = Number(element.customAttributes?.volume);
+	const value = element.customAttributes?.volume;
+	const raw = value?.trim() ? Number(value) : DEFAULT_VOLUME;
```
umairnadeem, line 23: good catch on the mute footgun (`Number("")` is 0, finite, clamps to volume 0 instead of defaulting to 10). fwiw `getLoops` right below has the same latent bug: blank coerces to 0 and only lands on the right answer because its min clamp (1) happens to equal its default. works today by coincidence, breaks the day someone copies the pattern for an attribute where default != min. maybe pull the blank-aware coercion into a shared helper (`coerceClamped(value, {min, max, fallback})`) so both getters go through it? fine as a fast-follow, not blocking.

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). correct fix and the blank/whitespace test cases are exactly the right coverage. one non-blocking thought on fixing the class of bug rather than the instance.

umairnadeem:
useless

PR #263, opened 2026-06-12
refactor: nightly maintainability 2026-06-12

umairnadeem:
## Summary
- Extract shared glass dropdown surface/item helpers with selected-state check column support
- Reuse the helpers across UI dropdowns in the canvas, character fields, transition panel, sortable insert menu, and copilot attach menu
- Switch duplicated dropdown selections to idiomatic Radix `onSelect` handlers while preserving existing prevent-close behavior for multi-select character picks

## Architecture changes
- Centralizes glass menu chrome in `GlassDropdownContent` and `GlassDropdownItem`, giving dropdown call sites a narrower styling/composition interface.
- Keeps per-site sizing, z-index, scrolling, and intentional visual deviations as explicit class overrides.

## Maintainability changes
- Removes duplicate menu-content and selected-check markup from seven UI modules.
- Deletes the local `EnumOption` implementation in favor of the shared glass item component.
- Keeps behavior and visuals unchanged; notably preserves the enum menu's existing `shadow-black/40` override.

## Complexity delta
- 7 files touched.
- 5 duplicate selected-option implementations collapsed into one shared item helper.
- 1 local component removed (`EnumOption`).
- Net diff: 127 insertions / 113 deletions, with duplicated branching/check rendering centralized.

## Validation
- [x] `npm run build`
- [x] `npm test -- --passWithNoTests`
- [x] CI parity: `npm run lint && npm run format:check && npm run typecheck && npm run knip && npm run test:coverage`

## Open PR overlap check
Considered open PRs:
- #262 touches `lib/video/__tests__/elementAttributes.test.ts`, `lib/video/elementAttributes.ts`
- #236 touches `ARCHITECTURE.md`, render/upload/validate API routes, and `lib/api/*`
- #231 touches Cartesia TTS, embeddings, voice similarity, and package files

This PR only changes UI dropdown components under `app/components/*`; it does not overlap those files, modules, or workflows.

## Skipped / notes
- Left `UserProfile`'s account menu alone because it uses separate card-style chrome, not this glass dropdown pattern.
- Did not normalize the enum dropdown `shadow-black/40` to the default `/8`, because that could be a visual behavior change.

PR #264, opened 2026-06-12
perf: nightly performance optimizations 2026-06-12

umairnadeem:
## Summary
- Moves the refine prompt's transient input state into a small `RefineComposer` child component.
- Keeps keystrokes in the inline refine box from rerendering the full post-prompt shell, Slate editor setup, player-position consumers, and video controls.
- Preserves existing submit/stop/loading behavior and prompt clearing semantics.

## Open PR overlap check
Considered open PRs before and after the diff:
- #263 touches GlassDropdown, SortableActions, AttributeBadge, CharactersPicker, character fields, TransitionPanel, ComposerCopilot.
- #262 touches video element attribute parsing/tests.
- #236 touches API route wrappers/render/upload/validate-code routes.
- #231 touches Cartesia TTS/embedding/package files.

This PR only changes `app/components/PostPromptView.tsx`, so it avoids those files and themes.

## Skipped
- Reverted a broader DnD sortable-items stabilization idea because #263 already touches the canvas DnD area and the overlap guard favors a cleaner non-overlapping target.
- Skipped bundle/code-splitting churn and cosmetic cleanup; no material runtime UX win for this pass.

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage`

### DIFF
--- a/app/components/PostPromptView.tsx
+++ b/app/components/PostPromptView.tsx
@@ -42,6 +42,31 @@ import { VideoLayoutProvider } from "./video/VideoLayoutContext";
 import editorStyles from "./Editor.module.css";
 import genStyles from "./styles/gen-button.module.css";
 
+function RefineComposer({
+	onSubmit,
+	onStop,
+	loading,
+}: {
+	onSubmit: (prompt: string) => void;
+	onStop: () => void;
+	loading: boolean;
+}) {
+	const [value, setValue] = useState("");
+	return (
+		<InlineCopilot
+			value={value}
+			onValueChange={setValue}
+			onSubmit={() => {
+				onSubmit(value);
+				setValue("");
+			}}
+			onStop={onStop}
+			loading={loading}
+			placeholder="Refine your script…"
+		/>
+	);
+}
+
 function PostPromptViewInner() {
 	const { loading: scriptLoading, stopGeneration } = useScriptControl();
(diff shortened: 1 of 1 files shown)


app/components/PostPromptView.tsx
```diff
@@ -42,6 +42,31 @@ import { VideoLayoutProvider } from "./video/VideoLayoutContext";
 import editorStyles from "./Editor.module.css";
 import genStyles from "./styles/gen-button.module.css";
 
+function RefineComposer({
```
umairnadeem, line 46: the move-state-down isolation is right, but this wrapper only exists because InlineCopilot insists on a controlled `value`/`onValueChange`. it has exactly one consumer (this file, checked master), so IMO push the draft state into InlineCopilot itself: internal `useState`, `onSubmit(prompt: string)`, clear after submit. same keystroke isolation, and you get to delete RefineComposer plus the controlled API entirely. one less layer of indirection, and the controlled flexibility is YAGNI with a single call site. not blocking, fine as a fast-follow.

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). solid perf instinct, moving the draft state down kills the per-keystroke rerender of the whole post-prompt shell (slate setup, player panels, the works). one suggestion inline on where that state could live instead.

PR #265, opened 2026-06-12
refactor: rethink openslop architecture 2026-06-12

umairnadeem:
## App understanding

OpenSlop is an open-source AI video creator: users compose a script on a Slate-based canvas (scenes containing narration/character/image/video/sound/music elements), generate assets per element through pluggable AI providers (Anthropic, Runware, ElevenLabs, Cartesia), refine the script via LLM edit-ops, and render the result with Remotion. Persistence and auth run on Supabase. The intended layering is:

```
UI (app/) → hooks → generation queue → connectors → gateways → API routes (app/api/v1) → providers
```

## From-scratch architecture comparison

Rebuilt from scratch, the layering above would survive almost unchanged — it is genuinely good (queue decouples async polling from UI, connectors/gateways are intentionally layered, validation is centralized in zod route schemas). The one thing a from-scratch design would do differently: the canvas **domain model** (OSML serialization, node creation, editor ops, element-type configs) would live in the domain layer from day one, because project serialization (`lib/project`), script refinement (`lib/script`), and video resolution (`lib/video`) all depend on it. Today that logic lived under `app/components/canvas/`, forcing 5 upward `lib → app` imports — the only real inversion in the codebase.

1. **Moved canvas domain logic into `lib/canvas/`** (git renames, no behavior change): `osmlSerializer`, `editorOps`, `nodeUtils`, `createCanvasNode`, `insertElement`, `elementConfigs`, `guards`, `parseXmlTag`, `constants`, `hydrateConnectorConfig`, `useOSMLSerializer` — plus their 7 test files. All `lib → app` imports are now gone (`lib/project/serialize.ts`, `lib/script/refine/applyOps.ts`, `lib/script/ScriptProvider.tsx`, `lib/video/resolve.ts`).
2. **Enforced the boundary**: new ESLint `no-restricted-imports` rule errors when any `lib/**` file imports from `app/` (covers both `@/app/**` and relative escapes; smoke-tested both). One-line note added to CLAUDE.md.
3. **Deduplicated `ZERO_WIDTH_SPACE`** in `lib/generation/getPromptText.ts`, now importable from `lib/canvas/constants`.
4. **Compat shims** at `app/components/canvas/utils/editorOps.ts` and `app/components/canvas/config/elementConfigs.ts` re-export only `setNodeAttrs` / `AttributeSpec` so the two files open PR #263 touches (`AttributeBadge.tsx`, `CharactersPicker.tsx`) stay byte-identical. Delete the shims and inline the `@/lib/canvas/*` imports once #263 lands.

UI-only utilities (`scrollToScene`, `characters`, `buildGenerationJob`, `metadataTags`) intentionally stayed in `app/`.

## Validation

- `npm run lint` ✅ (boundary rule verified to fire on a synthetic violation, then removed)
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm run knip` ✅
- `npm run test:run` ✅ 818/818 tests, 94 files
- `npm run build` ✅

## Best-practice review

`/simplify` was run on the diff (4 parallel review agents); it produced the `ZERO_WIDTH_SPACE` dedup and the relative-import ESLint glob hardening. `/vercel-composition-patterns`, `/web-design-guidelines`, and `/vercel-react-best-practices` were not run because this diff introduces no new components, UI, or data-fetching — it is module moves + import rewrites.

## Skipped items / risks

- **ARCHITECTURE.md** mentions of these paths were left untouched — open PR #236 owns that file; update after it lands.
- `useOSMLSerializer` is a React hook now in `lib/` — consistent with existing lib infrastructure (`ScriptProvider`, `GenerationQueueProvider`, `useProject`), so not split further.
- `lib/templates/templates.ts` (843-line data blob) and extracting `ConfigProvider`'s plugin wiring were considered and skipped as lower-leverage churn for this PR.

## Open PR overlap check

Considered open PRs #264 (PostPromptView/refine composer), #263 (GlassDropdown + canvas dropdown themes), #262 (`lib/video/elementAttributes`), #236 (API auth/route wrappers, ARCHITECTURE.md), #231 (Cartesia TTS/embedding, package.json). Final diff re-checked by filename against every file those PRs touch: **zero overlap**. The two #263 files that import moved modules are kept unchanged via the compat shims described above.


lib/video/resolve.ts
```diff
@@ -2,7 +2,7 @@ import type { CanvasElement } from "@/lib/canvas/types";
 import { getContentElements } from "@/lib/canvas/scenes";
 import { getPrimaryUrl } from "@/lib/connectors/assetUrl";
 import type { ElementSnapshot } from "@/lib/generation/queue";
-import { ELEMENT_CONFIGS } from "@/app/components/canvas/config/elementConfigs";
+import { ELEMENT_CONFIGS } from "@/lib/canvas/elementConfigs";
```
umairnadeem, line 5: now that elementConfigs is officially domain-layer, worth being honest about what moved with it: `icon: React.ReactNode`, `bgColor`, and the lucide-react import all live in lib/ now, and this render-path file pulls them in transitively even though it only reads `outputKind`. the purist fix is splitting presentational fields into an app-side lookup keyed by type, but that's two registries that can drift, which IMO is worse than the impurity. fine as is, just flagging it as the known loose end of the lib-is-domain story so nobody 'fixes' it later without weighing the split-brain cost.

app/components/canvas/utils/editorOps.ts
```diff
@@ -1,62 +1,3 @@
...
-			delete merged[key];
-		} else {
-			merged[key] = value;
-		}
-	}
-	Transforms.setNodes(editor, { customAttributes: merged }, { at: path });
-}
+// Compat re-export: keeps files touched by open PR #263 unchanged.
+// Fold into direct @/lib/canvas/editorOps imports once it lands.
+export { setNodeAttrs } from "@/lib/canvas/editorOps";
```
umairnadeem: verified the shim coverage: AttributeBadge and CharactersPicker are the only files left on the old paths (both touched by #263, no file overlap between the two PRs), and they only need `setNodeAttrs` plus the `AttributeSpec` type, so these two shims are exactly sufficient. nice conflict dodge. one ask: nothing fails if these linger after #263 lands, and new code could quietly start importing the shim path. maybe drop a note on #263 itself so whoever merges it remembers the 2-file codemod + shim delete.

umairnadeem (commented):
agent self-review pass (same account as author, fwiw). this is the right call and it's executed carefully: the lib -> app inversion was the only real layering violation in the codebase, the move is pure git renames with no behavior change, and the eslint no-restricted-imports rule means the boundary can't silently regress, which is what separates this from doc-only architecture PRs. verified the #263 compat story holds (details inline). two non-blocking thoughts.

PR #273, opened 2026-06-13
Format My Slop timestamps on the server

umairnadeem:
## Summary
- Fix React #418 hydration error on the My Slop list: the "Edited X ago" label was computing `toLocaleDateString()`/`Date.now()` during SSR, so a UTC server and a non-UTC/non-en-US browser produced different text. It only started firing once the first projects aged past 30 days and hit the locale-dependent branch.
- Compute the label once in the server component (`app/page.tsx`) and pass it to `ProjectsList` as a string, so the browser never recomputes it and there's nothing to mismatch.
- Move the formatter to `lib/project/relativeTime.ts` (server-only); delete the client `RelativeTime` component and its `useSyncExternalStore` hydration gate.
- Trade-off: labels are now a server snapshot, so they don't tick live, and >30-day dates render in en-US/UTC rather than each user's locale.

## Test plan
- [x] `npm run lint && npm run format:check && npm run typecheck && npm run knip` — clean
- [x] `npm run test:run` — 831 passing, including a new `lib/project/__tests__/relativeTime.test.ts`
- [x] `relativeTime` test passes under `TZ=Asia/Tokyo LC_ALL=en_GB.UTF-8` (env-independent output)
- [ ] Load the My Slop list logged in with a >30-day-old project; no #418 in the console


umairnadeem (commented):
clean approach, moving the computation server-side removes the `useSyncExternalStore` machinery entirely and the tests are solid. two small things.

app/components/projects/ProjectsList.tsx
```diff
@@ -8,12 +8,13 @@ import { createProject, deleteProject } from "@/lib/project/api";
 import type { ProjectRow } from "@/lib/project/api";
 import { ImageWithShimmer } from "@/lib/components/ImageWithShimmer";
 import UserProfile from "@/app/components/UserProfile";
-import { RelativeTime } from "./RelativeTime";
+
+export type ProjectListItem = ProjectRow & { editedLabel: string };
```
umairnadeem: nit: `lib/project/types.ts` already exists and is the natural home for data-shape types. exporting this from the UI component flips the dependency direction — `app/page.tsx` (a server component) ends up reaching *into* a client component file for a type, which is a bit backwards. move it to `lib/project/types.ts` or even inline it next to `relativeTime.ts`?

lib/project/relativeTime.ts
```diff
@@ -0,0 +1,12 @@
+// Server only to prevent hydration mismatches
```
umairnadeem: the comment documents intent but doesn't enforce it — if someone imports this in a client component down the road, the server-computed label and the client-computed label will silently differ and you're back to hydration mismatches. the next.js canonical fix is `import "server-only"` at the top of the file (one line, caught at build time). it's not installed here so either add the package or drop the comment and rely on code review. a comment that implies a constraint without enforcing it is a footgun.

PR #284, opened 2026-06-15
perf: nightly performance optimizations 2026-06-15

umairnadeem:
## Summary
- Skips Slate `set_selection`-only operations in the canvas `onChange` handler.
- Keeps React's mirrored document value focused on AST/content edits instead of cursor movement.
- Reduces editor-side rerenders/effect churn during clicks, arrow-key movement, and selection changes without changing document behavior.

## Why this matters
Slate stores selection on the editor; pure selection operations do not change the document tree. Before this pass, cursor movement still called `setValue(next)`, causing the canvas tree and document-dependent effects to churn for non-content changes. This applies Slate's recommended performance pattern so autosave/script sync/video layout consumers only see actual AST changes.

## Open PR overlap check
Considered current open PRs:
- #283 copilot input composition/state ownership (`PostPromptView`, copilot components)
- #282 story-mode LLM prompt correctness
- #280 canvas/generation helper relocation and character-name helpers
- #279 dependency security (`package.json`, `package-lock.json`)
- #278 required context consolidation
- #277 video asset prefetch readiness

This PR touches only `app/components/canvas/Canvas.tsx`, so it has no file overlap and a distinct Slate runtime-performance theme.

## Skipped
- Remotion `MotionLayer` transform memoization: skipped because frame changes every render frame, so caching would add indirection without a material win.
- Captions arithmetic memoization: skipped as marginal; the existing calculations are trivial and already behind stable inputs where meaningful.
- Bundle/dynamic-import churn: skipped because the runbook prioritizes runtime interaction responsiveness over initial-load micro-optimizations.

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm test -- --passWithNoTests` — 95 files / 836 tests passed
- `npm run test:coverage` — 95 files / 836 tests passed
- Claude best-practice review (`/simplify`, `/vercel-composition-patterns`, `/web-design-guidelines`, `/vercel-react-best-practices` lenses): PASS, no blockers

PR #293, opened 2026-06-17
perf: nightly performance optimizations 2026-06-17

umairnadeem:
## Summary
- Reduces per-frame work in the segmented video seek bar by replacing per-segment repeated prefix summing with a single linear prefix pass.
- Memoizes `SegmentedSeekBar`'s scrub segment projection so playback frame ticks do not allocate a new segment array unless scene segments or total duration change.

## Why this matters
`SegmentedSeekBar` updates with `usePlayerFrame`, so this path runs during playback and scrubbing. The change removes avoidable O(n²) segment math and per-frame segment object allocation on longer multi-scene videos while preserving identical seek/hover behavior.

## Open PR overlap check
Considered current open PRs and avoided all overlapping files/themes:
- #292 editor toolbar/post prompt/canvas provider refactor
- #291 character name deduplication
- #285 canvas media-result color/design-system work

Changed files are limited to `app/components/video/ScrubBar.tsx` and `app/components/video/SegmentedSeekBar.tsx`, with no overlap.

## Skipped
- Did not touch Slate selection-only guards or single-callback/state-setter micro-optimizations, per PR #284 feedback.
- Did not memoize frame-dependent Remotion transforms because `frame` is a true input.
- Did not hand-roll existing lodash/binary-search helpers or pursue bundle/TTFB-only churn.

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` (94 files / 842 tests passed)
- `npm test -- --passWithNoTests` (94 files / 842 tests passed)

### DIFF
--- a/app/components/video/SegmentedSeekBar.tsx
+++ b/app/components/video/SegmentedSeekBar.tsx
@@ -6,7 +6,7 @@ import type { VideoLayout } from "@/lib/video/types";
 import { usePlayerFrame } from "./usePlayerState";
 import { SeekTooltip } from "./SeekTooltip";
 import { findSegmentIndexAt, type SceneSegment } from "./useSceneSegments";
-import { ScrubBar, type ScrubHover, type ScrubSegment } from "./ScrubBar";
+import { ScrubBar, type ScrubHover } from "./ScrubBar";
 
 const clamp = (n: number, min: number, max: number) =>
 	Math.max(min, Math.min(max, n));
@@ -49,7 +49,7 @@ export function SegmentedSeekBar({
 
 	if (segments.length === 0) return null;
 
-	const scrubSegments: ScrubSegment[] = segments.map((seg) => ({
+	const scrubSegments = segments.map((seg) => ({
 		id: seg.sceneId,
 		basis: seg.duration / totalDurationSec,
 	}));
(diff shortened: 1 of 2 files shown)


app/components/video/SegmentedSeekBar.tsx
```diff
@@ -47,12 +47,16 @@ export function SegmentedSeekBar({
 		[],
 	);
 
-	if (segments.length === 0) return null;
+	const scrubSegments = useMemo<ScrubSegment[]>(
```
umairnadeem: afaict this memo is a ~no-op in practice. this component re-renders every frame via `usePlayerFrame(player)` up top, and `ScrubBar` is a plain `export function` (not `React.memo`), so it re-renders every frame regardless of whether `scrubSegments` keeps a stable ref. all the useMemo saves is one `.map` over the scene count (a handful), which is negligible.

same story for the ScrubBar prefix-sum change: correct, but n is tiny so the O(n^2)->O(n) gain is noise.

if the goal is actually fewer renders here, the lever is wrapping `ScrubBar` in `React.memo` + stabilizing the callbacks passed to it, then the stable `scrubSegments` ref pays off. as-is this is the kind of trivial-gain perf change thats easy to close as no-op. up to you.

umairnadeem (commented):
nice that you moved the early return below the new useMemo so Rules of Hooks stays happy. one concern on whether these actually buy anything perf-wise, inline.

PR #300, opened 2026-06-18
fix: cap toast height and make long messages scrollable

umairnadeem:
## What

Top-level toasts now cap their content area at `40vh` and scroll vertically when the message is longer. Previously a long toast message could overflow off-screen and become unreadable.

## How

Applied a `content` class via sonner's `toastOptions` in `AppToaster`. Scoping to the `content` slot (rather than the whole toast) keeps the icon and close button fixed while only the message scrolls.

```tsx
toastOptions={{
  classNames: { content: "max-h-[40vh] overflow-y-auto" },
}}
```

### DIFF
--- a/app/components/AppToaster.tsx
+++ b/app/components/AppToaster.tsx
@@ -10,6 +10,9 @@ export function AppToaster() {
 			theme={resolvedTheme === "dark" ? "dark" : "light"}
 			position="bottom-center"
 			richColors
+			toastOptions={{
+				classNames: { content: "max-h-[40vh] overflow-y-auto" },
+			}}
 		/>
 	);
 }


app/components/AppToaster.tsx
```diff
@@ -10,6 +10,9 @@ export function AppToaster() {
 			theme={resolvedTheme === "dark" ? "dark" : "light"}
 			position="bottom-center"
 			richColors
+			toastOptions={{
+				classNames: { content: "max-h-[40vh] overflow-y-auto" },
```
umairnadeem, line 14: fwiw the gemini bot is wrong here, `content` is a valid key on `ToastClassnames` as of sonner 2.0.7 (`content?: string` is right there in the dist types), and `tsc --noEmit` passes clean. so no compile error, and this typed approach is actually nicer than the raw `[data-content]` css the bot suggested since it targets the exact same node without leaking a global selector. lgtm as-is.

umairnadeem (commented):
clean little fix, lgtm. one note inline pushing back on the gemini bot's blocking claim so you don't chase a non-bug.

PR #301, opened 2026-06-18
fix: forward Cartesia TTS emotion

umairnadeem:
## Summary
- Forward supported TTS emotion controls to Cartesia via `generation_config.emotion`
- Keep Cartesia request construction typed against the SDK `GenerationRequest`
- Add regression coverage for emotion propagation

## Validation
- `npm test -- lib/providers/__tests__/cartesia-tts.test.ts` — 21 tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm test -- --passWithNoTests` — 94 files / 837 tests passed
- `npm run test:coverage` — 94 files / 837 tests passed; coverage report generated
- `npm run build`
- `npm run knip`

### DIFF
--- a/lib/providers/tts/cartesia.ts
+++ b/lib/providers/tts/cartesia.ts
@@ -205,6 +205,7 @@ export class CartesiaTTS extends BaseProvider<TTSGenerateParams, RawTTSResult> {
 				generation_config: {
 					speed: CARTESIA_SPEED[params.speed ?? "medium"],
 					volume: CARTESIA_VOLUME,
+					emotion: params.emotion || "neutral",
 				},
 			};
 			for await (const response of ws.generate(req)) {
--- a/app/api/v1/tts/route.ts
+++ b/app/api/v1/tts/route.ts
@@ -8,6 +8,7 @@ const schema = bodySchema(TTS_MODELS, {
 	voiceId: requiredVoiceId,
 	speed: z.enum(TTS_SPEEDS).optional(),
 	volume: z.number().optional(),
+	emotion: z.string().optional(),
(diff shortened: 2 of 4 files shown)


lib/providers/tts/cartesia.ts
```diff
@@ -38,6 +38,16 @@ const CARTESIA_SPEED: Record<TTSSpeed, number> = {
...
 };
 
+type CartesiaEmotion = NonNullable<
+	GenerationRequest["generation_config"]
+>["emotion"];
+
+function toCartesiaEmotion(
+	emotion: TTSGenerateParams["emotion"],
+): CartesiaEmotion | undefined {
+	return emotion === "joking/comedic" ? undefined : emotion;
```
umairnadeem: nit: this `"joking/comedic"` is the one `TTSEmotion` value cartesia's union doesn't include, but that why is totally opaque to a reader. two small things: use `TTSEmotion.Joking` instead of the raw string (params.emotion is `TTSEmotion`, so it reads as a real enum member and won't quietly stop matching if that string value is ever renamed), and maybe a one-liner like `// cartesia's emotion union omits joking/comedic`. otherwise someone hits this in 6mo with no idea why this single emotion gets dropped. not blocking.

umairnadeem (commented):
nice clean way to thread emotion through, and the type-driven approach is solid: `CartesiaEmotion` derives straight from the SDK union so if cartesia's supported set ever drifts from `TTSEmotion`, the else-branch return stops compiling and forces you to handle it. one small note inline on the magic string.

PR #303, opened 2026-06-18
refactor: rethink openslop architecture 2026-06-18

umairnadeem:
## App understanding

OpenSlop is a free, open-source AI video creator: a Slate-based storyboard editor where a prompt is streamed (OSML over SSE) into typed elements (narration, character, image, animated_image, clip, music, sound), each generated through a pluggable AI pipeline, then composed and rendered to MP4 via Remotion Lambda. Next.js 16 App Router + React 19, Supabase (auth/Postgres/RLS), Vercel Blob + Queue, Zustand for project state.

Main workflows: submit prompt → stream OSML into the canvas; generate stale elements (client queue → `/api/v1/{asset}` → jobs row → Vercel Queue worker → provider → Blob → poll); save/restore project snapshots; render the composition.

Architectural boundaries: the generation pipeline is a deliberate three-layer split — **connectors** (`lib/connectors/`, editor-facing, plugin-pipelined), **gateways** (`lib/gateway/`, thin HTTP clients to `/api/v1/*`), **providers** (`lib/providers/`, server-side vendor adapters). `lib/` is the domain layer and never imports from `app/`.

## From-scratch architecture vs. current

A clean-room design would keep this exact three-layer split (it is what lets providers/asset-types be swapped independently) and would have **one uniform way for a connector to obtain its gateway**, plus **one helper for the recurring "prepend a base system prompt" operation** in the LLM plugins. The current code was 90% there but had two pockets of drift:

1. **Gateway wiring diverged between connector families.** `BaseLLMConnector` already injected its gateway through the constructor, but every asset connector independently declared a `protected gateway` field and a constructor that newed it up from `config.baseUrl`. Same idea, two shapes.
2. **The same system-prompt merge was re-implemented four times** across `story-mode`, `script-mode`, `template-mode`, and `project-metadata` (two via `lodash.compact`).

## Implemented simplifications

- **Unified gateway construction.** `BaseAssetConnector` is now generic over its gateway type (default `AssetGateway<TParams>`, mirroring `BaseLLMConnector`) and receives the gateway as a constructor parameter property. The six OpenSlop asset connectors (image, animated_image, music, sfx, video, tts) drop their duplicated field + assignment and just pass the gateway up via `super(new XGateway(config.baseUrl), config)`. `BaseTTSConnector` threads the concrete gateway type through so `OpenSlopTTS` keeps its `searchVoices` typing without an unsafe field redeclaration. Both connector families now share one wiring convention.
[...]

PR #302, opened 2026-06-18
refactor: nightly maintainability 2026-06-18

umairnadeem:
## Summary
- Finished the `SimpleTooltip` migration for the remaining exact simple tooltip call sites.
- Replaced repeated raw `Tooltip` / `TooltipTrigger asChild` / `TooltipContent` boilerplate in editor and canvas controls with the existing wrapper.
- Preserved trigger elements, labels, disabled states, click handlers, and tooltip text.

## Architecture / maintainability changes
- Consolidates simple tooltip usage behind the narrow `SimpleTooltip` interface.
- Removes the remaining duplicate raw primitive pattern where no custom tooltip behavior was needed.
- Keeps raw tooltip primitives available only for provider/root wiring and future customized tooltip cases.

## Complexity delta
- 3 files changed.
- 21 net lines removed: 40 insertions / 61 deletions.
- Simplified three UI control render paths by replacing repeated wrapper structure with a single composition boundary.

## Validation
- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm run knip` ✅
- `npm run build` ✅
- `npm test -- --passWithNoTests` ✅ (94 files / 836 tests)
- `npm run test:coverage` ✅ (94 files / 836 tests; all files 77.04% statements)
- Independent review: passed; no behavior, accessibility, provider-ordering, or import-boundary issues found.

## Open PR overlap check
Considered open PRs:
- #301 `fix: forward Cartesia TTS emotion` touches Cartesia TTS provider/tests.
- #300 `fix: cap toast height and make long messages scrollable` touches `AppToaster` and toast scrolling.

This PR touches only `EditorToolbar`, `AudioPlayer`, and `CharacterEditModal` tooltip composition, so it does not overlap by file or theme.

## Skipped items
- Did not touch `proxy.ts` or third-party `<head>` scripts.
- Did not modify open-PR areas: Cartesia TTS emotion forwarding or toast height/scrolling.
- Did not add docs because the change is self-explanatory and docs would be low-signal.

PR #304, opened 2026-06-18
test: add playwright smoke tests 2026-06-18

umairnadeem:
## Summary
- Add a minimal Chromium-only Playwright smoke-test setup for the public unauthenticated onboarding surface.
- Add `npm run test:e2e`, Playwright artifact ignores, and a Vitest exclude so unit coverage does not pick up e2e specs.
- Wire smoke tests into CI after build/test with Chromium browser installation.

## Smoke flows covered
- `/` renders the public beta onboarding page and links to `/login`.
- `/login` renders the email magic-link login form.
- `/signup` renders the signup form with name and email fields.

## CI wiring
- CI installs the Chromium Playwright browser via `npx playwright install --with-deps chromium`.
- CI runs `npm run test:e2e` after the existing build and coverage steps, with the same public Supabase env used by build.
- Playwright starts the built Next app in CI via `npm run start -- -p <port>` and reuses a local server outside CI.

## Validation
- `npx playwright install chromium` — passed (browser already/fallback installed locally; host reports unsupported-OS fallback warning).
- `PLAYWRIGHT_PORT=3001 npm run test:e2e` — passed, 3/3 smoke tests. Used 3001 locally because port 3000 is occupied by this machine's Hermes WhatsApp bridge; CI/default remains 3000.
- `npm run lint` — passed.
- `npm run knip` — passed.
- `npm run build` — passed.
- `npm run test:coverage` — passed, 94 files / 836 tests.
- `npm test -- --passWithNoTests` — passed, 94 files / 836 tests.

## Logged-in-route skip rationale
- Logged-in/editor/project smoke tests are intentionally skipped because auth currently uses email-link/Gmail flows, not user/password auth, and the runbook explicitly scoped authenticated smoke tests out.

## Open PR overlap check
Considered currently open PRs and confirmed no file/theme overlap:
- #303 architecture/connectors/LLM prompt helper files — not touched.
- #302 tooltip composition in `EditorToolbar`, `AudioPlayer`, `CharacterEditModal` — not touched.
- #301 Cartesia TTS provider/tests — not touched.
- #300 toast scrolling in `AppToaster` — not touched.

## Follow-up risks / skipped items
- Kept smoke scope intentionally light and public-only; no screenshots, visual snapshots, sleeps, third-party service workflows, or logged-in routes.
- The `/` page includes server-side Supabase auth lookup before rendering unauthenticated onboarding; CI already provides the public Supabase env used by build, but that route remains the one smoke path most sensitive to Supabase availability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,17 @@ jobs:
       - name: Tests
         run: npm run test:coverage
 
+      - name: Resolve Playwright version
+        id: playwright-version
+        run: echo "version=$(npm ls @playwright/test --json | jq -r '.dependencies["@playwright/test"].version')" >> "$GITHUB_OUTPUT"
+
+      - name: Cache Playwright browsers
+        id: playwright-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ steps.playwright-version.outputs.version }}
+
       - name: Install Playwright browser
         run: npx playwright install --with-deps chromium
 
@@ -52,6 +63,16 @@ jobs:
         env:
           NEXT_PUBLIC_SUPABASE_URL: ${{ vars.NEXT_PUBLIC_SUPABASE_URL }}
           NEXT_PUBLIC_SUPABASE_ANON_KEY: ${{ vars.NEXT_PUBLIC_SUPABASE_ANON_KEY }}
+
+      - name: Upload Playwright report
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report
+          path: |
+            playwright-report/
+            test-results/
+          retention-days: 7
 
       - name: Bundle size report
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,15 @@ jobs:
       - name: Tests
         run: npm run test:coverage
 
+      - name: Install Playwright browser
+        run: npx playwright install --with-deps chromium
+
+      - name: Smoke tests
+        run: npm run test:e2e
+        env:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -44,6 +44,15 @@ jobs:
       - name: Tests
         run: npm run test:coverage
 
+      - name: Install Playwright browser
+        run: npx playwright install --with-deps chromium
+
+      - name: Smoke tests
+        run: npm run test:e2e
```
umairnadeem, line 51: the `trace: "on-first-retry"` you set in playwright.config only pays off if you can open the trace after a red run, but there's no `actions/upload-artifact` step here and `playwright-report/` + `test-results/` are gitignored. so when a smoke test fails in CI the trace/report gets torn down with the runner and you're debugging blind from just the `list` reporter output.

perhaps add an `if: ${{ failure() }}` upload-artifact step for those two dirs? non-blocking but it's the difference between a 2min repro and a head-scratcher next time one flakes.

.github/workflows/ci.yml
```diff
@@ -44,6 +44,15 @@ jobs:
       - name: Tests
         run: npm run test:coverage
 
+      - name: Install Playwright browser
+        run: npx playwright install --with-deps chromium
```
umairnadeem, line 48: nit: this re-downloads chromium on every CI run since there's no cache. caching `~/.cache/ms-playwright` keyed on the resolved playwright version would shave it off most runs. totally optional / fast-follow.

umairnadeem (commented):
nice first slice of e2e coverage. role/label-based queries are exactly right (resilient to markup churn), the vitest `exclude: e2e/**` keeps the two runners from colliding, and `forbidOnly`/`retries` are gated on CI correctly. verified all the selectors against the actual pages and they match. two non-blocking notes on the CI wiring below.

PR #309, opened 2026-06-19
refactor: share dropdown components between composer and canvas

umairnadeem:
## What

The composer was using a one-off `GlassDropdown`, while the canvas hand-rolled the same dropdown patterns in several places. This consolidates them into two shared, composition-based primitives built on the existing `DropdownMenu` primitive:

- **`SelectMenu`** — value + options list with a checkmark on the selected row. Used by the composer's mode / aspect-ratio / template selectors and the canvas `AttributeBadge` enum menu.
- **`ActionMenu`** — icon + label + `onSelect` action list. Used by the composer's `AttachMenu` and the canvas `SortableActions` insert menu.

Both take their trigger as `children` via Radix `asChild` (composition over config), and triggers use the **React 19 `ref`-as-prop** idiom instead of `forwardRef`.

`GlassDropdown` is deleted.

Also: give `PrePromptView` the same `dot-grid-bg` background as `PostPromptView`.

## Notes

- Dropdown surfaces now use the shared `DropdownMenu` default (`bg-popover` / `shadow-elevation-5`). `--card` and `--popover` are the same token, so the only visible shift is a slightly different shadow, now consistent across all menus.
- The canvas `EnumField` (nullable value + bordered field trigger) was intentionally left as-is — its semantics differ enough that folding it in would muddy the `SelectMenu` API.

## Verification

`lint`, `format:check`, `typecheck`, `knip`, and tests all pass.


components/ui/select-menu.tsx
```diff
@@ -0,0 +1,62 @@
...
+			<DropdownMenuTrigger asChild>{children}</DropdownMenuTrigger>
+			<DropdownMenuContent
+				side={side}
+				align={align}
+				className={cn("min-w-32", contentClassName)}
+			>
+				{options.map((option) => (
+					<DropdownMenuItem
+						key={option.value}
+						onClick={() => onChange(option.value)}
```
umairnadeem: nit: sibling `ActionMenu` uses `onSelect` here but `SelectMenu` uses `onClick`. both fire on keyboard activation since radix dispatches a click on Enter/Space, so not a bug, but `onSelect` is the canonical menu-item handler (and gives you preventDefault to keep the menu open if you ever need it). worth making the two new helpers consistent.

components/ui/select-menu.tsx
```diff
@@ -0,0 +1,62 @@
...
+						className="cursor-pointer py-1 text-[11px] text-muted-foreground"
+					>
+						<span className="flex w-3.5 shrink-0 items-center justify-center">
+							{option.value === value && (
+								<Check className="h-3 w-3 text-foreground" aria-hidden />
```
umairnadeem: nit: this unifies the selected check to `text-foreground`, but the old `AttributeBadge` used `text-accent` for its check (only GlassDropdown used foreground). so attribute badges quietly lose the violet check. totally fine if intended, just flagging the drift since it's easy to miss.

umairnadeem (commented):
clean DRY win. ActionMenu/SelectMenu collapse the 4 copy-pasted dropdown call sites nicely, and leaning on the base DropdownMenuContent defaults holds up: bg-popover == bg-card in both themes, shadow-elevation-5 and px-2 py-1.5 match the old overrides, so no visual drift. bonus, PillTrigger adds type="button" which the old GlassDropdown trigger was missing (latent form-submit footgun). couple tiny nits inline. only other thing: the dot-grid-bg add to PrePromptView is unrelated to the dropdown work, harmless but scope creep if you're keeping PRs tight.

PR #305, opened 2026-06-18
fix: preserve project after reload undo

umairnadeem:
## Summary
- Prevent project rehydration from recording Slate load operations in undo history.
- Extract the rehydrate operation into a focused helper so it can be covered without mounting the editor.
- Add regression coverage that a loaded project keeps the undo stack empty.

## Selected issue
Selected #266 because it is a small, clearly scoped correctness bug: after project reload, Ctrl+Z can undo the initial load and leave an empty project. The relevant code path was isolated to `useProjectRehydrate`, and the fix can be validated with a focused Slate history regression test.

## Validation
- `npm test -- app/components/canvas/hooks/useProjectRehydrate.test.ts --run` — passed (RED first: failed with missing `rehydrateProjectEditor`, then passed after implementation).
- `npm run lint` — passed.
- `npm run format:check` — passed.
- `npm run typecheck` — passed.
- `npm run knip` — passed.
- `npm run build` — passed.
- `npm run test:coverage` — passed, 95 files / 837 tests.
- `npm test -- --passWithNoTests` — passed, 95 files / 837 tests.

## Open PR overlap check
Checked currently open PRs before editing and again before staging:
- #304 touches CI, Playwright e2e setup, package files, and Vitest config — no overlap.
- #303 touches connector/gateway/LLM plugin architecture — no overlap.
- #302 touches tooltip composition in editor/canvas UI components — no overlap.
- #301 touches Cartesia TTS provider/tests — no overlap.
- #300 touches `AppToaster` toast scrolling — no overlap.

This PR only touches `app/components/canvas/hooks/useProjectRehydrate.ts` and its new focused test.

## Skipped issue categories
Skipped broader/product-context-heavy items such as avatar/style generation policy, post-generation upload UX, render lifecycle robustness, video export/upscaling, i18n, and major feature/control requests.

Closes #266

PR #312, opened 2026-06-19
fix: nightly bug fixes 2026-06-19

umairnadeem:
## Summary
- Add strict route-boundary validation for `/api/v1/tts/voices` search parameters.
- Reject malformed voice search query values before calling the TTS provider instead of silently dropping invalid filters.
- Keep permissive metadata parsing separate from API request validation.

## Bugs fixed
- `limit=abc` (and other malformed positive-integer limits) previously parsed to `undefined`, causing the route to return an unbounded/default voice search instead of a client error.
- Invalid voice search trait values now fail request validation at the route boundary rather than being silently ignored.

## Tests
- Updated `GET /api/v1/tts/voices` coverage to assert valid `limit` values are parsed and forwarded as numbers.
- Added regression coverage that invalid voice search parameters return `400` and do not call the provider.

## Validation
- `npm test -- app/api/v1/__tests__/routes.test.ts` ✅
- `npm run build` ✅
- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm run knip` ✅
- `npm run test:coverage` ✅

## Open PR overlap check
Considered open PRs #309, #305, #304, #303, and #302. This PR changes only:
- `app/api/v1/__tests__/routes.test.ts`
- `app/api/v1/tts/voices/route.ts`
- `lib/project/types.ts`

These files and the voice-search route-boundary validation theme do not overlap with the open dropdown, project rehydrate, Playwright smoke, connector architecture, or tooltip-maintainability PRs.

## Notes
- Claude Code was invoked in the required YOLO print-mode style for the core pass, but it timed out without leaving a diff; Hermes completed the focused fallback fix and independently validated it.

### DIFF
--- a/lib/project/types.ts
+++ b/lib/project/types.ts
@@ -10,6 +10,12 @@ import type { AspectRatio } from "@/lib/video/aspectRatio";
 import type { TransitionType } from "@/lib/video/transitions";
 
 const optionalString = z.string().min(1).optional().catch(undefined);
+const voiceSearchString = z.string().min(1).optional();
+const voiceSearchLimitSchema = z
+	.string()
+	.regex(/^[1-9]\d*$/, "limit must be a positive integer")
+	.transform(Number)
+	.optional();
 
 export const genderSchema = z.enum(TTS_GENDERS).optional().catch(undefined);
 export const languageSchema = z.enum(TTS_LANGUAGES).optional().catch(undefined);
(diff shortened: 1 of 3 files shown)


lib/project/types.ts
```diff
@@ -31,10 +37,19 @@ export const MetadataVoiceSchema = voiceTraitsSchema.extend({
 	resolvedVoiceId: optionalString,
 });
 
-export const voiceSearchParamsSchema = voiceTraitsSchema.extend({
-	query: optionalString,
-	name: optionalString,
-	limit: z.coerce.number().int().positive().optional().catch(undefined),
+const voiceSearchTraitsSchema = z.object({
```
umairnadeem, line 40: split-brain: `voiceSearchTraitsSchema` re-declares the exact same 6 trait fields as `voiceTraitsSchema` above, the only diff is strict vs `.catch(undefined)`. add a 7th trait enum later (emotion, whatever) and you have to remember to touch both objects or they silently drift. could derive both from one base so there's a single source of truth, e.g. keep a strict trait object as canonical and build the permissive metadata variant off it. not blocking but it's the kind of dupe that quietly rots.

lib/project/types.ts
```diff
@@ -10,6 +10,12 @@ import type { AspectRatio } from "@/lib/video/aspectRatio";
 import type { TransitionType } from "@/lib/video/transitions";
 
 const optionalString = z.string().min(1).optional().catch(undefined);
+const voiceSearchString = z.string().min(1).optional();
```
umairnadeem, line 13: footgun: `voiceSearchString` is `min(1)` with no `.catch`, so `?query=` (empty string, a super common shape when a client serializes an unset filter) now 400s instead of being treated as 'no query filter'. old schema caught it -> undefined. same hits `name`/`description`. if empty-means-absent is the intended UX, `.transform(s => s || undefined)` (or dropping min(1) for these search fields) keeps the old behavior while still rejecting genuinely bad input. up to you whether that empty-param shape is real for this API.

umairnadeem (commented):
tightening the route boundary is the right move. swapping the silent `.catch(undefined)` parse for `safeParse` + `badRequest` so `limit=abc` returns a 400 instead of a default unbounded search is a real fix, and the regression test asserting the provider isn't called is exactly what i'd want. two things worth a look below (one maintainability, one edge case), neither blocking.

umairnadeem:
useless

PR #315, opened 2026-06-19
fix: include reference images and all character avatars as style refs

### LINKED ISSUE #298: Fall back to uploaded character avatars for art-style description when no art-style references exist
## Problem

Uploaded character avatars never feed into the generated art-style description. The art-style block only comes from project-level reference images, so a user who uploads character references but no art-style references gets no art-style guidance at all.

- `reference-style.ts` reads project-level `referenceImages` and bails when empty: `if (referenceImages.length === 0) return prompt;` (`lib/connectors/llm/plugins/reference-style.ts:17-19`). Only when art-style refs exist does it derive an `Art style reference: ...` block.
- The separate `character-avatar-style.ts` plugin does look at uploaded character avatars (`avatarUploaded && avatarUrl`), but it describes each character's *appearance* for consistency ("preserve these appearance descriptions exactly"), explicitly per-character and "Do not describe the background" (`lib/connectors/llm/plugins/character-avatar-style.ts:19-43`). It is not an overall art-style description.

Net effect: upload character refs only, and the art style is never inferred from them.

## Suggested direction

When no art-style reference images are uploaded but uploaded character avatars exist, have `reference-style.ts` fall back to those character avatars as the source for the art-style description.

- In `reference-style.ts`, if `referenceImages` is empty, collect uploaded character avatar URLs (`metadata.characters[*].avatarUploaded && avatarUrl`) and, if any, pass them as `referenceImages` to the existing `gateway.generate` art-style describe call.
- Prefer uploaded avatars over generated ones: generated avatars were themselves produced from the style, so using them would be circular. Uploaded avatars are the real user-provided signal.
- Keep the project-level reference images as the primary source when present; the character-avatar path is a fallback only.

## Relevant code

- `lib/connectors/llm/plugins/reference-style.ts:17-29` -- art-style description, gated on project-level `referenceImages`
- `lib/connectors/llm/plugins/character-avatar-style.ts:17-22` -- existing pattern for reading uploaded character avatars (`avatarUploaded && avatarUrl`)
- `lib/project/types.ts:42-45` -- `MetadataCharacter` (`avatarUrl`, `avatarUploaded`)
- `lib/config/ConfigProvider.tsx:121-125` -- plugin chain order (`reference-style` then `character-avatar-style`)

## Related

- #243 (character/avatar generation correctness) -- covers the inverse direction (character avatars should respect art-style refs); this is art style inferred from character avatars.
- #248 (reusable character library + per-character style)

---

umairnadeem:
## Summary
- Feed the LLM reference-style plugin **uploaded** style sources: project reference images **and uploaded** character avatars.
- Previously uploaded character avatars were only a fallback used when no project reference images existed. Now they are always combined with project references into a single style-reference set.
- **Generated** character avatars are excluded: they were themselves produced from the style, so describing style off them feeds the model's own output back as ground truth (circular → style drift). See #298.

## Selected issue and scope
Selected #298 because it is a small, well-scoped correctness bug with clear acceptance criteria and an isolated code path (`lib/connectors/llm/plugins/reference-style.ts`). The fix is limited to style-reference input selection and focused regression tests.

## Behavior
- Style references = `referenceImages` + every uploaded `character.avatarUrl` (`avatarUploaded && avatarUrl`), deduped of empties via `lodash/compact`.
- Uploaded character avatars contribute even when no project-level references exist; generated avatars never do.
- Empty set (no reference images and no uploaded avatars) still short-circuits and returns the prompt unchanged.

## Related fix
- `character-avatar-style.ts` now preserves **all** character avatars (uploaded and generated), not just uploaded ones. A pregenerated avatar would otherwise be ignored by the script LLM, which would invent a new appearance and leave the existing avatar stale. Generated avatars reuse their stored appearance text; uploaded ones are described from the image. (This is the appearance-preservation path, distinct from art-style extraction above, so including generated avatars here is not circular.)

## Tests
- Reference images combine with uploaded avatars but exclude generated ones, in order.
- Uploaded character avatars are used as style references when no project reference images are present (the original #298 path).
- Generated-only avatars short-circuit (no gateway call), guarding against circular references.
- Character-avatar-style coverage: passthrough with no avatars, uploaded described from image, generated avatars reuse appearance text without a gateway call, mixed generated + uploaded, and the gateway-required guard.

## Validation
- `npm test -- lib/connectors --run` ✅
- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅

Closes #298

### DIFF
--- a/lib/connectors/llm/plugins/reference-style.ts
+++ b/lib/connectors/llm/plugins/reference-style.ts
@@ -19,8 +19,10 @@ export function createReferenceStylePlugin(projectId: string): LLMPlugin {
 				getProjectStore(projectId).getState();
 			const styleReferenceImages = compact([
 				...referenceImages,
-				...Object.values(metadata.characters).map(
-					(character) => character.avatarUrl,
+				...Object.values(metadata.characters).flatMap((character) =>
+					character.avatarUploaded && character.avatarUrl
+						? [character.avatarUrl]
+						: [],
 				),
 			]);
 			if (styleReferenceImages.length === 0) return prompt;
--- a/lib/connectors/llm/plugins/reference-style.ts
+++ b/lib/connectors/llm/plugins/reference-style.ts
@@ -19,10 +19,8 @@ export function createReferenceStylePlugin(projectId: string): LLMPlugin {
 				getProjectStore(projectId).getState();
 			const styleReferenceImages = compact([
 				...referenceImages,
-				...Object.values(metadata.characters).flatMap((character) =>
-					character.avatarUploaded && character.avatarUrl
-						? [character.avatarUrl]
-						: [],
+				...Object.values(metadata.characters).map((character) =>
+					character.avatarUploaded ? character.avatarUrl : undefined,
 				),
 			]);
 			if (styleReferenceImages.length === 0) return prompt;
--- a/lib/connectors/llm/plugins/reference-style.ts
+++ b/lib/connectors/llm/plugins/reference-style.ts
@@ -1,4 +1,5 @@
 import dedent from "dedent";
+import compact from "lodash/compact";
 import { getProjectStore } from "@/lib/project/store";
 import type {
 	LLMGenerateParams,
@@ -14,14 +15,20 @@ export function createReferenceStylePlugin(projectId: string): LLMPlugin {
 			prompt: string,
 			ctx?: PluginContext<LLMGenerateParams, LLMGenerateResult>,
 		) {
-			const referenceImages =
-				getProjectStore(projectId).getState().referenceImages;
-			if (referenceImages.length === 0) return prompt;
+			const { metadata, referenceImages } =
+				getProjectStore(projectId).getState();
+			const styleReferenceImages = compact([
+				...referenceImages,
+				...Object.values(metadata.characters).map(
+					(character) => character.avatarUrl,
+				),
+			]);
+			if (styleReferenceImages.length === 0) return prompt;
 			if (!ctx?.gateway)
 				throw new Error("reference-style plugin requires gateway context");
 			const { text: style } = await ctx.gateway.generate({
 				prompt: dedent`Vividly and concisely describe the visual art style of the attached reference image(s) in 1–2 concise sentences. Include ultra specific detail on character art style and overall art style.`,
-				referenceImages,
+				referenceImages: styleReferenceImages,
 				maxTokens: 4096,
 			});
 			return dedent`Art style reference: ${style}
--- a/lib/connectors/__tests__/character-avatar-style.test.ts
+++ b/lib/connectors/__tests__/character-avatar-style.test.ts
@@ -1,0 +1,117 @@
+import { describe, expect, it, vi } from "vitest";
+import { GatewayClient } from "@/lib/gateway/base";
+import { createCharacterAvatarStylePlugin } from "@/lib/connectors/llm/plugins/character-avatar-style";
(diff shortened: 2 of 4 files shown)


umairnadeem (commented):
clean, well-scoped fix that matches the issue's acceptance criteria, and nice that the regression test pins both the fallback and the generated-avatar exclusion. one DRY nit below.

lib/connectors/llm/plugins/reference-style.ts
```diff
@@ -14,14 +14,22 @@ export function createReferenceStylePlugin(projectId: string): LLMPlugin {
...
-			const referenceImages =
-				getProjectStore(projectId).getState().referenceImages;
-			if (referenceImages.length === 0) return prompt;
+			const { metadata, referenceImages } =
+				getProjectStore(projectId).getState();
+			const styleReferenceImages =
+				referenceImages.length > 0
+					? referenceImages
+					: Object.values(metadata.characters).flatMap((character) =>
+							character.avatarUploaded && character.avatarUrl
```
umairnadeem: the `avatarUploaded && avatarUrl` -> avatar url predicate now lives here AND in `character-avatar-style.ts:20`. two plugins encoding the same "what counts as an uploaded avatar" rule = drift risk (e.g. if you later add a url-validity check or rename the field, easy to update one and miss the other). perhaps pull a tiny `uploadedCharacterAvatars(projectId)` helper and share it?

also minor: the sibling guards `metadata.characters ?? {}` but this trusts it's defined. type says it's required and the store inits `{}`, so probably fine, but the two disagreeing is a small smell. a shared helper would settle it either way.

lib/connectors/llm/plugins/reference-style.ts
```diff
@@ -14,14 +14,22 @@ export function createReferenceStylePlugin(projectId: string): LLMPlugin {
...
+								? [character.avatarUrl]
+								: [],
+						);
```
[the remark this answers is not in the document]
umairnadeem: no

lib/connectors/llm/plugins/reference-style.ts
```diff
@@ -14,14 +15,20 @@ export function createReferenceStylePlugin(projectId: string): LLMPlugin {
...
 		) {
-			const referenceImages =
-				getProjectStore(projectId).getState().referenceImages;
-			if (referenceImages.length === 0) return prompt;
+			const { metadata, referenceImages } =
+				getProjectStore(projectId).getState();
+			const styleReferenceImages = compact([
+				...referenceImages,
+				...Object.values(metadata.characters).map(
+					(character) => character.avatarUrl,
```
umairnadeem: this now folds *generated* avatars into the style reference (no `avatarUploaded` gate), but #298 explicitly called that out as circular: generated avatars were themselves produced from the style, so describing style off them feeds the model's own output back in as ground truth -> style drift over iterations. the sibling `character-avatar-style.ts:20` still gates on `avatarUploaded && avatarUrl` for exactly this reason, so the two plugins now disagree on what counts as an authoritative source (split-brain).

was pulling in generated avatars here intentional? if yes, worth a line in the PR/issue on why circularity isn't a concern anymore (since you're closing #298 which says the opposite). if not, filter to `avatarUploaded && avatarUrl` like the sibling does and it'd settle the inconsistency too.

umairnadeem (commented):
scope expanded since my last pass, one new thing worth a look re: including generated avatars.

PR #318, opened 2026-06-20
Add export popover with resolution and background renders

umairnadeem:
## Summary
- Replace the player's hover-only render button with an **Export** button in the editor toolbar that opens a bottom-right popover.
- Popover lets you pick output resolution (480p–4K, default 1080p) and shows live render progress; the footer Export button morphs into a progress bar, then into a success state with file size, duration, and a Download button.
- Closing the popover mid-render hands off to a persistent bottom-right Sonner toast (progress + View to reopen, then download when ready) so renders continue in the background — never cancelled.
- Resolution maps to a new optional `scale` arg on `/render` (`renderMediaOnLambda`); composition base stays 1080p, so `scale = targetHeight / 1080`. `BASE_HEIGHT` is now the single source of truth in `lib/video/types.ts`.
- Render + popover state lives in a new `RenderProvider` (mounted in `CanvasProviders`) shared by the button and the toast.
- Updated the project loading skeleton to include the new Export button placeholder.

## Test plan
- [ ] Generate a script + elements until the player is ready; confirm the **Export** button appears right of "Generate All" and is disabled until ready.
- [ ] Open the popover (bottom-right), pick a resolution, click Export; verify progress advances and the spinner shows.
- [ ] Close the popover mid-render → persistent toast shows progress with **View**; click View to reopen — render is still going.
- [ ] On completion, verify the popover/toast show size + Download, and the downloaded file matches the chosen resolution (e.g. 4K → 3840×2160).
- [ ] Toggle light/dark mode and confirm the Export trigger and popover colors look correct.
- [ ] Confirm the old hover button no longer appears over the player.


app/components/video/ExportButton.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { Spinner } from "@/components/ui/spinner";
+import { formatBytes } from "@/lib/format";
+import { useScriptControl } from "@/lib/script/ScriptProvider";
+import { RESOLUTIONS, scaleForHeight } from "@/lib/video/resolutions";
+import { formatDuration } from "@/lib/video/timestamps";
+import { BASE_HEIGHT } from "@/lib/video/types";
+import { useRender } from "./RenderProvider";
+import { useLayout } from "./VideoLayoutContext";
+
+const triggerClass =
```
umairnadeem, line 29: these raw hexes go against DESIGN.md ("Use semantic tokens ... never raw hexes"), and you're already setting `variant="secondary"` then clobbering it with a bespoke pink ramp that maps to no token. that's drift waiting to happen, light/dark gets hand-maintained here forever instead of flipping with the ramps. can this just be `variant="secondary"` (or `accent`), or if export genuinely needs its own tint, add semantic tokens to globals.css and reference those?

lib/video/resolutions.ts
```diff
@@ -0,0 +1,11 @@
+import { BASE_HEIGHT } from "./types";
+
+export const RESOLUTIONS = [
+	{ label: "480p", height: 480 },
```
umairnadeem: 480p footgun: width = 1920 * (480/1080) = 853.33, which isn't even an integer, and h264 rejects odd dimensions, so this resolution may just fail the lambda render. 720/1440/2160 all land on even dims, 480 is the odd one out (literally). test plan only covers 4K afaict, worth actually exporting a 480p to confirm it doesn't blow up. easy fix is to key off explicit even w/h per entry instead of deriving from a fractional scale.

umairnadeem (commented):
nice refactor, moving render out of the player hover-button into a real Export control + the close-popover-hands-off-to-toast flow is a clean way to keep renders alive in the bg. `BASE_HEIGHT` as the single source of truth for the scale math is the right call too. couple things below, mainly the raw hexes and the 480p scale math.

lib/format.ts
```diff
@@ -0,0 +1,5 @@
+export function formatBytes(bytes: number): string {
+	if (bytes < 1024) return `${bytes} B`;
+	if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`;
+	return `${(bytes / (1024 * 1024)).toFixed(1)} MB`;
```
umairnadeem: nit: now that 4K is a headline option, renders can easily clear 1GB and this caps at MB, so you'll show stuff like `1843.0 MB`. since you're centralizing the helper anyway, maybe add a GB tier.

lib/video/resolutions.ts
```diff
@@ -0,0 +1,11 @@
+import { BASE_WIDTH } from "./types";
+
+export const RESOLUTIONS = [
+	{ label: "480p", width: 854, height: 480 },
```
umairnadeem, line 4: nit: `height` here is dead data, only `width` + `label` are ever read and `scaleForWidth` derives the scale from width alone. so width and height have to be kept in 16:9 sync by hand but only width actually drives the output, classic split-brain waiting to drift. either drop the height column, or compute it (`Math.round(width * BASE_HEIGHT / BASE_WIDTH)`) so it can't desync. up to you.

app/components/video/ExportButton.tsx
```diff
@@ -0,0 +1,171 @@
...
+							<span className="text-muted-foreground">Size</span>
+							<span className="font-mono tabular-nums">
+								{formatBytes(state.size)}
+							</span>
+						</div>
+						{layout && (
+							<div className="flex items-center justify-between text-xs">
+								<span className="text-muted-foreground">Duration</span>
+								<span className="font-mono tabular-nums">
+									{formatDuration(layout.totalDurationSec)}
```
umairnadeem: this reads `layout.totalDurationSec` live, but the whole point of the bg-render flow is you close the popover and keep editing. so if the script changes while a render is in flight, the done state shows the *current* duration, not the one that was actually rendered (size is correct since it comes from the render result). minor, but worth snapshotting duration at render() time alongside the layout if you want it accurate.

umairnadeem (commented):
follow-up after the address-review commit. the raw hexes, 480p dims, GB tier, in-popover disable and toast-unmount are all sorted, lgtm on those. two small new things below, neither blocking.

PR #322, opened 2026-06-20
Overlay polish + typography token & prop-driven sizing sweep

umairnadeem:
## Summary

Two layers of design-system polish on this branch:

### Overlay radius & elevation
- Overlays — modals, popovers, dropdowns, selects (and the select trigger) — share an 8px `rounded-md` corner instead of the larger `rounded-xl`/`rounded-2xl`/`rounded-3xl` mix.
- Popovers and menus drop from `shadow-elevation-5` to `shadow-elevation-3`; the base popover is borderless (shadow alone separates it).
- Export popover uses `bg-surface-elevated` to read slightly lighter than the page; settings popover keeps its hairline border.

### Typography tokens & sizing discipline
- Replace raw `text-[Npx]`/`text-xs`/`text-sm` arbitraries with named `@theme` font-size tokens (`text-badge`/`text-label`/`text-body`/…) across components and the shared `ui` primitives.
- Register the custom font-size tokens with `tailwind-merge` (`extendTailwindMerge`) so `cn()` no longer strips them when a class string also sets a text color. *(This was the root cause of sidebar/tooltip/dropdown labels rendering oversized.)*
- Apply `font-body` (Inter) to every editable input and the project title; point `body` / `.font-title` / `.font-body` at the next/font variables directly, since the `@theme inline` aliases aren't emitted as CSS variables.
- Size Select triggers and toolbar buttons via the `size` prop instead of hand-sizing with `h-*`/`text-*` in `className` (fixes the layout-panel dropdown label).
- Absorb the fixed positioning into `UserProfile` (all four callers wrapped it identically).
- Share one `GenerationStatus` type between `ElementSnapshot` and `GenerateButton`; replace the nested-ternary label with a small helper.
- DESIGN.md documents the type scale and button-sizing conventions.

## Test plan
- [ ] Overlays (export/settings popovers, Select, dropdown, Dialog/AlertDialog) — 8px corners, lighter shadow; settings keeps hairline border.
- [ ] Sidebar rail, tooltips, and menu items render at their intended small sizes (no oversized labels).
- [ ] Editable fields (script editor, character/attribute fields, copilot prompts, project title) render in Inter.
- [ ] Layout & Properties panel dropdowns show their selected label at the small size.
- [ ] Account avatar stays fixed top-left across pre-prompt, post-prompt, and projects views.
- [ ] Verify light + dark themes.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -28,6 +28,8 @@ One clean grotesque for UI and body: **Inter** (`--font-sans`). **IBM Plex Mono*
 
 The scale lives as `@theme` tokens in `app/globals.css`: use the `text-badge-xs` / `text-badge` / `text-label-xs` / `text-label` / `text-body` / `text-body-lg` / `text-heading-sm` / `text-heading` / `text-heading-lg` / `text-display` utilities — named font-sizes only (leading inherits; pair with `leading-*` / `font-medium` where a tighter line or heavier weight is wanted) — rather than hand-picking raw `text-xs` / `text-sm` or `text-[Npx]` arbitraries.
 
+In a dense row, a primary label and its secondary description share one size and weight (e.g. both `text-label`), differentiated by color alone — `text-foreground` for the label, `text-muted-foreground` for the description. Don't shrink or bold the label to set it apart.
+
 ## Spacing, radius, elevation
 
 4px base spacing. Radius via `--radius` (`0.625rem`); cards `rounded-xl`, overlays (modals, popovers, dropdowns, selects) and form controls (inputs, textareas, select triggers) `rounded-md`, pills `rounded-full`. Depth comes from token elevations `shadow-elevation-1/3/5/10`, not glow — popovers and menus sit at `shadow-elevation-3`.
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -24,7 +24,7 @@ Use semantic tokens (`bg-background`, `text-foreground`, `border-border`, `bg-ca
 
 ## Typography
 
-One clean grotesque for UI and body: **Inter** (`--font-sans`). **IBM Plex Mono** (`--font-mono`) for timecodes and durations. Scales are as follows: label 11–12px, body 14–16px, headings 18/24/36/48px. `.font-title` = Inter 600 with tight tracking; `.font-body` = Inter 400.
+UI and titles use **Sloptastic** (`--font-sans`, `.font-title`), a clean grotesque; long-form body text uses **Inter** (`--font-body`). **IBM Plex Mono** (`--font-mono`) for timecodes and durations. Two display faces for marketing surfaces only: **Instrument Serif** (`--font-serif`, hero headlines) and **Sentient** (`--font-sentient`, the onboarding wordmark). Scales are as follows: label 11–12px, body 14–16px, headings 18/24/36/48px. `.font-title` = Sloptastic 575 with tight tracking; `.font-body` = Inter 425.
 
 The scale lives as `@theme` tokens in `app/globals.css`: use the `text-badge-xs` / `text-badge` / `text-label-xs` / `text-label` / `text-body` / `text-body-lg` / `text-heading-sm` / `text-heading` / `text-heading-lg` / `text-display` utilities — named font-sizes only (leading inherits; pair with `leading-*` / `font-medium` where a tighter line or heavier weight is wanted) — rather than hand-picking raw `text-xs` / `text-sm` or `text-[Npx]` arbitraries.
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -28,7 +28,7 @@ One clean grotesque for UI and body: **Inter** (`--font-sans`). **IBM Plex Mono*
 
 ## Spacing, radius, elevation
 
-4px base spacing. Radius via `--radius` (`0.625rem`); cards `rounded-xl`, modals `rounded-3xl`, pills `rounded-full`. Depth comes from token elevations `shadow-elevation-1/3/5/10`, not glow.
+4px base spacing. Radius via `--radius` (`0.625rem`); cards `rounded-xl`, overlays (modals, popovers, dropdowns, selects) and form controls (inputs, textareas, select triggers) `rounded-md`, pills `rounded-full`. Depth comes from token elevations `shadow-elevation-1/3/5/10`, not glow — popovers and menus sit at `shadow-elevation-3`.
 
 ## Motion
 
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -26,6 +26,8 @@ Use semantic tokens (`bg-background`, `text-foreground`, `border-border`, `bg-ca
 
 One clean grotesque for UI and body: **Inter** (`--font-sans`). **IBM Plex Mono** (`--font-mono`) for timecodes and durations. Scales are as follows: label 11–12px, body 14–16px, headings 18/24/36/48px. `.font-title` = Inter 600 with tight tracking; `.font-body` = Inter 400.
 
+The scale lives as `@theme` tokens in `app/globals.css`: use the `text-badge-xs` / `text-badge` / `text-label-xs` / `text-label` / `text-body` / `text-body-lg` / `text-heading-sm` / `text-heading` / `text-heading-lg` / `text-display` utilities — named font-sizes only (leading inherits; pair with `leading-*` / `font-medium` where a tighter line or heavier weight is wanted) — rather than hand-picking raw `text-xs` / `text-sm` or `text-[Npx]` arbitraries.
+
 ## Spacing, radius, elevation
 
 4px base spacing. Radius via `--radius` (`0.625rem`); cards `rounded-xl`, overlays (modals, popovers, dropdowns, selects) and form controls (inputs, textareas, select triggers) `rounded-md`, pills `rounded-full`. Depth comes from token elevations `shadow-elevation-1/3/5/10`, not glow — popovers and menus sit at `shadow-elevation-3`.
@@ -41,3 +43,5 @@ One opt-in treatment: subtle film grain (`.grain`, theme-tuned `--grain-opacity`
 ## Components
 
 Idiomatic, shadcn-consistent primitives in `components/ui/` built with `class-variance-authority`, tokens only. Buttons: `default` (near-black inverse, for in-tool actions), `accent` (blurple, for hero/auth CTAs), plus `secondary`/`outline`/`ghost`/`icon`/`destructive`. Icons: `lucide-react`, thin (`strokeWidth` ≈1.5).
+
+Button **dimensions** come from the `size` prop (`sm`/`default`/`lg`/`icon`) defined in `buttonVariants` — never hand-size a button with `h-*`/`px-*`/`text-*` in `className`. Add a named `size` to the variant if you need a new one.
--- a/app/components/canvas/ProjectTitle.tsx
+++ b/app/components/canvas/ProjectTitle.tsx
@@ -47,7 +47,7 @@ export function ProjectTitle() {
 					if (e.key === "Enter") commit();
 					else if (e.key === "Escape") setEditing(false);
 				}}
-				className="mb-3 w-full bg-transparent font-body text-2xl font-semibold text-foreground outline-none"
+				className="mb-3 w-full bg-transparent font-body text-heading font-semibold text-foreground outline-none"
 				aria-label="Project title"
 			/>
 		);
@@ -55,7 +55,7 @@ export function ProjectTitle() {
 
 	return (
 		<div className="group mb-3 flex items-center gap-2">
-			<h1 className="font-body text-2xl font-semibold text-foreground">
+			<h1 className="font-body text-heading font-semibold text-foreground">
 				{title}
 			</h1>
 			<button
--- a/app/components/canvas/ProjectTitle.tsx
+++ b/app/components/canvas/ProjectTitle.tsx
@@ -47,7 +47,7 @@ export function ProjectTitle() {
 					if (e.key === "Enter") commit();
 					else if (e.key === "Escape") setEditing(false);
 				}}
-				className="mb-3 w-full bg-transparent text-2xl font-semibold text-foreground outline-none"
+				className="mb-3 w-full bg-transparent font-body text-2xl font-semibold text-foreground outline-none"
 				aria-label="Project title"
 			/>
 		);
@@ -55,7 +55,9 @@ export function ProjectTitle() {
 
 	return (
 		<div className="group mb-3 flex items-center gap-2">
-			<h1 className="text-2xl font-semibold text-foreground">{title}</h1>
+			<h1 className="font-body text-2xl font-semibold text-foreground">
+				{title}
+			</h1>
 			<button
 				type="button"
 				onClick={startEditing}
--- a/app/components/canvas/elements/GenerateButton.tsx
+++ b/app/components/canvas/elements/GenerateButton.tsx
@@ -1,15 +1,17 @@
 "use client";
 
-import { AlertCircle, Sparkles } from "@/components/ui/icon";
+import { AlertCircle, Hourglass, Sparkles } from "@/components/ui/icon";
 import { SimpleTooltip } from "@/components/ui/tooltip";
+import { Button } from "@/components/ui/button";
+import { Spinner } from "@/components/ui/spinner";
 import type { CanvasContentElement } from "@/lib/canvas/types";
-import { cn } from "@/lib/utils";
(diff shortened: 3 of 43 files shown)


components/ui/select.tsx
```diff
@@ -35,7 +35,7 @@ function SelectTrigger({
 		<SelectPrimitive.Trigger
 			data-slot="select-trigger"
 			className={cn(
-				"flex h-9 w-fit items-center justify-between gap-2 rounded-lg border border-border bg-input px-3 py-2 text-sm whitespace-nowrap outline-none transition-colors focus-visible:border-ring focus-visible:ring-2 focus-visible:ring-ring/50 disabled:cursor-not-allowed disabled:opacity-50 data-[placeholder]:text-muted-foreground [&_svg]:pointer-events-none [&_svg:not([class*='size-'])]:size-4 [&_svg]:shrink-0",
+				"flex h-9 w-fit items-center justify-between gap-2 rounded-md border border-border bg-input px-3 py-2 text-sm whitespace-nowrap outline-none transition-colors focus-visible:border-ring focus-visible:ring-2 focus-visible:ring-ring/50 disabled:cursor-not-allowed disabled:opacity-50 data-[placeholder]:text-muted-foreground [&_svg]:pointer-events-none [&_svg:not([class*='size-'])]:size-4 [&_svg]:shrink-0",
```
umairnadeem: the SelectTrigger is an input control, not an overlay, but it went `rounded-lg` -> `rounded-md` here while `Input` and `Textarea` are still `rounded-lg`. so a select sitting next to a text input in the same form now has mismatched corners, and DESIGN.md only documents the overlay radius (nothing about the trigger). intended to diverge from the other inputs, or should input/textarea move to `rounded-md` too so the form controls stay consistent? up to you, just don't love the split.

components/ui/popover.tsx
```diff
@@ -36,7 +36,7 @@ function PopoverContent({
 				align={align}
 				sideOffset={sideOffset}
 				className={cn(
-					"z-50 min-w-48 origin-(--radix-popover-content-transform-origin) rounded-xl border border-border bg-popover p-3 text-popover-foreground shadow-elevation-5 outline-none data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
+					"z-50 min-w-48 origin-(--radix-popover-content-transform-origin) rounded-md bg-popover p-3 text-popover-foreground shadow-elevation-3 outline-none data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
```
umairnadeem, line 39: dropping the border from the base affects all 3 consumers, not just the two you touched: ExportButton got `bg-surface-elevated` and ElementSettings got `border border-border` back, but `TextAttributePopover` got neither so it silently goes borderless-default (just shadow-elevation-3 over `bg-popover`). is the bare borderless look intended there too? per DESIGN.md borderless is the blessed default so probably fine, just flagging it's the one consumer that wasn't explicitly reviewed for the new look.

umairnadeem (commented):
nice tightening, the borderless-popover-relies-on-shadow direction reads clean and DESIGN.md got updated alongside (no doc drift). two things worth a look before merge, both non-blocking.

components/ui/dropdown-menu.tsx
```diff
@@ -42,7 +42,7 @@ function DropdownMenuContent({
 				data-slot="dropdown-menu-content"
 				sideOffset={sideOffset}
 				className={cn(
-					"z-50 max-h-(--radix-dropdown-menu-content-available-height) min-w-[8rem] origin-(--radix-dropdown-menu-content-transform-origin) overflow-x-hidden overflow-y-auto rounded-xl border bg-popover p-1 text-popover-foreground shadow-elevation-5 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
+					"z-50 max-h-(--radix-dropdown-menu-content-available-height) min-w-[8rem] origin-(--radix-dropdown-menu-content-transform-origin) overflow-x-hidden overflow-y-auto rounded-md border bg-popover p-1 text-popover-foreground shadow-elevation-3 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
```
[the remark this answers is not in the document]
umairnadeem: Done — added `border-border` here (and on SubContent) to match SelectContent/DialogContent.

components/ui/select.tsx
umairnadeem: Good catch on the form-control split. Rather than reverting the trigger back up to `rounded-lg`, I moved Input and Textarea down to `rounded-md` so all form controls stay consistent, and documented the form-control radius in DESIGN.md.

components/ui/popover.tsx
umairnadeem: Intended — borderless is the base-popover default, so TextAttributePopover taking just shadow-elevation-3 over bg-popover is the wanted look. Leaving it.

app/components/video/ExportButton.tsx
```diff
@@ -58,7 +58,7 @@ export function ExportButton() {
 				align="end"
 				side="top"
 				sideOffset={8}
-				className="w-80 p-4"
+				className="w-80 bg-surface-elevated p-4"
```
umairnadeem: heads up, `bg-surface-elevated` here is theme-dependent in a way that's easy to miss: in light mode `--surface-elevated` and `--popover` are both `#fff` so this override is a no-op, but in dark mode they diverge (`#372f32` vs `#251e21`). net effect is the export popover is the *only* popover lighter than the rest, and only in dark mode. either drop it back to the base `bg-popover` so it matches every other popover, or if export genuinely wants the elevated surface make it intentional in both themes. right now it reads like accidental drift more than a choice.

umairnadeem (commented):
one new thing after the latest round (form-control radius + border-border fixes all look good): the export popover's bg override behaves differently per theme, flagging below.

DESIGN.md
```diff
@@ -28,7 +28,7 @@ One clean grotesque for UI and body: **Inter** (`--font-sans`). **IBM Plex Mono*
 
 ## Spacing, radius, elevation
 
-4px base spacing. Radius via `--radius` (`0.625rem`); cards `rounded-xl`, modals `rounded-3xl`, pills `rounded-full`. Depth comes from token elevations `shadow-elevation-1/3/5/10`, not glow.
+4px base spacing. Radius via `--radius` (`0.625rem`); cards `rounded-xl`, overlays (modals, popovers, dropdowns, selects) and form controls (inputs, textareas, select triggers) `rounded-md`, pills `rounded-full`. Depth comes from token elevations `shadow-elevation-1/3/5/10`, not glow — popovers and menus sit at `shadow-elevation-3`.
```
umairnadeem, line 35: the policy now says overlays (dropdowns) = `rounded-md`, but `UserProfile`'s DropdownMenuContent still hardcodes `rounded-2xl` (app/components/UserProfile.tsx:63). so once this lands the account-menu dropdown is the one overlay that's way rounder than every select/popover/menu, i.e. the overlay migration is partial / split-brain with the doc. either drop that override so it inherits the new base, or if the profile menu genuinely wants the bigger radius call it out as an explicit exception here. (fwiw ElementContainer:90 + OnboardingCard are rounded-2xl/3xl vs the documented cards=`rounded-xl` too, but that's pre-existing and out of scope for this PR.)

umairnadeem (commented):
follow-up pass: just one new thing not covered in my earlier comments, the rest still lgtm.

lib/utils.ts
```diff
@@ -1,5 +1,32 @@
...
+// Register the custom @theme font-size tokens so tailwind-merge treats them as
+// font-sizes, not text-colors. Otherwise it strips e.g. `text-label-xs` when a
+// class string also sets a text color (`text-muted-foreground`), since it can't
+// tell the token apart from a color.
+const twMerge = extendTailwindMerge({
+	extend: {
+		classGroups: {
+			"font-size": [
+				{
+					text: [
```
umairnadeem, line 13: good catch registering these as font-sizes, the `text-label-xs` + `text-muted-foreground` strip would've been a nasty silent one. only thing: this list is now a second source of truth that has to mirror the `@theme --text-*` tokens in globals.css. add a `--text-foo` token over there and forget to add it here and `cn()` silently drops it whenever it sits next to a text color, i.e. the exact bug you just fixed, reappearing. worth a breadcrumb (`// keep in sync with --text-* in globals.css`) or deriving both from one shared const so they can't drift. non-blocking.

DESIGN.md
```diff
@@ -26,9 +26,13 @@ Use semantic tokens (`bg-background`, `text-foreground`, `border-border`, `bg-ca
 
 One clean grotesque for UI and body: **Inter** (`--font-sans`). **IBM Plex Mono** (`--font-mono`) for timecodes and durations. Scales are as follows: label 11–12px, body 14–16px, headings 18/24/36/48px. `.font-title` = Inter 600 with tight tracking; `.font-body` = Inter 400.
```
umairnadeem: while you're in the typography doc: this still says Inter is the UI/body font via `--font-sans`, but `--font-sans` resolves to `--font-sloptastic` now, so body + `.font-title` actually render Sloptastic. Inter is only `.font-body` these days. so the canonical doc names the wrong primary font (and the `.font-title = Inter 600` bit too). partly pre-existing, but this is the typography PR so feels like the spot to fix it. up to you.

umairnadeem (commented):
the typography token sweep reads clean and the tailwind-merge font-size registration is a genuinely good catch (that silent strip is easy to miss). two doc/maintainability split-brains from the latest round below, both non-blocking.

app/globals.css
```diff
@@ -351,7 +363,7 @@
 	}
 	body {
 		@apply bg-background text-foreground;
-		font-family: var(--font-sans);
+		font-family: var(--font-sloptastic), ui-sans-serif, system-ui, sans-serif;
```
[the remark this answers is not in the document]
umairnadeem: intentional. `--font-sloptastic` is a next/font local defined in layout.tsx so it resolves fine, and sloptastic is the UI/title font by design. the real mismatch was the doc, now fixed in DESIGN.md to name sloptastic (ui/title) + inter (body). not switching the font to inter.

app/globals.css
```diff
@@ -365,14 +377,14 @@
 }
 
 .font-title {
-	font-family: var(--font-sans);
+	font-family: var(--font-sloptastic), ui-sans-serif, system-ui, sans-serif;
```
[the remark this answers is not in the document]
umairnadeem: same as above. `.font-title` is sloptastic by design, so the doc got fixed rather than the code.

app/components/canvas/elements/character/VoicePicker.tsx
```diff
@@ -135,16 +135,16 @@ export function VoicePicker({
 								}`}
 							>
 								<div className="flex min-w-0 items-center gap-1.5">
-									<span className="min-w-0 flex-1 truncate text-xs font-medium text-foreground">
+									<span className="min-w-0 flex-1 truncate text-label text-foreground">
```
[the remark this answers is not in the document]
umairnadeem: intentional. the name matches descript label-1-regular here, which is regular weight, so name + description differ by color only. that dense-row pattern (color, not size/weight) is now documented in DESIGN.md. leaving regular weight.

app/components/canvas/elements/character/VoicePicker.tsx
```diff
@@ -154,7 +154,7 @@ export function VoicePicker({
 								</div>
 								{(voice.description || voice.previewUrl) && (
 									<div className="flex min-w-0 items-center justify-between gap-2">
-										<span className="min-w-0 flex-1 truncate text-[10px] leading-tight text-muted-foreground">
+										<span className="min-w-0 flex-1 truncate text-label text-muted-foreground">
```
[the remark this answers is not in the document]
umairnadeem: intentional 12px, not 10px. the description matches label-1-regular (12px) sitting next to the name, color-differentiated. keeping `text-label`.

app/components/copilot/InlineCopilot.tsx
```diff
@@ -87,7 +87,7 @@ export default function InlineCopilot({
 								onChange={(e) => onValueChange(e.target.value)}
 								onKeyDown={(e) => e.key === "Enter" && handleSubmit()}
 								placeholder={placeholderText}
-								className=" w-full bg-transparent text-sm text-foreground caret-accent placeholder:text-muted-foreground outline-none focus-visible:ring-2 focus-visible:ring-ring/30 focus-visible:rounded-sm"
+								className=" w-full bg-transparent font-body text-xs text-foreground caret-accent placeholder:text-muted-foreground outline-none focus-visible:ring-2 focus-visible:ring-ring/30 focus-visible:rounded-sm"
```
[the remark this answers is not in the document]
umairnadeem: took the `text-label` swap. skipped the `ring-accent-violet` part, that token does not exist in our theme (accent rings go through `ring-ring`), so it would just drop the ring.

app/components/canvas/ProjectTitle.tsx
```diff
@@ -47,15 +47,17 @@ export function ProjectTitle() {
 					if (e.key === "Enter") commit();
 					else if (e.key === "Escape") setEditing(false);
 				}}
-				className="mb-3 w-full bg-transparent text-2xl font-semibold text-foreground outline-none"
+				className="mb-3 w-full bg-transparent font-body text-2xl font-semibold text-foreground outline-none"
```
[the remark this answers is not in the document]
umairnadeem: done, `text-2xl` -> `text-heading`.

app/components/canvas/ProjectTitle.tsx
```diff
@@ -47,15 +47,17 @@ export function ProjectTitle() {
...
 				aria-label="Project title"
 			/>
 		);
 	}
 
 	return (
 		<div className="group mb-3 flex items-center gap-2">
-			<h1 className="text-2xl font-semibold text-foreground">{title}</h1>
+			<h1 className="font-body text-2xl font-semibold text-foreground">
```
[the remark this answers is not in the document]
umairnadeem: done, `text-2xl` -> `text-heading`.

app/components/canvas/panel/EditorSidebar.tsx
```diff
@@ -90,9 +90,9 @@ export function EditorSidebar() {
 
 	return (
 		<div className="flex shrink-0">
-			<nav className="flex w-14 shrink-0 flex-col items-center gap-1 pt-16 pr-0.5 mr-2 pl-1 lg:w-[72px] lg:pt-4">
+			<nav className="flex w-14 shrink-0 flex-col items-center gap-1 pt-16 mr-2 pl-1 lg:w-[72px] lg:pt-4">
```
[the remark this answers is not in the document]
umairnadeem: leaving as-is. could not reproduce a real overlap between the fixed avatar and the home rail item. if it collides on lg i will bump the padding in a follow-up.

app/components/canvas/elements/GenerateButton.tsx
```diff
@@ -1,49 +1,62 @@
...
+				<Spinner className="text-current" />
+			) : status === "queued" ? (
+				<Hourglass aria-hidden="true" />
+			) : (
+				<Sparkles aria-hidden="true" />
+			)}
+			{label}
+		</Button>
 	);
 }
```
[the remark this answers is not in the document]
umairnadeem: keeping the label always-on, it is the intended behavior from this sweep. if it overflows on compact cards i will add a `showLabel` prop in a follow-up.

app/components/video/ExportButton.tsx
umairnadeem: fixed. dropped `bg-surface-elevated` so the popover inherits `bg-popover` and matches every other popover in both themes.

DESIGN.md
umairnadeem: keeping the account menu at `rounded-2xl` as a deliberate exception, it is the one identity surface so the softer corner is wanted. the select/popover/menu overlays stay at the documented `rounded-md`.

DESIGN.md
umairnadeem: fixed. DESIGN.md now names sloptastic (ui/title) and inter (body), plus the two display faces (instrument serif, sentient), and the `.font-title` = inter 600 line is corrected.

PR #323, opened 2026-06-20
fix: nightly bug fixes 2026-06-20

umairnadeem:
## Summary
- Add strict route-boundary validation for `GET /api/v1/tts/voices` query parameters.
- Reject malformed voice search filters before invoking the TTS provider instead of silently dropping them via the permissive persisted-metadata parser.
- Remove the now-unused permissive voice search schema from `lib/project/types.ts` so API validation stays separate from metadata parsing.

## Bugs fixed
- `gender=robot`, `limit=abc`, and blank `limit` values previously parsed to `undefined` and still called the provider, returning a successful voice search with client filters silently ignored.
- Malformed public API requests now return `400` with actionable context and do not call `getTTSProvider().search()`.

## Tests added
- Added API route regressions covering malformed voice filters and blank numeric limits.
- Both tests assert the route returns `400` before the provider is called.

## Validation
- Verified RED first: `npm test -- --run app/api/v1/__tests__/routes.test.ts` failed on the two new regression tests before the fix (`200` instead of `400`).
- `npm test -- --run app/api/v1/__tests__/routes.test.ts` ✅
- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm run knip` ✅
- `npm run build` ✅
- `npm run test:coverage` ✅
- `npx playwright install --with-deps chromium` attempted, but this local cron environment cannot sudo-install OS deps (`sudo: a terminal is required`).
- `PLAYWRIGHT_PORT=3001 npm run test:e2e` ✅ (reused an already-running local Next dev server after default port 3000/alternate 3100 were occupied by an existing dev lock).

## Open PR overlap check
- Considered open PR #322 (`umair/overlay-radius-polish`): touches `DESIGN.md`, canvas/export UI, and `components/ui/*` overlay/input/select/textarea styling.
- This PR only touches `app/api/v1/tts/voices`, its API route tests, and `lib/project/types.ts`; no file or workflow overlap with #322.

## Skipped items / rationale
- Did not touch UI or visual files, so `DESIGN.md` decisions were not needed.
- Did not broaden this into a global validation refactor; the change is scoped to the real TTS voice-search route bug.

### DIFF
--- a/app/api/v1/tts/voices/route.ts
+++ b/app/api/v1/tts/voices/route.ts
@@ -1,14 +1,56 @@
 import { NextRequest, NextResponse } from "next/server";
+import { z } from "zod";
 import { getTTSProvider } from "@/lib/api/providers";
+import { badRequest } from "@/lib/api/response";
 import { withApiAccess } from "@/lib/api/with-auth";
-import { voiceSearchParamsSchema } from "@/lib/project/types";
+import {
+	TTS_ACCENTS,
+	TTS_AGES,
+	TTS_GENDERS,
+	TTS_LANGUAGES,
+	TTS_PITCHES,
+} from "@/lib/connectors/tts/enums";
+
+const nonBlankString = z.string().min(1);
+const queryLimitSchema = z
+	.union([z.number(), z.string()])
+	.transform((value) =>
+		typeof value === "string" && value.trim() === "" ? NaN : Number(value),
+	)
+	.refine((value) => Number.isFinite(value), {
+		message: "must be a finite number",
+	})
+	.refine((value) => Number.isInteger(value) && value > 0, {
+		message: "must be a positive integer",
+	})
+	.optional();
+
+const voiceSearchQuerySchema = z
+	.object({
+		gender: z.enum(TTS_GENDERS).optional(),
+		age: z.enum(TTS_AGES).optional(),
+		pitch: z.enum(TTS_PITCHES).optional(),
+		accent: z.enum(TTS_ACCENTS).optional(),
+		description: nonBlankString.optional(),
+		language: z.enum(TTS_LANGUAGES).optional(),
+		query: nonBlankString.optional(),
+		name: nonBlankString.optional(),
+		limit: queryLimitSchema,
+	})
+	.strict();
 
 export async function GET(request: NextRequest) {
 	return withApiAccess("Voice search", async () => {
-		const params = voiceSearchParamsSchema.parse(
+		const parsed = voiceSearchQuerySchema.safeParse(
 			Object.fromEntries(request.nextUrl.searchParams),
 		);
-		const voices = await getTTSProvider().search(params);
+		if (!parsed.success) {
+			return badRequest(
+				`Invalid voice search parameters: ${parsed.error.issues[0]?.message ?? "unknown error"}`,
+			);
+		}
+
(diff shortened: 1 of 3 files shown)


app/api/v1/tts/voices/route.ts
```diff
@@ -1,14 +1,56 @@
...
+	)
+	.refine((value) => Number.isFinite(value), {
+		message: "must be a finite number",
+	})
+	.refine((value) => Number.isInteger(value) && value > 0, {
+		message: "must be a positive integer",
+	})
+	.optional();
+
+const voiceSearchQuerySchema = z
```
umairnadeem, line 28: this moved the voice-search validation out of the domain layer (`lib/project/types.ts`) and into the route, and in doing so it re-declares the trait field list inline. now there are two sources of truth: `voiceTraitsSchema` (gender/age/pitch/accent/description/language) in lib already encodes exactly these fields off the same enums, and lib even exports `genderSchema`/`ageSchema`/etc. add a `speed` trait to `voiceTraitsSchema` tomorrow and this endpoint silently won't filter on it.

afaict the only reason you couldn't reuse the lib schemas directly is the `.catch(undefined)` (silently ignore) vs the new reject-with-400 semantics you want here. so rather than copy the shape into `app/`, perhaps keep `voiceSearchQuerySchema` in lib next to the traits and build it from the bare `z.enum(TTS_*)` building blocks (no `.catch`), then just import it. keeps domain validation in the domain layer (CLAUDE.md: shared/domain logic lives in `lib/`) and the route stays thin. up to you, not blocking.

umairnadeem (commented):
nice to see the silent `.catch(undefined)` swallowing replaced with a real 400 + the two regression tests, good direction. one structural nit below (gemini already flagged the `.strict()` footgun and the limit-schema verbosity, both worth taking).

umairnadeem (commented):
nice that this fails loudly with a 400 + test coverage instead of the old silent `.catch(undefined)`. one new thing on top of my earlier split-brain note: the `.strict()` is a behavior change worth a sanity check.

app/api/v1/tts/voices/route.ts
```diff
@@ -1,14 +1,56 @@
...
+		age: z.enum(TTS_AGES).optional(),
+		pitch: z.enum(TTS_PITCHES).optional(),
+		accent: z.enum(TTS_ACCENTS).optional(),
+		description: nonBlankString.optional(),
+		language: z.enum(TTS_LANGUAGES).optional(),
+		query: nonBlankString.optional(),
+		name: nonBlankString.optional(),
+		limit: queryLimitSchema,
+	})
+	.strict();
```
umairnadeem, line 40: heads up, `.strict()` is a behavior change that's easy to miss. the old `voiceSearchParamsSchema` (`voiceTraitsSchema.extend(...)`) was default-strip, so unknown query params got silently ignored. now ANY extra param 400s: `?limit=5&_=1699` (cache buster), a stray `utm_*`, or any param a future client tacks on all hard-fail. strict does fail loudly on typos like `gendr=male` (which silently returned unfiltered before), so there's a real tradeoff, but for an HTTP query string strip is usually the safer default. intended? if so a one-liner why would help, otherwise drop `.strict()`. not blocking.

umairnadeem:
overengineered

PR #338, opened 2026-06-21
fix: nightly security audit 2026-06-21

umairnadeem:
## Summary
- Bumped the Remotion package family from `4.0.477` to `4.0.481` to pull in patched Remotion/renderer/studio dependencies, including `ws@8.21.0`.
- Ran `npm audit fix` without force, updating transitive vulnerable packages including `@babel/core`, `hono`, `js-yaml`, and `undici` through safe non-major lockfile changes.
- No source-code hardening changes were made; this PR is dependency-only and behavior-preserving.

## Vulnerability delta
- Before: `npm audit --json` reported 15 vulnerabilities: 1 low, 3 moderate, 11 high.
- After: `npm audit --json` reports 2 vulnerabilities: 0 low, 2 moderate, 0 high, 0 critical.
- Remaining: `uuid` via `@runware/sdk-js` has `fixAvailable: false`. The advisory targets v3/v5/v6 calls with a `buf` argument; `@runware/sdk-js` imports `v4 as uuidv4` and `validate`, so forcing an override to a major uuid version would be speculative and higher-risk.

## Validation
- `npm audit` — exits nonzero only for the remaining 2 moderate `uuid`/`@runware/sdk-js` advisories described above.
- `npm run lint` — passed.
- `npm run format:check` — passed.
- `npm run typecheck` — passed.
- `npm run knip` — passed.
- `npm run build` — passed.
- `npm test -- --passWithNoTests` — passed, 846 tests / 96 files.
- `npm run test:coverage` — passed, 846 tests / 96 files.
- `npm run test:e2e` — passed, 3 tests. Initial run hit a stale local listener on port 3000; after clearing it, the Playwright run passed.

## Open PR overlap check
Considered open PRs:
- #337 (`refactor: nightly maintainability 2026-06-21`) touches canvas/video UI component refactors.
- #336 (`fix: nightly bug fixes 2026-06-21`) touches video element attribute tests/helpers.

This PR changes only `package.json` and `package-lock.json`, so it does not overlap with the files or themes in those PRs.

## Skipped
- No source changes for headers/CSP/routes/Supabase/secrets: high-signal scan did not identify a concrete exploitable source issue outside the dependency advisories.
- Did not force a `uuid` override because npm reports no fix through `@runware/sdk-js`, and the vulnerable API shape is not reachable in its current use.

PR #337, opened 2026-06-21
refactor: nightly maintainability 2026-06-21

umairnadeem:
## Summary
- Extracts a shared `CloseButton` UI primitive for repeated close-button markup.
- Migrates dialog, character edit modal, export popover, and export toast close buttons to the shared primitive.
- Keeps positioning, icon sizing, and focus styles at each call site so behavior and UI intent remain unchanged.

## Architecture / maintainability
- Centralizes the common close-button contract (`type="button"`, `aria-label="Close"`, rounded-md spacing, muted text, hover state, X icon) in `components/ui/close-button.tsx`.
- Leaves call-site-specific responsibilities (Radix dialog wrapping, modal close interception, popover/toast dismissal, focus styling, layout positioning) at the call sites.
- Reduces drift risk from four hand-rolled close buttons while avoiding a broader UI migration.

## Complexity delta
- 5 files touched.
- Net effect: one small reusable primitive, 4 duplicated close-button implementations replaced.
- Diff is 40 insertions / 31 deletions; repeated class/ARIA/button boilerplate is now defined once.

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm test -- --passWithNoTests` (96 files / 846 tests passed)
- `npm run test:coverage` (96 files / 846 tests passed)
- `PLAYWRIGHT_PORT=3108 npm run test:e2e` (3 passed)
- Independent fresh-context review passed with no security or logic concerns.

## Open PR overlap check
- Considered open PR #336 (`fix: nightly bug fixes 2026-06-21`), which touches `lib/video/elementAttributes.ts` and `lib/video/__tests__/elementAttributes.test.ts` for video volume parsing/default behavior.
- This PR touches only close-button UI primitive usage in `components/ui`, character edit modal chrome, and export UI chrome; no overlapping files or themes.

## Skipped / notes
- No `.claude/commands/*.md` or `.claude/skills/*/SKILL.md` guidance files were present in this checkout.
- Avoided broader design-token or route-validation changes because recent maintainer feedback called out low-value/overengineered churn and behavior-change risk.

PR #339, opened 2026-06-21
fix: preserve preview playhead after edits

### LINKED ISSUE #334: Preview playhead resets to start after canvas edits (player remounts on every edit)
## Problem

Editing the canvas resets the preview **playhead back to the start**. If you've scrubbed/played to the middle of a video and then make any canvas edit, the player jumps to frame 0 — losing your place and forcing you to scrub back to where you were to review a change in context.

## Root cause

The Remotion player is **remounted** on every canvas edit. `VideoPanel` renders `<VideoPreview key={playerKey} layout={layout} />` (`app/components/video/VideoPanel.tsx:28`), and `playerKey` is `` `${layoutKey}-${resultVersion}` `` (`app/components/video/useVideoLayout.ts:38`). Any canvas edit bumps `layoutKey` (and/or `resultVersion`), which changes the React `key`, tearing down `VideoPreview` and mounting a fresh `<Player>` (`app/components/video/VideoPreview.tsx`). A new `PlayerRef` starts at frame 0, so the playhead position is lost.

The `key` is presumably there to force the player to pick up changed composition props (`durationInFrames`/`inputProps`). The position isn't captured/restored across that remount.

## Proposed direction

Preserve the playhead's last position across canvas edits:

1. **Capture** the current frame (`player.getCurrentFrame()`) before the remount and **restore** it after the new player mounts via `player.seekTo(frame)` (the seek path already exists — `PlayerControlContext.tsx:32`, `SegmentedSeekBar.tsx:83`).
2. **Clamp** the restored frame to the new composition length (`layout.totalFrames - 1`), since an edit can shorten the video — don't seek past the end.
3. Preferably **avoid the full remount** where possible: feed changed `inputProps`/`durationInFrames` to the existing `<Player>` instead of re-keying it, so position (and play/pause state) survive naturally. If the remount is unavoidable, fall back to the capture-and-restore approach above.
4. Consider preserving play/pause state too, so an edit mid-playback doesn't silently pause.

## Relevant code

- `app/components/video/VideoPanel.tsx:28` — `<VideoPreview key={playerKey} … />` (remount trigger)
- `app/components/video/useVideoLayout.ts:38` — `playerKey = \`${layoutKey}-${resultVersion}\``
- `app/components/video/VideoPreview.tsx` — `<Player>` mount; `setPlayer` ref wiring
- `app/components/video/usePlayerState.ts:40` — `usePlayerFrame` / `getCurrentFrame()` (position read)
- `app/components/video/PlayerControlContext.tsx:32`, `app/components/video/SegmentedSeekBar.tsx:83` — existing `seekTo()` usage for restore

## Acceptance criteria

- [ ] After a canvas edit, the preview stays at (or near) the previous playhead position instead of jumping to the start.
- [ ] If the edit shortens the video, the playhead clamps to the new end rather than overrunning.

---

umairnadeem:
## Summary
- Preserves the preview player's last frame across the existing keyed remount after canvas/layout edits.
- Clamps restored frames to the new layout duration so shortening a video does not seek past the end.
- Adds focused regression coverage for remembering, restoring, and clamping preview frames.

## Selected issue
Selected #334 because it is a small, well-scoped correctness bug with a clear root cause and acceptance criteria. The fix is limited to preview player remount state and does not require auth, secrets, external services, or broader product decisions.

## Validation
- `npm run test:run -- app/components/video/__tests__/playerFrameRestore.test.ts` (RED first: failed before helper existed, then passed after implementation)
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage`
- `npm run test:e2e` initially failed because port 3000 was occupied; retried with `PLAYWRIGHT_PORT=3107 npm run test:e2e` and it passed.

## Open PR overlap check
Checked open PRs before editing and again before commit:
- #338 touches `package.json` / `package-lock.json` security audit files.
- #337 touches export/dialog UI files (`ExportButton`, `ExportToast`, dialog/close-button, character modal).
- #336 touches `lib/video/elementAttributes.ts` and its tests.

This PR touches `VideoPanel`, `VideoPreview`, and focused frame-restore tests/helper only, so it does not overlap by file or theme.

## Skipped issue categories
Skipped broad/product-heavy or larger issues such as public feed/social features, BYOK/gateway integrations, storyboard/refine UX, art-style editing, reference-image fidelity, licensing/legal process, audio crossfade behavior, and multi-part generation/render lifecycle issues.

Closes #334


app/components/video/VideoPreview.tsx
```diff
@@ -68,6 +74,15 @@ export function VideoPreview({ layout }: { layout: VideoLayout }) {
...
 	}, [player, registerPlayer]);
+	useEffect(() => {
+		if (!player) return;
+
+		restorePlayerFrame(player, restoreFrameRef, layout.totalFrames);
+
+		return () => {
+			rememberPlayerFrame(player, restoreFrameRef);
+		};
+	}, [layout.totalFrames, player, restoreFrameRef]);
```
umairnadeem: `layout.totalFrames` in the deps means this effect isn't only restore-on-remount, it also fires remember -> seekTo on the *live* player whenever duration changes without a remount. playerKey is `${layoutKey}-${resultVersion}`, but aspect ratio / transition toggles recompute layout (and totalFrames) without bumping it, so no remount. repro: play the preview, switch aspect ratio mid-play, this snapshots getCurrentFrame() then seekTo's it back on the same player.

harmless-ish today (clamps to ~the same frame), but it's a surprising side effect for an effect that's really there to survive the `key` remount. if restore is meant to key off player identity only, dropping totalFrames from the deps works, on a real remount the component has already re-rendered with the new layout so the closure still has fresh totalFrames for the clamp. costs an exhaustive-deps disable though, so up to you whether it's worth it.

umairnadeem (commented):
nice one. pulling the restore logic into a pure, injectable helper (FrameReader/FrameSeeker + clampRestoredFrame) makes it trivially testable, and lifting the ref above the `key`-remount boundary in VideoPanelBody is exactly the right place for it. the regression-shaped tests are a good call too. one subtlety on the effect deps below, non-blocking.

PR #346, opened 2026-06-21
Label element type icon chips on wider cards

umairnadeem:
## What

Element cards in the canvas previously showed only a colored icon for each media type. This expands that chip into a **media-type pill** that reveals the type name (Narration, Character, Image, Animated image, Clip, Sound, Music) beside the icon — but only when there's room.

The reveal keys off the **card's own width** via a container query (`@container` on the card → `@sm:inline` on the label), not the viewport, since the element card is just the left half of the row (card | divider | preview). On narrow cards the pill collapses back to the original 28px icon-only square with no layout shift.

## Design

- Label font: **Sloptastic** (`font-sans`), weight 425 — not bold
- Size: **`text-label-xs`** (11px), the design system's small-label token, with baked 16px leading — understated
- Color: inherits the element's `--media-*` tint, inside the existing `bg-media-*/15` pill at `rounded-lg`
- Added a `colorClass` field to `ElementConfig` so the pill tints icon + label from one media-color token

Scope is `ElementContainer` only — the insert "+" menu already labels types and `CompactElement` already shows element text.

## Verification

- CI checks pass locally: lint, format, typecheck, knip, build; 846/846 unit tests
- Verified live in the running app: pills show labels in matching media colors on wide cards, collapse to 28px squares when narrowed, and resolve dark-mode tints correctly


lib/canvas/elementConfigs.tsx
```diff
@@ -99,6 +101,7 @@ export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 			<Voice size={16} strokeWidth={1.5} className="text-media-narration" />
 		),
 		iconBgClass: "bg-media-narration/15",
+		colorClass: "text-media-narration",
```
umairnadeem, line 102: split-brain: the media color now lives in two places per entry, on the icon's own `className="text-media-narration"` (line 101) and again here in `colorClass`. lucide icons stroke with `currentColor`, so the parent span's colorClass already colors the icon, the per-icon className is redundant now. easy to change one and forget the other. perhaps drop the color className off the icons and let colorClass be the single source of truth?

umairnadeem (commented):
nice, the @container approach for the label pill is the right call (no JS width measuring). one maintainability nit below.

lib/canvas/elementConfigs.tsx
```diff
@@ -95,10 +97,9 @@ export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 		connector: "tts",
 		outputKind: "audio",
 		label: "Narration",
-		icon: (
-			<Voice size={16} strokeWidth={1.5} className="text-media-narration" />
-		),
+		icon: <Voice size={16} strokeWidth={1.5} />,
```
umairnadeem, line 100: single-sourcing the tint onto colorClass is right, but it only holds if *every* icon render site applies colorClass, and the insert menu doesn't. INSERT_OPTIONS in SortableContent.tsx maps `{key,label,icon,iconBgClass}` (no colorClass), and SortableActions wraps the icon in `${option.iconBgClass} ...` only. so post-this-commit the add-item menu icons lose their media tint and fall back to the menu item's `text-muted-foreground`, ie a gray icon sitting on a colored bg bubble. fix: carry colorClass through INSERT_OPTIONS and add it to that wrapper span too.

app/components/video/ExportButton.tsx
```diff
@@ -59,7 +59,7 @@ export function ExportButton() {
 				align="end"
 				side="top"
 				sideOffset={8}
-				className="w-80 p-3 font-medium"
+				className="w-80 p-3 font-medium bg-surface-elevated"
```
umairnadeem: drive-by unrelated to type pills (and not in the commit msg). PopoverContent already sets bg-popover, so this overrides it to surface-elevated. those tokens are identical in light but diverge in dark (popover #251e21 vs surface-elevated #41393c), so this quietly makes just the export popover a lighter surface than every other popover in dark mode. intentional? if not, drop it; if yes, probably its own PR.

umairnadeem (commented):
thanks for the colorClass single-source cleanup, that's exactly the fix, and the explicit w-7 square is more robust than relying on padding+icon summing to 28px. two things from the latest commit: the refactor missed one icon render site, and an unrelated export-popover bg change rode along.

components/ui/popover.tsx
```diff
@@ -36,7 +36,7 @@ function PopoverContent({
 				align={align}
 				sideOffset={sideOffset}
 				className={cn(
-					"z-50 min-w-48 origin-(--radix-popover-content-transform-origin) rounded-md bg-popover p-3 text-popover-foreground shadow-elevation-3 outline-none data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
+					"z-50 min-w-48 origin-(--radix-popover-content-transform-origin) rounded-md bg-surface-elevated p-3 text-popover-foreground shadow-elevation-3 outline-none data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
```
umairnadeem, line 39: the "unify popover/menu surface" goal isn't actually met yet. this moves PopoverContent (and SelectContent) to bg-surface-elevated, but the real menus, ActionMenu/SelectMenu, render through DropdownMenuContent which is still bg-popover (so are dialog, alert-dialog, toolbar, ExportToast). light mode hides it (#fdfcfc vs #fff) but dark mode is #41393c vs #251e21, a visible step. so a settings popover ends up a different grey than the insert ActionMenu sitting right next to it. either bump dropdown-menu.tsx to bg-surface-elevated too (probably the dialog/toolbar family as well for real consistency), or popover/menu aren't unified, they've just diverged onto two surface tokens.

umairnadeem (commented):
the colorClass threading through InsertOption and the size-6/rounded-md insert chips lgtm. one thing on the surface unify though, see inline.

components/ui/dropdown-menu.tsx
```diff
@@ -42,7 +42,7 @@ function DropdownMenuContent({
 				data-slot="dropdown-menu-content"
 				sideOffset={sideOffset}
 				className={cn(
-					"z-50 max-h-(--radix-dropdown-menu-content-available-height) min-w-[8rem] origin-(--radix-dropdown-menu-content-transform-origin) overflow-x-hidden overflow-y-auto rounded-md border border-border bg-popover p-1 text-popover-foreground shadow-elevation-3 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
+					"z-50 max-h-(--radix-dropdown-menu-content-available-height) min-w-[8rem] origin-(--radix-dropdown-menu-content-transform-origin) overflow-x-hidden overflow-y-auto rounded-md border border-border bg-popover p-1 font-body text-popover-foreground shadow-elevation-3 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
```
umairnadeem, line 45: per DESIGN.md (line 27): UI + titles use Sloptastic (`--font-sans`), Inter (`font-body`) is for *long-form body text* only. dropdowns/selects are UI chrome, not body copy, so flipping them to Inter cuts against the design system, and CLAUDE.md says deviations need explicit sign-off.

the "match input/textarea" rationale is a false friend imo: inputs hold typed user text (legit body), menus are chrome. also it's half-applied as overlay surfaces go: popover.tsx (touched in this same PR) and tooltip both stay on the default Sloptastic, so the floating surfaces now disagree with each other.

either leave menus on the default font, or if Inter-for-chrome is actually the intent, get it blessed in DESIGN.md and apply it to all portaled surfaces in one place rather than sprinkling `font-body` per-component.

umairnadeem (commented):
one more on the font commit (b13601c). flagging a DESIGN.md deviation, otherwise the surface-unification direction is good.

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -230,18 +227,15 @@ function CharacterEditDialogBody({
 			</div>
 
 			<DialogFooter className="shrink-0">
-				<button
+				<Button
 					type="button"
+					variant="destructive"
```
umairnadeem: heads up, this drops the two-state visual on the confirm-delete guard. before, the unconfirmed state was neutral (border/muted) and only went red on `confirmDelete`. now it's `variant="destructive"` (solid red) in *both* states, so the only thing `confirmDelete` still drives is the label text. the escalation that signaled 'are you sure' is gone, and the first click already looks like the destructive action.

perhaps: keep the unconfirmed state low-emphasis and reserve red for confirm, e.g. `variant={confirmDelete ? "destructive" : "outline"}` (or ghost). up to you if always-red is the intended look now.

app/components/canvas/elements/ElementContainer.tsx
```diff
@@ -100,9 +96,12 @@ export function ElementContainer({
 						contentEditable={false}
 					>
 						<span
-							className={`flex size-7 shrink-0 items-center justify-center rounded-lg ${config.iconBgClass}`}
+							className={`flex h-7 w-7 @sm:w-auto shrink-0 items-center justify-center @sm:justify-start gap-1.5 rounded-lg @sm:px-1.5 ${config.iconBgClass} ${config.colorClass}`}
 						>
 							{config.icon}
+							<span className="hidden font-body text-label-xs @sm:inline">
```
umairnadeem, line 102: nit: this flips the type-pill label to `font-body` (Inter), but per the DESIGN.md you updated in this same PR, Inter is for body text + form controls (inputs/selects/dropdowns) while UI labels stay on Sloptastic (`font-sans`). a storyboard type chip is UI chrome, not a form control, so this reads as a small doc-vs-code split-brain. either revert to font-sans or widen the DESIGN.md rule if you want short UI labels on Inter too.

umairnadeem (commented):
took another pass over the two new commits (font/hover polish + CloseButton unify). the CloseButton extraction is clean and nicely reuses focus-ring + the new --surface-hover token, and threading that token through light/dark/@theme is the right single-source. two notes below, one's a UX question on the delete button.

components/ui/close-button.tsx
```diff
@@ -0,0 +1,25 @@
...
+			className={cn(
+				"inline-flex h-6 w-6 items-center justify-center rounded-md transition-colors hover:bg-surface-hover hover:text-foreground focus-ring",
+				className,
+			)}
+			{...props}
+		>
+			<X size={size} />
+		</button>
+	);
+}
```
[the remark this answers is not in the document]
umairnadeem: were on react 19

PR #348, opened 2026-06-22
Retain characters when converting image to animated image

umairnadeem:
## Problem

The **Animate** button converts an `image` element into an `animated_image` via a refine `set` op that changes the element `type`. `applySet` rebuilt the node from the new type's defaults and replaced `customAttributes` wholesale — silently dropping the character cast (`characters`) assigned to the shot.

## Fix

Add a small declarative policy — `lib/canvas/preservedAttributes.ts` — as the single source of truth for *which attributes survive a type change*: a map of attribute → the element types that use it. An attribute is preserved when the **target** type still uses it.

- Cross-type authoring intent (the cast) carries over: `characters` is shared by `image` ↔ `animated_image`.
- Stale generation-result metadata (`url`, `durationSec`) and connector config (`model`/`provider`) are **not** carried — they're correctly reset/re-derived for the new type by `createCanvasNode`.
- Type-specific config (`motion`, etc.) is reset to the new type's defaults.

`applySet` feeds the preserved attributes through `createCanvasNode`'s existing `attrs` merge, so the refine layer stays thin and an explicit attr in the op still wins.

## Tests

- New `preservedAttributes.test.ts`: carries within the group, drops when the target is outside it, ignores non-listed keys like `url`, handles missing attributes.
- New `applyOps` case asserting `characters` survives `image → animated_image` while a stale `url` is dropped — also the first coverage of `applySet`'s type-change branch.

All CI checks (lint, format, typecheck, knip) pass; 851 tests green.

### DIFF
--- a/lib/canvas/preservedAttributes.ts
+++ b/lib/canvas/preservedAttributes.ts
@@ -1,0 +1,24 @@
+import type { CanvasContentElement, CanvasElementType } from "./types";
+
+type ElementTypeGroup = readonly CanvasElementType[];
+
+/**
+ * A map of attribute name and the types within which it's preserved
+ */
+const PRESERVED_ATTRIBUTE_TYPES: Record<string, ElementTypeGroup> = {
+	characters: ["image", "animated_image"],
+};
+
+export function preservedAttributes(
+	source: CanvasContentElement,
+	targetType: CanvasElementType,
+): Record<string, string> {
+	const attrs = source.customAttributes ?? {};
+	const preserved: Record<string, string> = {};
+	for (const [attribute, types] of Object.entries(PRESERVED_ATTRIBUTE_TYPES)) {
+		if (attrs[attribute] && types.includes(targetType)) {
+			preserved[attribute] = attrs[attribute];
+		}
+	}
+	return preserved;
+}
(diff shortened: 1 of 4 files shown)


lib/canvas/preservedAttributes.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export function preservedAttributes(
+	source: CanvasContentElement,
+	targetType: CanvasElementType,
+): Record<string, string> {
+	const attrs = source.customAttributes ?? {};
+	const preserved: Record<string, string> = {};
+	for (const [attribute, types] of Object.entries(PRESERVED_ATTRIBUTE_TYPES)) {
+		if (attrs[attribute] && types.includes(targetType)) {
```
umairnadeem, line 19: this only gates on the *target* type being in the group, not the source. harmless today since `characters` only ever lives on image/animated_image anyway, but the map reads as "preserve X when moving *between* these types". once PRESERVED_ATTRIBUTE_TYPES grows, a key could teleport onto the target from an unrelated source type that happens to carry the same attr name. maybe also check `types.includes(source.type)`? up to you.

umairnadeem (commented):
nice, pulling preservation into a pure table-driven helper with thorough tests, and dropping stale attrs (url) on the type swap is the right move. one forward-looking question inline, not blocking.

PR #349, opened 2026-06-22
Let users set clip and animated image duration

umairnadeem:
## Summary

Adds a **4–15s duration dropdown** to `clip` and `animated_image` elements.

- **Clip** duration was displayed but read-only — it's now editable.
- **Animated image** had no duration control at all, and the value was dropped before the video step — it's now settable and carried through to generation.

## Changes

- `lib/canvas/elementConfigs.tsx` — duration attribute/control for clip and animated image
- `lib/connectors/animated_image/plugins/animated-image-chain.ts` — propagate duration into the video step (+ test)
- `lib/connectors/types.ts` — supporting type

### DIFF
--- a/lib/canvas/elementConfigs.tsx
+++ b/lib/canvas/elementConfigs.tsx
@@ -54,6 +54,7 @@ const VOLUME_OPTIONS = [
 ] as const;
 const EMOTION_OPTIONS = Object.values(TTSEmotion);
 const CAPTIONS_OPTIONS = ["on", "off"] as const;
+const DURATION_OPTIONS = Array.from({ length: 12 }, (_, i) => String(i + 4));
 
 const volumeSpec = (color: string): AttributeSpec => ({
 	color,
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
nice, reusing durationSpec for both clip and animated_image (and threading it through the chain stash) keeps this tight. one correctness thing on the string-vs-number boundary below.

lib/canvas/elementConfigs.tsx
```diff
@@ -54,6 +54,7 @@ const VOLUME_OPTIONS = [
 ] as const;
 const EMOTION_OPTIONS = Object.values(TTSEmotion);
 const CAPTIONS_OPTIONS = ["on", "off"] as const;
+const DURATION_OPTIONS = Array.from({ length: 12 }, (_, i) => String(i + 4));
```
umairnadeem, line 57: these options are strings ("4".."15"), and the value flows through untouched: getGenerationInputs spreads customAttributes into the generate params, so duration reaches the connector as the string "5". but AnimatedImageGenerateParams/VideoGenerateParams type it as `number`, and runware consumes it raw, `duration: params.duration || 5` goes to videoInference and `durationSec: params.duration ?? 5` lands in metadata. so it's a type lie and a string actually hits the runware API + the durationSec metadata (typed number). the tests pass `duration: 8` as a number so they don't catch the real string path. coerce with Number() somewhere on the way in (volume already does exactly this via elementAttributes.ts), otherwise this silently ships strings to the provider.

umairnadeem: coerced to number on route

PR #350, opened 2026-06-22
fix: nightly bug fixes 2026-06-22

umairnadeem:
## Summary
- Makes `GET /api/v1/tts/voices` validate the `limit` query parameter before calling the TTS provider.
- Returns a client-facing 400 for blank, malformed, non-integer, or non-positive limits instead of silently dropping invalid values or treating blank input as `0`.
- Keeps persisted metadata parsing untouched; this change is limited to the API route boundary.

## Bugs fixed
- The voice search API used the permissive persisted metadata schema at a public route boundary. Because `z.coerce.number().optional().catch(undefined)` was used for `limit`, requests like `?limit=+`, `?limit=not-a-number`, or `?limit=0` could silently proceed without a valid limit instead of failing loudly.

## Tests added / areas covered
- Added route regression coverage for:
  - valid string limit coercion and provider forwarding (`limit=5`)
  - blank/whitespace limit (`limit=+`)
  - malformed limit (`limit=not-a-number`)
  - non-positive limit (`limit=0`)

## Validation
- `npm test -- app/api/v1/__tests__/routes.test.ts` ✅
- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm run build` ✅
- `npm run knip` ✅
- `npm run test:coverage` ✅
- `npm run test:e2e` initially hit port 3000 in use; retried with `PLAYWRIGHT_PORT=3107 npm run test:e2e` ✅

## Open PR overlap check
Considered open PRs #349, #348, #346, #339, #338, and #337. This PR only changes `app/api/v1/tts/voices/route.ts` and `app/api/v1/__tests__/routes.test.ts`, which do not overlap with the files or themes in those PRs.

## Runner note
The required Claude Code core pass was invoked in YOLO print mode, but it produced no output or diff after a bounded wait and was killed as unavailable for this run. I completed a narrow Hermes-driven TDD correctness pass instead, then independently validated the result.

### DIFF
--- a/app/api/v1/tts/voices/route.ts
+++ b/app/api/v1/tts/voices/route.ts
@@ -1,14 +1,35 @@
 import { NextRequest, NextResponse } from "next/server";
+import { z } from "zod";
 import { getTTSProvider } from "@/lib/api/providers";
+import { badRequest } from "@/lib/api/response";
 import { withApiAccess } from "@/lib/api/with-auth";
 import { voiceSearchParamsSchema } from "@/lib/project/types";
 
+const voiceSearchQuerySchema = voiceSearchParamsSchema.extend({
+	limit: z
+		.string()
+		.transform((value) => (value.trim() === "" ? NaN : Number(value)))
+		.refine((value) => Number.isFinite(value), {
+			message: "limit must be a finite number",
+		})
+		.refine((value) => Number.isInteger(value), {
+			message: "limit must be an integer",
+		})
+		.refine((value) => value > 0, { message: "limit must be positive" })
+		.optional(),
+});
+
 export async function GET(request: NextRequest) {
 	return withApiAccess("Voice search", async () => {
-		const params = voiceSearchParamsSchema.parse(
+		const parsed = voiceSearchQuerySchema.safeParse(
 			Object.fromEntries(request.nextUrl.searchParams),
 		);
-		const voices = await getTTSProvider().search(params);
+		if (!parsed.success) {
+			return badRequest(
+				parsed.error.issues[0]?.message ?? "Invalid voice search query",
+			);
+		}
+		const voices = await getTTSProvider().search(parsed.data);
 		return NextResponse.json({ voices });
 	});
 }
--- a/app/api/v1/__tests__/routes.test.ts
+++ b/app/api/v1/__tests__/routes.test.ts
@@ -465,7 +465,7 @@ describe("API routes", () => {
 			]);
 
 			const req = makeRequest(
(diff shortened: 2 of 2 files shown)


umairnadeem (commented):
nice, turning a silently-swallowed bad `limit` into a real 400 is the right call (and the custom messages + tests are thorough). one structural thing below.

app/api/v1/tts/voices/route.ts
```diff
@@ -1,14 +1,35 @@
 import { NextRequest, NextResponse } from "next/server";
+import { z } from "zod";
 import { getTTSProvider } from "@/lib/api/providers";
+import { badRequest } from "@/lib/api/response";
 import { withApiAccess } from "@/lib/api/with-auth";
 import { voiceSearchParamsSchema } from "@/lib/project/types";
 
+const voiceSearchQuerySchema = voiceSearchParamsSchema.extend({
```
umairnadeem, line 8: split-brain on `limit`: the canonical `voiceSearchParamsSchema` (lib/project/types.ts:37) already defines it as `z.coerce.number().int().positive().optional().catch(undefined)`, and this extend now shadows that with a stricter reimpl. so there are two definitions of "valid limit", and the base one's `.catch(undefined)` (the silent swallow you're trying to kill) is just dead weight here.

since this route is the *only* consumer of `voiceSearchParamsSchema`, simplest is to fix it at the source: drop `.catch(undefined)` on the base and parse it directly, rather than wrapping it. if the only reason for the manual `.transform`+refines is the nice error strings, you could keep those on the base too. either way one source of truth for what a valid limit is.

app/api/v1/tts/voices/route.ts
```diff
@@ -1,14 +1,35 @@
...
 export async function GET(request: NextRequest) {
 	return withApiAccess("Voice search", async () => {
-		const params = voiceSearchParamsSchema.parse(
+		const parsed = voiceSearchQuerySchema.safeParse(
 			Object.fromEntries(request.nextUrl.searchParams),
 		);
-		const voices = await getTTSProvider().search(params);
+		if (!parsed.success) {
+			return badRequest(
+				parsed.error.issues[0]?.message ?? "Invalid voice search query",
```
umairnadeem, line 29: nit: `?? "Invalid voice search query"` is dead, on a `safeParse` failure `issues` is always non-empty so `issues[0]` is guaranteed. can drop the fallback (or keep it if you'd rather not rely on that invariant, up to you).

umairnadeem:
useless

PR #351, opened 2026-06-22
perf: nightly performance optimizations 2026-06-22

umairnadeem:
## Summary
- Optimizes the playback transport seekbar hot path by memoizing scene-to-scrub segment projection across frame ticks.
- Switches scene lookup for frame/hover updates from a linear scan to binary search over sorted scene segment end times.
- Preserves existing boundary behavior covered by `useSceneSegments` tests: empty segments return `-1`, boundaries advance to the next segment, and out-of-range times clamp to the nearest available segment.

## Why this matters
The bottom transport bar receives Remotion `frameupdate` events during playback. Active-scene lookup and seekbar rendering are therefore on a per-frame UI path; reducing repeated O(n) scans and allocations keeps playback controls snappier as project scene counts grow.

## Open PR overlap check
Considered open PRs #350, #349, #348, #346, #339, #338, and #337. This PR only touches `app/components/video/SegmentedSeekBar.tsx` and `app/components/video/useSceneSegments.ts`, which are outside the listed open-PR file/theme overlap areas.

## Skipped
- Slate selection-only/state-setter guards and one-off callback tweaks: too trivial per prior maintainer feedback.
- Bundle/TTFB/dynamic import churn: no clear in-app runtime UX impact for this pass.
- Open PR overlap areas, including ExportButton/VideoPreview/package deps/API routes/connectors currently in flight.

## Validation
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage`
- `PLAYWRIGHT_PORT=3001 npm run test:e2e`

Note: plain `npm run test:e2e` initially failed locally because port 3000 is occupied by the Hermes WhatsApp bridge (`bridge.js --port 3000`), so the same Playwright suite was rerun successfully on port 3001.

### DIFF
--- a/app/components/video/SegmentedSeekBar.tsx
+++ b/app/components/video/SegmentedSeekBar.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import type { PlayerRef } from "@remotion/player";
-import { useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { clamp } from "@/lib/utils";
 import type { VideoLayout } from "@/lib/video/types";
 import { usePlayerFrame } from "./usePlayerState";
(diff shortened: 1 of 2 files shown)


app/components/video/SegmentedSeekBar.tsx
```diff
@@ -45,28 +45,35 @@ export function SegmentedSeekBar({
...
+				basis: seg.duration / totalDurationSec,
+			})),
+		[segments, totalDurationSec],
+	);
 
-	const scrubSegments = segments.map((seg) => ({
-		id: seg.sceneId,
-		basis: seg.duration / totalDurationSec,
-	}));
+	const onHoverChange = useCallback(
```
umairnadeem, line 57: afaict this `useCallback` is dead optimization: `ScrubBar` isn't memoized and `onHoverChange` isn't a dep of any memo/effect (it's just called directly in ScrubBar's pointer handlers), so stabilizing its identity changes nothing. and `onScrub`/`onScrubStart`/`onScrubEnd`/`children` are still fresh every render anyway. since this component re-renders every frame via `usePlayerFrame`, the whole ScrubBar subtree re-renders per frame regardless. to actually "reduce frame work" you'd either isolate the frame-driven `value`/`progress` into a thin leaf so the segments+handlers subtree doesn't re-render, or memo ScrubBar + stabilize *all* its props. the `scrubSegments` useMemo does save the per-frame `.map()` alloc which is legit, but on its own it barely moves the needle here.

app/components/video/useSceneSegments.ts
```diff
@@ -29,11 +29,15 @@ export function findSegmentIndexAt(
 	timeSec: number,
 ): number {
 	if (segments.length === 0) return -1;
-	for (let i = 0; i < segments.length; i++) {
-		const seg = segments[i];
-		if (timeSec < seg.start + seg.duration) return i;
+	let lo = 0;
```
umairnadeem, line 32: nit: binary search is correct (segment ends are monotonic) but N here is scene count, a handful, and this only runs on hover, not per frame. trading the dead-simple linear scan for non-obvious bisection cleverness has no measurable win and is less grokkable. up to you, but i'd keep the linear scan.

umairnadeem (commented):
nice that the hooks got hoisted above the `segments.length === 0` early return so rules-of-hooks stays clean. one concern though: the memo/callback here mostly don't buy what the commit title promises, details inline.

umairnadeem:
overengineered. negligble impact

PR #352, opened 2026-06-22
fix: allow inline character removal in composer

### LINKED ISSUE #340: Inline character removal in the hero composer (no edit-modal round-trip)
## Problem

In the hero composer, there's no quick way to **remove a character**. Character assets render as avatar tiles (`ComposerAssets`) that only expose an **edit** action — to delete a character you have to open the `CharacterEditModal` and hit its delete button, which is several clicks for a common action.

Notably, reference-image tiles in the same row already get an inline remove (`onRemove` → X overlay via `AssetTile`), but character tiles only get `onEdit`. The pattern to fix this already exists.

- "Want to easily remove characters in composer hero, instead have to open edit modal and hit delete which isn’t ideal"

## Relevant code

- `app/components/copilot/ComposerAssets.tsx:44-53` — character `AssetTile`s pass only `onEdit`; reference tiles at `:54-62` pass `onRemove` (the inline-X pattern to reuse)
- `app/components/canvas/elements/AssetTile.tsx` — supports `onRemove` (X overlay) already
- `app/components/canvas/elements/character/CharacterEditModal.tsx:232-245` — the current delete path (modal-only)
- `app/components/copilot/ComposerCopilot.tsx:204,310-313` — wires `onEditCharacter` / edit modal

## Suggested direction

Surface an inline remove (X overlay) on character tiles in the composer, mirroring how reference images are removed via `AssetTile`'s `onRemove`. Keep the edit modal for editing; removal shouldn't require opening it.

---

## Requirement: identical to reference-image removal

The character removal experience must be **identical** to the reference-image removal UX, not merely similar: the same X-overlay affordance, placement, hover/interaction states, and confirmation behavior that `AssetTile`'s `onRemove` already gives reference tiles. Wire character tiles through the same `onRemove` path so the two stay consistent by construction rather than diverging.

umairnadeem:
## Summary
- Adds inline character removal from the hero composer asset row, mirroring the existing reference-image remove affordance.
- Keeps the character edit action available on the tile while showing a separate X remove control when both edit and remove actions exist.
- Adds focused regression coverage for removing only the selected composer character while preserving other composer assets.

## Selected issue
Selected #340 because it is a small, well-scoped UI behavior gap with clear acceptance criteria and an obvious existing pattern (`AssetTile` `onRemove`) to reuse. It does not require auth, secrets, external services, or broader product decisions.

## Validation
- `npm run test:run -- app/components/copilot/__tests__/ComposerAssets.test.ts` — RED first failed with `removeComposerCharacter is not a function`; GREEN passed after implementation.
- `npm run lint` — passed.
- `npm run knip` — passed.
- `npm run build` — passed.
- `npm run test:coverage` — passed, 847 tests / 97 files.
- `npm run test:e2e` — initial run failed because port 3000 was already in use.
- `PLAYWRIGHT_PORT=3109 npm run test:e2e` — passed, 3 tests.

## Open PR overlap check
Checked open PRs before editing and again before commit:
- #351 touches video seekbar/scene segment performance.
- #350 touches `/api/v1/tts/voices` route validation.
- #349 touches clip/animated-image duration config and connector types.
- #348 touches canvas preserved attributes/refine type-change behavior.
- #346 touches element type pills/element config/export popover styling.
- #339 touches video preview playhead restoration.
- #338 touches dependency audit files.
- #337 touches shared close-button/dialog/export/character-modal chrome.

This PR touches only `AssetTile`, `ComposerAssets`, and a focused composer-assets test, so it is non-overlapping by file and theme.

## Skipped issue categories
Skipped broad/product-heavy or larger issues such as public feed/social features, BYOK/gateway integrations, fullscreen transport controls, global destructive-button restyling, art-style editing, reference-image upload-after-generation, licensing/legal process, audio crossfades, and multi-part generation/render lifecycle issues.

Closes #340

### DIFF
--- a/app/components/canvas/elements/AssetTile.tsx
+++ b/app/components/canvas/elements/AssetTile.tsx
@@ -28,6 +28,8 @@ export function AssetTile({
 	const initial = name?.trim().charAt(0).toUpperCase();
 	const fallbackContent =
 		fallback === "icon" || !initial ? <Icon className="h-5 w-5" /> : initial;
+	const showActions = status !== "generating";
+	const showFullRemove = onRemove && !onEdit;
 	return (
 		<div className="group/tile flex w-16 flex-col gap-1 sm:w-20">
 			<div className="relative aspect-square overflow-hidden rounded-md border border-border bg-card">
@@ -63,7 +65,7 @@ export function AssetTile({
 						<Icon className="h-3 w-3" />
 					</div>
 				)}
-				{onEdit && status !== "generating" && (
+				{onEdit && showActions && (
 					<button
 						type="button"
 						onClick={onEdit}
@@ -73,7 +75,7 @@ export function AssetTile({
 						<Pencil className="h-3.5 w-3.5" />
 					</button>
 				)}
-				{onRemove && status !== "generating" && (
+				{showFullRemove && showActions && (
 					<button
 						type="button"
 						onClick={onRemove}
@@ -81,6 +83,16 @@ export function AssetTile({
 						className="absolute inset-0 flex items-center justify-center bg-background/70 text-foreground opacity-0 transition-opacity group-hover/tile:opacity-100 focus:opacity-100 focus:outline-none"
 					>
 						<X className="h-3.5 w-3.5" />
+					</button>
+				)}
+				{onRemove && onEdit && showActions && (
+					<button
+						type="button"
+						onClick={onRemove}
+						aria-label={`Remove ${name ?? "asset"}`}
+						className="absolute right-1 top-1 z-10 flex size-5 items-center justify-center rounded-md bg-card text-foreground opacity-0 shadow-sm ring-1 ring-border transition-opacity hover:bg-button-hover group-hover/tile:opacity-100 focus:opacity-100 focus:outline-none"
+					>
+						<X className="h-3 w-3" />
 					</button>
 				)}
 			</div>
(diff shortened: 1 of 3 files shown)


app/components/canvas/elements/AssetTile.tsx
```diff
@@ -83,6 +85,16 @@ export function AssetTile({
 						<X className="h-3.5 w-3.5" />
 					</button>
 				)}
+				{onRemove && onEdit && showActions && (
+					<button
+						type="button"
+						onClick={onRemove}
+						aria-label={`Remove ${name ?? "asset"}`}
+						className="absolute right-1 top-1 z-10 flex size-5 items-center justify-center rounded-md bg-card text-foreground opacity-0 shadow-sm ring-1 ring-border transition-opacity hover:bg-button-hover group-hover/tile:opacity-100 focus:opacity-100 focus:outline-none"
```
umairnadeem, line 93: this corner chip is basically the idle type-badge (line 64) in a second state: same `right-1 top-1 size-5 rounded-md bg-card shadow-sm ring-1 ring-border`, just opacity-0 + z-10 + an X instead of the icon. for a character with an avatar both render and the X chip happens to land exactly on top of the type badge, which only works by geometry coincidence. tweak the badge size/pos in one place and they silently desync. perhaps pull the shared chip classes into a const (or a tiny `CornerChip`) so position+size are single-sourced? not blocking.

umairnadeem (commented):
nice that the remove button is gated behind `showActions` so you can't nuke a character whose avatar is still mid-generation (avoids the in-flight job resurrecting it via `setCharacter(name, {...existing})`). one small structural nit below; the wrapper + focus-ring stuff the bot already flagged so won't pile on.

umairnadeem:
too complex, need to have a consistent removal experience with reference images too

PR #364, opened 2026-06-24
fix: suppress character nameplates leaking into scene images

### LINKED ISSUE #354: Character nameplate badges leak from avatars into scene images (and uploaded refs lack nameplates)
## Problem

Generated character avatars bake a nameplate into the image on purpose: the `character-avatar` plugin appends `A small rectangular nameplate at the bottom of the frame reads "<name>" ...` to the avatar prompt (`lib/connectors/image/plugins/character-avatar.ts:16`). That nameplate is how the image model tells one referenced character from another.

The problem is that those same avatars are then passed straight through as reference images when generating scene images: `character-references` maps each referenced character's `avatarUrl` into `referenceImages` (`lib/connectors/image/plugins/character-references.ts:18-24`). The image model copies the baked-in nameplate into the scene, so the character's name badge leaks into the final rendered images.

There's also an inconsistency for uploaded characters: an uploaded image is stored to `avatarUrl` directly with `avatarUploaded: true` and no nameplate (`app/components/canvas/elements/character/CharacterEditModal.tsx:85-88`). So uploaded refs carry no name label at all, while generated ones do, which weakens the model's ability to bind the right reference to the right character when multiple characters are in a scene.

- "character avatar badges leak into images"

## Proposed direction

1. **Suppress the nameplate in scene images.** When a scene image/animated-image prompt includes character reference images, append a negative instruction (e.g. "no character nameplate, no name labels or text overlays on the characters") so the model doesn't reproduce the badge baked into the reference avatars. Natural home is the `character-references` plugin (it's the one that injects the avatars as references) or the shared scene image prompt builder.
2. **Add a nameplate to uploaded character images.** Make uploaded character refs consistent with generated avatars so the model can bind name -> reference across both. Render/composite the character name as a small nameplate onto the uploaded image at upload time, or pass the name alongside the uploaded ref the same way generated avatars carry it.

## Relevant code

- `lib/connectors/image/plugins/character-avatar.ts:16` - bakes the nameplate into generated avatars
- `lib/connectors/image/plugins/character-references.ts:18-24` - passes avatar URLs as `referenceImages` for scene generation (where the leak surfaces; natural place for the negative instruction)
- `app/components/canvas/elements/character/CharacterEditModal.tsx:85-88` - uploaded image -> `avatarUrl` with `avatarUploaded: true`, no nameplate today
- `lib/upload/useImageUpload.ts` - upload hook (where an upload-time nameplate composite would hook in)

## Related

- #243 (character/avatar generation correctness) - same files, but that issue is about the avatar dependency graph + staleness; this is about the nameplate text leaking and uploaded-vs-generated ref consistency.
- #268 (character appearance drifts between scenes) - adjacent consistency concern; cleaner reference images help binding.

umairnadeem:
## Problem

Generated character avatars intentionally bake a nameplate into the image (`A small rectangular nameplate at the bottom of the frame reads "<name>" ...`) so the image model can tell referenced characters apart. Those same avatars are then passed straight through as `referenceImages` when generating scene images, so the model copies the baked-in nameplate into the scene and the character's name badge leaks into the final rendered images.

## Fix

Append a `No nameplates` negative instruction in the `character-references` plugin whenever it attaches reference avatars to a generation. This is the natural home — it's the plugin that injects the avatars as references, and `buildImagePlugins` (which includes it) is reused by the animated-image chain, so both scene images and animated images are covered by one change.

The instruction is only appended when reference images are actually present, so prompts without character refs are untouched.

## Tests

Updated `character-references.test.ts` cases that resolve reference images to expect the appended suffix; the no-reference cases stay unchanged, locking in the conditional.

Checks run locally: lint, format, typecheck, knip, and the full connector test suite (123 passed).

Closes #354

### DIFF
--- a/lib/connectors/image/plugins/character-references.ts
+++ b/lib/connectors/image/plugins/character-references.ts
@@ -19,9 +19,12 @@ export function createCharacterReferencesPlugin(
 				.map((name) => chars[name]?.avatarUrl)
 				.filter(Boolean);
 
+			if (referenceImages.length === 0) return rest;
+
 			return {
 				...rest,
-				...(referenceImages.length > 0 && { referenceImages }),
+				prompt: `${rest.prompt}. No nameplates`,
+				referenceImages,
 			};
 		},
 	};
--- a/lib/connectors/__tests__/character-references.test.ts
+++ b/lib/connectors/__tests__/character-references.test.ts
@@ -48,7 +48,7 @@ describe("character-references plugin", () => {
 		});
 
 		expect(result).toEqual({
-			prompt: "Red meets Granny",
+			prompt: "Red meets Granny. No nameplates",
 			referenceImages: ["https://img/red.png", "https://img/granny.png"],
 		});
 	});
@@ -85,7 +85,7 @@ describe("character-references plugin", () => {
 		});
 
 		expect(result).toEqual({
-			prompt: "Hello",
+			prompt: "Hello. No nameplates",
 			referenceImages: ["https://img/alice.png", "https://img/bob.png"],
 		});
 	});
@@ -102,7 +102,7 @@ describe("character-references plugin", () => {
 		});
 
 		expect(result).toEqual({
-			prompt: "Hello",
+			prompt: "Hello. No nameplates",
 			referenceImages: ["https://img/alice.png"],
 		});
 	});
@@ -118,7 +118,7 @@ describe("character-references plugin", () => {
 		});
 
 		expect(result).toEqual({
-			prompt: "Hello",
+			prompt: "Hello. No nameplates",
 			referenceImages: ["https://img/alice.png"],
 		});
 	});


lib/connectors/image/plugins/character-references.ts
```diff
@@ -19,9 +19,12 @@ export function createCharacterReferencesPlugin(
 				.map((name) => chars[name]?.avatarUrl)
 				.filter(Boolean);
 
+			if (referenceImages.length === 0) return rest;
+
 			return {
 				...rest,
-				...(referenceImages.length > 0 && { referenceImages }),
+				prompt: `${rest.prompt}. No nameplates`,
```
umairnadeem, line 26: split-brain worth calling out: the nameplate gets *baked into* the avatar over in character-avatar.ts:16 (`A small rectangular nameplate ... reads "${name}"`), and here we beg the downstream model to ignore it. two prompts encoding the same concept in opposition. if someone tweaks/removes the avatar nameplate wording, this `. No nameplates` suffix silently becomes a no-op or a lie, and nothing links them.

root-cause fix imo is to not pollute the reference image in the first place: does the avatar actually need a baked-in text label? if it's just for gallery identification, render it clean and overlay the name in the UI instead. then references stay nameplate-free and you can drop this suffix entirely.

secondary nit: appending a negation to a positive prompt is weak steering for most image models, esp when the reference literally *shows* the nameplate, so this may not reliably suppress it. not blocking, but the source-side fix sidesteps the reliability question too.

umairnadeem (commented):
nice that the fix lands in buildImagePlugins so the animated-image chain is covered too, and the tests track the new prompt suffix. one structural thing below, otherwise lgtm.

PR #371, opened 2026-06-25
refactor: nightly maintainability 2026-06-25

umairnadeem:
## Summary
Two maintainability improvements: extracts a shared `TooltipIconButton` primitive and consolidates a duplicated ElevenLabs audio-format constant.

### 1. Shared `TooltipIconButton` primitive (`components/ui/icon-button.tsx`)
5 call sites had inline `<Tooltip>` → `<IconButton>` wrappers that duplicated the same 15-line pattern (tooltip, icon-button imports, size/color/disabled wiring). The new `TooltipIconButton` component wraps the common pattern, defaults `ariaLabel` to `label`, and surfaces the same `IconButton` props.

**Changed files:**
- `components/ui/icon-button.tsx` — new `TooltipIconButton` export
- `app/components/video/BottomTransportBar.tsx` — dropped local `TooltipButton`; 3 call sites use shared primitive
- `app/components/canvas/elements/AudioPlayer.tsx` — replaced inline wrapper
- `app/components/canvas/elements/PlayFromHereButton.tsx` — replaced inline wrapper
- `app/components/canvas/elements/character/VoicePicker.tsx` — replaced inline wrapper
- `app/components/canvas/elements/character/CharacterEditModal.tsx` — replaced inline wrapper

### 2. `ELEVENLABS_AUDIO_FORMAT` constant
The `{ codec: "mp3", sampleRate: 44100, bitrateKbps: 128 }` block was byte-identical in both ElevenLabs audio providers.

**Changed files:**
- `lib/providers/elevenlabs.ts` — exported `ELEVENLABS_AUDIO_FORMAT`
- `lib/providers/music/elevenlabs.ts` — references shared constant
- `lib/providers/sfx/elevenlabs.ts` — references shared constant

## Complexity delta
- 9 files touched
- 77 insertions / 102 deletions (net -25 lines)
- 5 duplicated Tooltip+IconButton patterns → 1 shared primitive
- 1 duplicated audio-format literal → 1 shared constant

## Deliberately skipped
- **fetchJson/postJson consolidation** (4 copies) — error-message formats differ per call site; unifying would change error surfaces (behavior change).
- **Shared 600_000 timeout constant** — poll deadline and Runware client timeout are separate concerns; forcing a shared constant adds coupling.
- **.catch(undefined) schema fallbacks** — intentional resilience against untrusted LLM-generated XML attributes; changing them would be a behavior change.

## Open PR overlap check
No open PRs at time of branching. No overlap risk.

## Validation
[...]

PR #372, opened 2026-06-25
refactor: rethink openslop architecture 2026-06-25

umairnadeem:
## App Understanding

OpenSlop is an AI creative-media studio: a SlateJS storyboard editor where a prompt streams OSML over SSE into canvas elements (scenes, characters, image/clip/animated/music/sound/narration). Stale elements are queued client-side and generated via `/api/v1/*` routes (connector → gateway → provider), with results cached in Vercel Blob and the composition rendered to MP4 through Remotion Lambdas. State persists per-project in Supabase. Main workflows: submit prompt → generate assets → arrange/refine on the canvas → preview → export.

## From-Scratch Architecture

The existing three-layer generation pipeline (connectors / gateways / providers) plus the route factory and Zustand stores is already the right shape and was recently audited clean. A from-scratch build would keep that. The only genuine drift is in the **app/canvas UI tree**, where a small amount of plumbing exists that a clean build wouldn't: a module-constant renderer threaded as a prop, and a duplicated overlay-button.

## Implemented Simplifications

1. **Remove the `renderElement` prop-drill in the canvas dnd tree.** `renderCanvasElement` is a module-level constant, but it was passed `Canvas → SortableScene/SortableContent → SortableItem` purely to be invoked at the leaf. `SortableItem` now imports it directly — matching how `DragOverlay` already consumes it — and the prop is dropped from three component interfaces (`SortableItem`, `SortableScene`, `SortableContent`). The import also moves out of `Canvas`, where it was no longer used. Fewer cross-tree props, one consistent way to reach the renderer. Verified acyclic (`dnd → elements/ElementContainer` only; nothing in `elements/` imports `SortableItem`).
2. **Dedupe `AssetTile` overlay buttons.** The edit/remove hover buttons shared an identical 90-char className and structure; extracted a local `OverlayButton` to remove the duplication and the sync hazard.

## Open PR Overlap Check

- Considered the only open PR: **#371** (`refactor: nightly maintainability 2026-06-25`).
- #371 touches `BottomTransportBar`, `AudioPlayer`, `PlayFromHereButton`, `CharacterEditModal`, `VoicePicker`, `icon-button`, and the three ElevenLabs providers, plus the shared `TooltipIconButton` / audio-format-constant themes.
- This PR touches `Canvas.tsx`, `SortableContent.tsx`, `SortableItem.tsx`, `SortableScene.tsx`, `AssetTile.tsx` — **zero file overlap** and none of #371's themes (no TooltipIconButton wrapper, no ElevenLabs constants).

## Validation

| Check | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass (no new issues) |
| `npm run build` | pass |
| `npm run test:run` | pass (851/851) |

## Skipped Items / Risks

Exploration surfaced many candidates that were deliberately **not** taken because they were speculative, risky, or trivial:
- **Move pure helpers to `lib/` "for reuse"** (`formatAttributeValue`, `getInitial`, `findSegmentIndexAt`, `refineScript`, `useResize`): each has a single consumer today, or — like `findSegmentIndexAt` — is *already* a shared export living in a hook file. Moving them is speculative abstraction (against CLAUDE.md) with no current second caller.
- **Merge `useMetadataSync` + `useScriptSync`**: they solve distinct concerns (store metadata vs editor text); the claimed "redundant subscription" perf win is illusory under React batching, and merging reduces clarity.
- **Collapse `ActiveSceneContext`'s value/setter split**: that split is a deliberate re-render-isolation pattern; collapsing it risks a perf regression.
- **Merge video contexts / extract player-action helpers**: the video context boundaries are intentional and the overlapping logic largely lives in `BottomTransportBar` (off-limits, #371). No safe, non-churny win here.

Risk on the shipped changes is minimal: no behavior change, pure plumbing/dedup, full green CI.

### DIFF
--- a/app/components/canvas/elements/AssetTile.tsx
+++ b/app/components/canvas/elements/AssetTile.tsx
@@ -5,6 +5,27 @@ import { useQueueSelector } from "@/lib/generation/GenerationQueueProvider";
 import { ImageWithShimmer } from "@/lib/components/ImageWithShimmer";
 import { GenerationIndicator } from "./GenerationIndicator";
 
+function OverlayButton({
+	icon: Icon,
+	label,
+	onClick,
+}: {
+	icon: LucideIcon;
+	label: string;
+	onClick: () => void;
+}) {
+	return (
+		<button
+			type="button"
+			onClick={onClick}
+			aria-label={label}
+			className="absolute inset-0 flex items-center justify-center bg-background/70 text-foreground opacity-0 transition-opacity group-hover/tile:opacity-100 focus:opacity-100 focus:outline-none"
+		>
+			<Icon className="h-3.5 w-3.5" />
+		</button>
+	);
+}
+
 export function AssetTile({
 	name,
 	previewUrl,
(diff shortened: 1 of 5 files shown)


umairnadeem (commented):
clean little refactor, dropping the renderCanvasElement prop-drill so SortableItem reaches it the same way DragOverlay already does is the right call, and it's verified acyclic. one forward-looking note on the OverlayButton extraction below. (the focus-ring a11y point is already covered by gemini, not repeating it.)

app/components/canvas/elements/AssetTile.tsx
```diff
@@ -64,24 +85,18 @@ export function AssetTile({
...
-						className="absolute inset-0 flex items-center justify-center bg-background/70 text-foreground opacity-0 transition-opacity group-hover/tile:opacity-100 focus:opacity-100 focus:outline-none"
-					>
-						<Pencil className="h-3.5 w-3.5" />
-					</button>
+					/>
 				)}
 				{onRemove && status !== "generating" && (
-					<button
-						type="button"
+					<OverlayButton
```
umairnadeem, line 95: small footgun now that this is a reusable component: both overlay buttons are `absolute inset-0`, so a tile passed both `onEdit` and `onRemove` would stack them and only the last one (remove) is reachable, edit gets clobbered. no caller does that today, but extracting OverlayButton makes it look composable when it silently isn't. either position them side by side or just leave a breadcrumb that they're mutually exclusive. non-blocking.

PR #377, opened 2026-06-26
refactor: nightly maintainability 2026-06-26

umairnadeem:
## What

Split the 1265-line template monolith (`lib/templates/templates.ts`) into one file per template, with a thin barrel that re-assembles the `TEMPLATES` array. Seven template files extracted:
- `pov-life.ts`
- `sleep-story.ts`
- `finance-tips.ts`
- `true-crime.ts`
- `pov-financial-lifestyle.ts`
- `kids-animated.ts`
- `psychology-of.ts`

Shared types (`Template`, `TemplateShowcase`) and the `templateAsset` helper moved to `lib/templates/types.ts`. The barrel (`templates.ts`) assembles `TEMPLATES`, exposes `getTemplateById`, and re-exports types — all existing consumer imports are unchanged.

## Architecture changes

- **Module boundary**: Template definitions are now self-contained modules rather than entries in a huge array literal. Each template file exports a single `Template` object, making it easy to add/edit/remove templates without touching the barrel.
- **Type extraction**: `Template` and `TemplateShowcase` types extracted to `types.ts`, decoupled from the template data.

## Maintainability

- Reduced `templates.ts` from 1265 lines to a thin barrel (~30 lines).
- Each template is now a focused ~100–200 line file with clear responsibility.
- No behavior change — template content is byte-for-byte identical (verified: all template IDs match, narration line counts are 271 → 271).

## Files changed (9)

All changes confined to `lib/templates/` — no overlap with open PRs.

## Open PR overlap check

Considered open PRs #376, #373, #372, #371. None touch `lib/templates/`. This PR is non-overlapping.

## Validation

- `npm run build` ✓
- `npm test` — 851 tests, 97 files ✓
- `npm run lint` ✓
- `npx tsc --noEmit` ✓
- `knip` (dead code) ✓
- `format:check` ✓

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #376, opened 2026-06-26
fix: nightly bug fixes 2026-06-26

umairnadeem:
## Summary
- Fix blank/whitespace volume and loops attributes being coerced to `0` by `Number("")` instead of falling back to defaults.
- Add optional-chaining guard on `character.appearance?.trim()` in the character-avatar-style plugin to prevent a crash when appearance is missing at runtime.
- Harden `optionalCoercedNumber` Zod schema to reject zero and negative values for image dimensions and video durations.

## Bugs fixed
1. **Blank/whitespace volume → 0**: `Number("")` coerces empty strings to `0`, so persisted blank or whitespace-only volume attributes would mute media instead of defaulting to `10`. Fixed with a `parseFiniteNumber` helper that treats blank/whitespace/missing values as `undefined`, letting `getVolume` and `getLoops` apply their own defaults.
2. **Crash on missing character appearance**: `character.appearance.trim()` threw when `appearance` was `undefined` on characters with avatars. Added `?.` to fall through to the gateway gracefully.
3. **Zero/negative API dimensions/durations accepted**: `optionalCoercedNumber` previously allowed `0` and negative values, which are semantically invalid for width/height/duration. Now rejects with `"must be a positive number"`.

## Tests added / areas covered
- `lib/video/__tests__/elementAttributes.test.ts`: blank/whitespace volume regression (2 new tests)
- `lib/api/__tests__/request-schema-fields.test.ts`: positive/missing/blank/zero/negative/non-finite values for `optionalCoercedNumber` and `optionalDurationSeconds` (7 tests)
- `lib/connectors/llm/plugins/__tests__/character-avatar-style.test.ts`: stored appearance reuse, missing-appearance fallthrough, no-avatar no-op (3 tests)

## Validation
- `npm test -- --passWithNoTests`: 99 files, 862 tests — all pass
- `npm run build`: pass
- `npm run lint`: pass
- `npm run format:check`: pass
- `npm run typecheck`: pass (1 pre-existing unrelated error in character-avatar-style test regarding `GatewayClient` export)
- `npm run knip`: pass

## Open PR overlap check
Considered open PRs:
- #371 (`refactor: nightly maintainability 2026-06-25`): touches BottomTransportBar, AudioPlayer, PlayFromHereButton, CharacterEditModal, VoicePicker, icon-button, and ElevenLabs providers. Zero file overlap.
[...]

PR #373, opened 2026-06-25
fix: prevent delete button from overlapping Settings icon on element cards

umairnadeem:
## Summary

Fix the delete/Settings button overlap on element cards. The delete button (absolutely positioned `top-2 right-2`) could overlap the Settings icon when header badges filled to the right edge or at narrow container widths. Adding `pr-9` (36px) to the header row reserves space matching the delete button's footprint, keeping Settings fully visible and clickable.

## Selected issue

**#362** — "Delete button can overlap and cover the Settings icon in the element card header"

- **Why it was safe/small**: Single-line CSS change (`pr-9`), no behavior or logic changes, no new components. The delete button's position and the header row's flex layout are well-understood, and the 36px reserved space exactly matches the button's occupied area (`right-2` 8px + `h-7` 28px).
- **Acceptance criteria met**:
  - Delete button never overlaps Settings icon at any container width or badge count ✓
  - No layout shift on hover (permanent reserved space, not hover-conditional) ✓
  - Settings icon stays fully clickable while card is hovered ✓

## Changes

- `app/components/canvas/elements/ElementContainer.tsx`: Added `pr-9` to the header row's className (line 96) to reserve space for the absolutely-positioned delete button.

## Validation

| Check | Result |
|-------|--------|
| `npm run lint` | pass |
| `npm run format:check` | pass |
| `npm run typecheck` | pass |
| `npm run knip` | pass (no new issues) |
| `npm run build` | pass |
| `npm run test:run` | pass (97 files, 851 tests) |

## Open PR overlap check

- **#371** (`umair/nightly-maintainability-20260625-0107`): Touches `BottomTransportBar`, `AudioPlayer`, `PlayFromHereButton`, `CharacterEditModal`, `VoicePicker`, `icon-button`, ElevenLabs providers. Zero file overlap with `ElementContainer.tsx`.
- **#372** (`umair/nightly-architecture-rethink-20260625-0302`): Touches `Canvas.tsx`, `SortableContent.tsx`, `SortableItem.tsx`, `SortableScene.tsx`, `AssetTile.tsx`. Zero file overlap with `ElementContainer.tsx`.

## Skipped candidates

- **#317** (Apache-2.0 license): Skipped — legal/compliance decision requiring external contributor sign-off. Runbook excludes legal/compliance judgment.
- **#259** (configurable concurrency): Skipped — gated behind unlaunched BYOK feature; requires product decisions about feature gating.

Closes #362

PR #378, opened 2026-06-26
refactor: rethink openslop architecture 2026-06-26

umairnadeem:
## App understanding

OpenSlop is an AI creative-media studio: a SlateJS storyboard editor + canvas where users describe a story/script/template, the LLM connector streams **OSML** into the canvas, and a generation pipeline (connectors → gateways → providers) renders each storyboard element (image, animated image, clip, narration, music, sound, character) into assets, finally composited to an MP4 via Remotion Lambda. State lives in Zustand (`lib/project/store.ts`) + Slate; generation is queued client-side and polled. Auth/persistence is Supabase with RLS; assets go to Vercel Blob.

**Architectural boundaries** (already audited clean in prior passes): `lib/` is the domain layer and never imports from `app/`; the connector/gateway/provider split is intentional and stays. So this pass targets the **UI layer**, where the real remaining duplication lives.

Asset management (characters, narrator, reference images) is surfaced in two unrelated component trees — the editor's left **assets panel** (`AssetsSection`) and the **composer** (`ComposerCopilot` + `ComposerAssets`). A from-scratch design would own each asset-edit interaction (open create / edit a character / edit the narrator) and each store mutation (remove a reference image) in exactly one place, and let both surfaces call it. The current code instead repeated that wiring verbatim in both trees.

1. **`useAssetEditDialogs` hook** (`app/components/canvas/elements/character/useAssetEditDialogs.tsx`) — owns the create-character / edit-character / edit-narrator dialog state and rendering. Previously `AssetsSection` (3 `useState` + 3 modal renders) and `ComposerCopilot` (the same 3 `useState` + the same 3 modal renders) carried byte-identical boilerplate. Both now call the hook and drop the modal imports. Follows the codebase's existing "hook returns JSX" convention (`useImageUpload` → `inputElement`).
2. **`removeReferenceImage(index)` store action** (`lib/project/store.ts`) — replaces two copies of `setReferenceImages(referenceImages.filter((_, j) => j !== i))` (in `ComposerAssets` and `AssetsSection`) with one store mutation. `ComposerAssets` no longer needs `useConfig`/`getProjectStore` for the removal.

Net: existing files lose 69 lines, gain 33; logic consolidated into one small hook + one store action. No behavior change.

## Validation

- `typecheck` ✅
- `lint` ✅
- `format:check` ✅
- `knip` ✅ (no dead code)
- `build` ✅
- `test:run` ✅ — 852 tests pass (added a store test for `removeReferenceImage`)

## Skipped items (deliberately out of scope)

- Generic attribute/zod-schema centralization and connector/gateway/provider consolidation — prior nightly audits marked these clean; avoided per "no speculative abstractions" guidance.
- Unifying the `AssetsSection` vs `ComposerAssets` tile rows — they differ meaningfully (New-character button + collapsible header vs. upload shimmers), so merging the rows risked behavior change for marginal gain.
- Duplicated `useGenerate()` calls across sibling element components — these are independent Zustand selector subscriptions; collapsing them risks re-render behavior changes.
- Magic-number centralization in video components — cosmetic.

## Open PR overlap check

Considered open PRs #371, #372, #373, #376, #377. This PR touches only `AssetsSection.tsx`, `ComposerCopilot.tsx`, `ComposerAssets.tsx`, a new `useAssetEditDialogs.tsx`, `lib/project/store.ts`, and `lib/project/__tests__/store.test.ts` — none of which appear in those PRs' file lists (#372 covers `AssetTile.tsx` but not `AssetsSection.tsx`; #376 covers `app/api/v1/__tests__/routes.test.ts` and `lib/video`/`lib/api`/`lib/connectors` files, not `lib/project/store.ts`). Non-overlapping.

### DIFF
--- a/app/components/canvas/elements/character/useAssetEditDialogs.tsx
+++ b/app/components/canvas/elements/character/useAssetEditDialogs.tsx
@@ -6,9 +6,8 @@ import { NarratorEditModal } from "./NarratorEditModal";
 import { NewCharacterDialog } from "./NewCharacterDialog";
 
 /**
- * Owns the create-character / edit-character / edit-narrator dialog flow shared
- * by the editor's assets panel and the composer. Returns imperative openers plus
- * the rendered dialogs (mirrors the `useImageUpload` → `inputElement` pattern).
+ * Returns rendered dialogs as JSX (mirrors the `useImageUpload` → `inputElement`
+ * pattern) plus imperative openers.
  */
 export function useAssetEditDialogs() {
 	const [creating, setCreating] = useState(false);
@@ -39,7 +38,7 @@ export function useAssetEditDialogs() {
 
 	return {
 		openCreateCharacter: () => setCreating(true),
-		editCharacter: setEditingName,
+		editCharacter: (name: string) => setEditingName(name),
 		openNarrator: () => setEditingNarrator(true),
 		dialogs,
 	};
--- a/app/components/canvas/elements/character/useAssetEditDialogs.tsx
+++ b/app/components/canvas/elements/character/useAssetEditDialogs.tsx
@@ -1,0 +1,46 @@
+"use client";
+
+import { useState } from "react";
+import { CharacterEditModal } from "./CharacterEditModal";
+import { NarratorEditModal } from "./NarratorEditModal";
+import { NewCharacterDialog } from "./NewCharacterDialog";
+
+/**
+ * Owns the create-character / edit-character / edit-narrator dialog flow shared
+ * by the editor's assets panel and the composer. Returns imperative openers plus
+ * the rendered dialogs (mirrors the `useImageUpload` → `inputElement` pattern).
+ */
+export function useAssetEditDialogs() {
+	const [creating, setCreating] = useState(false);
+	const [editingName, setEditingName] = useState<string | undefined>();
+	const [editingNarrator, setEditingNarrator] = useState(false);
+
+	const dialogs = (
+		<>
+			<NewCharacterDialog
+				open={creating}
+				onOpenChange={setCreating}
+				onCreated={(name) => {
+					setCreating(false);
+					setEditingName(name);
+				}}
+			/>
+			<CharacterEditModal
+				open={editingName !== undefined}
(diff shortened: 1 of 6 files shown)


app/components/canvas/elements/character/useAssetEditDialogs.tsx
```diff
@@ -0,0 +1,46 @@
...
+			<NarratorEditModal
+				open={editingNarrator}
+				onOpenChange={setEditingNarrator}
+			/>
+		</>
+	);
+
+	return {
+		openCreateCharacter: () => setCreating(true),
+		editCharacter: setEditingName,
```
umairnadeem: small consistency nit: the other two openers are wrapped (`() => setX(true)`) but `editCharacter` hands out the raw `setEditingName`, so it leaks the full `Dispatch<SetStateAction<string|undefined>>` signature, callers can pass an updater fn or `undefined` and it'd silently work. maybe `editCharacter: (name: string) => setEditingName(name)` to match the siblings and keep the public surface a plain `(name) => void`. up to you.

app/components/canvas/elements/character/useAssetEditDialogs.tsx
```diff
@@ -0,0 +1,46 @@
+"use client";
+
+import { useState } from "react";
+import { CharacterEditModal } from "./CharacterEditModal";
+import { NarratorEditModal } from "./NarratorEditModal";
+import { NewCharacterDialog } from "./NewCharacterDialog";
+
+/**
+ * Owns the create-character / edit-character / edit-narrator dialog flow shared
```
umairnadeem: nit: first sentence mostly restates the hook name. the keeper here is the `useImageUpload`→`inputElement` breadcrumb (explains the unusual hook-returns-JSX shape), the rest is inferable. could trim to that one line. non-blocking.

umairnadeem (commented):
clean little consolidation, nice that you added a store test for `removeReferenceImage` and the `useImageUpload`→`inputElement` precedent makes the hook-returns-JSX pattern fit right in. just two non-blocking nits below.

PR #381, opened 2026-06-26
fix: use filled square variant for refine copilot bar Stop icon

### LINKED ISSUE #380: Use the filled square variant for the refine copilot bar's Stop icon while loading
## Problem

The refine copilot bar's **Stop** button (shown while a refine is loading) uses an **outline** square instead of a filled one, so the stop affordance reads as hollow.

The call site (`app/components/copilot/InlineCopilot.tsx:98`) is:

```tsx
icon={<Square className="h-3 w-3 fill-current" />}
```

In this codebase icons are **mask-painted spans**, not `<svg>` (`components/ui/icon.tsx:23-46`), so `fill-current` does nothing — it's a leftover from the lucide migration. `Square` maps to `--square-icon` (`components/ui/icons.css:640`), which is an outline glyph. Hence the button looks hollow despite the intent to be filled.

## Reported by (user feedback)

- Use the filled variant for the stop icon in the refine copilot bar when it's loading.

## Fix (the filled glyph already exists)

`components/ui/icons.css:641` already defines `--square-filled-icon` (and `:243` defines a solid `--stop-icon`). There's just no export for it.

- Add `export const SquareFilled = icon("square-filled");` in `components/ui/icon.tsx` (next to `Square`, line 99).
- In `InlineCopilot.tsx:98`, use `<SquareFilled className="h-3 w-3" />` and drop the dead `fill-current`.

## Acceptance criteria

- [ ] The Stop button in the refine bar renders a filled square while loading.
- [ ] The dead `fill-current` class is removed.

---

solarrezaei11:
## Summary

Fixes #380

The Stop button in the refine copilot bar was rendering a hollow square instead of a filled one. The issue had two parts:

1. No export existed for the filled variant — added `export const SquareFilled = icon("square-filled")` to `components/ui/icon.tsx` next to `Square`
2. `InlineCopilot.tsx` was using `<Square className="h-3 w-3 fill-current" />` — switched to `<SquareFilled className="h-3 w-3" />` and removed the dead `fill-current` class (leftover from the lucide migration, has no effect on mask-painted span icons)

The filled glyph (`--square-filled-icon`) already existed in `icons.css` — this just wires it up.

## Changes
- `components/ui/icon.tsx` — export `SquareFilled`
- `app/components/copilot/InlineCopilot.tsx` — use `SquareFilled`, drop `fill-current`

### DIFF
--- a/components/ui/icon.tsx
+++ b/components/ui/icon.tsx
@@ -97,6 +97,7 @@ export const SlidersAlt = icon("sliders-alt");
 export const SlidersAltFill = icon("sliders-alt-fill");
 export const Sparkles = icon("magic");
 export const Square = icon("square");
+export const SquareFilled = icon("square-filled");
 export const Trash2 = icon("trash");
 export const User = icon("user");
 export const UserPlus = icon("user-plus");
--- a/components/ui/icon.tsx
+++ b/components/ui/icon.tsx
@@ -96,7 +96,6 @@ export const SlidersHorizontal = icon("sliders");
 export const SlidersAlt = icon("sliders-alt");
 export const SlidersAltFill = icon("sliders-alt-fill");
 export const Sparkles = icon("magic");
-export const Square = icon("square");
 export const SquareFilled = icon("square-filled");
 export const Trash2 = icon("trash");
 export const User = icon("user");
--- a/app/components/copilot/InlineCopilot.tsx
+++ b/app/components/copilot/InlineCopilot.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { type ReactNode, useEffect, useState } from "react";
-import { CornerDownLeft, Square, Wand2 } from "@/components/ui/icon";
+import { CornerDownLeft, SquareFilled, Wand2 } from "@/components/ui/icon";
 import OrbLoader from "../OrbLoader";
 import { ActionButton } from "./ActionButton";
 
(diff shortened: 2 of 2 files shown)


components/ui/icon.tsx
```diff
@@ -97,6 +97,7 @@ export const SlidersAlt = icon("sliders-alt");
 export const SlidersAltFill = icon("sliders-alt-fill");
 export const Sparkles = icon("magic");
 export const Square = icon("square");
```
umairnadeem: nit, non-blocking: InlineCopilot was the only consumer of `Square`, so after this swap `export const Square` + the `--square-icon` var are now dead. might as well garbage collect them here (fwiw `--square-s-icon` in icons.css is already a byte-identical dupe of `--square-icon`). up to you if you'd rather keep the full icon set around as a library.

umairnadeem (commented):
nice, this is the right fix. these are mask-image icons (`--icon-mask`), so the old `fill-current` was a no-op and `Square` was actually rendering the outline ring, not a solid stop square. `--square-filled-icon` already lives in icons.css so this just wires it up. one heads up: #382 is a byte-for-byte dup of this PR (same 2 files, same change), opened ~15 min later, so one of them should get closed to avoid a double-merge.

umairnadeem (approved):
thank you for contributing!

PR #382, opened 2026-06-26
Use filled square variant for refine copilot bar Stop icon

### LINKED ISSUE #380: Use the filled square variant for the refine copilot bar's Stop icon while loading

WaterXiao-git:
Fix the Stop button in the refine copilot bar showing a hollow outline square instead of a filled square while loading.

- Add SquareFilled icon export
- Use SquareFilled in InlineCopilot.tsx
- Remove dead fill-current class

Closes #380

### DIFF
--- a/components/ui/icon.tsx
+++ b/components/ui/icon.tsx
@@ -97,6 +97,7 @@ export const SlidersAlt = icon("sliders-alt");
 export const SlidersAltFill = icon("sliders-alt-fill");
 export const Sparkles = icon("magic");
 export const Square = icon("square");
+export const SquareFilled = icon("square-filled");
 export const Trash2 = icon("trash");
 export const User = icon("user");
 export const UserPlus = icon("user-plus");
--- a/app/components/copilot/InlineCopilot.tsx
+++ b/app/components/copilot/InlineCopilot.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { type ReactNode, useEffect, useState } from "react";
-import { CornerDownLeft, Square, Wand2 } from "@/components/ui/icon";
+import { CornerDownLeft, SquareFilled, Wand2 } from "@/components/ui/icon";
 import OrbLoader from "../OrbLoader";
 import { ActionButton } from "./ActionButton";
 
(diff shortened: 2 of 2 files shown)


umairnadeem (commented):
heads up, this is a byte-for-byte dup of #381 (same 2 files, identical diff), which was opened ~15 min earlier. left the substantive review over there. the change itself is correct (the old `fill-current` was a no-op on these mask-image icons), but one of these two PRs should get closed so we don't double-merge.

components/ui/icon.tsx
```diff
@@ -97,6 +97,7 @@ export const SlidersAlt = icon("sliders-alt");
 export const SlidersAltFill = icon("sliders-alt-fill");
 export const Sparkles = icon("magic");
 export const Square = icon("square");
+export const SquareFilled = icon("square-filled");
```
umairnadeem, line 100: heads up, this is a dup of #381 (same fix). one diff worth noting: InlineCopilot was the only consumer of `Square` (line 99), so after this swap that export is dead. #381 garbage-collected `Square` (+ the `--square-icon` css var); this keeps it dangling. if 382 is the one that lands, worth gc'ing those too, otherwise probably close in favor of #381.

umairnadeem (commented):
clean little swap, the filled-square stop icon is the right call.

umairnadeem:
@WaterXiao-git this was merged in https://github.com/openslop/openslop/pull/381. please feel free to take on another issue! happy to have contributions from you

PR #383, opened 2026-06-26
Coding conventions: rename to CONVENTIONS.md + enforce across codebase

umairnadeem:
## Summary
Establishes the repo's coding conventions as a canonical doc and enforces them across the codebase.

### Docs
- Rename `CONSTITUTION.md` → `CONVENTIONS.md` (deduped, tightened, no LLM-isms).
- Make `AGENTS.md` canonical (vendor-neutral) with `CLAUDE.md` as a symlink; `AGENTS.md` imports `@CONVENTIONS.md`.

### Enforcement (from a full CONVENTIONS.md audit)
- **HIGH**: fix render-phase store mutation in `ProjectEditor` (`useState` initializer).
- **DRY**: extract `useActiveSegmentIndex` hook and `AuthFooterLink` component.
- **Types**: add `AssetConnectorType`, tighten the connector contract, drop an unsafe cast; remove non-null assertions and ban them via ESLint.
- **Backend**: add `parseSearchParams` + `withPublic` API tiers; `validate-code` now fails loudly (500) on RPC errors instead of masking as 401.
- **lib**: extract `replaceNodeType` in `applyOps`; trim redundant comments.
- **Design system**: sweep raw text sizes → named tiers; add `--overlay`/`--on-media` tokens; compose icon buttons onto `buttonVariants`; elevation shadows; `font-sentient` utility.

### Verification
lint, format, typecheck, knip, build, and all 851 tests pass locally.


lib/api/with-auth.ts
```diff
@@ -31,3 +41,7 @@ export const withApiAccess = (label: string, run: RouteBody) =>
 	);
 
 export const withSession = (label: string, run: RouteBody) => guard(label, run);
+
+// Public routes: no auth, but still get the uniform error envelope.
+export const withPublic = (label: string, run: () => Promise<Response>) =>
```
umairnadeem, line 46: `runGuarded` echoes `stringifyError(error)` straight back to the client (line 18), and `stringifyError` is `JSON.stringify(serializeError(error))` so that's the full message + stack. for `withApiAccess`/`withSession` the caller is at least authed, but `withPublic` now hands that to *anonymous* callers. validate-code is public, so an anon who trips an RPC/infra error gets a 500 body like `validate-code failed: {"message":"rpc failure","stack":...}`. mild info disclosure. maybe public routes return a generic `${label} failed` and keep the serialized detail to the server-side `logger.error` only?

umairnadeem (commented):
nice consolidation pulling the error envelope into `runGuarded` so public routes get the same uniform handling. one security-flavored thing on the new `withPublic` path, inline.

AGENTS.md
```diff
@@ -0,0 +1,30 @@
...
+- After completing code changes, run the following to review against best practices:
+  - `/simplify`
+  - `/vercel-composition-patterns`
+  - `/web-design-guidelines`
+  - `/vercel-react-best-practices`
+
+## Project structure
+
+- Use the `@/*` path alias for imports that traverse 2+ directory levels (e.g. `@/lib/connectors/types`); keep single-level relative imports (`../`) as-is.
+- `lib/` is the domain layer and must never import from `app/` (enforced by ESLint `no-restricted-imports`). Logic needed by both layers belongs in `lib/`.
```
umairnadeem: split-brain: this `lib/` never-imports-`app/` rule now lives in both here and CONVENTIONS.md (`Dependencies point one way...ESLint-enforced`). the whole point of this PR is making CONVENTIONS.md canonical, so having AGENTS.md re-encode the same architecture rules means they can drift. maybe thin this down to the workflow/tooling bits + the `@/*` alias (which is the one rule *not* in CONVENTIONS.md) and let CONVENTIONS.md own the rest? otherwise two sources of truth again.

app/projects/[id]/ProjectEditor.tsx
```diff
@@ -28,7 +28,8 @@ export default function ProjectEditor({
 	initialGeneration: Record<string, ElementSnapshot>;
 	user: User;
 }): ReactNode {
-	applyStoreSnapshot(getProjectStore(projectId), initialStore);
+	// Hydrate the store once, before children render and without re-running each render.
+	useState(() => applyStoreSnapshot(getProjectStore(projectId), initialStore));
```
Comment (umairnadeem) [inline app/projects/[id]/ProjectEditor.tsx:32]: heads up: `useState(() => ...)` runs the initializer once per mounted instance only. ProjectEditor isn't keyed by projectId in page.tsx, and App Router reuses the instance across `/projects/[id]` navigations (no remount on param change). so going A->B updates the projectId/initialStore props but never re-runs hydration, and `getProjectStore(B)` is a fresh unhydrated store -> B renders with empty metadata/referenceImages. the old every-render call masked this via the `hydrated` guard. if full-reload-per-project isn't the assumption, `key={projectId}` on ProjectEditor (or hydrating inside getProjectStore) is the safer fix.

umairnadeem (commented):
nice cleanup overall, pulling the error envelope into `runGuarded` + tightening `ConnectorType` to `AssetConnectorType` are both solid. two new things below worth a look (the with-auth `stringifyError` leak from the earlier pass still stands too). non-blocking aside: the `file-issue` skill feels orthogonal to a conventions-rename PR, might be cleaner to split it out.

PR #388, opened 2026-07-01
refactor: use destructive text/hover treatment for delete buttons

### LINKED ISSUE #341: Use a consistent red/destructive treatment for delete buttons everywhere
## Problem

Delete / destructive actions don't use a consistent red treatment across the app. The shared `DeleteButton` renders with a neutral `bg-muted` and only turns destructive on hover (`hover:text-destructive`), while the character editor uses its own custom rose/red delete button. There's a `destructive` Button variant available but it isn't used. The result is that "delete" doesn't read as red/destructive everywhere, which is inconsistent and reduces clarity for an irreversible action.

- "Delete button color should be red everywhere"

## Relevant code

- `components/ui/delete-button.tsx:7-23` — neutral `bg-muted hover:text-destructive` (not red by default)
- `components/ui/button.tsx:7-38` — defines a `destructive` variant that's effectively unused
- `app/components/canvas/elements/DeleteButton.tsx` — canvas wrapper around the UI delete button
- `app/components/canvas/elements/character/CharacterEditModal.tsx:232-245` — bespoke rose/red delete button (divergent styling)
- Usages: `app/components/canvas/elements/SceneContainer.tsx:83`, `ElementContainer.tsx:95`, `app/components/projects/ProjectsList.tsx:100-103`

## Suggested direction

Standardize all delete/destructive controls on a single red destructive treatment (the `destructive` Button variant / `text-destructive` token), so delete reads as red consistently. Fold the bespoke character-modal delete button into the shared component rather than re-styling per call site.

---

natan497:
## Summary
- Standardize the shared `DeleteButton` (`components/ui/delete-button.tsx`) on the destructive/red treatment (`text-destructive` + `hover:bg-destructive/10` / `dark:hover:bg-destructive/20`) instead of the neutral `bg-muted hover:text-destructive` styling
- This matches the destructive treatment already established in `dropdown-menu.tsx`, and since all delete affordances (`ElementContainer`, `SceneContainer`, `ProjectsList`) share this one component, they all get consistent red delete buttons for free
- The character editor's delete button was already fixed in a prior PR (#322) to use the `destructive` Button variant, so no change was needed there
- **Follow-up:** in `ProjectsList`, the delete button overlays an arbitrary user-uploaded project thumbnail with no other backdrop, so dropping the always-on `bg-muted` there made the red icon low-contrast against light/red-toned/busy thumbnails. Added a card-colored ring/shadow backing there (same pattern as the upload-image button in `CharacterEditModal`) so it stays visible over any thumbnail

Fixes #341

## Test plan
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run test:run` (851 tests passing)
- [x] Verified in a running dev server that the button renders in `--destructive`/`--red-500` (`#e62324`)
- [x] Verified the `ProjectsList` contrast fix against light, red-toned, and dark swatches — icon stays legible in all three

### DIFF
--- a/components/ui/delete-button.tsx
+++ b/components/ui/delete-button.tsx
@@ -10,7 +10,10 @@ export function DeleteButton({
 	return (
 		<TooltipIconButton
 			label="Delete"
-			className={cn("bg-muted hover:text-destructive", className)}
+			className={cn(
+				"text-destructive hover:bg-destructive/10 dark:hover:bg-destructive/20",
+				className,
+			)}
 			{...props}
 		>
 			<Trash2 className="h-4 w-4" strokeWidth={1.5} />
--- a/components/ui/delete-button.tsx
+++ b/components/ui/delete-button.tsx
@@ -11,7 +11,7 @@ export function DeleteButton({
 		<TooltipIconButton
 			label="Delete"
 			className={cn(
-				"text-destructive hover:bg-destructive/10 dark:hover:bg-destructive/20",
+				"bg-muted text-destructive hover:bg-destructive/10 dark:hover:bg-destructive/20",
 				className,
 			)}
 			{...props}
--- a/app/components/projects/ProjectsList.tsx
+++ b/app/components/projects/ProjectsList.tsx
@@ -100,7 +100,6 @@ export default function ProjectsList({
 										<DeleteButton
 											ariaLabel={`Delete ${project.name}`}
 											onClick={() => handleDelete(project.id)}
-											className="bg-card shadow-sm ring-1 ring-border"
 										/>
 									</div>
 								</div>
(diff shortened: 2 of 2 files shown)


umairnadeem:
@natan497 thank you for your contribution!

components/ui/delete-button.tsx
```diff
@@ -10,7 +10,10 @@ export function DeleteButton({
 	return (
 		<TooltipIconButton
 			label="Delete"
-			className={cn("bg-muted hover:text-destructive", className)}
+			className={cn(
+				"text-destructive hover:bg-destructive/10 dark:hover:bg-destructive/20",
```
umairnadeem: nit: can we add back in bg-muted here for consistency with other icon buttons?

umairnadeem (changes requested):
lgtm, one small nit

umairnadeem (approved):

PR #389, opened 2026-07-02
feat: inline character removal in the hero composer

### LINKED ISSUE #340: Inline character removal in the hero composer (no edit-modal round-trip)

natan497:
## Summary
- Character tiles in the hero composer (`ComposerAssets`) now expose an inline remove affordance — no more edit-modal round-trip to delete a character
- Per review, the affordance is the **corner-cross pattern** from the element-container character pills: a small circular X pinned outside the tile's top-right corner, revealed on hover/focus. Extracted into a shared `RemoveCrossButton` used by both `CharacterPill` and `AssetTile`, so the deletion affordance stays consistent across the app by construction
- Deletion routes through a shared `lib/project/deleteCharacter` helper (discard pending avatar job → remove from store), called by both the composer and `CharacterEditModal`, so the two delete paths cannot drift
- Since character deletion is unrecoverable (no undo/redo for project assets), the inline remove asks for confirmation via a new generic `ConfirmDeleteDialog` (built on the existing `AlertDialog` pattern). Project deletion in the gallery — previously an instant hard delete from a hover-revealed button — now goes through the same dialog. Canvas element/scene deletion stays instant since Slate history makes it Ctrl+Z-recoverable; pill removal stays instant since it is scene-scoped and re-addable
Screenshot 2026-07-02 201331

- `AssetTile` overlay buttons gained hover feedback and the standard `focus-ring` (inset) for keyboard navigation, per review

Fixes #340

## Test plan
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run test:run` (852 tests passing)
- [x] Verified in a running dev server: corner-cross reveal on hover/focus, confirm dialog on character delete, project-delete confirmation in the gallery, element/scene delete still instant + undoable

### DIFF
--- a/app/components/copilot/ComposerAssets.tsx
+++ b/app/components/copilot/ComposerAssets.tsx
@@ -1,7 +1,9 @@
 "use client";
 
+import { useState } from "react";
 import { Mic, Palette, User } from "@/components/ui/icon";
 import { AssetTile } from "@/app/components/canvas/elements/AssetTile";
+import { DeleteCharacterDialog } from "@/app/components/canvas/elements/character/DeleteCharacterDialog";
 import { characterAvatarElementId } from "@/lib/project/ensureCharacterAvatars";
 import { useConfig } from "@/lib/config/ConfigProvider";
 import { useGenerationQueue } from "@/lib/generation/GenerationQueueProvider";
@@ -26,8 +28,15 @@ export function ComposerAssets({
 	const narration = useProject((s) => s.metadata.narration);
 	const removeReferenceImage = useProject((s) => s.removeReferenceImage);
 
+	const [deletingName, setDeletingName] = useState<string>();
+
 	const hasNarration = Object.keys(narration).length > 0;
 	const characterEntries = Object.entries(characters);
+
+	const confirmDelete = () => {
+		if (deletingName) deleteCharacter(projectId, queue, deletingName);
+		setDeletingName(undefined);
+	};
 
 	return (
 		<div className="flex flex-wrap gap-2 pb-2">
@@ -47,7 +56,8 @@ export function ComposerAssets({
 					Icon={User}
 					elementId={characterAvatarElementId(name)}
 					onEdit={() => onEditCharacter(name)}
-					onRemove={() => deleteCharacter(projectId, queue, name)}
+					onRemove={() => setDeletingName(name)}
+					removeAffordance="corner"
 				/>
 			))}
 			{referenceImages.map((url, i) => (
@@ -65,6 +75,13 @@ export function ComposerAssets({
 					className="aspect-square w-16 shrink-0 rounded-md shimmer-surface sm:w-20"
 				/>
 			))}
+			<DeleteCharacterDialog
+				name={deletingName}
+				onOpenChange={(open) => {
+					if (!open) setDeletingName(undefined);
+				}}
+				onDelete={confirmDelete}
+			/>
 		</div>
 	);
 }
--- a/app/components/copilot/ComposerAssets.tsx
+++ b/app/components/copilot/ComposerAssets.tsx
@@ -3,6 +3,7 @@
 import { Mic, Palette, User } from "@/components/ui/icon";
 import { AssetTile } from "@/app/components/canvas/elements/AssetTile";
 import { characterAvatarElementId } from "@/lib/project/ensureCharacterAvatars";
+import { useGenerationQueue } from "@/lib/generation/GenerationQueueProvider";
 import { useProject } from "@/lib/project/useProject";
 
 interface ComposerAssetsProps {
@@ -16,13 +17,20 @@ export function ComposerAssets({
 	onEditCharacter,
 	onEditNarrator,
 }: ComposerAssetsProps) {
+	const queue = useGenerationQueue();
 	const referenceImages = useProject((s) => s.referenceImages);
 	const characters = useProject((s) => s.metadata.characters);
 	const narration = useProject((s) => s.metadata.narration);
 	const removeReferenceImage = useProject((s) => s.removeReferenceImage);
+	const removeCharacterFromStore = useProject((s) => s.removeCharacter);
 
 	const hasNarration = Object.keys(narration).length > 0;
 	const characterEntries = Object.entries(characters);
+
+	const removeCharacter = (name: string) => {
+		queue.discard(characterAvatarElementId(name));
+		removeCharacterFromStore(name);
+	};
 
 	return (
 		<div className="flex flex-wrap gap-2 pb-2">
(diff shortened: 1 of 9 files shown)


app/components/copilot/ComposerAssets.tsx
```diff
@@ -31,6 +33,11 @@ export function ComposerAssets({
 			.getState()
 			.setReferenceImages(referenceImages.filter((_, j) => j !== index));
 
+	const removeCharacter = (name: string) => {
```
umairnadeem: this is now the 2nd copy of "how to delete a character" (`CharacterEditModal.handleDelete` does the same `queue.discard(characterAvatarElementId(name))` + `removeCharacter(name)`). split-brain: the day deletion needs more cleanup (e.g. purging the character from scenes/OSML refs) someone updates one site and the other silently leaks orphaned data.

maybe pull it into a small shared helper in lib, e.g. `deleteCharacter(projectId, queue, name)`, and call it from both? keeps it a plain fn (no hook needed) so the modal + composer stay in sync.

umairnadeem (commented):
nice, routing inline removal through the exact same discard+removeCharacter path as the modal is the right call, and folding the overlay buttons into a flex row is clean. one split-brain worth tightening below. (gemini's focus-ring note on the overlay buttons is also fair, wont dupe it.)

umairnadeem:
Thanks for this @natan497! Heads up — this now conflicts with `master`. We just merged #378 (`refactor: rethink openslop architecture 2026-06-26`), which consolidated the asset-tile wiring across the editor and composer: it added a `removeReferenceImage(index)` action to the project store and reworked `ComposerAssets.tsx` so it no longer pulls in `useConfig`/`getProjectStore` for removals.

Since your PR also touches `ComposerAssets.tsx` (and `AssetTile.tsx`), it's now out of date and GitHub shows it as conflicting. Could you rebase onto the latest `master`? While you're at it, it'd be great to route the inline character removal through the same store-action pattern #378 introduced (rather than reaching for `getProjectStore` directly), to keep the two surfaces consistent.

Let me know if you hit any snags with the rebase!

app/components/copilot/ComposerAssets.tsx
```diff
@@ -42,6 +47,7 @@ export function ComposerAssets({
 					Icon={User}
 					elementId={characterAvatarElementId(name)}
 					onEdit={() => onEditCharacter(name)}
+					onRemove={() => deleteCharacter(projectId, queue, name)}
```
umairnadeem: footgun: the inline X deletes immediately, no confirm. `removeCharacter` is a hard `delete` on the store with no undo/history, so one fat-finger on the X permanently nukes the character + avatar. the modal path guards this with a two-step confirm (Delete -> Confirm delete), so we've got inconsistent destructiveness for the same action.

maybe route the inline remove through the same confirm affordance, or at least make it undoable? re-typing a character prompt from scratch is a lot more painful than re-adding a reference image. up to you whether that's blocking.

umairnadeem (commented):
nice, the shared `deleteCharacter` helper is exactly the fix, both sites stay in sync now and the focus-ring/hover cleanup on the overlay buttons lgtm. one new thing on the inline path below.

umairnadeem:
For the inline "delete character" action, I'd prefer we match the pattern used for character badges in the element container headers: a cross icon pinned to the top-right corner of the item, rather than an inline delete button.

Rationale: this makes accidental deletes much less likely. That matters here because we don't currently support undo/redo history for project assets, so a mis-click is unrecoverable. Reusing the corner-cross pattern also keeps the deletion affordance consistent across the app.

app/components/copilot/ComposerAssets.tsx
```diff
@@ -59,6 +75,16 @@ export function ComposerAssets({
 					className="aspect-square w-16 shrink-0 rounded-md shimmer-surface sm:w-20"
 				/>
 			))}
+			<ConfirmDeleteDialog
+				open={deletingName !== undefined}
+				onOpenChange={(open) => {
+					if (!open) setDeletingName(undefined);
+				}}
+				title={`Delete ${deletingName}?`}
```
umairnadeem, line 83: nit: title flashes on close. AlertDialogContent has a `zoom-out-95`/fade exit anim so it stays mounted while animating out, but `confirmDelete`/onOpenChange already cleared `deletingName`, so for that ~150ms the header renders `Delete undefined?`. ProjectsList has the same thing (`Delete ?` via `deleting?.name`).

maybe latch the name so it survives the close (keep the last non-empty value), or drive the title off state that only clears after the anim. not blocking.

umairnadeem (commented):
the shared `deleteCharacter` helper + confirm gate lands both of my earlier notes cleanly, and reusing the blessed `focus-ring` utility (with `ring-inset` on the tile overlay so it isnt clipped) is the right call. one small new thing below.

natan497:
> the shared `deleteCharacter` helper + confirm gate lands both of my earlier notes cleanly, and reusing the blessed `focus-ring` utility (with `ring-inset` on the tile overlay so it isnt clipped) is the right call. one small new thing below.

Just pushed fix. Thanks for your check.
I was a bit busy due to my work. Sorry for the delay.
If this one is merged, I can move to more complex task since I can now almost understand the project.
Thanks again

umairnadeem (approved):
looks great. thanks for the revisions!

Issue #370, opened 2026-06-25
Configurable provider & model per element (character/narration modals + per-element provider)


umairnadeem:
## Problem

There's no UI to choose the **provider** and **model** for an element. Generation already reads `customAttributes.provider` (defaulting to `"openslop"`) and a `defaultModel` from config, and the `ConnectorRegistry` is structured to hold multiple providers/models per connector type — but nothing lets a user actually pick them. In particular:

- The **character edit modal** and **narration edit modal** expose voice settings only; no provider/model controls.
- `ModelBadge` displays the provider/model read-only and is never editable for these element types because provider/model aren't in `visibleAttributes`.
- Provider can't be set per element from any edit surface.

This is **foundational for BYOK (#331)**: once users can bring third-party providers (ElevenLabs, Fal.ai, Anthropic/OpenAI, etc.), they need a place to select which provider/model each element uses. The wiring (storage + dispatch) already exists; the selection UI is the missing layer.

## Reported by (verbatim feedback)

- "model and provider should be configurable on the character edit modal and narration modal. also providers should be configurable per element"

## What already exists (so this is UI-only, not new plumbing)

- Provider/model are already stored in `customAttributes` and read at generation time.
- `ConnectorRegistry` already enumerates `models[]` and a `defaultModel` per provider per connector type.
- Reusable selector primitives exist (`AttributeBadge` enum editor, `EnumField`, read-only `ModelBadge`).

## Relevant code

- `app/components/canvas/elements/character/CharacterEditModal.tsx:38-120` — character modal (appearance + `VoiceSection`); add provider/model controls
- `app/components/canvas/elements/character/NarratorEditModal.tsx:15-48` — narration modal (`VoiceSection` only); add provider/model controls
- `app/components/canvas/elements/character/VoiceMetadataFields.tsx:15-104` — `VoiceSection`; uses `EnumField` already, natural host for the dropdowns
- `app/components/canvas/elements/ModelBadge.tsx:12-30` — currently read-only display; make editable / mirror selection
- `app/components/canvas/elements/AttributeBadge.tsx:26-102` + `.../fields.tsx:77-124` (`EnumField`) — reusable enum/dropdown editors to drive the pickers
- `lib/canvas/elementConfigs.tsx:107-246` — per-type config; `provider`/`model` not in `visibleAttributes`, so no edit surface today
- `lib/generation/buildGenerationJob.ts:8-26` — reads `customAttributes.provider` (default `"openslop"`); already consumes a per-element provider
[...]

Issue #331, opened 2026-06-21
BYOK: bring-your-own-key + third-party gateway (ElevenLabs/Speechify/Gemini TTS, Fal.ai, Anthropic/OpenAI LLMs)


umairnadeem:
## Problem

BYOK (bring-your-own-key) is the **core point of the app**, but there's no way for users to plug in their own provider keys and pick which models/voices/engines to use. Today every connector resolves to a single `openslop` provider — the factory's `PROVIDERS` map is `{ llm: { openslop }, tts: { openslop }, image: { openslop }, ... }` (`lib/connectors/factory.ts`), all routed through the openslop-hosted gateway (`lib/gateway/openslop/*`). There is **no third-party gateway** for talking to external providers directly with a user-supplied key.

Several existing issues already assume BYOK exists but nothing tracks building it:

- #259 (queue concurrency) is explicitly gated "post-BYOK".
- #253 wants ElevenLabs voices as a TTS option.

## Proposed direction

Build BYOK end to end: let users supply their own keys and select providers/models per connector type, routed through a **third-party gateway** (the missing layer) parallel to the existing openslop gateway.

**Providers to support (initial):**

- **TTS:** ElevenLabs, Speechify, Gemini TTS (alongside existing Cartesia)
- **Image/Video:** Fal.ai (alongside existing Runware)
- **LLM:** Anthropic, OpenAI (model selection within each)

**Architecture (respect the existing layering — connectors → gateway → providers; do not flatten it, and keep room for multiple impls per type):**

1. **Third-party gateway layer** (`lib/gateway/`): a sibling to `lib/gateway/openslop/` that calls external providers directly using the user's key, per connector type (llm/tts/image/animated_image/video/sfx/music).
2. **New connector providers**: register additional `ProviderKey`s in `lib/connectors/factory.ts` (e.g. `anthropic`, `openai`, `elevenlabs`, `speechify`, `gemini`, `fal`) backed by the third-party gateway, each under its type's `lib/connectors/<type>/<provider>/` dir like the current `openslop/` impls.
3. **Key management**: secure per-user storage of provider keys + a settings UI to enter keys and select active provider/model per type. Keys must never be logged or sent anywhere but the provider.
4. **Selection UX**: provider + model/voice pickers wired into `ConfigProvider` / `connectorConfig` so the rest of the app picks them up via the existing `getDefaultConnector` path.

## Relevant code

[...]

PR #393, opened 2026-07-06
refactor: resolve element attributes by (connector, provider, model)

### LINKED ISSUE #357: Define configurable attributes per (connector, provider, model) via a class hierarchy, not hardcoded per element type
## Problem

Configurable element attributes (duration, volume, motion, captions, emotion, speed, loops, videoPrompt) are hardcoded **per element type** in `ELEMENT_CONFIGS.visibleAttributes` / `defaultAttributes`. They are keyed only by `CanvasElementType`, which maps 1:1 to a `ConnectorType`. They are completely independent of provider and model.

Different models support different attributes and value ranges. OpenSlop video supports `duration` 4-15s; a future BYOK model (e.g. Veo) may support a different duration set, different params, or entirely different settings. There is currently no place for a provider or model to declare its own attribute set.

The architecture is already provider- and model-aware: `ProviderKey`, `ConnectorConfig` with `apiKey`/`baseUrl` for BYOK, per-model model maps, and a factory keyed by `(type, provider)`. Attribute schemas never followed that structure, so configuration does not scale as we add models and providers via BYOK.

## Relevant code (current state)

- `lib/canvas/elementConfigs.tsx`: `ELEMENT_CONFIGS`, `visibleAttributes` / `defaultAttributes`, `AttributeSpec` / `AttributeEdit`. The single source today, keyed by element type only.
- `lib/canvas/types.ts`: `CanvasElementType`, `DURATION_OPTIONS`, `customAttributes` on `CanvasContentElement`.
- `lib/connectors/types.ts`: `ConnectorType`, `ProviderKey` (`"openslop"` only today), `ConnectorConfig` (`apiKey`/`baseUrl` = BYOK), per-type generate params.
- `lib/connectors/factory.ts`: `PROVIDERS` map (`type` -> `provider` -> constructor), `createConnector`.
- `lib/connectors/<type>/openslop/models.ts`: model string maps (friendly name -> backend id), e.g. `VIDEO_MODELS`.
- `lib/connectors/base.ts`, `asset-base.ts`, `<type>/connector.ts`: the connector class hierarchy.
- `lib/config/ConfigProvider.tsx`: `ConnectorRegistry`, `models[]` per `(type, provider)`.
- `lib/canvas/createCanvasNode.ts`: applies `defaultAttributes` on node creation.
- `lib/canvas/hydrateConnectorConfig.ts`: stamps `model` + `provider` into `customAttributes`.
- `lib/generation/getGenerationInputs.ts` + `generateForElement.ts`: spread `customAttributes` into `connector.generate`.
- `app/components/canvas/elements/ElementContainer.tsx` (`ElementSettings`), `AttributeBadge.tsx`, `ModelBadge.tsx`: the UI consumers.
- `lib/canvas/preservedAttributes.ts`: existing per-type-change attribute preservation (precedent for reconciliation).

## Proposed direction

Resolve the attribute schema by `(connectorType, provider, model)` and co-locate per-model schemas with the provider/model implementations via a clean, fluent, extendable class hierarchy.

1. **Attribute primitives** in `lib/connectors/attributes/`: `AttributeControl` (keep the current two kinds, `enum` and `text`), `AttributeDef { key, label, control, default }`, and `AttributeSchema` as an ordered collection. A small fluent builder for composition: `extend(base)`, `override(key, partial)`, `add(def)`, `remove(key)`, `build()`.
2. **Per-connector-type base schema** in `lib/connectors/<type>/attributes.ts`, holding the current per-type attribute sets. These double as the OpenSlop defaults.
3. **Elevate model definitions**: each model carries `{ name, backendId, schema }`, composed from the type base via the builder. OpenSlop uses the base unchanged; future BYOK models override/add. Keep `models: string[]` in `ConnectorConfig` for compatibility.
4. **Class hierarchy exposes resolution**: a static `attributesFor(model)` on the connector classes resolves from the provider's model registry, defaulting to the type base schema. Provider subclasses supply their own via their model defs.
5. **Single resolver** `resolveAttributeSchema(type, provider, model)` used by UI, node creation, and reconciliation, looking up via the factory's constructor.
6. **Slim `ELEMENT_CONFIGS`** to presentation only (`connector`, `outputKind`, `icon`, `label`, `colorClass`, `iconBgClass`, `placeholder`). Remove `visibleAttributes` / `defaultAttributes`; derive both from the resolved schema.
7. **Remove dead presentation**: `AttributeSpec.color` is unused for enum pills (only the `videoPrompt` text pill in `TextAttributePopover` reads it). Normalize that pill to the standard secondary style and drop the per-attribute `color` field. `icon` stays (ModelBadge synthesizes it).
8. **Reconcile on model/provider change**: when `ModelBadge` switches model (or the element type changes), recompute the schema, drop unknown keys, and fill new defaults. Extend the `preservedAttributes` pattern.
9. **UI**: `ElementSettings` + `AttributeBadge` consume the resolved schema for the element's current `(provider, model)` instead of `config.visibleAttributes`. `ModelBadge` is unchanged except for triggering reconciliation.

## Acceptance criteria

- [ ] Attribute schemas resolve by `(connectorType, provider, model)`; element type is no longer the sole key.
- [ ] Per-model attribute sets live with the model/provider definitions via a class hierarchy + fluent builder.
- [ ] Adding a new model or provider with different attributes requires only a model definition (no edits to UI or `ELEMENT_CONFIGS`).
- [ ] OpenSlop behavior is unchanged: identical attributes, defaults, and options as today.
- [ ] Dead presentation fields removed.
- [ ] Switching a model reconciles attributes (drops unsupported keys, fills new defaults).
- [ ] Lint, format, typecheck, and tests pass; existing attribute tests updated.

## Non-goals

- No `range`/slider control kind yet (enum + text only).
- No new provider or BYOK model shipped in this change; only the extension points so future BYOK config is trivial.
- No backend or gateway changes.

## Related

- BYOK config already exists via `ConnectorConfig.apiKey` / `baseUrl`; this unblocks per-model configuration for those future providers.

---

natan497:
## Summary

Closes #357

Resolves configurable element attributes by **(connectorType, provider, model)** via a class hierarchy, instead of hardcoding them per element type in `ELEMENT_CONFIGS`.

Previously `visibleAttributes`/`defaultAttributes` were keyed only by `CanvasElementType`, even though the real generation dimension is provider- and model-aware. This caused real duplication (`narration` and `character` carried byte-identical `tts` attribute blocks) and left no place for a provider or model to declare its own attribute set — so config didn't scale as BYOK models are added.

## What changed

- **New `lib/connectors/attributes/`** — `AttributeSchema` with a fluent builder (`extend` / `add` / `override` / `remove` / `build`) plus derived `visibleAttributes` / `defaultAttributes`; `common.ts` for attribute leaves shared across types (volume, motion, duration, loops); `reconcile.ts` for model-switch reconciliation.
- **Per-connector-type base schemas** (`lib/connectors/<type>/attributes.ts`) co-located with the connector implementations. The single `tts` schema now backs both `narration` and `character`.
- **Class hierarchy resolution** — each base connector exposes `static attributesFor(model)`; `llm` inherits an empty default. `resolveAttributeSchema(type, provider, model)` on the factory resolves via the same `PROVIDERS` map `createConnector` uses.
- **Slimmed `ELEMENT_CONFIGS`** to presentation only; `createCanvasNode`, `ElementSettings`, and `ModelBadge` derive attributes from the resolved schema. `hydrateConnectorConfig` folded into `createCanvasNode`.
- **Model-switch reconciliation** — changing a model drops attributes the new schema doesn't support and fills its new defaults, without touching unrelated keys (`model`, `provider`, `characters`).

Adding a new model or provider with different attributes now requires only a model/schema definition — no edits to the UI or `ELEMENT_CONFIGS`.

## Acceptance criteria

- [x] Attribute schemas resolve by `(connectorType, provider, model)`; element type is no longer the sole key
- [x] Per-model attribute sets live with the model/provider definitions via a class hierarchy + fluent builder
- [x] Adding a new model/provider requires only a model definition (no UI / `ELEMENT_CONFIGS` edits)
- [x] OpenSlop behavior unchanged — identical attributes, defaults, and options
- [x] Dead presentation fields removed (`AttributeSpec.color` was already absent)
- [x] Switching a model reconciles attributes (drops unsupported keys, fills new defaults)
- [x] Lint, format, typecheck, and tests pass; attribute tests added/updated

## Test plan

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run test:run` (862 tests passing; +10 new covering the schema builder, resolver, and reconciliation)

## Notes for reviewers

OpenSlop ships one model per connector type today, so the builder's `extend`/`override` methods and the reconciliation path have no visible in-app effect yet — they're the extension points for future BYOK models (#331, #370). Both are covered by unit tests using synthetic schemas.


app/components/canvas/elements/ElementContainer.tsx
```diff
@@ -31,7 +35,15 @@ function ElementSettings({
 	element: CanvasContentElement;
 	config: ElementConfig;
 }) {
-	const entries = Object.entries(config.visibleAttributes);
+	const { connectorConfig } = useConfig();
+	const { model, provider } = element.customAttributes ?? {};
+	const fallback = getDefaultConnector(connectorConfig, config.connector);
+	const schema = resolveAttributeSchema(
```
umairnadeem, line 51: heads up, this can throw now. resolveAttributeSchema does `if (!Ctor) throw` on an unknown provider, and here `provider` comes straight off persisted `element.customAttributes`. old code just read `config.visibleAttributes` which could never blow up. so an element saved with a provider that later got renamed/removed from the registry crashes the settings popover on render instead of degrading. the `?? fallback.provider` only catches nullish, not invalid. maybe fall back to `fallback.provider` when the stored provider isn't in the registry?

app/components/canvas/elements/ModelBadge.tsx
```diff
@@ -26,5 +26,31 @@ export function ModelBadge({ element }: { element: CanvasContentElement }) {
 	}, [connector, connectorConfig, model, provider]);
 
 	if (!spec) return null;
-	return <AttributeBadge element={element} attrKey="model" spec={spec} />;
+
+	const reconcileForModel = (next: string) => {
```
umairnadeem, line 37: afaict this whole reconcile path is inert today: `attributesFor` ignores `model` everywhere, so oldSchema and newSchema are always identical for the same (connector, provider), and reconcileAttributes returns `{}`. fine to build the seam ahead of per-model schemas, just flagging it's currently a no-op so nobody assumes model-switch reconciliation actually runs. also when it does light up: reconcile only diffs key presence, not enum-value validity, so `emotion:'happy'` that's valid on model A but not model B survives the switch. worth handling when per-model sets actually land.

lib/connectors/attributes/schema.ts
```diff
@@ -0,0 +1,90 @@
...
+/** An ordered, immutable set of attribute definitions for a connector type/model. */
+export class AttributeSchema {
+	private constructor(private readonly defs: readonly AttributeDef[]) {}
+
+	static from(defs: readonly AttributeDef[]): AttributeSchema {
+		return new AttributeSchema(defs);
+	}
+
+	/** Layer `overrides` on top of `base` (same-key defs in overrides win, new keys append). */
+	static merge(
```
umairnadeem: YAGNI-ish: `merge`/`extend`/`add`/`override`/`remove`/`build` are all test-only right now, nothing in prod calls them (only `from`/`keys`/`visibleAttributes`/`defaultAttributes` are actually used). that's a lot of combinator API + tests to carry for a shape that today is just `AttributeSchema.from([...])`. i'd trim to what's used and add these back when a connector genuinely needs to compose schemas. `build()` returning `this` especially is dead weight.

lib/connectors/base.ts
```diff
@@ -20,6 +21,15 @@ export abstract class BaseConnector<
 	abstract readonly type: ConnectorType;
 	protected plugins: ConnectorPlugin[];
 
+	/**
+	 * Attribute schema for this connector type/model. Types with no element-settings
+	 * UI (e.g. llm) inherit this empty default. For per-model sets, override on the
+	 * provider subclass and branch on `model` (e.g. `OpenSlopTTS.attributesFor`).
```
umairnadeem: nit: this points at `OpenSlopTTS.attributesFor` as the example but that override doesn't exist anywhere yet (no provider subclass overrides attributesFor, and `_model` is ignored in every impl). a breadcrumb to code that isn't there reads as stale/aspirational. either drop the real per-model example in when it lands, or soften to 'branch on `model`' without naming a specific class.

app/components/canvas/elements/AttributeBadge.tsx
```diff
@@ -21,13 +21,16 @@ interface AttributeBadgeProps {
 	attrKey: string;
 	spec: AttributeSpec;
 	hideLabel?: boolean;
+	/** Extra attrs to merge alongside the new value (e.g. schema reconciliation on model change). */
+	onSelect?: (next: string) => Record<string, string | null>;
```
umairnadeem: small nit: `onSelect` reads like a side-effecting event handler but it's really 'compute extra attrs to merge'. something like `deriveExtraAttrs`/`extraAttrsFor` would signal it returns a value rather than performs an action.

umairnadeem (commented):
nice refactor. moving the attribute schema onto the connector classes and resolving via `resolveAttributeSchema(type, provider, model)` is a clean way to gut the elementConfigs god-object, and the test coverage on schema/reconcile is solid. main thing i'd push on: there's a decent amount of speculative surface here (the whole `model` dimension + the fluent-builder API) that's currently inert, plus one resilience regression in the settings popover. details inline.

app/components/canvas/elements/ElementContainer.tsx
```diff
@@ -31,7 +35,21 @@ function ElementSettings({
...
 	config: ElementConfig;
 }) {
-	const entries = Object.entries(config.visibleAttributes);
+	const { connectorConfig } = useConfig();
+	const { model, provider } = element.customAttributes ?? {};
+	const fallback = getDefaultConnector(connectorConfig, config.connector);
+	// A persisted provider that's since been renamed/removed from the registry
+	// must not reach resolveAttributeSchema, which throws for unknown providers.
+	const resolvedProvider =
+		provider && provider in connectorConfig[config.connector]
```
umairnadeem: the guard checks `connectorConfig` membership, but the throw you're actually dodging lives in `resolveAttributeSchema`, which keys off the separate `PROVIDERS` map in factory.ts. those are two independently-maintained registries (both openslop-only today, but multi-provider is on the roadmap). so a provider thats present in connectorConfig yet dropped from PROVIDERS still sails past this check and crashes in resolveAttributeSchema, which is basically the exact 'removed from the registry' case this was meant to fix. cleaner to guard against the map that actually throws, e.g. expose an `isKnownProvider(type, provider)` from factory.ts, or just don't throw in resolveAttributeSchema and fall back there. not blocking while they're in sync, just flagging the split-brain before they drift.

umairnadeem (commented):
nice follow-ups, the deriveExtraAttrs rename and trimming the unused builder API both landed clean. one thing on the new provider guard.

lib/connectors/factory.ts
```diff
@@ -42,6 +43,14 @@ const PROVIDERS: Record<
 	video: { openslop: OpenSlopVideo },
 };
 
+/** Whether `provider` has a registered constructor for `type` — the same map `createConnector`/`resolveAttributeSchema` key off. */
+export function isKnownProvider(
+	type: ConnectorType,
+	provider: string,
+): provider is ProviderKey {
+	return provider in PROVIDERS[type];
```
umairnadeem: SG/non-blocking: since this is a defensive guard over persisted `customAttributes`, `in` walks the prototype chain, so a stored provider of `constructor`/`toString`/`hasOwnProperty` returns true here and sails through. then resolveAttributeSchema does `PROVIDERS[type][provider]` = `Object.prototype.constructor` (truthy, so no throw) and calls `.attributesFor` on it -> TypeError, the exact popover crash this guards against. `Object.hasOwn(PROVIDERS[type], provider)` sidesteps it. realistic provider names never hit this so not blocking, but it's cheap to close.

umairnadeem (commented):
isKnownProvider closes the split-brain nicely, guarding against the map that actually throws. one small hardening note inline.

umairnadeem (approved):
lgtm - very clean. thanks for the revisions!

Issue #269, opened 2026-06-12
Image generation failures: missing prompts, content-policy rejections, misspelled in-image text


umairnadeem:
## Problem

Three image-generation failure/quality modes, all currently fixed only by manual prompt rework:

- **Missing prompts:** the script occasionally produces image nodes with no prompt, so the image can't be generated. Workaround was refining the script and asking for a prompt for the specific scene.
- **Content-policy rejections:** the generated prompt sometimes trips the provider's content policy and the user gets a red error in place of the image; slightly rewording the prompt fixes it, but the app doesn't retry or reword automatically.
- **Misspelled text:** text rendered inside generated images is sometimes totally misspelled.

## Reported by (verbatim feedback)

- "Occasionally the prompt on some images were missing, so the image couldn't be generated. fixed by refining the script asking for the prompt to work on specific scene"
- "also sometimes I received a red prompt image telling that the prompt generated could violate some policies, fixed by changing the prompt slightly"
- "also some texts are totally misspelled"

## Relevant code

- `lib/generation/getPromptText.ts`, `lib/generation/getGenerationInputs.ts` — no empty-prompt validation; empty prompts flow straight to the connector
- `lib/providers/image/runware.ts` — provider call; policy rejections surface as raw errors
- `lib/connectors/base.ts` — `errorMessage()` / error propagation
- `app/components/canvas/elements/preview/overlays.tsx` — red error banner UI

## Suggested direction

- Validate prompts at script-generation time; auto-(re)generate a prompt for any image node that's missing one instead of failing.
- On a policy rejection, auto-reword the prompt via the LLM and retry once before surfacing a clear, actionable error.
- Discourage in-image text in prompts (or overlay real text instead) to avoid misspellings.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #394, opened 2026-07-07
feat: bring-your-own-image for image elements

### LINKED ISSUE #374: Use your own uploaded image as a scene's visual (bring-your-own-image, bypass generation)
## Problem

Every image node's visual comes only from AI generation — the result is `AssetResult.imageUrl`, set by `generateForElement` and read by the preview. There is **no way to use an image the user already has as the actual scene visual**. The only upload path treats images as project-level *style references* (`referenceImages`), which influence generation but never become an element's output.

A common, painful case: AI-generated **in-image text comes out garbled** (see #269), and the user has a correct version of that image — but can't drop it in.

- "That would be great to be able to upload our images, some images with text I generated, the text is a complete mess 😰"

## What's missing (no existing path)

- No field marking an image element as user-provided / storing a source image URL (`CanvasContentElement.customAttributes` has nothing for this).
- No way to set an element's `result` from an uploaded URL — results only come from `generateForElement` via the queue; there is no "set result" API.
- No UI affordance to "upload instead of generate" on an image node.

## Relevant code

- `lib/upload/useImageUpload.tsx`, `lib/upload/uploadImage.ts`, `app/api/upload/image/route.ts` — upload returns a hosted `{ url }`, immediately reusable
- `lib/connectors/types.ts:59-65` — `AssetResult` (`imageUrl`); the shape an uploaded image would populate
- `lib/generation/queue.ts:18-25,234-245,268-308` — `ElementSnapshot.result`; `restoreResult` requires matching gen inputs, so a direct set-result path is needed
- `app/components/canvas/elements/OutputPreview.tsx:40-49`, `lib/connectors/assetUrl.ts` — preview reads `result.imageUrl`
- `lib/canvas/createCanvasNode.ts`, `lib/canvas/elementConfigs.tsx`, `lib/canvas/editorOps.ts` — element creation + `customAttributes`

## Suggested direction

- Add an "Upload image" affordance on image elements that stores the uploaded hosted URL as the element's source and sets `result.imageUrl` directly, bypassing generation.
- Persist a `sourceImageUrl` / "user-provided" state so the node isn't clobbered by Generate All / regeneration unless the user opts in.
- Reuse the existing `useImageUpload` + `/api/upload/image` path; only the set-result mechanism on the queue is new.

## Acceptance criteria

- [ ] User can upload their own image into an image node and have it shown as that scene's visual.
- [ ] An uploaded image is not overwritten by Generate All / regeneration unless the user chooses to regenerate.
- [ ] The uploaded image flows through to the final render.

## Related

- #269 (garbled in-image text) — this is the direct workaround users are asking for
- #249, #297 (uploading reference images) — distinct: those feed *style*, this sets the actual visual

---

natan497:
## Summary

Closes #374

Lets users upload their own image into an `image` element and have it become that scene's visual, bypassing AI generation entirely. Motivating case: AI-generated in-image text (see #269) often comes out garbled, and users want to drop in a correct version of the image instead of fighting the prompt.

## What changed

- **`GenerationQueue.commitResult`** (`lib/generation/queue.ts`) — writes a result directly into the queue's state (result, `resultInputs`, provenance), clears `error`, resets `status` to `idle`, and records it in `history` so a later prompt edit-and-revert can restore it via the existing `restoreResult`. `handleJobSuccess` delegates to it too, so there's one write path whether the result came from a finished job or a caller handing one in. The caller passes provenance as `{ connectorType, uploaded }` and must `cancel()` any in-flight job first, or a late-resolving job can clobber the committed result. `connectorType` matters because `pickThumbnailUrl` filters on it — without it an uploaded-only project's card stays blank.
- **`uploaded` provenance on `ElementSnapshot`** — marks a result the user supplied rather than one a job produced. `useGenerateAll` skips uploads, so unrelated input drift (project aspect ratio, a tagged character's avatar) can't make Generate All silently regenerate over one. `isStaleResult` stays a pure inputs comparison, so the "prompt changed" staleness badge keeps telling the truth on an uploaded element — same split the character path already uses (`!character.avatarUploaded && isStaleResult(...)`). Provenance is stored in `history` next to the result it describes, so a prompt revert restores both together and can't launder an upload into a regeneratable result. The flag persists through autosave/reload and survives `resetToIdle` (a "Stop All").
- **The skip is gated on the queue entry actually belonging to the element's current connector** (`snap.connectorType === ELEMENT_CONFIGS[el.type].connector`). Animate rewrites an element's type in place and keeps its id, so a stale entry — its `uploaded` flag included — can be provenance for a connector the element no longer uses. Comparing connectors catches that in one check for any type change, not just the uploaded case, and doesn't require every reader of `snap.uploaded` to separately know Animate is the reason a leftover entry can outlive its element's type.
- **A failed generation no longer destroys the existing result** — `handleJobError` now sets only `error`, keeping `result`/`uploaded`. Previously a failed regenerate over an upload nulled the result, and since `history` isn't persisted and `restoreResult` can't fire once `result` is nil, the blob URL was gone for good. To keep that from turning a visible failure into a silent one, errors now render alongside a result too — as a compact `ErrorBadge` rather than the placeholder's full-bleed `ErrorMessage`, since over a live asset a transient failure is a status, not a modal (the takeover both hid the image and swallowed clicks meant for the player underneath). `error` is also dropped on rehydrate, like `status`/`seconds` — a past session's failure shouldn't come back sitting on a good result.
- **`error` now carries its own provenance** (`errorInputs`), mirroring `resultInputs`. Keeping the result on failure exposed a latent bug: generate at A, edit to B, fail, revert to A left the failure sitting on a result that was current and correct (the character modal even ships a Revert button that triggers it). `errorForInputs(snapshot, currentInputs)` derives the visible error, so editing away from the inputs a failure was thrown for retires it. Clearing on `!isStale` instead would have been a regression — a same-prompt regenerate that fails is also not stale, so the failure would vanish before the user saw it.
- **Both error surfaces are announced and reachable** — the badge truncates, so its trigger is a real focusable `<button>` carrying the message in its accessible name, with the message in a live region. The placeholder path (the *more* common failure — an element that never generated has no result) gets the same treatment. Copy confirmation lives in a polite `role="status"` region rather than swapping the trigger's `aria-label`, so the button keeps its name and copying can't re-announce the error assertively.
- **`UploadImageButton`** (`lib/upload/UploadImageButton.tsx`, new) — shared presentational upload button used by both the canvas `image`-element flow and the character-avatar upload, replacing two near-identical hand-rolled implementations. Shows a spinner and changes its accessible name while uploading. It reads its `onUpload` callback from a ref at commit time: an upload outlives the render that started it, so the captured callback would otherwise commit against stale inputs (badging a fresh upload as stale, and silently clobbering character edits made while it was in flight).
- **`ElementUploadButton`** (new) — wires it to the queue for `image` elements. Narrowed to `{ type: "image" }` via the new `isImageElement` guard, so the compiler (not just the call-site gate) enforces that it can't run for a non-image element.
- **Removed `queue.setError`** — its only caller was `useGenerate.generate()`'s empty-prompt guard, unreachable because `GenerateButton` already ships `disabled={!hasPrompt || status !== "idle"}` (CONVENTIONS: *"Define errors out of existence where you can"*).
- **`next.config.ts`** — `images.remotePatterns` now derives hostnames from `NEXT_PUBLIC_BLOB_URL` in addition to the committed fixtures bucket, so an optimized `<Image>` (e.g. `ProjectsList` thumbnails) doesn't 404 on deploys using their own blob store. Parses with `new URL` in a try/catch (works on the Node floor Next 16 supports; `URL.parse` needs 20.18+) and throws an error naming the env var, and rejects non-http(s) rather than casting the protocol.

Scoped to `image` elements only — `animated_image` generation is one atomic call that chains still-image gen into video gen, so there's no clean seam to inject an uploaded still frame into half that chain. Animating an uploaded image therefore regenerates the still from the prompt; making it animate the user's image is a feature, not part of this change.

## Acceptance criteria (from #374)

- [x] User can upload their own image into an image node and have it shown as that scene's visual.
- [x] An uploaded image is not overwritten by Generate All / regeneration unless the user chooses to regenerate — enforced in `useGenerateAll`, covered for the real-prompt-with-attribute-drift case and the revert-through-history case, plus the Animate-conversion case that must *not* be skipped.
- [x] The uploaded image flows through to the final render — `lib/video/resolve.ts` only checks `snapshot.result`, no provenance check.
- [x] An uploaded-only project (never generated) still gets a project-grid thumbnail — covered end-to-end through the real `pickThumbnailUrl`.
- [x] A failed regenerate never loses the upload.

## Test plan

- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run typecheck`
- [x] `npm run knip`
- [x] `npm run build`
- [x] `npm run test:run` (878 tests passing)
- [x] Manual verification: uploaded a PNG on an `image` element, confirmed it renders immediately, Generate switches to Regenerate, the upload survives Generate All after an aspect-ratio change and across a prompt edit-and-revert, the project-grid thumbnail shows without ever generating, and it persists across reload.


natan497:
Thanks for the review — fixed the three real issues:

- `UploadImageButton` is now gated at the `ElementContainer` call site (`element.type === "image" &&`) instead of mounting unconditionally and early-returning internally.
- `setManualResult` now calls `this.cancel(elementId)` first, so an in-flight generation job cannot resolve afterwards and silently overwrite a just-uploaded image. Added a test covering this.
- `ImageWithShimmer`'s unoptimized path now forwards standard img props (width/height/sizes/style/etc.) instead of dropping them; Next-only props (quality, priority, placeholder, blurDataURL, etc.) stay routed to `next/image` only in the optimized branch.

On the `tooltip` prop comment: not applied — this repo's `Button` component (`components/ui/button.tsx:46-63`) has a custom `tooltip` prop that wraps children in `SimpleTooltip` internally, so this is not a bug. Same pattern already used unmodified in `AnimateButton.tsx` and `GenerateButton.tsx`.

umairnadeem (commented):
nice feature, and the queue side is well done: setManualResult faithfully mirrors handleJobSuccess (history write + same snapshot shape) and cancelling the in-flight job to avoid a late generate clobbering the upload is exactly the race i'd worry about, with a test to match. two things on the ImageWithShimmer split inline.

lib/components/ImageWithShimmer.tsx
```diff
@@ -3,21 +3,86 @@
...
+	if (unoptimized) {
+		return (
+			<>
+				{/* eslint-disable-next-line @next/next/no-img-element */}
+				<img
+					{...props}
+					src={typeof src === "string" ? src : undefined}
+					alt={alt}
+					className={cn(fill && "absolute inset-0 h-full w-full", className)}
+					onLoad={handleLoad}
```
umairnadeem: the plain `` path relies solely on `onLoad` to clear the shimmer, but that's the one thing next/image doesn't: it uses a ref callback that checks `if (img.complete)` and fires the load handler manually, precisely because for an already-cached src the `load` event may never fire. so here: revisit a scene whose uploaded/mock image is already in the browser cache, `onLoad` doesn't fire, `loaded` stays false, and the `absolute inset-0` Skeleton sits on top of the image forever. and cached is the common case for exactly these unoptimized fixtures. mirror the guard, e.g. a ref callback that does `if (el?.complete) setLoaded(true)`.

lib/components/ImageWithShimmer.tsx
```diff
@@ -3,21 +3,86 @@
...
+				quality={quality}
+				preload={preload}
+				priority={priority}
+				placeholder={placeholder}
+				blurDataURL={blurDataURL}
+				overrideSrc={overrideSrc}
+				onLoadingComplete={onLoadingComplete}
+				layout={layout}
+				objectFit={objectFit}
+				objectPosition={objectPosition}
```
umairnadeem: this hand-maintained allowlist is a bit of a maintenance trap: you destructure ~14 props only to strip them off the ``, then re-list every one back onto `<Image>` (which `{...props}` would already carry). two failure modes: a new next-only prop lands in `...props` and leaks onto the DOM `` as an invalid attr, and it's just a lot of noise to keep in sync. simpler: destructure only what you actually special-case (`src/alt/onLoad/unoptimized/fill/className`), keep the rest in `...props`, pass `{...props}` straight to `<Image>`, and for the img omit a small `NEXT_ONLY` key list. up to you, not blocking.

lib/components/ImageWithShimmer.tsx
```diff
@@ -1,23 +1,89 @@
...
+export function ImageWithShimmer({
+	alt,
+	onLoad,
+	unoptimized,
+	fill,
+	className,
+	src,
+	...props
+}: ImageProps) {
 	const [loaded, setLoaded] = useState(false);
```
umairnadeem, line 11: nit, non-blocking: `loaded` is a plain bool that only ever flips true, so it never resets when `src` changes in place. MediaWithSkeleton renders this without a `key`, so on regenerate the new image swaps in with no shimmer. fwiw umair's #399 just fixed the exact sibling case for the video branch (`loadedVideoSrc === src`) to re-show the shimmer during reload, so this path is now inconsistent with that intent. cheap fix if you want to match: `const [loadedSrc, setLoadedSrc] = useState<string|null>(null)` and derive `const loaded = loadedSrc === src`. up to you though, showing the old image until the new one paints instead of a shimmer flash is arguably fine.

umairnadeem (commented):
the follow-ups landed well: the ref `complete`-on-mount check kills the stuck-shimmer-on-cached-image case, and centralizing the next-only props into a named list reads much better than the destructure pile. one small consistency note below, non-blocking.

lib/generation/queue.ts
```diff
@@ -231,6 +231,27 @@ export class GenerationQueue {
 		this.notify();
 	}
 
+	setManualResult(
```
umairnadeem: `setManualResult` is basically `handleJobSuccess` minus the abort check: same history.set, same `update` payload. two things point the same direction:

1. **dedupe** it. extract a dumb private `commitResult(elementId, result, inputs)` that does the history write + the `update`, and have both call it. `handleJobSuccess` = abort check → `commitResult`; this path = `cancel` → `commitResult`.
2. **altitude:** ideally the queue doesn't know "manual" vs generated, it just stores a result someone handed it. the only upload-specific bit here is the leading `this.cancel()`, which is really a caller concern (upload wants to kill any in-flight gen first). so keep `commitResult` provenance-agnostic and let the button do `queue.cancel(id); queue.commitResult(...)`. keeps the queue as dumb as possible.

up to you on exact shape, but the copy-paste + the `manual` naming both smell like the same thing.

app/components/canvas/elements/UploadImageButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+	const status = useQueueSelector(
+		(q) => q.getElementSnapshot(element.id).status,
+	);
+	const metadata = useProjectStore(projectId, (s) => s.metadata);
+
+	const { openPicker, uploading, inputElement } = useImageUpload({
+		onUpload: ([url]) => {
+			if (!url) return;
+			const inputs = getGenerationInputs(element, metadata);
+			queue.setManualResult(
```
umairnadeem: heads up we now have two "upload your own image instead of generating" UIs: this one, and the avatar upload in `CharacterEditModal`. and they persist differently, the character path writes the url into metadata (`avatarUrl` + `avatarUploaded` flag) and `queue.discard()`s the entry, this one writes it as the queue `result` via `setManualResult`. so "is this image user-uploaded?" is answered two different ways now, classic split-brain that'll drift.

the shells are near-identical too (both are `useImageUpload` + `ImagePlus` + uploading state, just different button chrome + a different `onUpload` target). worth pulling the presentational upload button into one shared component that takes `onUpload`, so there's a single upload affordance both call sites use. not asking you to merge the two *persistence* models here (image elements have no metadata slot, so queue-result is reasonable for them), but the divergence deserves a shared component at minimum, or an issue tracking it.

lib/components/ImageWithShimmer.tsx
```diff
@@ -1,23 +1,93 @@
...
+	// A cached image may never fire `load` (next/image does this same check).
+	// Keyed on `src` so a swap re-runs the check against the new image.
+	const checkAlreadyLoaded = useCallback(
+		(img: HTMLImageElement | null) => {
+			if (img?.complete) setLoadedSrc(src);
+		},
+		[src],
+	);
+
+	if (unoptimized) {
```
umairnadeem: do these need to ride along with BYOI? an uploaded image renders through the exact same `MediaWithSkeleton` → `unoptimized`+`fill` path a generated image does (has since #214), so the `EncodingError` noise and the cached-image stuck-shimmer are both pre-existing and orthogonal to this feature. only the `loadedSrc` swap-in-place re-show looks actually motivated by uploading over an existing result.

3 behavior changes (the plain-`` branch, the `complete` ref check, the src-keyed shimmer) landing with zero tests on this component makes me a bit nervous, esp the cached/`complete` path. either split the two orthogonal fixes into their own PR (keeps this one about the feature), or add a couple tests for the unoptimized branch + the swap-in-place reshow. wdyt?

umairnadeem (commented):
nice feature, and the in-flight-job `cancel()` so a gen that resolves after the upload can't silently clobber it is a genuinely good catch (and well tested). a few structural things worth talking through before merge:

- **is all of it needed?** mostly yes, imo. image elements have no metadata slot the way characters do, and there was no public queue API to inject a result, so *some* queue-write path is legit (not redundant with the character-modal upload). but it can be a lot leaner: `setManualResult` duplicates `handleJobSuccess` and teaches the queue about "manual" (see inline). and the `ImageWithShimmer` changes are orthogonal pre-existing fixes riding along (see inline).
- **split-brain:** we now have two different "upload your own image" UIs that persist provenance two different ways. worth unifying the button at least (inline on `UploadImageButton`).

natan497:
> nice feature, and the in-flight-job `cancel()` so a gen that resolves after the upload can't silently clobber it is a genuinely good catch (and well tested). a few structural things worth talking through before merge:
> 
> * **is all of it needed?** mostly yes, imo. image elements have no metadata slot the way characters do, and there was no public queue API to inject a result, so _some_ queue-write path is legit (not redundant with the character-modal upload). but it can be a lot leaner: `setManualResult` duplicates `handleJobSuccess` and teaches the queue about "manual" (see inline). and the `ImageWithShimmer` changes are orthogonal pre-existing fixes riding along (see inline).
> * **split-brain:** we now have two different "upload your own image" UIs that persist provenance two different ways. worth unifying the button at least (inline on `UploadImageButton`).

Thanks for your careful review. I just pushed the fix again.

app/components/canvas/elements/ElementUploadButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+	const busy = status === "generating" || status === "queued";
+
+	return (
+		<UploadImageButton
+			className="shrink-0"
+			disabled={busy}
+			onUpload={(url) => {
+				// Kill any in-flight generation so a late result can't clobber the upload.
+				queue.cancel(element.id);
+				queue.commitResult(
```
umairnadeem, line 34: this commits a result but never sets `connectorType`, and `pickThumbnailUrl` filters on exactly that:

```ts
// lib/project/thumbnail.ts
if (snap.connectorType !== "image" && snap.connectorType !== "animated_image") continue;
```

so: new project, drop an image element, upload your own image, never hit generate. autosave writes `thumbnail_url: null` and the project card stays blank even though the project has a visual. confirmed locally, `commitResult` leaves `connectorType: null` and `pickThumbnailUrl` returns null for that snapshot.

the invariant that broke is subtle: before this PR `result != null` implied a job ran, and `startJob` set `connectorType` on the way in. `commitResult` is the first path that produces a result without a job, so it has to carry the type itself. cheapest fix is to have the caller hand it in (it already knows: `ELEMENT_CONFIGS[element.type].connector`, same as `buildGenerationJob`), e.g. `commitResult(id, result, inputs, connectorType)`.

side note, this is now the second unwritten obligation on `commitResult` callers (cancel first, and set the type). worth a line in the doc comment on the queue method so the next caller doesn't miss both.

lib/blob.ts
```diff
@@ -1,4 +1,7 @@
-// Public Vercel Blob store hosting committed mock/template fixtures. Kept as a
-// literal (not NEXT_PUBLIC_BLOB_URL) so these fixtures resolve regardless of env.
+// Public Vercel Blob store hosting committed mock/template fixtures. Defaults
+// to the repo's committed public bucket (not NEXT_PUBLIC_BLOB_URL, which is
+// each deployer's own store) so these fixtures resolve regardless of env;
+// override via NEXT_PUBLIC_MOCK_BLOB_URL only if you need your own fixtures.
 export const BLOB_BASE_URL =
+	process.env.NEXT_PUBLIC_MOCK_BLOB_URL ??
```
umairnadeem: nit: `NEXT_PUBLIC_MOCK_BLOB_URL` isn't in `.env.example`, where every other `NEXT_PUBLIC_*` knob is listed. undiscoverable config is how you end up with tribal knowledge, either add it (commented out, like the optional provider keys) or drop it, since nothing in the repo sets it today and it's unrelated to BYOI anyway. up to you.

umairnadeem (commented):
the three follow-ups all landed well: `commitResult` is exactly the altitude i was after (queue stores a result, caller owns the cancel), the shared `UploadImageButton` kills the duplicated shell, and the new ImageWithShimmer tests hit the paths i was nervous about (cached/`complete`, swap-in-place). fwiw i sanity checked the swap case in real chrome, `complete` does flip to false synchronously when src is reassigned, so the ref check can't clear the shimmer early. 

one real bug found on the upload path though (project thumbnail), inline. plus a small nit.

lib/generation/queue.ts
umairnadeem: nit: now that we call `notify` in `commitResult`, we can remove this notify here

lib/generation/queue.ts
```diff
@@ -231,6 +231,29 @@ export class GenerationQueue {
 		this.notify();
 	}
 
+	// Store a result regardless of where it came from — a finished job or a
```
umairnadeem: nit: would rather trim the comment here since it doesn't add much context that can't already be inferred from the code

lib/upload/UploadImageButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+	variant?: "toolbar" | "icon";
+	className?: string;
+}) {
+	const { openPicker, uploading, inputElement } = useImageUpload({
+		onUpload: ([url]) => {
+			if (url) onUpload(url);
+		},
+	});
+
+	if (variant === "icon") {
```
umairnadeem: nit: would rather remove the icon variant altogether for simplicity. the toolbar variant is perfectly fine (and perhaps even preferable) even for the character edit modal

lib/upload/UploadImageButton.tsx
```diff
@@ -0,0 +1,68 @@
+"use client";
+
+import { Button } from "@/components/ui/button";
+import { ImagePlus, Loader2 } from "@/components/ui/icon";
+import { TooltipIconButton } from "@/components/ui/icon-button";
+import { useImageUpload } from "@/lib/upload/useImageUpload";
+
+// The one "upload your own image instead of generating" affordance. What
```
umairnadeem: nit: remove redundant comment

umairnadeem (commented):
few small nits on the recent follow-ups, all non-blocking (comment trimming, a stray notify, icon variant simplification)

umairnadeem (changes requested):

lib/components/ImageWithShimmer.tsx
```diff
@@ -1,23 +1,93 @@
...
+				className={className}
 				alt={alt}
-				onLoad={(event: SyntheticEvent<HTMLImageElement>) => {
-					onLoad?.(event);
-					setLoaded(true);
-				}}
+				onLoad={handleLoad}
 			/>
 			{!loaded && (
 				<Skeleton className="absolute inset-0 animate-none shimmer-surface" />
```
umairnadeem: if you take the drop-`unoptimized` route, everything here (the `loadedSrc`/`checkAlreadyLoaded` machinery + the `NEXT_ONLY_IMAGE_PROPS` denylist + the whole `` fork) exists only for the unoptimized path and collapses back to the pre-PR version:

```suggestion
import { useState, type SyntheticEvent } from "react";
import Image, { type ImageProps } from "next/image";
import { Skeleton } from "@/components/ui/skeleton";

/**
 * `next/image` with a shimmer overlay that hides once the image loads.
 */
export function ImageWithShimmer({ alt, onLoad, ...props }: ImageProps) {
	const [loaded, setLoaded] = useState(false);
	return (
		<>
			<Image
				{...props}
				alt={alt}
				onLoad={(event: SyntheticEvent<HTMLImageElement>) => {
					onLoad?.(event);
					setLoaded(true);
				}}
			/>
			{!loaded && (
				<Skeleton className="absolute inset-0 animate-none shimmer-surface" />
```

also delete the new `ImageWithShimmer.test.tsx` cases, they only cover the `` path that goes away.

umairnadeem (commented):
nice work on the byoi flow. cancelling any in-flight gen before `commitResult` in `setManualResult` is exactly right for the upload-vs-generation race, and the coverage on it is solid.

one push on the "also included" image-pipeline fix though: the `unoptimized` -> plain `` fork in `ImageWithShimmer` is a lot of new surface (a second render path + a `NEXT_ONLY_IMAGE_PROPS` denylist you have to keep in sync with next's api) to route around a `decode()` warning that next already swallows (`.catch(() => {})` in `handleLoading`), so afaict it's cosmetic at worst.

the only reason those call sites pass `unoptimized` at all is that `next.config.ts` hard-codes the mock blob hostname, so a deployer's own `NEXT_PUBLIC_BLOB_URL` store isn't allowlisted, and it duplicates the blob url that `lib/blob.ts` + `lib/api/asset-bundle.ts` already read from env (split-brain). that's already biting today, independent of this PR: `ProjectsList` renders thumbnails (generation `imageUrl`s on that same store) through `next/image` WITHOUT `unoptimized`, so on any deploy whose blob store isn't the committed bucket the projects grid already throws `hostname is not configured`.

if we env-drive `remotePatterns` we can drop `unoptimized` at the call sites, revert this whole component to the ~15-line original (suggestion inline), AND that existing ProjectsList breakage goes away for free. only real cost is generated/uploaded imgs now go through the optimizer (vercel image bill + a `/_next/image` hop for constantly-changing, low-view assets); if that bill is a concern, the other option is keep `unoptimized` and JUST env-drive `next.config`. either way next.config needs fixing. i'd lean toward dropping it. verified both locally, renders + shimmer still work.

`next.config.ts` (not in this diff, so pasting here):

```ts
import { BLOB_BASE_URL } from "./lib/blob";

// allowlist the deployer's own store + the committed fixtures bucket
const blobHostnames = Array.from(
  new Set(
    [process.env.NEXT_PUBLIC_BLOB_URL, BLOB_BASE_URL]
      .filter((url): url is string => Boolean(url))
      .map((url) => new URL(url).hostname),
  ),
);

// then in images.remotePatterns:
//   ...blobHostnames.map((hostname) => ({ protocol: "https" as const, hostname })),
//   { hostname: "picsum.photos" },
```

and drop the one `unoptimized` line from `MediaWithSkeleton.tsx` + `AssetTile.tsx`.

lib/upload/UploadImageButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+				size="sm"
+				tooltip="Upload your own image"
+				className={className}
+				disabled={uploading || disabled}
+				onMouseDown={(e) => e.preventDefault()}
+				onClick={openPicker}
+			>
+				<ImagePlus aria-hidden="true" />
+				<span className="hidden sm:inline">
+					{uploading ? "Uploading…" : "Upload"}
```
umairnadeem, line 43: the toolbar variant has basically no uploading feedback, while the icon variant right above swaps in a `Loader2`. two gaps:

- the only visual cue here is the label text, and it's `hidden sm:inline`. so on a narrow viewport you tap Upload, pick a 4mb file, and the button just dims (`disabled:opacity-50`) for a few seconds with zero indication anything is happening. `useImageUpload` only toasts on *error*, so there's no global progress affordance backstopping it either.
- `Button` sets `aria-label={tooltip}` ("Upload your own image"), and aria-label wins over inner text for the accessible name, so the "Uploading…" swap is never announced to a screen reader at all, even on desktop. static name the whole time.

cheapest fix is to just mirror what the icon branch already does:

```suggestion
				{uploading ? (
					<Loader2 className="animate-spin" aria-hidden="true" />
				) : (
					<ImagePlus aria-hidden="true" />
				)}
				<span className="hidden sm:inline">
					{uploading ? "Uploading…" : "Upload"}
```

(that suggestion replaces the `<ImagePlus>` line too, fwiw.) plus `aria-live="polite"` on the span, or swap the `tooltip` value while uploading so the name actually changes.

this matters more if you take my other note and drop the icon variant, since toolbar then becomes the *only* upload affordance in the app. also fwiw master just did this exact fix elsewhere (b04f54b "announce spinner loading state"), so there's a pattern to copy.

umairnadeem (commented):
no new commits since my last pass, so the earlier findings still stand (esp the `connectorType` thumbnail bug). one new thing i missed on the shared button while re-reading it, inline.

lib/blob.ts
umairnadeem: correction on my part, ignore the first half of this. `NEXT_PUBLIC_MOCK_BLOB_URL` *is* in `.env.example` in this PR (commented out, right under `NEXT_PUBLIC_BLOB_URL`), i misread. sorry for the noise. the only bit that still stands is whether the mock-blob knob belongs in the BYOI PR at all vs its own, and thats genuinely up to you, not blocking.

app/components/canvas/elements/ElementUploadButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+					element.id,
+					{ imageUrl: url, durationSec: 0 },
+					getGenerationInputs(element, metadata),
```
umairnadeem, line 37: binding the upload to `getGenerationInputs(element, metadata)` makes it *pretend* it was generated from the current prompt+attributes. but there's no provenance bit anywhere on the snapshot, and `isStaleResult` is purely `!isEqual(currentInputs, resultInputs)`, so the upload goes stale on input drift that has nothing to do with it.

for `image`, `attributes` = `imageDims` (project aspect ratio) + `characterAvatars` (avatar urls of characters tagged in that scene). so:

1. image element with prompt "wizard holding a sign that says WELCOME", generate, text comes out garbled (the #269 case from your description)
2. upload a fixed png over it
3. change the project aspect ratio, or re-roll the avatar of a character tagged in that scene
4. hit Generate All → `inputs.prompt` is truthy, `isStaleResult` is now true → job enqueued → `handleJobSuccess` → `commitResult` silently replaces your upload with a fresh AI gen

that's acceptance criterion #2 ("not overwritten by Generate All unless the user chooses to regenerate") failing on the exact motivating flow. the no-prompt case is safe only because `inputs.prompt` is falsy and the filter short-circuits before staleness is consulted, which is the half the new test pins.

fix imo is the provenance bit, since the queue genuinely cannot tell "result came from a job" from "result was handed in" today: add `uploaded: true` (or `source: "generated" | "uploaded"`) to `ElementSnapshot`, set it in `commitResult`, and have `isStaleResult` short-circuit `false` for it. persists for free through autosave, drops the bogus stale indicator on uploads too, and lines the image path up with the character path which already carries `avatarUploaded`.

app/components/canvas/__tests__/useGenerateAll.test.ts
```diff
@@ -256,6 +256,35 @@ describe("useGenerateAll", () => {
 		expect(jobs).toHaveLength(0);
 	});
 
+	it("does not include an uploaded image (no prompt) in Generate All", async () => {
```
umairnadeem, line 259: glad this exists, but it pins the easy half. the element has no prompt, so `inputs.prompt` is falsy and useGenerateAll's filter bails before `isStaleResult` is ever consulted. the interesting case (and the one the PR description leads with) is an image element that *does* have a prompt, generated garbled text, then got an upload over the top. that one goes through the staleness check and can get regenerated out from under the user, see my note on ElementUploadButton.

worth a second case: non-empty prompt, `resultInputs.attributes` drifted from current (e.g. a different aspect ratio), assert the uploaded element is still excluded. today that'd fail, which is kind of the point.

umairnadeem (commented):
the follow-ups all landed well: `commitResult` carrying `connectorType` kills the thumbnail bug, env-driving `remotePatterns` + reverting the `` fork is exactly the direction i was hoping for (the pre-existing ProjectsList breakage goes away for free), and the shared `UploadImageButton` with the swapped tooltip covers the a11y note. queue side reads clean now, `commitResult` is at the right altitude.

one real bug left though, and it's on the flow your description leads with: an uploaded image can still get clobbered by Generate All. inline, plus a next.config edge case and a typing nit.

next.config.ts
```diff
@@ -1,4 +1,15 @@
 import type { NextConfig } from "next";
+import { BLOB_BASE_URL } from "./lib/blob";
+
+// Must track the same blob URLs lib/blob.ts + lib/api/asset-bundle.ts read —
+// an unlisted hostname 404s in an optimized <Image> (e.g. ProjectsList).
+const blobHostnames = Array.from(
+	new Set(
+		[process.env.NEXT_PUBLIC_BLOB_URL, BLOB_BASE_URL]
+			.filter((url): url is string => Boolean(url))
+			.map((url) => new URL(url).hostname),
```
umairnadeem: two edges here, both from `new URL()` running at config load:

- it throws on anything that isn't an absolute url, so a scheme-less `NEXT_PUBLIC_BLOB_URL` (pasting `foo.public.blob.vercel-storage.com` straight out of the vercel dashboard, which is how it's shown) takes down `next build` AND `next dev` with a bare `TypeError: Invalid URL` and zero breadcrumb pointing at the env var. `.env.example` ships it empty with no sample value so it's an easy fat finger, and `AssetBundle` is tolerant of the same value (it just string-concats), so nothing else in the app complains.
- `protocol: "https"` is hardcoded while the hostname is env-driven. point it at an http store (local minio / self-host) and it parses fine here but then silently never matches the remotePattern.

both fall out of just using the parsed url and failing loudly:

```ts
const blobPatterns = [process.env.NEXT_PUBLIC_BLOB_URL, BLOB_BASE_URL]
  .filter((u): u is string => Boolean(u))
  .map((u) => {
    const parsed = URL.parse(u);
    if (!parsed) throw new Error(`Invalid blob URL (check NEXT_PUBLIC_BLOB_URL): ${u}`);
    return { protocol: parsed.protocol.slice(0, -1), hostname: parsed.hostname };
  });
```

(then dedupe on hostname as you do now). non-blocking, but the opaque build crash is worth 3 lines.

app/components/canvas/elements/ElementUploadButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { getGenerationInputs } from "@/lib/generation/getGenerationInputs";
+import { useProjectStore } from "@/lib/project/store";
+import { UploadImageButton } from "@/lib/upload/UploadImageButton";
+import { ELEMENT_CONFIGS } from "@/lib/canvas/elementConfigs";
+import type { CanvasContentElement } from "@/lib/canvas/types";
+
+export function ElementUploadButton({
+	element,
+}: {
+	element: CanvasContentElement;
```
umairnadeem: SG, but it's a hard-to-break thing: the prop is the broad `CanvasContentElement`, yet the body hardcodes an image-shaped result (`{ imageUrl }`) while deriving `connectorType` *dynamically* from `element.type`. those two disagree with each other. the only thing keeping it honest is the `element.type === "image"` gate at the ElementContainer call site, which the compiler knows nothing about.

render this for a narration element and you'd commit `{imageUrl}` under connectorType `"tts"`, then `getPrimaryUrl` reads `audioUrl` → undefined → `resolveElements` silently drops it from the render. narrow the prop to the image element type, or just pass `"image"` literally (a component that only ever commits an `imageUrl` isn't generic), so the invariant lives in the type rather than in tribal knowledge at one call site.

umairnadeem (commented):
the follow-ups landed well: `uploaded` on the snapshot is the right provenance model, the `isImageElement` narrowing kills the type/behavior gap on `connectorType`, and failing loudly on a bad blob URL is exactly right. one correctness hole left though: `history` doesn't carry the provenance bit, so a prompt revert can launder an upload back into a "generated" result and Generate All is free to clobber it again. details inline.

lib/generation/queue.ts
```diff
@@ -231,6 +238,31 @@ export class GenerationQueue {
...
+		elementId: string,
+		result: AssetResult,
+		inputs: GenerationInputs,
+		connectorType: AssetConnectorType,
+		uploaded: boolean,
+	): void {
+		const key = serializeInputs(inputs);
+		const elHistory =
+			this.history.get(elementId) ?? new Map<string, AssetResult>();
+		elHistory.set(key, result);
```
umairnadeem, line 244: the upload goes into `history` here, but `history` is `Map<string, Map<string, AssetResult>>`, ie result only, no provenance. `restoreResult` then writes the cached result back **without touching `uploaded`**, so the flag desyncs from the thing it's describing.

repro:
1. element has prompt A, user uploads → `history[A] = upload`, `uploaded: true`
2. user edits prompt to B, hits Generate (allowed, the button isn't stale-gated) → `handleJobSuccess` → `commitResult(..., uploaded: false)`
3. user reverts the prompt back to A → `useGenerate`'s effect sees stale, calls `restoreResult(A)` → result is the **upload** again, but `uploaded` is still `false` from step 2
4. change the aspect ratio → `isStaleResult` says stale → Generate All silently regenerates over the user's upload. acceptance criterion #2, right back where we started.

same hazard in reverse: an element that's currently `uploaded: true` restoring a *generated* result inherits `uploaded: true` and becomes permanently immune to staleness.

fix is to make history store what it's actually caching:

```ts
private history = new Map<string, Map<string, { result: AssetResult; uploaded: boolean }>>();
```

and have `restoreResult` patch `uploaded` from the cached entry. that way provenance travels with the result instead of being a sidecar bit that only happens to be right on the happy path.

lib/generation/queue.ts
```diff
@@ -22,13 +22,18 @@ export type ElementSnapshot = {
...
 
 export function isStaleResult(
 	snapshot: ElementSnapshot,
 	currentInputs: GenerationInputs,
 ): boolean {
 	if (isNil(snapshot.result)) return false;
+	// An uploaded result isn't a stand-in for some prompt/attributes combo, so
+	// unrelated input drift (aspect ratio, a tagged character's avatar) must
+	// not make it eligible for Generate All to silently replace.
+	if (snapshot.uploaded) return false;
```
umairnadeem: the short-circuit is in the right *spirit* but it's got a wider blast radius than the requirement. `isStaleResult` has three consumers, and you only want to change one of them:

- `useGenerateAll` → yes, must skip uploads. this is the actual criterion.
- `ElementStaleIndicator` → now never shows the "Stale" badge on an uploaded element. but if i upload, then edit the prompt, the badge's own copy ("Prompt changed, regenerate to update") is *literally true* and i now get zero signal that the visible image no longer matches my prompt.
- `useGenerate`'s restore effect → uploads can never restore from history.

afaict the predicate's actual job is "do the inputs still match what produced this result", and for an upload that's just... not a meaningful question, so answering `false` overloads it. cleaner to keep `isStaleResult` pure and put the skip where the requirement lives:

```ts
// useGenerateAll.ts
const needsGen = inputs.prompt && !snap.uploaded && (!snap.result || isStaleResult(snap, inputs));
```

one extra term, and the badge keeps telling the truth. (does interact with the history comment above, so probably worth doing both together.)

next.config.ts
```diff
@@ -1,4 +1,29 @@
...
+import { BLOB_BASE_URL } from "./lib/blob";
+
+// Must track the same blob URLs lib/blob.ts + lib/api/asset-bundle.ts read —
+// an unlisted hostname 404s in an optimized <Image> (e.g. ProjectsList).
+const blobPatterns = Array.from(
+	new Map(
+		[process.env.NEXT_PUBLIC_BLOB_URL, BLOB_BASE_URL]
+			.filter((url): url is string => Boolean(url))
+			.map((url) => {
+				const parsed = URL.parse(url);
```
umairnadeem: nit: `URL.parse` is Node >= 22.1 / 20.18. no `engines` field in package.json and next 16 supports node 20.9+, so a contributor on 20.9-20.17 gets `TypeError: URL.parse is not a function` at config load, which is the opposite of the friendly error you're going for two lines down. either pin `engines: { node: ">=22" }` or just `try { new URL(url) } catch { throw ... }`, which works everywhere.

SG on the same line: `parsed.protocol.slice(0, -1) as "http" | "https"` is a lie for any other scheme. an `ftp://` typo casts fine and dies later inside next with a much worse message.

app/components/canvas/elements/ElementUploadButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+			disabled={busy}
+			onUpload={(url) => {
+				// Kill any in-flight generation so a late result can't clobber the upload.
+				queue.cancel(element.id);
+				queue.commitResult(
+					element.id,
+					{ imageUrl: url, durationSec: 0 },
+					getGenerationInputs(element, metadata),
+					"image",
+					true,
```
umairnadeem: nit: `commitResult(id, result, inputs, "image", true)` is a boolean trap at the call site, `true` reads as nothing. and the two new params are load-bearing enough that queue.ts needs a comment explaining they exist. perhaps an options object:

```ts
queue.commitResult(element.id, { imageUrl: url, durationSec: 0 }, getGenerationInputs(element, metadata), {
  connectorType: "image",
  uploaded: true,
});
```

self-documenting, and then the second half of the doc comment on `commitResult` ("and pass connectorType and uploaded") can go, since it's just restating the signature. the cancel-first obligation is the only part of that comment worth keeping.

lib/generation/queue.ts
```diff
@@ -320,14 +355,19 @@ export class GenerationQueue {
 			seconds: 0,
 			result: null,
 			error: errorMessage(err),
+			// A failed (re)generation leaves no result, so it's no longer an
```
umairnadeem: this makes a failed regenerate destroy the user's upload, permanently.

repro, and it's the flow the PR description leads with:
1. image element generates garbled text, user uploads a fixed png over it (`result` = upload, `uploaded: true`)
2. later they tweak the prompt and hit the per-element Generate (allowed, the button isn't stale-gated)
3. provider 500s / rate-limits -> `handleJobError` -> `result: null, uploaded: false`
4. the uploaded image is gone from the element. `history` isn't persisted, and `restoreResult` can't fire anyway (`isStaleResult` returns false when `result` is nil). re-uploading from disk is the only way back.

the blob is still sitting in storage, the app just forgot the URL. that's a lot worse than losing a generated result, which you can always re-roll.

fwiw `result: null` on error is pre-existing, this PR just raises the stakes on it by making the result something the user can't reproduce. imo on failure keep `result` and `uploaded` as-is and only set `error` (the UI already renders an error alongside a result). that also makes the `uploaded: false` line here unnecessary, since the flag stays attached to a result that's still there.

separately, on `setError` (line 244): its only caller is `useGenerate.generate()`'s empty-prompt guard, and `GenerateButton` already ships `disabled={!hasPrompt || status !== "idle"}`, so that branch is unreachable. the new `uploaded: false` + its dedicated test are hardening dead code. `CONVENTIONS.md` § Principles: *"Define errors out of existence where you can. Design the API so the edge case can't arise in the first place"* -- the button already did that, so the destructive `setError` can just go.

lib/generation/queue.ts
```diff
@@ -236,9 +275,13 @@ export class GenerationQueue {
 		const cached = this.history.get(elementId)?.get(key);
 		if (!cached) return false;
 		this.update(elementId, {
-			result: cached,
+			result: cached.result,
 			error: null,
 			resultInputs: inputs,
+			// Provenance travels with the result — restoring an upload keeps
```
umairnadeem: convention, meta across this commit: `CONVENTIONS.md` § Structure and reuse says *"Comments are a last resort, not a habit. The code shows what it does... Don't narrate the code."* this commit added six.

this one narrates the two lines directly under it (`connectorType: cached.connectorType, uploaded: cached.uploaded` already says "provenance travels with the result"). same for `// uploaded describes the current result; clearing the result clears it.` above `uploaded: false` on 246, and the `// Never regenerate over a user's upload` block in `useGenerateAll`, which restates `!snap.uploaded`.

the one worth keeping is the `HistoryEntry` block (line ~52), because *why* history stores provenance rather than bare results is the non-obvious invariant a reader can't recover from the code. strip the rest, and the multi-line explainers inside the new tests too (the test names already carry them).

umairnadeem (commented):
both follow-ups landed clean. `history` carrying `{result, connectorType, uploaded}` closes the laundering hole properly (provenance travels with the thing it describes instead of being a sidecar bit), the `rr4` test is exactly the repro i had in mind, `isStaleResult` is a pure inputs comparison again so the staleness badge keeps telling the truth, and the options object kills the boolean trap. `new URL` + the protocol check in next.config is right.

one reachable data-loss path left, and it's on the motivating flow. plus a comment-bloat note.

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -47,18 +47,26 @@ function CancelButton({
...
+	error,
+}: GenerationState & { error: string | null }) {
 	return (
-		<GenerationIndicator
-			status={status}
-			seconds={seconds}
-			className="absolute top-2 left-2 z-10"
-		/>
+		<>
+			{error && <ErrorMessage message={error} />}
```
umairnadeem: keeping the result on failure is right, but the error rendering that comes with it has no exit. `error` is only cleared by a new `enqueue` or a `commitResult`, and `snapshot()` persists it (`if (result || error)`), so:

1. upload/generate an image, hit Regenerate, provider 500s
2. red box now sits centered over a perfectly good image
3. reload → still there

the only ways out are regenerate-until-success or `discard()` (which nukes the image you were trying to keep). for the placeholder case that was fine, there was nothing underneath. over a result it's different: the failure is transient, the asset isn't.

perhaps a `queue.clearError(id)` + a dismiss X on `ErrorMessage` when a result is present, or render it as a small corner badge (like `GenerationIndicator`) instead of an inset-0 takeover. up to you which.

app/components/canvas/elements/preview/results.tsx
```diff
@@ -27,6 +29,7 @@ export function AudioResult({
 				/>
 			)}
 			<AudioPlayer key={src} src={src} />
+			{error && <ErrorMessage message={error} />}
```
umairnadeem: `ErrorMessage` is `absolute inset-0 ... items-center justify-center` with a `pointer-events-auto` box in the middle, and it's now layered over live controls:

- here it lands on top of the `AudioPlayer` row (`min-h-16`), so after a failed regenerate you can't hit play anymore
- same story in `MediaPreview`, which renders video with `videoInteractive` — the box eats the center click

afaict that makes a transient provider error permanently disable playback of an asset that's still perfectly good. same root as the other comment, but worth calling out that it's not just cosmetic occlusion, it's swallowing clicks.

app/components/canvas/elements/preview/results.tsx
```diff
@@ -8,14 +8,16 @@ import {
...
+import { ErrorMessage, PlaceholderOverlay, ResultOverlay } from "./overlays";
 
 export function AudioResult({
 	src,
 	status,
 	seconds,
+	error,
 }: GenerationState & {
 	src: string;
+	error: string | null;
```
umairnadeem: nit: `error: string | null` is now hand-declared in three signatures (`AudioResult`, `MediaPreview`, `ResultOverlay`) while `PlaceholderProps` already extends `GenerationState` with the exact same field. every preview component takes it now, so it may as well live in `GenerationState` in `status.ts` and these three intersections collapse. small, non-blocking.

umairnadeem (commented):
re-read CLAUDE.md + CONVENTIONS.md against the new commits, clean against both (and the comment strip in b837e8b is exactly the right call, what's left all carries a why). both of my last findings are properly fixed: `handleJobError` keeps the result now, and 59c35ea clearing `error` on enqueue closes the obvious "stale error sticks around on retry" hole i was about to raise.

the new stuff is all in the "errors now render on top of a result" surface, three notes below, none blocking.

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -47,14 +48,45 @@ function CancelButton({
...
+}) {
 	return (
-		<GenerationIndicator
-			status={status}
-			seconds={seconds}
-			className="absolute top-2 left-2 z-10"
-		/>
+		<SimpleTooltip label={message}>
+			<span
+				role="status"
```
umairnadeem: the badge fixes the click-swallowing, but it quietly makes the error message unreachable for anyone not on a mouse.

`truncate` means the full text only lives in the radix tooltip, and radix fires that on hover/focus. a bare `<span>` isn't focusable, so keyboard-only and touch users get a red pill reading `failed to generate: provider e...` and no way to see the rest. the placeholder path still renders the full `ErrorMessage` with the copy button, so this only bites on elements that already have a result, which is exactly the flow the PR is built for (regen over an upload fails).

also `role="status"` is a polite live region, `role="alert"` is the one for an error.

perhaps: make the trigger a `<button>` instead of a `<span>`. it's focusable (tooltip now works on tab), and you can hang `ErrorMessage`'s existing `handleCopy` off the click so the badge keeps the copy affordance the full-bleed version had. cheap, and it means the two error UIs stay behaviorally equivalent, just different sizes.

umairnadeem (commented):
re-read against the new commit (8708c04). the badge is the right shape for the fix: `handleJobError` always pairs `error` with `status: "idle"` and `enqueueAll` nulls it on retry, so the badge and the `GenerationIndicator` can never fight over `top-2 left-2`. hoisting `error` onto `GenerationState` kills the three hand-declared copies i moaned about, and dropping the persisted error on rehydration closes the no-exit hole properly. one a11y regression that came along for the ride, inline.

natan497:
> re-read against the new commit ([8708c04](https://github.com/openslop/openslop/commit/8708c04a4c8e93b57cd301ace4997bbac50870c8)). the badge is the right shape for the fix: `handleJobError` always pairs `error` with `status: "idle"` and `enqueueAll` nulls it on retry, so the badge and the `GenerationIndicator` can never fight over `top-2 left-2`. hoisting `error` onto `GenerationState` kills the three hand-declared copies i moaned about, and dropping the persisted error on rehydration closes the no-exit hole properly. one a11y regression that came along for the ride, inline.

Nice catch btw, I just pushed the fix. Hope I can finalize this issue today 😅

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -78,6 +68,57 @@ function ErrorMessage({ message }: { message: string }) {
...
+export function ErrorBadge({
+	message,
+	className,
+}: {
+	message: string;
+	className: string;
+}) {
+	const { copied, copy } = useCopyMessage(message);
+	return (
+		<span role="alert" className={className}>
```
umairnadeem: the new wrapper span moves the width constraint off the flex item, so the badge stops truncating in `AudioResult`.

before, the badge itself was the flex child and carried `max-w-[calc(100%-1rem)]`, so the `%` resolved against the audio row and long errors ellipsised. now `AudioResult` passes `shrink-0` to this span (results.tsx:31), the span is the flex item with auto width and no max, and `max-w-full` on the Badge just resolves against that auto width. net: a long provider error sizes to max-content and gets clipped by the row's `overflow-hidden` instead of truncating.

repro: fail an audio gen with a long message ("provider returned 429: rate limit exceeded for model ...") and the badge runs off the card edge, no ellipsis.

perhaps: `<span role="alert" className={cn("min-w-0 max-w-full", className)}>` and drop `shrink-0` at the AudioResult call site so the span can actually shrink.

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -78,6 +68,57 @@ function ErrorMessage({ message }: { message: string }) {
...
+			<SimpleTooltip label={copied ? "Copied" : message}>
+				<Badge asChild variant="destructive" className="max-w-full focus-ring">
+					<button
+						type="button"
+						onClick={copy}
+						aria-label={`Generation failed: ${message}. Copy error message`}
```
umairnadeem: SG: aria-label is static, so the copy confirmation is mouse-only here. sighted+SR users get "Copied" in the tooltip, keyboard/SR users press enter and hear nothing. `ErrorMessage` right below does flip its label on `copied`, so the two copy affordances in the same file disagree. cheapest fix is the same trick: `aria-label={copied ? "Copied" : \`Generation failed: ${message}. Copy error message\`}`.

also minor: Radix puts the tooltip on `aria-describedby`, so the message is now in both the name and the description and gets read twice. not blocking, but if you're touching it anyway, tooltip label could just be "Copy error message" / "Copied" and let the name carry the error text.

umairnadeem (commented):
re-read against 191e527. making the badge a real focusable trigger is the right fix (truncated text + tooltip-only message was unreachable by keyboard/SR), and folding the copy state into `useCopyMessage` instead of duplicating the timer is nice. one layout regression from the wrapper span though, see inline.

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -78,6 +68,57 @@ function ErrorMessage({ message }: { message: string }) {
...
+					status={status}
+					seconds={seconds}
+					className="absolute top-2 left-2 z-10"
+				/>
+			)}
+		</>
+	);
+}
+
+function ErrorMessage({ message }: { message: string }) {
```
umairnadeem: the a11y fix only landed on half the error surface. `ErrorBadge` got `role="alert"` + `aria-label="Generation failed: ..."`, but `ErrorMessage` (the placeholder path) is untouched: no live region, and its only accessible name is still "Copy error message".

and the placeholder is the *more* common failure. an element that's never generated has no result, so a provider blowup renders here, not in the badge. a screen reader user just gets silence and a mystery copy button.

smallest fix: put `role="alert"` on the outer div and fold the message into the button's aria-label the same way you did for the badge.

app/components/canvas/elements/preview/__tests__/resultError.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+}));
+
+const { AudioResult, MediaPreview } = await import("../results");
+
+const render = (node: React.ReactNode) =>
+	renderToStaticMarkup(<TooltipProvider>{node}</TooltipProvider>);
+
+// The shimmer skeleton is legitimately inset-0, so scope to the error element.
+const errorClass = (html: string) => {
+	const match = html.match(/class="([^"]*bg-destructive[^"]*)"/);
```
umairnadeem: afaict `errorClass` grabs the wrong node, so the `not.toContain("inset-0")` assertions are vacuous.

the first `class=` containing `bg-destructive` is the `<Badge asChild>` **button** (destructive variant = `bg-destructive text-destructive-foreground`). the positioning className you actually care about lives on the outer `role="alert"` span, which has no `bg-destructive`. and the badge base is `inline-flex w-fit`, so that button could never be `inset-0` no matter what anyone does to `ErrorBadge`.

net: pass `className="absolute inset-0"` from `ResultOverlay` and this test still goes green, ie the exact regression it exists to catch sails through. anchor on the alert span instead, something like `html.match(/role="alert"[^>]*class="([^"]*)"/)`.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md again on this head, clean against both. the badge-over-result rework is the right call and dropping the persisted `error` on rehydration (with a test) is a nice catch. two things on the new commits below, both about the a11y/test half of it. fwiw the red CI is the fork-secrets `/login` prerender thing, not you.

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -78,6 +69,63 @@ function ErrorMessage({ message }: { message: string }) {
...
+	return (
+		// Constrained here, not at the call sites: this span is the flex item, so
+		// it's the one that has to shrink for the message to truncate.
+		<span role="alert" className={cn("min-w-0 max-w-full", className)}>
+			<SimpleTooltip label={copied ? "Copied" : message}>
+				<Badge asChild variant="destructive" className="max-w-full focus-ring">
+					<button
+						type="button"
+						onClick={copy}
+						aria-label={
```
umairnadeem: this fixes the mouse-only confirmation i moaned about, but it does it by nuking the button's name. for the 1.5s `copied` window the accessible name is just `"Copied"`, so a SR user who tabs onto the badge in that window hears "Copied, button" and gets neither the error nor what the button does. and since this sits inside `role="alert"` (assertive), flipping the label mutates the live region, so every copy click re-announces assertively.

perhaps: keep the name stable and let a separate polite region carry the confirmation, e.g. `aria-label={`Generation failed: ${message}. Copy error message`}` unchanged, plus a sibling `<span role="status" className="sr-only">{copied ? "Copied" : ""}</span>` outside the alert span. `ErrorMessage` at :139 does the same toggle, so worth fixing both the same way while you're in here.

app/components/canvas/elements/preview/__tests__/resultError.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+		const html = render(
+			<AudioResult
+				src="https://example.com/a.mp3"
+				status="idle"
+				seconds={0}
+				error="the provider rejected this request for a reason it described at considerable length"
+			/>,
+		);
+		expect(alertClass(html)).toContain("min-w-0");
+		expect(alertClass(html)).not.toContain("shrink-0");
```
umairnadeem: `CONVENTIONS.md` § Structure and reuse says *"Test behavior, not internals."* asserting on the tailwind class string is about as internal as it gets: this passes/fails on how the truncation is spelled, not on whether the message truncates.

the `not.toContain("shrink-0")` half can't ever regress either, afaict, since nothing in `ErrorBadge` adds `shrink-0` and `AudioResult` no longer passes it. it's asserting the absence of a class that no code path produces.

not blocking, and i get that truncation is annoying to assert in jsdom (no layout). if it can't be tested for real i'd rather drop it than pin the classnames, up to you.

umairnadeem (commented):
re-read CLAUDE.md + CONVENTIONS.md against e020e6b0, clean against both. moving `min-w-0 max-w-full` onto the span that's actually the flex item is the right fix, and that's one of the rare comments that genuinely earns its place (the shrink rule is not guessable from the markup). two things on the new commit below, plus my two from 191e5272 are still open (the `role="alert"` only landing on `ErrorBadge` and not `ErrorMessage`, and `errorClass` grabbing the wrong node).

app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -194,28 +181,27 @@ function CharacterEditDialogBody({
...
-							className="absolute left-2 top-2 z-10 rounded-full bg-card shadow-sm ring-1 ring-border"
-						>
-							{uploading ? (
-								<Loader2 className="h-3.5 w-3.5 animate-spin" />
-							) : (
-								<ImagePlus className="h-3.5 w-3.5" />
-							)}
-						</TooltipIconButton>
+						<UploadImageButton
+							className="absolute right-2 top-2 z-10 bg-card shadow-elevation-1 ring-1 ring-border"
```
umairnadeem: moving this to `right-2 top-2` parks it exactly on top of `CancelButton`, which is `absolute right-2 top-2 z-10` by default (overlays.tsx `OverlayButton`). same z-index, and this renders after `MediaPlaceholder`, so the upload button wins the paint order.

repro: character with no avatar, hit Generate, hover the placeholder while it's generating. the X is under a size-`sm` "Upload" button, so you can't cancel.

afaict you moved it off `left-2` because `GenerationIndicator` lives there, which is fair. two ways out: put it at `bottom-2 right-2`, or thread `cancelClassName` (PlaceholderOverlay already takes it, `MediaPlaceholder` just doesn't forward it) and push the X down. it's also the same overlap bug natan is fixing in #395, worth not re-introducing it here.

app/components/canvas/hooks/useGenerateAll.ts
```diff
@@ -19,8 +20,14 @@ export function useGenerateAll(editor: Editor) {
 			.filter((el) => {
 				const inputs = getGenerationInputs(el, metadata);
 				const snap = queue.getElementSnapshot(el.id);
+				// Only an image element can hold an upload, so on anything else the
+				// flag is a leftover from an in-place type change (image -> animated)
+				// and describes a result this element can no longer use.
+				const protectedUpload = snap.uploaded && isImageElement(el);
```
umairnadeem: right diagnosis, but this fixes it at the reader. `CONVENTIONS.md` § Principles: *"Solve the class of problem, not the instance"* and *"Almost no special-case, one-off wiring"*.

the class of problem is: Animate rewrites the element type in place keeping the id, so the whole queue entry (result, resultInputs, connectorType, uploaded) is now provenance for an element that no longer exists. the queue already knows this, queue.ts:51 literally says *"the uploaded bit must travel with the result it describes, not desync from it"*, and it stores `connectorType` alongside it for exactly that reason. so the general signal is already sitting right there: `snap.connectorType !== connectorFor(el.type)` means "this result was produced for a different element", and it covers `uploaded`, `result` and `resultInputs` in one go.

`uploaded && isImageElement(el)` instead encodes "only image elements can hold uploads" as tribal knowledge in one consumer, and the next reader of `snap.uploaded` gets to rediscover it. either compare provenance here, or clear the entry at the type-change site in applyOps.

app/components/canvas/hooks/useGenerateAll.ts
```diff
@@ -19,8 +20,14 @@ export function useGenerateAll(editor: Editor) {
...
 				const shouldGenerate =
-					inputs.prompt && (!snap.result || isStaleResult(snap, inputs));
+					inputs.prompt &&
+					!protectedUpload &&
```
umairnadeem: worth thinking through what happens now that this element *does* generate: `animated-image-chain.ts` `afterGenerate` animates `result.imageUrl` from the still-image generation it just ran, there's no path that reuses an existing result. so upload your own image, hit Animate, hit Generate All, and the still gets regenerated from the prompt and your upload is gone.

before this commit it was skipped entirely (also broken, just quieter). is regenerating the intended behavior for now? in a PR called bring-your-own-image, "Animate silently throws away the image you brought" feels like the thing people will hit first. fine as a fast-follow if the chain needs to learn to take a `frameImage`, but i'd at least not leave it undocumented.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md + DESIGN.md again, incremental pass over the last two commits (e22fa63, 77b68ac). the polite `role="status"` split is the right call, that's the correct way to keep the button's name and still confirm the copy. two things i'd fix before merge: the avatar upload button now sits on top of the cancel button, and the `protectedUpload` fix patches the desync at the read site instead of at the source.

app/components/canvas/elements/preview/__tests__/resultError.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+		(_name, node) => {
+			const html = render(node());
+			expect(html).toMatch(/<button[^>]*aria-label="Generation failed: boom\./);
+			expect(html).toMatch(/role="status"/);
+		},
+	);
+
+	// The placeholder is the more common failure: an element that never generated
+	// has no result, so a provider blowup lands here rather than on the badge.
+	it("MediaPlaceholder announces the error and names its copy button", () => {
```
umairnadeem: the truncation test (`"AudioResult error can shrink, so a long message truncates"`) got dropped in this commit. that one was pinning the exact regression you fixed two commits ago (min-w-0 on the flex item, no shrink-0), and now nothing guards it, so a future call site can re-break the badge width and CI stays green. keep it alongside the new cases.

also the `it.each` placeholder case and this test both assert the same `aria-label="Generation failed: ..."` on the placeholder, so one of them is redundant. and `expect(alertClass(html)).toBeDefined()` is a tautology (the helper throws if there's no match), so it asserts nothing beyond what the next line already does.

app/components/canvas/elements/preview/overlays.tsx
```diff
@@ -78,25 +69,97 @@ function ErrorMessage({ message }: { message: string }) {
...
+					aria-label={`Generation failed: ${message}. Copy error message`}
+					className="mt-px shrink-0 rounded-md text-destructive-foreground transition-colors focus-ring"
 				>
 					{copied ? (
 						<Check className="h-3.5 w-3.5" />
 					) : (
 						<Copy className="h-3.5 w-3.5" />
 					)}
 				</button>
+				<span role="status" className="sr-only">
```
umairnadeem: `ErrorBadge` and `ErrorMessage` now hold the same three things independently: `useCopyMessage`, the `Generation failed: ${message}. Copy error message` label string, and this sr-only `role="status"` region. `CONVENTIONS.md` § Structure and reuse: *"Each piece of knowledge gets one home."* the a11y contract for "copy this error" is knowledge, and it's now in two homes, which is how the last round's fix landed on only half the surface. a small `<CopyErrorButton message={...} />` covers both.

same breath on comments (third time i've raised this on this PR, sorry): the diff adds three comment blocks to this file. *"Comments are a last resort, not a habit... Don't narrate the code."* the one on the `role="status"` span is narrating what `role="status"` + `sr-only` already say to anyone reading it. if the shared component exists, none of them need to.

app/components/canvas/hooks/useGenerateAll.ts
```diff
@@ -19,8 +20,16 @@ export function useGenerateAll(editor: Editor) {
 			.filter((el) => {
 				const inputs = getGenerationInputs(el, metadata);
 				const snap = queue.getElementSnapshot(el.id);
+				// A type change (e.g. Animate) rewrites the element in place and
+				// keeps its id, so the queue entry can be provenance for a connector
+				// this element no longer uses -- that invalidates all of it, not just
+				// the uploaded flag.
+				const producedForCurrentType =
+					snap.connectorType === ELEMENT_CONFIGS[el.type].connector;
```
umairnadeem: this treats "no connectorType" the same as "wrong connectorType", and those are really different claims. `connectorType` only started getting stamped at `enqueueAll` in #393 (2026-07-11), so every element generated before that has a persisted entry with no `connectorType` key at all. the constructor spreads the row in raw (`{...snap, status: "idle", ...}`, queue.ts:104) with no `EMPTY_SNAPSHOT` merge, and `getElementSnapshot` only falls back to `EMPTY_SNAPSHOT` when the id has *no* entry, so `snap.connectorType` lands as `undefined`, not `null`.

so: `undefined === "image"` -> `producedForCurrentType` false -> `shouldGenerate` short-circuits to true on the `!producedForCurrentType` arm, and `isStaleResult` never gets consulted.

repro: open any project whose assets were generated before #393 shipped, change nothing, hit Generate All -> every element with a prompt regenerates. before this commit those were skipped (`isStaleResult` reads `resultInputs`, which *is* persisted on old rows, so it correctly said not-stale). that's real provider spend, and it replaces results the user was happy with.

afaict the invalidating signal is a *known* mismatch, not an absent claim:

```suggestion
				const foreignConnector =
					!isNil(snap.connectorType) &&
					snap.connectorType !== ELEMENT_CONFIGS[el.type].connector;
				const shouldGenerate =
					inputs.prompt &&
					(foreignConnector ||
						(!snap.uploaded && (!snap.result || isStaleResult(snap, inputs))));
```

keeps the Animate fix intact (there `connectorType` is present and genuinely different), stops punishing legacy rows, and reads closer to what you actually mean. it also flattens the double-negative nesting a bit, which `CONVENTIONS.md` § Principles would thank you for (*"Every unit stays as dumb and simple as possible"*).

app/components/canvas/__tests__/useGenerateAll.test.ts
```diff
@@ -256,6 +261,122 @@ describe("useGenerateAll", () => {
...
+			// (left over from before the type change) reveals it's unusable.
+			resultInputs:
+				id === "a"
+					? {
+							prompt: "wizard sign",
+							attributes: {
+								width: 2560,
+								height: 1440,
+								videoWidth: 1280,
+								videoHeight: 720,
```
umairnadeem: nice case, but its premise is load-bearing and unasserted, so it can rot into a green tautology.

the whole point (per your comment right above) is *"this result is NOT stale under isStaleResult, only the connector mismatch reveals it's unusable"*. that only holds while these four magic numbers keep matching whatever the default 16:9 aspect ratio injects for an `animated_image`. change the default ratio, or add a field to the animated_image attributes, and `resultInputs` no longer equals `currentInputs` -> `isStaleResult` returns true -> the element gets enqueued for the *staleness* reason -> `expect(jobs).toHaveLength(1)` still passes, and the connector check it exists to cover is no longer exercised by anything.

cheapest fix is to just pin the premise so it fails loudly instead of silently going vacuous:

```ts
const snap = getElementSnapshotSpy("a");
expect(isStaleResult(snap, getGenerationInputs(el, metadata))).toBe(false);
```

or, if wiring the real inputs up here is more trouble than it's worth, assert it the cheap way and let the test say what it means: build `resultInputs` from the same helper the hook uses rather than hand-copying the numbers out of it.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md again on b0e8d58, clean against both. both of my last findings are properly fixed: the avatar button moved to `bottom-2`, which is genuinely unclaimed in both the MediaPreview and MediaPlaceholder paths (CancelButton owns `right-2 top-2`, the badge and GenerationIndicator own `top-2 left-2`), and invalidating by connector instead of `isImageElement` is the right generalization. the extra test for a *generated* leftover that happens to look non-stale is a good catch on your part, that case would have been stuck forever under the old check.

one problem with the new predicate though: it reads a missing `connectorType` as a mismatch, which regenerates the world on any project last generated before #393. details inline, plus a note on the new test's premise.

still open from the last pass (not re-commenting, just not forgetting): upload -> Animate -> Generate All now regenerates the still and throws away the image you brought. that flow got *louder* with this commit, since the entry is now always invalidated on a type change. worth an answer even if the answer is "fast-follow".

app/components/canvas/__tests__/useGenerateAll.test.ts
umairnadeem: nvm, feel free to ignore

umairnadeem:
heya, mea culpa on this one. our review drove way too much scope creep here. the whole error-handling + badge + a11y stack snowballed out of one "keep the result on failure" note and kept growing, well past what #374 actually asked for.

mind resetting the branch back to b1771b13f and force-pushing (`git reset --hard b1771b13f && git push --force-with-lease`)? that lops off the creep and keeps the actual feature. once it's there this lgtm and we'll approve. sorry for the churn 🙏

umairnadeem (approved):
looks great, thank you for the revisions!

PR #395, opened 2026-07-08
Fix canvas interaction bugs: Enter duplicating nodes, cancel/toggle overlap

### LINKED ISSUE #242: Canvas interaction bugs (Enter duplicates node, still/video overlay overlap, lost uploads on mode switch)
## Problem

Several distinct canvas interaction bugs:

- **Pressing Enter in an element prompt creates a duplicate node** instead of a newline/submit.
- **On animated-image elements, the cancel-generation button overlaps the still/video selection toggle** (the "still/video overlay overlaps").
- **Uploaded images are lost when switching modes** (story / script / template).

- "Pressing enter in prompts creates a duplicate node"
- "Cancel generation button overlaps with still/video selection for animated image elements"
- "still/video overlay overlaps"
- "Maintain uploaded images when switching modes"

## Relevant code

- `app/components/canvas/Canvas.tsx` — Slate key handling (Enter)
- `app/components/canvas/elements/preview/AnimatedImagePreview.tsx` — still/video toggle + generation/cancel overlay layout & z-index
- `lib/upload/useImageUpload.tsx` + `lib/config/ConfigProvider.tsx` — upload state across mode switches

## Suggested direction

- Intercept Enter so it doesn't insert a duplicate node.
- Reposition / conditionally render the cancel button vs the still/video toggle so they don't overlap.
- Persist uploaded images across mode changes.

---

natan497:
## Summary

Closes #242

- Pressing Enter in an element prompt now inserts a newline instead of splitting the block into a duplicate node. Guards against modifier combos (Ctrl/Cmd/Alt+Enter, left alone for generation shortcuts), IME composition (CJK input commits via Enter), and bubbled keydowns from non-editable canvas chrome (Cancel button, insert-element menu, settings popover) — a plain Enter on one of those controls now activates it instead of being hijacked into a stray newline.
- The animated-image cancel-generation button no longer overlaps the still/video toggle (moved below it via a `cancelClassName` override threaded through `MediaPlaceholder` → `PlaceholderOverlay` → `CancelButton`).

An earlier version of this PR also preserved user-uploaded reference images across a template switch. That was reverted: wiping reference images on template apply is the intended, correct behavior (applying a template is a reset-to-template action), not a bug — so this PR no longer touches `applyTemplate.ts`'s behavior beyond an unrelated cleanup (routing the last few hardcoded template asset URLs through the existing `templateAsset()` helper).

## Test plan

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run format:check`
- [x] `npm run test:run` (863 tests passing)
- [x] Manual verification:
  - Pressed Enter in an element prompt — element count unchanged, newline inserted (no duplicate node)
  - Tabbed to the Cancel-generation button on a generating element and pressed Enter — button activates (cancels), no stray newline lands in the story
  - Triggered a generation and confirmed via DOM measurement the cancel button no longer overlaps the still/video toggle
  - Applied a template — reference images are replaced by the new template's own images (intended wipe-on-apply behavior)


lib/templates/applyTemplate.ts
```diff
@@ -5,8 +5,15 @@ export function applyTemplate(projectId: string, templateId: string) {
 	const template = getTemplateById(templateId);
 	if (!template) return;
 	const project = getProjectStore(projectId).getState();
+	const userUploadedImages = project.referenceImages.filter(
+		(url) => !project.templateReferenceImages.includes(url),
```
umairnadeem: heads up: `templateReferenceImages` is a second source of truth that only lives in memory. `extractStoreSnapshot` / `ProjectStoreSnapshot` (storeSnapshot.ts) persist only `metadata` + `referenceImages`, so after an autosave + reload it comes back as `[]` while `referenceImages` still holds `[oldTemplate imgs + user uploads]`. the next template switch then makes this filter treat the *whole* list (incl. the previous template's own images) as user uploads and carries them forward, and they keep accumulating on every switch after that. exactly the leak the old wipe-everything code avoided. fix: add `templateReferenceImages` to the snapshot so it survives reload. the new tests don't catch this since they never cross the snapshot boundary.

umairnadeem (commented):
the enter-duplication and cancel/toggle-overlap fixes read clean, and +1 for adding the switch-templates test case. one correctness thing on the reference-image preservation inline.

lib/templates/applyTemplate.ts
natan497: Thanks for your review. I totally missed it. I just pushed the fix

app/components/canvas/Canvas.tsx
```diff
@@ -39,7 +39,12 @@ export default function Canvas({
 
 	const handleKeyDown = useCallback(
 		(event: KeyboardEvent<HTMLDivElement>) => {
-			if (event.shiftKey && event.key === "Enter") {
+			if (
+				event.key === "Enter" &&
```
umairnadeem: this now grabs every plain Enter, but there's no IME guard. for CJK/IME users Enter is how you commit a composition candidate, and this fires preventDefault + insertText("\n") on that commit, so they get a stray newline (and a clobbered composition) every time they accept a suggestion. the old `shiftKey` requirement masked it. mirror the standard guard and bail early on `event.nativeEvent.isComposing || event.keyCode === 229`.

umairnadeem (commented):
the templateReferenceImages persistence fix looks right now, the snapshot round-trip test nails exactly the leak i flagged, and the cancel offset is clean. one i18n footgun on the broadened Enter handler inline.

lib/templates/applyTemplate.ts
```diff
@@ -5,8 +5,15 @@ export function applyTemplate(projectId: string, templateId: string) {
 	const template = getTemplateById(templateId);
 	if (!template) return;
 	const project = getProjectStore(projectId).getState();
+	const userUploadedImages = project.referenceImages.filter(
```
umairnadeem: follow-up on the same leak, different data path: this only self-heals for stores saved *after* this ships. projects saved before the `templateReferenceImages` column existed come back from supabase with it absent, and `applyStoreSnapshot` guards `if (snapshot?.templateReferenceImages)` so it stays the initial `[]`. so the first template switch on any pre-existing project sees an empty `templateReferenceImages`, treats its *current* template's own default images as user uploads, and carries them forward once, same accumulation you fixed for the fresh path.

self-corrects after that one switch so not blocking, but worth a note (or a one-time backfill) if there's meaningful existing project data out there.

umairnadeem (commented):
persistence fix + IME guard both landed clean, and the snapshot round-trip test nails the leak. one residual migration edge inline, non-blocking.

lib/templates/templates.ts
```diff
@@ -543,10 +550,10 @@ export const TEMPLATES: Template[] = [
...
 			"Semi-realistic digital comic illustration, cel-shaded with bold ink outlines, muted earthy palette, cinematic dramatic lighting, gritty detailed textures, expressive characters, vertical 9:16 composition, Rockstar Games concept art style",
 		referenceImages: [
-			"https://mqzeech9ugknls54.public.blob.vercel-storage.com/assets/upload/template/true-crime-1",
-			"https://mqzeech9ugknls54.public.blob.vercel-storage.com/assets/upload/template/true-crime-2",
-			"https://mqzeech9ugknls54.public.blob.vercel-storage.com/assets/upload/template/true-crime-3",
-			"https://mqzeech9ugknls54.public.blob.vercel-storage.com/assets/upload/template/true-crime-4",
+			templateAsset("true-crime-1"),
+			templateAsset("true-crime-2"),
+			templateAsset("true-crime-3"),
+			templateAsset("true-crime-4"),
```
umairnadeem, line 549: nice routing these through `templateAsset()`. one straggler: this same template's `showcase.image` a few lines down (line 567) is still the raw hardcoded `https://mqzeech9...vercel-storage.com/...` literal, and it's the only `blob.vercel-storage.com` string left in the whole file. every other template's `showcase.image` already uses `templateAsset()`. since the point of this change is making the base URL env-overridable, that literal quietly defeats it for true-crime. convert it to `templateAsset("true-crime-5")` too?

umairnadeem (commented):
backfill for pre-field snapshots + the path-segment match both look right, and the tests cover the env-override case nicely. one half-migrated straggler flagged inline.

lib/templates/applyTemplate.ts
umairnadeem: the goal here (keep user uploads, drop the *previous* template's images on switch) is good, but do we need to persist `templateReferenceImages` in the project store to get it? that field leaks a template concern into the store and creates two sources of truth for "is this a template image": the stored array *and* `isTemplateAsset`, which now have to stay in sync.

and you already trust `isTemplateAsset` enough to reconstruct provenance for **every** pre-existing project in the snapshot backfill. if it's reliable enough for the migration, it's reliable enough to just *be* the mechanism. i verified it can't false-positive: user uploads land under `/upload/user/...` (`AssetBundle.upload("upload", "user", ...)`), templates under `/assets/upload/template/`, so the url check cleanly separates them.

so applyTemplate can derive uploads by url and the stored field disappears entirely:
```ts
const userUploaded = project.referenceImages.filter((url) => !isTemplateAsset(url));
project.reset();
project.setReferenceImages([...userUploaded, ...template.referenceImages]);
```
that deletes the store field + setter, the snapshot field + round-trip, the `reset()` clearing, and the whole backfill branch (+ its 4 tests). one source of truth instead of two.

fwiw pure-additive (just appending `template.referenceImages` with no filter, as one might first reach for) isn't enough on its own — switching template A→B would leak A's images forever. the url filter is exactly what lets you drop the prior template's images while keeping uploads, without teaching the store what a template is.

umairnadeem (commented):
the preserve-user-uploads behavior is the right call, and the backfill-for-old-snapshots care is thoughtful. but i think the whole `templateReferenceImages` store field can go away, see inline on `applyTemplate.ts` — you already wrote the primitive that makes it redundant.

umairnadeem (commented):
following up on my earlier comment with a stronger take, see inline.

lib/templates/applyTemplate.ts
umairnadeem: actually, stepping back past my earlier comment: wiping existing uploaded images on template apply is intended behavior, not a bug. applying a template is a reset-to-template action, so preserving user uploads across it is a non-goal. the old `wipes reference images set outside the template before applying` test was encoding the design correctly.

so rather than a leaner implementation of preservation, i'd drop this whole part of the PR: revert `applyTemplate` to `project.setReferenceImages(template.referenceImages)`, keep the old test, and none of the new surface (`templateReferenceImages` store field + setter, snapshot field + round-trip, `reset()` clearing, backfill branch + its tests) needs to exist. the other two fixes in this PR (Enter newline, cancel/toggle overlap) are unrelated and fine to keep.

app/components/canvas/Canvas.tsx
```diff
@@ -39,10 +40,9 @@ export default function Canvas({
 
 	const handleKeyDown = useCallback(
 		(event: KeyboardEvent<HTMLDivElement>) => {
-			if (event.shiftKey && event.key === "Enter") {
-				event.preventDefault();
-				editor.insertText("\n");
-			}
+			if (!shouldInsertNewlineOnEnter(event)) return;
```
umairnadeem: the IME guard is right, but broadening this from shift+Enter to *any* Enter has a side effect. `onKeyDown` sits on `<Editable>`, and every element preview renders inside it (SortableActions, the overlay Cancel button, MediaToggle, the upload buttons). nothing in between stops propagation, so a plain Enter on a focused button in the canvas now bubbles up here, hits `preventDefault()` (which cancels the button's click activation) and drops a stray `\n` into the editor at the last selection.

repro: tab to the Cancel-generation X on a generating image, hit Enter. nothing cancels, and a newline lands in your story instead. same for the insert-element menu. fwiw `OverlayButton` is already styled `focus-visible:opacity-100`, so keyboard reach is clearly intended there.

cheap fix, bail when the keydown didn't originate in editable text:
```ts
if ((event.target as HTMLElement).closest("button,[role='button'],input,textarea,select,a")) return;
```
or `closest('[contenteditable="false"]')`, since SortableItem already wraps its actions in one.

app/components/canvas/elements/preview/__tests__/overlays.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+// The GenerationIndicator sibling also carries a "top-2" class, so scope
+// the assertion to the cancel button itself rather than the whole markup.
+const cancelButtonClass = (html: string) => {
+	const match = html.match(/aria-label="Cancel generation" class="([^"]*)"/);
+	if (!match) throw new Error("cancel button not found in rendered markup");
+	return match[1];
+};
+
+describe("PlaceholderOverlay cancelClassName", () => {
```
umairnadeem: afaict this tests code the PR doesn't change. `PlaceholderOverlay` already accepts `cancelClassName` and forwards it to `CancelButton` on master (overlays.tsx isn't in the diff), so both cases here pass on master verbatim. meanwhile what you *did* change has no coverage: `MediaPlaceholder` forwarding `cancelClassName` through, and `AnimatedImagePreview` passing `top-10` so the X clears the MediaToggle. drop the `{...props}` spread or the `top-10` and this stays green.

perhaps render `AnimatedImagePreview` with no urls + status generating and assert the cancel button lands at `top-10`? that pins the actual regression. also the regex scrape of `renderToStaticMarkup` is brittle, `@testing-library/react` + `getByLabelText("Cancel generation")` reads a lot better if it's already in deps.

lib/canvas/keyboardGuards.ts
```diff
@@ -0,0 +1,20 @@
+export type EnterKeyEvent = {
+	key: string;
+	ctrlKey: boolean;
+	metaKey: boolean;
+	altKey: boolean;
+	keyCode: number;
+	nativeEvent: { isComposing?: boolean };
+};
+
+// IME composition (CJK, etc.) also commits via Enter — don't hijack that
```
umairnadeem: nit: 5 lines of comment on a 4 line function. the IME / keyCode 229 bit is a genuine why and worth keeping, but "Modifier combos (Ctrl/Cmd/Alt+Enter) are left alone for generation shortcuts" just restates `!event.ctrlKey && !event.metaKey && !event.altKey` two lines down. trim to the IME why.

lib/templates/templates.ts
```diff
@@ -2,8 +2,10 @@ import dedent from "dedent";
 import { BLOB_BASE_URL } from "@/lib/blob";
 import type { MetadataCharacter, MetadataVoice } from "@/lib/project/types";
 
+const TEMPLATE_ASSET_PATH = "/assets/upload/template/";
```
umairnadeem: tiny cruft from the revert: `TEMPLATE_ASSET_PATH` now has exactly one caller, two lines below it. it earned its keep when `isTemplateAsset` shared it, but that's gone. inline it back into the template literal?

umairnadeem (commented):
good call reverting the template-preservation piece, the PR is much tighter now and the revert is clean (no `templateReferenceImages` residue left anywhere). two real ones on the new commits: the broadened Enter handler swallows button activation inside the canvas, and the new overlay test covers code this PR doesn't actually touch.

app/components/canvas/Canvas.tsx
```diff
@@ -39,10 +40,13 @@ export default function Canvas({
 
 	const handleKeyDown = useCallback(
 		(event: KeyboardEvent<HTMLDivElement>) => {
-			if (event.shiftKey && event.key === "Enter") {
-				event.preventDefault();
-				editor.insertText("\n");
-			}
+			// Keydowns on non-editable chrome (Cancel button, insert menu, etc.)
+			// bubble up here too — don't hijack their Enter activation.
+			if ((event.target as HTMLElement).closest('[contenteditable="false"]'))
```
umairnadeem: guard is right in spirit, three small things:

1. it runs `closest()` on **every** keydown, before the cheap `shouldInsertNewlineOnEnter` check. that's a DOM tree walk per keystroke while typing. flip the order so the Enter check short-circuits first.
2. `closest('[contenteditable="false"]')` is an indirect proxy for the actual invariant. `event.target.isContentEditable` is the direct one, and it's default-deny: a chrome element that someone forgets to mark `contentEditable={false}` still won't get its Enter hijacked. as written, any new toolbar button that misses the attr silently regresses this.
3. it lives here, outside `shouldInsertNewlineOnEnter`, so the tested pure helper no longer holds the whole "should we hijack Enter" rule. two places decide now, and the new half isn't covered by keyboardGuards.test.

perhaps fold it in:

```ts
// keyboardGuards.ts
export function shouldInsertNewlineOnEnter(event: EnterKeyEvent): boolean {
  if (event.nativeEvent.isComposing || event.keyCode === 229) return false;
  if (!(event.target as HTMLElement).isContentEditable) return false;
  ...
}
```

then Canvas.tsx goes back to the one-liner and the rule stays testable in one place.

umairnadeem (commented):
all three follow-ups landed clean: the chrome guard fixes the Enter-on-button hijack, the cancelClassName test now actually targets the code this PR changes, and TEMPLATE_ASSET_PATH is inlined. one thing on the new guard, non-blocking.

natan497:
@umairnadeem thanks for your review.
I'm gonna fix the rest of your comments early tomorrow. Today. I'm gonna be off soon.
Thanks again.

lib/canvas/keyboardGuards.ts
```diff
@@ -0,0 +1,24 @@
+export type EnterKeyEvent = {
+	key: string;
+	ctrlKey: boolean;
+	metaKey: boolean;
+	altKey: boolean;
+	keyCode: number;
+	nativeEvent: { isComposing?: boolean };
+	target: EventTarget | null;
```
umairnadeem: `target` isn't nullable on the thing you actually pass in: react's `KeyboardEvent<T>` extends `SyntheticEvent` whose `target: EventTarget` is non-optional, and a real keydown always has one. so the `| null` here, the `?.` on line 23, and the "does not throw or fire when the target is missing" test are all guarding a state that can't happen, and the test only passes because the hand-rolled `EnterKeyEvent` type let you construct it.

drop all three: `target: EventTarget`, `(event.target as HTMLElement).isContentEditable`, and the test goes with it. one less impossible branch to reason about.

lib/canvas/keyboardGuards.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+export function shouldInsertNewlineOnEnter(event: EnterKeyEvent): boolean {
+	// IME composition (CJK, etc.) also commits via Enter — don't hijack that
+	// keydown into a newline, or it clobbers the composition. `keyCode === 229`
+	// is the long-standing fallback for browsers that don't set isComposing.
+	if (event.nativeEvent.isComposing || event.keyCode === 229) return false;
+	if (event.key !== "Enter" || event.ctrlKey || event.metaKey || event.altKey)
+		return false;
+	// Only hijack Enter inside editable text. This handler sits on <Editable>
```
umairnadeem: nit, and i know i've beaten this drum on this exact function: 9 lines of comment on 6 lines of code. `CONVENTIONS.md` § Structure and reuse: *"Comments are a last resort, not a habit... A comment earns its place only by capturing what the code can't"*.

the portal bit is a genuine why (it's why this isn't `closest()`), it just doesn't need 5 lines. something like `// Radix portals out of the editor DOM but still bubbles through the React tree, so closest() can't see it. isContentEditable is portal-proof.` carries the same load. the "Only hijack Enter inside editable text" opener restates the return.

umairnadeem (commented):
the fold landed exactly right: the whole "should we hijack Enter" rule is back in one tested pure function, `isContentEditable` is default-deny so a new toolbar button that forgets an attr can't silently regress it, and the cheap key check now short-circuits before the DOM read. the Radix-portal note is a good catch too, `closest()` genuinely couldn't have seen those. two small ones, neither blocking.

umairnadeem (commented):
re-read CLAUDE.md + CONVENTIONS.md against the head branch, clean against both now. the last two commits close out everything i had open: impossible null-target branch and its test are gone, the Enter rule lives entirely in `shouldInsertNewlineOnEnter` so the pure helper still owns the whole decision, and the comment is down to the two genuine whys (IME/229, portal-proofing). no new findings from me, lgtm.

fwiw the red CI is the fork-secrets thing (prerendering `/login` blows up on a missing supabase url/key), not your diff. lint/typecheck/tests are green.

natan497:
> re-read CLAUDE.md + CONVENTIONS.md against the head branch, clean against both now. the last two commits close out everything i had open: impossible null-target branch and its test are gone, the Enter rule lives entirely in `shouldInsertNewlineOnEnter` so the pure helper still owns the whole decision, and the comment is down to the two genuine whys (IME/229, portal-proofing). no new findings from me, lgtm.
> 
> fwiw the red CI is the fork-secrets thing (prerendering `/login` blows up on a missing supabase url/key), not your diff. lint/typecheck/tests are green.

Thanks for your kind check on my works and sorry for making you to add recurring feedbacks.
In the previous fix commits, I did focused on the issue you just mentioned and didn't check what the PR claims and the conventions.
From now on, I will keep building more precisely

umairnadeem (commented):
re-read CLAUDE.md + CONVENTIONS.md on this head, clean against both. the last two commits do exactly what i asked: `target` is non-nullable now (and the dead null-target test is gone with it), `isComposing` is required so the type matches the real event instead of pretending it might be missing, and the comments are down to the two genuine whys (IME/keyCode 229, radix portals defeating `closest()`). no new findings from me, lgtm.

only thing holding this is the red CI, which is the fork-secrets `/login` prerender failure, not your code.

app/components/canvas/Canvas.tsx
umairnadeem: heads up, this is intended behavior, not a bug. splitting a node in two on Enter is something users rely on: press enter mid-node to split a narration/dialog block, then drag foreground elements into the gap. can we drop the `shouldInsertNewlineOnEnter` change so plain Enter keeps splitting the node? once that's reverted this lgtm.

umairnadeem (commented):
one blocker inline, then this is good to merge. the enter-to-split behavior this PR removes is actually intended (users split narration/dialog nodes with it), so we need that kept. the cancel/toggle overlap fix and the templateAsset cleanup are both fine.

umairnadeem (approved):
thank you, much cleaner now!

PR #396, opened 2026-07-09
Fail animated-image chain at the still-image step, not video

### LINKED ISSUE #386: Animated image proceeds after still-image gen fails, then surfaces a cryptic frameImages error
## Problem

When an `animated_image` element's **still-image generation fails but still yields a URL** (an errored/partial/unreachable asset), the video chain proceeds to animate it anyway. The guard in `afterGenerate` only checks that *some* `imageUrl` is present, not that the image actually generated successfully:

```ts
if (!result.imageUrl) {
  throw new Error("animated_image chain expected an imageUrl from the still-image generation");
}
// ...proceeds regardless of whether that imageUrl is a valid, fetchable image
const videoResult = await video.generate({ prompt, frameImages: [result.imageUrl], ... });
```

The video provider (Runware) then rejects the unusable frame and returns a low-level error about `inputs.frameImages`. That raw provider JSON is surfaced **verbatim** in the red error banner, so the user sees a cryptic message that points at `frameImages` instead of the real cause — the **upstream image generation failed**.

Two distinct defects:
1. **Wrong failure point.** We should fail at the still-image step, not inadvertently chain into video generation with a bad frame.
2. **Cryptic error.** The user gets a raw provider error with no actionable meaning.

> anyone know why i keep getting this error on a scene im trying to generate:
> ```json
> {"error":{"code":"invalidValueUploadFailed","message":"Processing parameter 'inputs.frameImages' failed. Try again later.","parameter":"inputs.frameImages","type":"string","taskUUID":"6b526b1f-3b96-4347-b379-efbc359afade"}}
> ```

## Relevant code

- `lib/connectors/animated_image/plugins/animated-image-chain.ts:44-66` — `afterGenerate` only guards `!result.imageUrl`, then calls `video.generate({ frameImages: [result.imageUrl] })` regardless of image validity
- `lib/providers/video/runware.ts:40` — passes `frameImages` straight to Runware; an unusable frame comes back as `invalidValueUploadFailed`
- `lib/connectors/base.ts:52-54` — rethrows the raw provider error unchanged
- `app/components/canvas/elements/preview/overlays.tsx:108-114` — red banner renders the raw provider `message` verbatim (`ErrorMessage`)

## Suggested direction

Solve the class of problem: validate before chaining, and translate provider errors.

1. **Gate the chain on a genuinely successful still image** — not just URL presence. If the image step errored or produced an invalid/unfetchable asset, abort with a clear error (e.g. "Couldn't animate: the source image failed to generate") and don't call `video.generate`.
2. **Map provider errors to actionable messages.** Detect the `frameImages`/`invalidValueUploadFailed` class and surface a human-readable cause that points at the still image, keeping the raw payload behind the existing "copy error" affordance.
3. *(Optional)* retry the still image once before failing the element.

## Acceptance criteria

- [ ] If still-image generation fails, the animated_image element fails at the image step and does **not** attempt video generation with a bad frame.
- [ ] The surfaced error is human-readable and identifies the real cause (image generation failed), with raw details still copyable.
- [ ] A scene that previously showed the raw `invalidValueUploadFailed` error now shows an actionable message.

## Related

- #269 (image-generation failure modes & error UX) — same surface; this is the animated-image chaining case
- #310 (closed: animated images time out, still image never sent) — adjacent animated-image failure mode
- #375 (animate from an uploaded frame) — shares the `frameImages` path

---

natan497:
## Summary

Closes #386

When an `animated_image` element's still-image generation succeeds but yields a frame the video provider can't actually use, the chain proceeded into video generation anyway. The video provider (Runware) then rejected it with a raw `invalidValueUploadFailed`/`frameImages` JSON error, surfaced verbatim to the user with no indication the real cause was the still image.

- The `video.generate(...)` call in the chain plugin is now wrapped so that a provider failure matching the frame-image error class (`frameImages`/`invalidValueUploadFailed`) is rewritten into a human-readable message ("Couldn't animate: the source image failed to generate. Try regenerating the image.") instead of the raw provider JSON. The raw detail is preserved on `Error.cause` for server/console debugging without being painted into the user-facing banner. Unrelated video-provider errors (rate limits, auth, model outage) pass through unchanged.
- Fixed a related gap: `RunwareVideo.toVideoJob` never copied the provider's error detail onto poll results, so async (poll-time) frame-image failures reached the client as a generic "Video generation failed" and could never be classified. The error now flows through on both a resolved-but-failed status item and a hard-rejected status query.

An earlier version of this fix also added a client-side HEAD reachability check before chaining into video generation. Self-review found that check unsafe (cross-origin HEAD against a freshly-uploaded Blob URL is vulnerable to propagation-delay false negatives, had no timeout/cancellation, and was largely redundant with the catch-side translation) and it was removed in favor of the catch-based approach above.

## Test plan

- [x] `npm run lint`, `npm run typecheck`, `npm run format:check`, `npm run test:run` (856 tests passing)
- [x] Added/updated unit tests covering: frame-image errors get the friendly message with raw detail on `cause`, unrelated video errors pass through unchanged, `RunwareVideo` maps provider errors on both the resolved-failed-item and rejected-status-query poll paths
- [x] Self-reviewed via 8-angle finder + per-candidate verification pass; addressed the confirmed correctness/dead-code findings, documented the trade-offs on the two left open (see PR discussion)

### DIFF
--- a/app/api/queues/asset-generate/route.ts
+++ b/app/api/queues/asset-generate/route.ts
@@ -1,4 +1,5 @@
 import { handleCallback } from "@vercel/queue";
+import { serializeError } from "serialize-error";
 import { getJobHandler } from "@/lib/api/job-handlers";
 import type { AssetQueueMessage } from "@/lib/api/jobs";
 import { loadJobForProcessing, updateJob } from "@/lib/api/jobs";
@@ -26,10 +27,9 @@ export const POST = handleCallback<AssetQueueMessage>(
 				// A human message: job.error surfaces verbatim in the failure banner
 				// via rowView, so it must never be a JSON/stack dump.
 				error: humanErrorMessage(error, "Generation failed"),
-				// The raw error rides alongside so rowView can surface structured
-				// detail (e.g. Runware's {error:{code,parameter}}) for a job that
-				// failed at submit, before any live provider poll.
-				metadata: { ...job.metadata, errorDetail: error },
+				// Structured detail for rowView (e.g. Runware's {error:{code,parameter}}),
+				// serialized circular-safe so this updateJob can't itself throw mid-catch.
+				metadata: { ...job.metadata, errorDetail: serializeError(error) },
 			});
 			throw error;
 		}
--- a/app/api/queues/asset-generate/route.ts
+++ b/app/api/queues/asset-generate/route.ts
@@ -24,6 +24,10 @@ export const POST = handleCallback<AssetQueueMessage>(
 			await updateJob(jobId, {
 				status: "failed",
 				error: stringifyError(error),
+				// Keep the raw error alongside the stringified message so rowView can
+				// surface structured detail (e.g. Runware's {error:{code,parameter}})
+				// for a job that failed at submit, before any live provider poll.
+				metadata: { ...job.metadata, errorDetail: error },
 			});
 			throw error;
 		}
--- a/app/api/queues/asset-generate/route.ts
+++ b/app/api/queues/asset-generate/route.ts
@@ -3,7 +3,7 @@ import { getJobHandler } from "@/lib/api/job-handlers";
 import type { AssetQueueMessage } from "@/lib/api/jobs";
 import { loadJobForProcessing, updateJob } from "@/lib/api/jobs";
 import { logger } from "@/lib/api/logger";
-import { stringifyError } from "@/lib/errors";
+import { humanErrorMessage } from "@/lib/errors";
 
 export const POST = handleCallback<AssetQueueMessage>(
 	async ({ jobId, connectorType }) => {
@@ -23,10 +23,12 @@ export const POST = handleCallback<AssetQueueMessage>(
 			logger.error(error, `Job ${jobId} failed`);
 			await updateJob(jobId, {
 				status: "failed",
-				error: stringifyError(error),
-				// Keep the raw error alongside the stringified message so rowView can
-				// surface structured detail (e.g. Runware's {error:{code,parameter}})
-				// for a job that failed at submit, before any live provider poll.
+				// A human message: job.error surfaces verbatim in the failure banner
+				// via rowView, so it must never be a JSON/stack dump.
+				error: humanErrorMessage(error, "Generation failed"),
+				// The raw error rides alongside so rowView can surface structured
+				// detail (e.g. Runware's {error:{code,parameter}}) for a job that
+				// failed at submit, before any live provider poll.
 				metadata: { ...job.metadata, errorDetail: error },
 			});
 			throw error;
--- a/lib/api/handlers/video.ts
+++ b/lib/api/handlers/video.ts
@@ -29,8 +29,19 @@ export const videoHandler: JobHandler<VideoGenerateParams, VideoMetadata> = {
 		}
 		if (status === "failed") {
 			const error = upstream.metadata?.error ?? "Video generation failed";
-			await updateJob(job.id, { status, error });
-			return { jobId: job.id, status, result: null, error };
+			const errorDetail = upstream.metadata?.errorDetail;
+			await updateJob(job.id, {
+				status,
+				error,
+				// Merge (not replace) so providerJobId survives; persist errorDetail
+				// so a later rowView re-fetch of this failed job still carries it.
+				// `!= null` so an explicit null (possible across the JSON boundary)
+				// isn't persisted as a useless errorDetail: null.
+				...(errorDetail != null && {
+					metadata: { ...job.metadata, errorDetail },
+				}),
+			});
+			return { jobId: job.id, status, result: null, error, errorDetail };
 		}
 		return { jobId: job.id, status, result: null, error: null };
 	},
--- a/lib/api/handlers/video.ts
+++ b/lib/api/handlers/video.ts
@@ -1,3 +1,4 @@
+import { serializeError } from "serialize-error";
 import type { VideoGenerateParams } from "@/lib/connectors/types";
 import type { JobPoll, JobStatus } from "@/lib/gateway/base";
 import type { VideoProviderResponse } from "@/lib/providers/video/base";
@@ -29,14 +30,16 @@ export const videoHandler: JobHandler<VideoGenerateParams, VideoMetadata> = {
 		}
 		if (status === "failed") {
 			const error = upstream.metadata?.error ?? "Video generation failed";
-			const errorDetail = upstream.metadata?.errorDetail;
+			const rawDetail = upstream.metadata?.errorDetail;
+			// Circular-safe, same guard as the submit path in the queue worker —
+			// this value is JSON.stringified into jsonb and the HTTP response.
+			const errorDetail =
+				rawDetail != null ? serializeError(rawDetail) : undefined;
 			await updateJob(job.id, {
 				status,
 				error,
 				// Merge (not replace) so providerJobId survives; persist errorDetail
 				// so a later rowView re-fetch of this failed job still carries it.
-				// `!= null` so an explicit null (possible across the JSON boundary)
-				// isn't persisted as a useless errorDetail: null.
 				...(errorDetail != null && {
 					metadata: { ...job.metadata, errorDetail },
 				}),
--- a/lib/connectors/animated_image/plugins/animated-image-chain.ts
+++ b/lib/connectors/animated_image/plugins/animated-image-chain.ts
@@ -3,7 +3,6 @@ import type { ConnectorRegistry } from "@/lib/config/ConfigProvider";
 import { getDefaultConnector } from "@/lib/config/connectorUtils";
 import { createConnector } from "@/lib/connectors/factory";
 import { buildImagePlugins } from "@/lib/connectors/image/plugins/imageChain";
-import { stringifyError } from "@/lib/errors";
 import type {
 	AnimatedImageGenerateParams,
 	AssetResult,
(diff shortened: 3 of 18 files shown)


lib/connectors/animated_image/plugins/animated-image-chain.ts
```diff
@@ -18,6 +19,19 @@ type Stashed = {
 	duration?: number;
 };
 
+const STILL_IMAGE_FAILED_MESSAGE =
+	"Couldn't animate: the source image failed to generate. Try regenerating the image.";
+
+// Marks a video-provider failure as belonging to the still-image frame we
+// handed it, rather than an unrelated video-generation failure (rate limit,
+// auth, model outage, ...) that shouldn't be misattributed to the image step.
+const FRAME_IMAGE_FAILURE_MARKERS = ["frameimages", "invalidvalueuploadfailed"];
```
umairnadeem: SG, non-blocking: driving user copy off a lowercased substring match is a bit brittle. `invalidValueUploadFailed` really means "the video provider couldn't accept the frame", which can be a transient upload hiccup, not necessarily "the source image failed to generate" (the still did generate, you have its url), so "Try regenerating the image" can be the wrong remedy. runware already hands you structured `error.code` / `error.parameter` (`inputs.frameImages`) and you're flattening it to a string then re-parsing it here. matching the structured code/param upstream would be sturdier than stringify+includes. works today, just flagging.

umairnadeem (commented):
nice, splitting task-failure from connection-failure in the runware poll (retry the latter, fail the former) is exactly the right call and well tested. one non-blocking thought on the frame-failure heuristic inline.

lib/connectors/animated_image/plugins/animated-image-chain.ts
```diff
@@ -18,6 +18,32 @@ type Stashed = {
...
+	};
+};
+
+// Marks a video-provider failure as belonging to the still-image frame we
+// handed it, rather than an unrelated video-generation failure (rate limit,
+// auth, model outage, ...) that shouldn't be misattributed to the image step.
+// Matches Runware's structured error.code/error.parameter directly rather
+// than string-matching the serialized error, which is sturdier against
+// unrelated fields that happen to mention "frameImages" in passing.
+function isFrameImageFailure(err: unknown): boolean {
```
umairnadeem, line 32: following up on my earlier note: matching structured `error.code`/`error.parameter` is the right call, but afaict the structured shape never reaches here on the real path. `video.generate()` is a `BaseAssetConnector`, and for an async runware job the frameImages upload failure surfaces at poll time: `RunwareVideo._poll` catches it and flattens it to `metadata.error = stringifyError(err)` (a string), then `asset-base.ts` does `throw new Error(completed.error)`. so `afterGenerate` catches a plain `Error` whose `.error` is undefined, `isFrameImageFailure` returns false, and the friendly message never fires.

the new tests pass because they mock `createConnector().generate` to reject with the raw `{error:{...}}` object, which the real connector only produces on a *submit-time* rejection, not the async poll path this feature is for. so it's green but wouldn't work in prod afaict.

options: preserve the structured error through asset-base (throw with `{cause}` carrying the payload) so it survives to this layer, or do the frameImages->friendly translation down in the runware/poll layer where the structure still exists, or go back to matching the string message. lmk if i'm missing a submit-time path here.

umairnadeem (commented):
the structured-matching move (from my last note) is the right instinct, but i think it doesn't actually fire on the real path. one thing inline.

lib/connectors/animated_image/plugins/animated-image-chain.ts
natan497: Thanks for your review @umairnadeem 
I totally missed it. I'd follow the structured error path.
Just pushed the fix.

lib/providers/video/runware.ts
```diff
@@ -1,18 +1,36 @@
...
 	return {
 		url: video.videoURL,
 		metadata: {
 			jobId: video.taskUUID,
 			status: video.status as VideoJobStatus,
+			...(video.error != null && {
+				error:
+					typeof video.error === "string"
+						? video.error
+						: stringifyError(video.error),
```
umairnadeem: heads up, i think this regresses the banner for a plain `clip`/video element. before this PR `metadata.error` was undefined on a failed video, so videoHandler fell back to a clean "Video generation failed". now `error` = `stringifyError(video.error)` = a raw JSON dump (`{"code":"invalidValueUploadFailed","parameter":"inputs.frameImages","taskUUID":...}`), and that string renders verbatim in the red `ErrorMessage` overlay for anything that isn't the animated-image chain (the chain overrides with a friendly message, direct video doesn't).

since you just added `errorDetail` to carry the raw payload, `error` should stay a human string. maybe `error: typeof video.error === "string" ? video.error : (video.error?.message ?? "Video generation failed")` and keep the full object on `errorDetail`. same shape applies to the `stringifyError(err)` in the catch path below.

umairnadeem (commented):
the errorDetail threading looks solid now, and this closes the gap i flagged earlier (structured detail actually reaches isFrameImageFailure). one new thing on the user-facing side below.

lib/providers/video/runware.ts
natan497: Good catch. Thanks for the careful review

app/api/queues/asset-generate/route.ts
```diff
@@ -23,7 +23,13 @@ export const POST = handleCallback<AssetQueueMessage>(
...
 			await updateJob(jobId, {
 				status: "failed",
-				error: stringifyError(error),
+				// A human message: job.error surfaces verbatim in the failure banner
+				// via rowView, so it must never be a JSON/stack dump.
+				error: humanErrorMessage(error, "Generation failed"),
+				// The raw error rides alongside so rowView can surface structured
+				// detail (e.g. Runware's {error:{code,parameter}}) for a job that
+				// failed at submit, before any live provider poll.
+				metadata: { ...job.metadata, errorDetail: error },
```
umairnadeem: new since my last pass: `errorDetail: error` stashes the raw error object into jsonb metadata, and supabase-js JSON.stringifies it on write. if `error` is ever non-serializable (circular ref, e.g. a socket/fetch error from a connection-level failure), this `updateJob` throws inside the catch, which masks the real failure and leaves the job stuck in `processing` (never marked failed). the old `stringifyError` path was circular-safe on purpose. you already import `serializeError` in errors.ts, so `errorDetail: serializeError(error)` buys the safety back and still preserves the `{error:{code,parameter}}` shape the chain classifier reads. non-blocking but cheap.

umairnadeem (commented):
the follow-ups all landed well: humanErrorMessage keeps the banner clean, errorDetail threads the structured payload, and the runware video-banner regression i flagged earlier is gone. one non-blocking serialization footgun on the new errorDetail line below.

lib/api/handlers/video.ts
```diff
@@ -29,8 +29,19 @@ export const videoHandler: JobHandler<VideoGenerateParams, VideoMetadata> = {
...
+			const errorDetail = upstream.metadata?.errorDetail;
+			await updateJob(job.id, {
+				status,
+				error,
+				// Merge (not replace) so providerJobId survives; persist errorDetail
+				// so a later rowView re-fetch of this failed job still carries it.
+				// `!= null` so an explicit null (possible across the JSON boundary)
+				// isn't persisted as a useless errorDetail: null.
+				...(errorDetail != null && {
+					metadata: { ...job.metadata, errorDetail },
```
umairnadeem, line 44: heads up, non-blocking: you serialized the submit path in route.ts but this poll-failure write is the same class and didn't get the treatment. `errorDetail` here comes straight out of runware `_poll` (`video.error`, typed `unknown`, or the caught `err` that only passed an `isApiError` *shape* check, not an acyclic one) and lands directly in jsonb via `updateJob`. if it ever carries a circular/unserializable value, supabase-js JSON.stringify throws right here inside the `failed` branch, exactly the mid-catch throw you just guarded route.ts against. wrap this one in `serializeError` too so both persistence paths stay symmetric?

umairnadeem (commented):
nice, serialize-error is the right call for the submit path and the circular-error test locks it in. one gap: the poll-failure persistence path didn't get the same guard, flagged inline.

lib/api/handlers/video.ts
natan497: Good catch. Thanks.

umairnadeem (commented):
the poll-path serialize landed, and the circular-errorDetail test is exactly the right regression to pin (asserts both the persisted jsonb and the HTTP-response copy stringify cleanly). one leftover below on where that guard should live.

lib/api/handlers/video.ts
```diff
@@ -29,8 +30,21 @@ export const videoHandler: JobHandler<VideoGenerateParams, VideoMetadata> = {
 		}
 		if (status === "failed") {
 			const error = upstream.metadata?.error ?? "Video generation failed";
-			await updateJob(job.id, { status, error });
-			return { jobId: job.id, status, result: null, error };
+			const rawDetail = upstream.metadata?.errorDetail;
+			// Circular-safe, same guard as the submit path in the queue worker —
+			// this value is JSON.stringified into jsonb and the HTTP response.
+			const errorDetail =
+				rawDetail != null ? serializeError(rawDetail) : undefined;
```
umairnadeem, line 37: small gripe, non-blocking: we now import `serialize-error` directly in two places (here + `app/api/queues/asset-generate/route.ts:32`), and the only thing binding them is the comment saying "same guard as the submit path". that's a split-brain waiting to happen: the next handler that persists an `errorDetail` won't know the invariant exists, and the failure path throws mid-catch when it forgets.

`lib/errors.ts` is already the canonical home for this stuff and already imports serialize-error. perhaps hoist it:

```ts
export const errorDetail = (error: unknown): unknown => serializeError(error);
```

then both sites just call `errorDetail(...)` and the why lives in one jsdoc next to `stringifyError`/`humanErrorMessage` instead of being copy-pasted prose in each caller.

umairnadeem:
while attempting to repro the error, it seems like this error was due to a transient downstream failure within Runware (and not due to us not handling the error). We already __do__ fail-fast for animated images when the image generation fails: https://github.com/openslop/openslop/blob/7bb27d351f59017778009232772fd73403d0e070/lib/connectors/asset-base.ts#L47

closing this issue since it's redundant

PR #399, opened 2026-07-11
fix: nightly bug fixes 2026-07-11

umairnadeem:
## Summary

Nightly correctness bug-fix pass. Three real bugs found and fixed with regression tests.

### 1. Volume attribute: blank/whitespace silently mutes (`lib/video/elementAttributes.ts`)
`Number("")` and `Number("   ")` both coerce to `0`, so a blank or whitespace `volume` custom attribute on an element was silently interpreted as volume 0 (muted) instead of the default 10. Added an explicit null/blank/whitespace guard before the `Number()` coercion. This is a real user-facing bug — clearing a volume field could accidentally mute the element's audio in the final render.

### 2. LLM route: unbounded `maxTokens` and `temperature` (`app/api/v1/llm/route.ts`, `lib/api/request-schema-fields.ts`)
The LLM API route accepted any number value for `maxTokens` and `temperature` without bounds checking. Out-of-range values (negative, zero, fractional maxTokens, or temperature outside [0,1]) would pass through to the provider and surface as opaque provider errors. Extracted an `optionalLlmSampling` schema with `.int().positive()` for maxTokens and `.min(0).max(1)` for temperature, so invalid values fail fast at the route boundary with a 400.

### 3. MediaWithSkeleton: stale skeleton on src change (`app/components/canvas/elements/MediaWithSkeleton.tsx`)
When the video `src` changed (e.g., regenerating an animated element), the `videoLoaded` state wasn't reset, so the loading skeleton never reappeared — the old video frame remained frozen while the new source loaded. Replaced the boolean `videoLoaded` state with a `loadedVideoSrc` string, deriving `videoLoaded` from `loadedVideoSrc === src`. When `src` changes, `videoLoaded` automatically becomes false until the new source fires `onLoadedData`. No effect or render-phase state mutation needed.

## Tests added

- `lib/api/__tests__/request-schema-fields.test.ts` — validates `optionalLlmSampling`: accepts in-range values, rejects temperature outside [0,1], rejects non-positive/fractional maxTokens
- `lib/video/__tests__/elementAttributes.test.ts` — added case for empty and whitespace volume defaulting to 10 instead of 0

## Validation

- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm test -- --passWithNoTests` ✅ (100 files, 869 tests)
- `npm run build` ✅

## Open PR overlap check

Considered open PRs #396 (animated-image chain errors), #395 (canvas interaction bugs), and #394 (bring-your-own-image). This PR touches only `app/api/v1/llm/route.ts`, `MediaWithSkeleton.tsx`, `lib/api/request-schema-fields.ts`, `lib/video/elementAttributes.ts`, and their tests — zero file or theme overlap with any open PR.

## Skipped items / rationale

- **TTS voices route validation (#350)** — the unmerged nightly PR's concern about permissive `.catch(undefined)` on optional query params was reviewed; the current schema's behavior is correct (invalid optional params become undefined, which is the same as not-provided). No user-facing bug.
- **No broader Zod schema tightening sweeps** — kept changes scoped to bugs with real user-visible correctness impact per the runbook constraints.
- **No style/cosmetic/refactor-only changes** — all changes are correctness fixes with regression tests.

### DIFF
--- a/lib/api/request-schema-fields.ts
+++ b/lib/api/request-schema-fields.ts
@@ -37,6 +37,17 @@ export const optionalReferenceImages = {
 	referenceImages: z.array(referenceImageUrlOrDataUri).optional(),
 } as const;
 
+/**
+ * LLM sampling controls, bounded to the ranges the underlying providers accept
+ * so an out-of-range client value is a 400 at the boundary rather than a
+ * provider-side 400 re-surfaced as a 500. `temperature` is constrained to
+ * [0, 1]; `maxTokens` must be a positive integer.
+ */
+export const optionalLlmSampling = {
+	maxTokens: z.number().int().positive().optional(),
+	temperature: z.number().min(0).max(1).optional(),
+} as const;
+
 export const optionalFrameImages = {
 	frameImages: z.array(referenceImageUrlOrDataUri).optional(),
 } as const;
--- a/app/api/v1/llm/route.ts
+++ b/app/api/v1/llm/route.ts
@@ -2,15 +2,17 @@ import { NextResponse } from "next/server";
 import { z } from "zod";
 import { getLLMProvider } from "@/lib/api/providers";
 import { bodySchema, createApiRouteHandler } from "@/lib/api/route-handler";
-import { optionalReferenceImages } from "@/lib/api/request-schema-fields";
+import {
+	optionalLlmSampling,
+	optionalReferenceImages,
+} from "@/lib/api/request-schema-fields";
 import { createSSEStreamResponse } from "@/lib/api/sse";
 import { LLM_MODELS } from "@/lib/connectors/llm/openslop/models";
 
 const schema = bodySchema(LLM_MODELS, {
 	systemPrompt: z.string().optional(),
 	thinkingLevel: z.string().optional(),
-	maxTokens: z.number().optional(),
-	temperature: z.number().optional(),
+	...optionalLlmSampling,
 	...optionalReferenceImages,
 	stream: z.boolean().optional(),
 });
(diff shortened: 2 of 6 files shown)


lib/api/request-schema-fields.ts
```diff
@@ -37,6 +37,17 @@ export const optionalReferenceImages = {
...
 } as const;
 
+/**
+ * LLM sampling controls, bounded to the ranges the underlying providers accept
+ * so an out-of-range client value is a 400 at the boundary rather than a
+ * provider-side 400 re-surfaced as a 500. `temperature` is constrained to
+ * [0, 1]; `maxTokens` must be a positive integer.
+ */
+export const optionalLlmSampling = {
+	maxTokens: z.number().int().positive().optional(),
```
umairnadeem, line 47: nit, non-blocking: the bound is one-sided. temperature is fully clamped to [0,1] (correct, only model is claude-opus-4-8), but maxTokens has no upper cap, so a client passing e.g. maxTokens: 1e9 still sails past the schema and 500s at the provider, which is exactly the "opaque provider error re-surfaced as a 500" case the docstring says this closes. perhaps add a `.max(<model output limit>)` so both ends fail fast at the boundary? up to you if you'd rather not hardcode the ceiling here.

umairnadeem (commented):
clean little bug-fix pass, and nice that the MediaWithSkeleton fix derives videoLoaded off `loadedVideoSrc === src` instead of the stuck-bool pattern. one non-blocking note on the maxTokens bound below.

umairnadeem:
useless

PR #400, opened 2026-07-11
refactor: rethink openslop architecture 2026-07-11

umairnadeem:
## App understanding

OpenSlop is an AI creative-media studio: a prompt streams OSML markup into a SlateJS canvas, users generate media assets (image/video/music/sfx/voice) per element through a three-layer pipeline (Connectors → Gateways → Providers), and projects save/restore to Supabase with Remotion-Lambda rendering. Per-project editor state lives in a Zustand store (`lib/project/store.ts`), read through selector hooks and mutated through its actions (`updateMetadata`, `setCharacter`, …).

## From-scratch architecture notes

Rebuilt today, the store seam would follow one rule consistently: **UI reads through selector hooks and writes through named `lib/` functions — never by reaching into the store's nested update shape inline.** The read side already honors this (`useAspectRatio`, `useTransitionType`, `deriveProjectName`), and most writes do too (`deleteCharacter(projectId, …)`, `applyTemplate(projectId, …)` are plain `lib/` functions taking `projectId`). The gap was a handful of UI components that still re-stated the store's `{ videoSettings: { … } }` / `{ title }` update shape by hand — the one read/write asymmetry left in the metadata path.

Closed that asymmetry so **no UI component knows the store's metadata write shape**:

- `setAspectRatio(projectId, ratio)` next to `useAspectRatio`.
- `setTransitionType(projectId, type)` next to `useTransitionType`.
- `setProjectTitle(projectId, title)` next to `deriveProjectName`.
- Repointed `ComposerCopilot`, `PropertiesPanel`, and `ProjectTitle` at the writers; dropped the now-unused `getProjectStore` imports.
- Added `videoSettings` and `projectName` write-seam tests.

Behavior is preserved; this is a pure coupling reduction (UI no longer imports the store to mutate it) that mirrors the pattern the rest of the app already follows.

## Validation

- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run test:run` — 871 passed (100 files)
- `npm run knip` — clean

## Deliberately skipped

Evaluated and rejected as out of scope for a *safe* nightly (per the repo's own conventions):

- **Extract a `shouldGenerate`/`isElementStale` helper** (`useGenerate` / `useGenerateAll` / `CharacterEditModal`). The three sites genuinely differ — `useGenerate` already holds `currentInputs` for `restoreResult`, `useGenerateAll` composes a "prompt && (!result || stale)" policy, and `CharacterEditModal` builds a synthetic element with an extra `!avatarUploaded` guard. A single helper would double-compute inputs or only partially fit — a speculative abstraction below the rule-of-three bar.
- **`useCharacterAvatar` extraction from `CharacterEditModal`** — high leverage but a behavior-sensitive state-machine extraction from a 249-line component; too risky for an unattended nightly.
- **Composer-input / asset-tile de-duplication** — rule-of-two only; and the asset-tile / reference-image surfaces overlap in-flight image-upload work (PR #394). Left untouched.
- **Adopting `useProject(selector)` at ~9 raw `useConfig()+useProjectStore` sites** — mechanical churn, not architecture.

All open-PR-owned files (PRs #394/#395/#396/#399) were left untouched.

### DIFF
--- a/lib/project/projectName.ts
+++ b/lib/project/projectName.ts
@@ -1,6 +1,11 @@
+import { getProjectStore } from "./store";
 import type { Metadata } from "./types";
 
 export function deriveProjectName(metadata: Metadata | undefined): string {
 	const title = metadata?.title?.trim();
 	return title && title.length > 0 ? title : "Untitled";
+}
+
+export function setProjectTitle(projectId: string, title: string) {
+	getProjectStore(projectId).getState().updateMetadata({ title });
 }
--- a/app/components/canvas/ProjectTitle.tsx
+++ b/app/components/canvas/ProjectTitle.tsx
@@ -3,7 +3,8 @@
 import { useState } from "react";
 import { Pencil } from "@/components/ui/icon";
 import { useConfig } from "@/lib/config/ConfigProvider";
-import { getProjectStore, useProjectStore } from "@/lib/project/store";
+import { setProjectTitle } from "@/lib/project/projectName";
+import { useProjectStore } from "@/lib/project/store";
 import { useScriptControl } from "@/lib/script/ScriptProvider";
 
 export function ProjectTitle() {
@@ -25,7 +26,7 @@ export function ProjectTitle() {
 	const commit = () => {
 		const next = draft.trim();
 		if (next && next !== title) {
-			getProjectStore(projectId).getState().updateMetadata({ title: next });
+			setProjectTitle(projectId, next);
 		}
 		setEditing(false);
 	};
--- a/app/components/canvas/panel/PropertiesPanel.tsx
+++ b/app/components/canvas/panel/PropertiesPanel.tsx
@@ -8,9 +8,11 @@ import {
 	SelectValue,
 } from "@/components/ui/select";
 import { useConfig } from "@/lib/config/ConfigProvider";
-import { getProjectStore } from "@/lib/project/store";
 import { TRANSITION_TYPES, type TransitionType } from "@/lib/video/transitions";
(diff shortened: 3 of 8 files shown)


umairnadeem (commented):
colocating the write helpers next to their read hooks (setAspectRatio by useAspectRatio, etc.) is a nice read/write pairing, and the tests are tight. one design nit inline since this PR is explicitly an architecture pass.

lib/project/projectName.ts
```diff
@@ -1,6 +1,11 @@
+import { getProjectStore } from "./store";
 import type { Metadata } from "./types";
 
 export function deriveProjectName(metadata: Metadata | undefined): string {
 	const title = metadata?.title?.trim();
 	return title && title.length > 0 ? title : "Untitled";
 }
+
+export function setProjectTitle(projectId: string, title: string) {
```
umairnadeem, line 9: small gripe, non-blocking: `projectName.ts` was a pure, dependency-free derivation module (`deriveProjectName` takes metadata in, returns a string). dropping `setProjectTitle` here drags a `getProjectStore` dependency + a side-effecting write into it, and it's the odd one out vs the other two setters that live beside their read hook (there's no `useProjectTitle` to pair with). also worth noting all three setters are single-call extractions, so there's no dedup win here, just indirection + testability. if the goal is the write-next-to-read pattern, maybe a `useProjectTitle`/`projectTitle.ts` home would be more consistent, otherwise this is fine to leave. up to you.

umairnadeem:
useless

PR #401, opened 2026-07-11
feat: allow uploading reference images after story generation

umairnadeem:
## Summary

Closes #297 — lets users upload reference images from the Assets sidebar after the story is generated, not just from the pre-prompt composer.

The edit/narration modals and character asset tiles already lived in the sidebar post-prompt, but the reference-image upload affordance was gated behind the pre-prompt `ComposerCopilot`. A user who accidentally removed a reference image had no way to re-add it.

## What changed

- **`AssetsSection.tsx`** (+30/−2): Added an "Upload" button after the reference-image tiles that reuses the same `useImageUpload` hook already in use by `ComposerCopilot`. The button follows the same dashed-border style as the "New" character button in the same section. During upload, it shows a spinner. The hidden `<input type="file">` is rendered alongside it.

## Validation

- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm run test:run` ✅ (99 files, 865 tests)

## Open PR overlap check

Considered open PRs #400, #399, #396, #395, and #394. This PR touches only `AssetsSection.tsx`, which is not modified by any open PR. The store import (`getProjectStore`) is read-only; the store file is not modified.

## Skipped items / rationale

- **#397** (remove LucideIcon shims) — maintainer previously rejected similar cleanup as "useless fluff" on #376
- **#342** (Done button in modals) — UI enhancement requiring broader modal design decisions; not clearly scoped enough for unattended
- **#259** (configurable queue concurrency) — explicitly "post-BYOK"; BYOK is an XL issue still open
- All other open issues are size M/L/XL with broad feature/product scope beyond the small-issue criteria

PR #402, opened 2026-07-12
fix: nightly bug fixes 2026-07-12

umairnadeem:
## Summary

Nightly correctness bug-fix pass. One real bug found and fixed with regression test.

### 1. RunwareImage: PNG content-type mismatch (`lib/providers/image/runware.ts`)

The Runware API defaults to JPEG output when `outputFormat` is omitted, but the provider was labeling the returned bytes as PNG (`contentType: "image/png"`, `filename: "output.png"`). This caused a content-type mismatch where JPEG bytes were served as `image/png`. The fix pins `outputFormat: "PNG"` on the Runware API call so the bytes, filename, and content type are consistent.

## Tests added

- `lib/providers/__tests__/runware-image.test.ts` — new test verifying:
  - `outputFormat: "PNG"` is present in the Runware API request
  - The files passed to `AssetBundle.upload` carry the correct PNG filename and content type

## Validation

- `npm run lint` ✅
- `npm run format:check` ✅
- `npm run typecheck` ✅
- `npm test -- --passWithNoTests` ✅ (99 files, 866 tests)
- `npm run build` ✅

## Open PR overlap check

Considered open PRs #401, #400, #399, #396, #395, and #394. This PR touches only `lib/providers/image/runware.ts` and `lib/providers/__tests__/runware-image.test.ts` — zero file or theme overlap with any open PR. (#396 touches `lib/providers/video/runware.ts` and `lib/providers/__tests__/runware-video.test.ts` — different files, different provider.)

## Skipped items / rationale

- **No other changes made** — Claude Code timeout at 80 turns after finding this single real bug. Additional provider audits (other Runware call sites) were skipped: RunwareVideo is already covered by PR #396's error-handling work.
- **No broader cleanups** — kept changes scoped to the real content-type mismatch per the runbook constraints.
- **No style/cosmetic/refactor-only changes.**

Generated with Hermes Agent (scheduled cron)

### DIFF
--- a/lib/providers/image/runware.ts
+++ b/lib/providers/image/runware.ts
@@ -39,6 +39,7 @@ export class RunwareImage extends BaseProvider<
 				width: params.width || 2848,
 				height: params.height || 1600,
 				outputType: "base64Data",
+				outputFormat: "PNG",
 				numberResults: 1,
 				referenceImages: params.referenceImages,
 			});
--- a/lib/providers/__tests__/runware-image.test.ts
+++ b/lib/providers/__tests__/runware-image.test.ts
@@ -16,6 +16,7 @@ vi.mock("@runware/sdk-js", () => ({
 	},
 }));
 
+import { AssetBundle } from "@/lib/api/asset-bundle";
 import { RunwareImage } from "../image/runware";
 
 describe("RunwareImage", () => {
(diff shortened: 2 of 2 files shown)


lib/providers/image/runware.ts
```diff
@@ -39,6 +39,7 @@ export class RunwareImage extends BaseProvider<
 				width: params.width || 2848,
 				height: params.height || 1600,
 				outputType: "base64Data",
+				outputFormat: "PNG",
```
umairnadeem, line 42: nit, non-blocking: this pins the request to `PNG`, but the returned `format: "png"` on line 52 is a *separate* literal, and that return value is what drives `filename`/`contentType` in `toFiles`. so the image format now lives in two places that have to stay in lockstep. that's the exact split-brain that caused this bug (request said JPG-default, files said png). if someone later flips this to WEBP and forgets line 52, you're right back to a mismatch. perhaps hoist a single `const OUTPUT_FORMAT = "png"` and derive both the request (`.toUpperCase()`) and the return from it, so they can't drift. up to you.

umairnadeem (commented):
clean, well-scoped fix and nice regression test that asserts both the request `outputFormat` and the resulting filename/contentType. one small consistency note inline.

umairnadeem:
useless

PR #407, opened 2026-07-13
fix: nightly bug fixes 2026-07-13

umairnadeem:
## Summary

Nightly correctness bug hunt and test coverage pass.

## Open PR overlap check

Reviewed open PRs #395 (canvas-interaction-bugs) and #394 (bring-your-own-image). This PR does not touch any files or modules covered by those PRs. No overlap.

## Bugs fixed

1. **Asset queue: failed jobs were silently skipped instead of retried** — The route handler returned early for `"failed"` jobs, treating them as terminal. But the queue redelivers failed jobs after the rethrow, so they must stay eligible for reprocessing. Only `"completed"` is truly terminal now.

2. **Video render polling: no abort mechanism, stale closure on unmount** — Added AbortController support so the polling loop is cleanly cancelled on reset or unmount. fetch and wait now accept a signal. Added proper useEffect cleanup and fixed stale closure.

3. **Supabase middleware: session cookie dropped on auth-route redirect** — When getUser() refreshes the session, new cookies live on supabaseResponse. If the middleware then redirects (authenticated user visiting login page), those cookies were dropped. Now forwarded onto the redirect.

4. **Cartesia TTS: infinite pagination guard** — Added safety break if the API returns an empty page or stale cursor.

5. **XML serialization: unescaped quotes/angle brackets in OSML attributes** — Raw quote or `>` in a user prompt silently truncates the attribute on round-trip. Added XML entity escaping on serialization and unescaping on parsing.

6. **XML tag parsing: split-on-whitespace bug** — Old approach split tag string on spaces, potentially including `<scene` (with `<` prefix) in the tag name. Switched to regex extraction.

7. **OSML stream parser: missing `updated` flag on open tags** — Callers polling the parser didn't know new content arrived.

8. **Volume/loops: Number("") footgun silently mutes audio** — Blank value for volume or loops would silently mute audio or set loops to 0. Now treats blank strings as undefined, falling back to correct defaults.

9. **Layout key: scene grouping ignored** — Moving an element between scenes produced the same key even though the resolved layout would differ. Now incorporates scene membership.

10. **Poll job: unnecessary upstream polling for terminal jobs** — Added isTerminal short-circuit to avoid unnecessary upstream calls.

11. **SSE reader: unreleased stream on early consumer break** — Added reader.cancel() in finally block.

12. **PostPromptView.tsx: dead import and stale function signature** — Cleaned up unused getContentElements import, updated getLayoutKey call site.

## Tests added

- **layoutKey.test.ts** — Restructured to use scene-wrapped elements. Added 3 new test cases: element moves between scenes (key changes), scenes split (key changes), orphan elements ignored (consistent with resolver).

- **asset-generate.test.ts** — Updated "skips failed jobs" to "reprocesses a failed job" with assertions that the handler is invoked and result stored.

## Validation

- `npm test -- --passWithNoTests`: 100 files, 869 tests — all passing
- `npm run build`: compiles successfully
- `npm run typecheck`: no errors
- `npm run lint`: no errors
- `npm run format:check`: all files pass

### DIFF
--- a/app/api/queues/asset-generate/route.ts
+++ b/app/api/queues/asset-generate/route.ts
@@ -8,7 +8,9 @@ import { stringifyError } from "@/lib/errors";
 export const POST = handleCallback<AssetQueueMessage>(
 	async ({ jobId, connectorType }) => {
 		const job = await loadJobForProcessing(jobId);
-		if (job.status === "completed" || job.status === "failed") return;
+		// Only `completed` is final. A `failed` job is redelivered by the queue
+		// after the rethrow below, so it must stay eligible for reprocessing.
+		if (job.status === "completed") return;
 
 		const handler = getJobHandler(connectorType);
 		if (!handler) {
--- a/app/components/video/useRendering.ts
+++ b/app/components/video/useRendering.ts
@@ -1,6 +1,6 @@
 "use client";
 
-import { useCallback, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import type { ProgressResponse } from "@/app/api/render/progress/route";
 import { stringifyError } from "@/lib/errors";
 import type { VideoLayout } from "@/lib/video/types";
@@ -19,13 +19,29 @@ type RenderState =
 
 const POLL_INTERVAL_MS = 5000;
 
-const wait = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
+const wait = (ms: number, signal: AbortSignal) =>
+	new Promise<void>((resolve, reject) => {
+		const timer = setTimeout(resolve, ms);
+		signal.addEventListener(
+			"abort",
+			() => {
+				clearTimeout(timer);
+				reject(signal.reason);
+			},
+			{ once: true },
+		);
+	});
 
-async function postJSON<T>(url: string, body: unknown): Promise<T> {
+async function postJSON<T>(
+	url: string,
+	body: unknown,
+	signal: AbortSignal,
+): Promise<T> {
 	const res = await fetch(url, {
 		method: "POST",
 		headers: { "content-type": "application/json" },
 		body: JSON.stringify(body),
+		signal,
 	});
 	if (!res.ok) throw new Error(await res.text());
 	return res.json() as Promise<T>;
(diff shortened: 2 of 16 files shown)


app/api/queues/asset-generate/route.ts
```diff
@@ -8,7 +8,9 @@ import { stringifyError } from "@/lib/errors";
 export const POST = handleCallback<AssetQueueMessage>(
 	async ({ jobId, connectorType }) => {
 		const job = await loadJobForProcessing(jobId);
-		if (job.status === "completed" || job.status === "failed") return;
+		// Only `completed` is final. A `failed` job is redelivered by the queue
+		// after the rethrow below, so it must stay eligible for reprocessing.
+		if (job.status === "completed") return;
```
umairnadeem, line 13: this makes the row lie during a retry. the catch below writes `status: "failed"` and *then* rethrows to trigger redelivery, so between attempts the DB says failed while the queue still has the job in flight.

and the consumer already treats that as terminal: `awaitCompletion` in lib/connectors/asset-base.ts stops on `status === "failed"` and throws `Generation failed`. so the caller has already given up by the time attempt 2 runs. we burn an upstream generation (real $) and write a result nobody is listening for.

two ways out imo:
- don't mark it failed until retries are exhausted. leave it `processing` on a transient throw and only write `failed` on the last attempt (vercel queue exposes the attempt count on the callback).
- or make `failed` genuinely non-terminal end to end, which means teaching awaitCompletion/pollJob to keep waiting while attempts remain.

first one is way less blast radius. as it stands the retry is a no-op from the user's POV, just more expensive.

lib/gateway/base.ts
```diff
@@ -6,6 +6,10 @@ export abstract class GatewayClient<TParams = unknown, TResult = unknown> {
 
 export type JobStatus = "pending" | "processing" | "completed" | "failed";
 
+/** A terminal job has a settled result; it must never be re-processed or re-polled upstream. */
```
umairnadeem, line 9: split-brain: this says `failed` "must never be re-processed", and the queue route two files over now deliberately re-processes failed jobs (with its own hand-rolled `job.status === "completed"` check instead of this helper). two sources of truth for what terminal means.

whoever comes along next and "simplifies" route.ts to use `isTerminal` reintroduces the exact bug this PR is fixing. whichever way the thread above lands, the definition should live in one place and route.ts should call it.

lib/canvas/osmlSerializer.ts
```diff
@@ -12,7 +13,7 @@ export function getElementText(element: ParsedElement): string {
 function serializeElement(element: CanvasContentElement): string {
 	const attributes = element.customAttributes ?? {};
 	const attrString = Object.entries({ id: element.id, ...attributes })
-		.map(([key, value]) => ` ${key}="${value}"`)
+		.map(([key, value]) => ` ${key}="${escapeXmlAttribute(value)}"`)
 		.join("");
 	return `<${element.type}${attrString}>${getElementText(element)}</${element.type}>\n`;
```
umairnadeem, line 18: escaping is half-migrated: attributes get `escapeXmlAttribute`, but the element body (`getElementText`) is still interpolated raw, and that's the field the user actually types into.

repro: narration text `she gasped <softly> and left` -> serialize -> parseOSML -> TAG_PATTERN matches `<softly>` and appendNext pushes a bogus generic node, text after it lands on the wrong element. same failure class the PR is fixing, just the more likely one.

fixing it properly means escaping on serialize *and* unescaping in `updateCurrent`, which is a bigger change than this PR. fine as a fast-follow, but worth calling out now, because as-is the escaping gives false confidence that the round-trip is safe.

lib/canvas/parseXmlTag.ts
```diff
@@ -1,13 +1,16 @@
+import { unescapeXmlEntities } from "./xmlEntities";
+
+const ATTRIBUTE_PATTERN = /([\w-]+)="([^"]*)"/g;
```
umairnadeem, line 3: SG: module-level `g` regex + manual `lastIndex = 0` is stateful global you have to remember to reset. the old local-const version was equally fast (V8 caches the literal) and can't be footgunned by a re-entrant call. up to you.

also fwiw the PR description says the old split-on-whitespace could leave `<scene` in the tag name, but the caller passes `match[1]` from `/<([^<>/][^<>]*?)>/` so the `<` was never in the string. this is a clean-up, not a bug fix. `/^\S*/` would keep a leading `<` too.

app/components/video/useRendering.ts
```diff
@@ -19,59 +19,96 @@ type RenderState =
 
 const POLL_INTERVAL_MS = 5000;
 
-const wait = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
+const wait = (ms: number, signal: AbortSignal) =>
+	new Promise<void>((resolve, reject) => {
+		const timer = setTimeout(resolve, ms);
+		signal.addEventListener(
```
umairnadeem, line 25: nit: on the happy path the abort listener is never removed (`once` only fires on abort), so every poll tick leaves a dead closure on the signal. 20 min render at 5s = ~240 of them.

perhaps: `AbortSignal.any` or just `signal.removeEventListener("abort", onAbort)` in the resolve path.

umairnadeem (commented):
nice haul, and the layoutKey scene-membership fix + the tests around it are the good kind of nightly: real bug, real repro, tests that would've caught it.

two things i'd want resolved before merge: the failed-job retry change (#1) doesn't line up with how the rest of the stack treats `failed`, and the XML escaping (#5) only covers attributes so the round-trip is still broken for element text. rest is nits.

meta: this adds ~9 new comments and several just restate the adjacent code (`// An empty page or a cursor that doesn't advance would page forever.` above the exact check, the 3-liner in sse.ts explaining that a `finally` shouldn't rethrow). CONVENTIONS says self-documenting code, comments only where the why can't be carried by the code. the `Number("")` one in elementAttributes and the queue-semantics one in route.ts earn their keep, most of the others don't imo.

lib/canvas/xmlEntities.ts
```diff
@@ -0,0 +1,28 @@
...
+	lt: "<",
+	gt: ">",
+};
+
+/**
+ * OSML attribute values are delimited by `"` and terminated by `>`, so a raw
+ * quote or angle bracket in a prompt silently truncates or drops the attribute
+ * when the script round-trips through the DB and back into the parser.
+ */
+export const escapeXmlAttribute = (value: string): string =>
```
umairnadeem, line 21: the escaping fix is the headline change of a PR literally branded `nightly-bugs-tests`, and it ships with zero tests. `xmlEntities.ts` is a new module with no test file, and neither `osmlSerializer.test.ts` nor `parseXmlTag.test.ts` got touched (only `asset-generate.test.ts` + `layoutKey.test.ts` did).

that matters because `parseXmlTag` got substantially rewritten here: the attr regex now scans the whole tag body instead of the post-tag-name slice, attr names widened `\w+` -> `[\w-]+`, trailing-slash strip moved, plus entity unescaping. the existing test cases (`image`, `character name="Lyra"`, `image/`, `music`) exercise exactly none of that.

CONVENTIONS.md § Structure and reuse: *"Test behavior, not internals. Cover the seams (pure helpers and module boundaries) so refactors stay safe."* this is the seam.

a single serialize -> parse round-trip over `& " < >` would (a) pin the escape/unescape idempotence the whole DB round-trip now silently leans on, and (b) have caught the half-migrated body escaping i flagged in `osmlSerializer.ts` before i had to. worth a hyphenated-attr-name case in `parseXmlTag.test.ts` too.

umairnadeem (commented):
second pass over this one (no new commits since my last review, so the earlier findings all still stand). re-read CLAUDE.md + CONVENTIONS.md against the head, no new rule violations. one gap i missed the first time, inline.

umairnadeem:
overengineered

PR #404, opened 2026-07-12
chore: remove residual LucideIcon references

umairnadeem:
## Summary

Closes #397 — removes all residual `LucideIcon` type references now that the app owns its icon library via `components/ui/icon.tsx`. Pure mechanical rename: `LucideIcon` → `IconComponent`. Also removes the stale `lucide-react` entry from `optimizePackageImports` in `next.config.ts`.

## Changes

- **`components/ui/icon.tsx`**: Removed the `LucideIcon` compat type alias (callers use `IconComponent` directly). Updated strokeWidth comment.
- **`components/ui/media-toggle.tsx`**: `LucideIcon` → `IconComponent` in import and `MediaToggleOption` interface.
- **`lib/connectors/attributes/schema.ts`**: `LucideIcon` → `IconComponent` in import and `AttributeSpec` interface.
- **`app/components/canvas/panel/EditorSidebar.tsx`**: `LucideIcon` → `IconComponent` in import and type annotations.
- **`app/components/canvas/elements/GenerationIndicator.tsx`**: `LucideIcon` → `IconComponent` in import and `ICONS` record type.
- **`app/components/canvas/elements/AssetTile.tsx`**: `LucideIcon` → `IconComponent` in import and type annotations.
- **`next.config.ts`**: Removed `"lucide-react"` from `optimizePackageImports`.

## Not changed (intentionally)

- `*Icon` suffixed aliases (`CheckIcon`, `ChevronDownIcon`, etc.) still used by `select.tsx` and `dropdown-menu.tsx` — retained.
- `strokeWidth` no-op prop still accepted by many call sites — retained until call sites are migrated off it.

## Validation

| Check | Result |
| --- | --- |
| `grep -rn "LucideIcon" --include="*.ts" --include="*.tsx"` | No matches |
| `grep -rn "lucide" next.config.ts` | No matches |
| `npm run lint` | ✅ pass |
| `npm run format:check` | ✅ pass |
| `npm run typecheck` | ✅ pass |
| `npm test -- --passWithNoTests` | ✅ 99 files, 865 tests |
| `npm run build` | ✅ pass |
| `npm run knip` | ✅ pass |

## Open PR overlap check

Considered open PRs #403, #402, #401, #400, #399, #396, #395, and #394. This PR touches only the 7 files listed above — zero file or theme overlap with any open PR.

## Skipped complex/product-context items

- All other open issues are size M/L/XL with broad feature/product scope beyond the small-issue criteria.
[...]

PR #408, opened 2026-07-13
refactor: nightly maintainability 2026-07-13

umairnadeem:
## Summary

Nightly maintainability + docs pass. No behavior changes. Net −18 lines across 6 files.

## Open PR overlap check

Reviewed open PRs #407 (nightly bug fixes), #395 (canvas-interaction-bugs), and #394 (bring-your-own-image). This PR does not touch any files or modules covered by those PRs. No overlap.

## Architecture changes

None — this pass found the architecture layer already clean (plugin piping, connector registry, video layout builder all in good shape).

## Maintainability changes

1. **Finished the Lucide icon migration** (`components/ui/icon.tsx`, `select.tsx`, `dropdown-menu.tsx`). PR #404 removed `LucideIcon` references but left the compatibility layer behind: five aliases (`CheckIcon`, `ChevronDownIcon`, `ChevronRightIcon`, `ChevronUpIcon`, `CircleIcon`) that existed purely so the two vendored shadcn files could keep their original Lucide import names. Rewrote those call sites to canonical names and deleted the aliases — one name per icon. Also removed six icon exports that nothing imports (`BookOpen`, `Film`, `EyeOff`, `PanelLeft`, `PanelRight`, `PanelTop`).

2. **Removed dead `onReady` prop from `Waveform`** (`lib/components/Waveform.tsx`). Declared and invoked in two places internally, but `AudioPlayer` is the only caller and never passes it. Shrunk the `exhaustive-deps` suppression justification with it.

3. **Made `ComposerCopilot` placeholder contract explicit** (`app/components/copilot/ComposerCopilot.tsx`, `app/components/ComposerHero.tsx`). The prop was typed `ReactNode` and then `typeof`-sniffed back apart inside the component into a native-placeholder path and an overlay path. Now two typed props (`placeholder?: string`, `placeholderOverlay?: ReactNode`) with visibility rules stated on each.

4. **Exhaustive mode label lookup** (`app/components/copilot/ComposerCopilot.tsx`). Replaced `MODE_OPTIONS.find(o => o.value === mode)?.label` with an exhaustive `Record<Mode, string>` that `MODE_OPTIONS` derives from. Adding a `Mode` now fails to compile instead of silently rendering an empty label.

5. **Hoisted doubled `getTemplateById`** in template pill render.

## Complexity delta

- Files changed: 6 (−18 net lines: +34 / −52)
- Removed 5 redundant icon aliases and 6 dead icon exports
- Removed 1 dead callback prop (2 call sites + 1 suppression comment)
- Replaced `typeof`-based prop sniffing with explicit typed props
- Replaced find-with-fallback with exhaustive Record lookup

## Skipped items with reasons

- **`lib/providers/llm/mock.ts` table-driven dispatch**: The four named predicates in an if-chain are already clear; converting to an array of `{predicate, response}` adds indirection to four branches without making anything easier to change.
- **Waveform mask-style consolidation**: The webkit/standard duplication should collapse into a CSS class, but the same mask boilerplate is duplicated in `preview/results.tsx`, which belongs to open PR #395. Migrating only the Waveform half would leave a half-done extraction — worth doing as one change once #395 lands.
- **`lib/connectors/types.ts` and `ConfigProvider.tsx`**: Read both and found no non-cosmetic improvement. The plugin-piping, connector registry, and layout builder are already clean.
- **`.claude/commands/` and `.claude/skills/`**: Neither directory exists in this repo (only `.claude/settings.local.json`).

## Validation

- `npm run build` — compiles successfully
- `npm test -- --passWithNoTests` — 100 files, 866 tests, all passing
- `npm run lint` — no errors
- `npm run typecheck` — no errors
- `npm run format:check` — all files pass


app/components/ComposerHero.tsx
```diff
@@ -39,12 +39,9 @@ export default function ComposerHero() {
...
 				}}
-				placeholder={
-					mode === "script" ? (
-						INPUT_SCRIPT_PLACEHOLDER
-					) : (
-						<AnimatedPlaceholder active />
-					)
+				placeholder={mode === "script" ? INPUT_SCRIPT_PLACEHOLDER : undefined}
+				placeholderOverlay={
+					mode === "script" ? undefined : <AnimatedPlaceholder active />
```
umairnadeem: while you're purging dead props: `active` here is the twin of the `onReady` you just killed on `Waveform`. this is the only call site and it's always `<AnimatedPlaceholder active />`, and the overlay only mounts when `!hasText && !showPill`, so the `if (!active) return;` guard + the `[active, clear]` dep in the component are unreachable. drop the prop and let mount/unmount be the on/off switch (which it already is).

app/components/copilot/ComposerCopilot.tsx
```diff
@@ -156,14 +160,18 @@ interface ComposerCopilotProps {
 	value: string;
 	onValueChange: (value: string) => void;
 	onSubmit: () => void;
-	placeholder?: ReactNode;
+	/** Native textarea placeholder, shown by the browser whenever the field is empty. */
```
umairnadeem: splitting the `typeof` sniff into two props is right, but the old `ReactNode` union enforced either/or by construction and two optional props don't: pass both and the browser paints the native placeholder *and* the overlay div in the same grid cell, two strings stacked on top of each other. only one caller today so not a live bug, just a footgun the previous shape didn't have. cheapest fix is probably to gate the native one on `!placeholderOverlay` at the `<textarea>` so the overlay always wins.

also nit: the doc comment on `placeholder` restates the type and the name, and the one below duplicates the `!hasText && !showPill` condition 60 lines down (which is where it'll go stale). i'd strip both.

umairnadeem (commented):
the icon-alias purge and the `Record<Mode, string>` swap are both the good kind of nightly: one name per icon, and adding a `Mode` now fails to compile instead of rendering an empty pill label. two things on the placeholder refactor inline, neither blocking.

components/ui/icon.tsx
```diff
@@ -37,31 +37,23 @@ function icon(name: string): IconComponent {
 
 export const AlertCircle = icon("alert");
 export const ArrowLeft = icon("arrow-left");
```
umairnadeem, line 39: the icon exports are gone but their css halves aren't, so this dead-code purge only got half of each icon.

the docblock right above says an icon is an `icon("<name>")` export **plus** a `--<name>-icon` var in `icons.css`. you deleted the six exports (`BookOpen`, `Film`, `EyeOff`, `PanelLeft`, `PanelRight`, `PanelTop`) and left all six vars behind: `--sidebar-left-icon` (icons.css:110), `--sidebar-right-icon` (111), `--eye-off-icon` (131), `--layout-alt-3-icon` (199), `--film-icon` (309), `--book-open-icon` (432). zero consumers anywhere at head, ~2.9kb of base64 data-uri still shipping in the global stylesheet.

the half-migration is the real cost, not the bytes: next person to grep `--film-icon` finds a live-looking token and assumes the icon is still a thing. drop those six lines from `icons.css` too.

umairnadeem (commented):
second pass on this one (no new commits since my last review, so the earlier findings still stand). one new thing below: the icon purge only did half the job.

app/components/copilot/ComposerCopilot.tsx
```diff
@@ -181,16 +189,14 @@ export default function ComposerCopilot({
 		});
 	const showPill = mode === "template" && selectedTemplateId !== undefined;
 	const hasText = value.trim().length > 0;
+	const selectedTemplate = selectedTemplateId
+		? getTemplateById(selectedTemplateId)
+		: undefined;
```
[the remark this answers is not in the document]
umairnadeem: stop it gemini

components/ui/icon.tsx
umairnadeem: we still want these icons

PR #409, opened 2026-07-13
refactor: rethink openslop architecture 2026-07-13

umairnadeem:
Nightly architecture rethink. Two things had more than one home; now they have one.

## App understanding

openslop is an AI storyboard/video creator. A user prompts an LLM, which streams OSML back over SSE; a parser inserts elements (image, animated image, clip, narration, character, sound, music) into a Slate canvas grouped into scenes. Elements are generated through a client queue that fans out to `/api/v1/{asset}` job routes, which enqueue to a Vercel Queue whose workers call the vendor providers and upload to Blob. The resolved elements are assembled into a timeline and rendered with Remotion Lambda.

Layering is `connectors` (what the editor calls, plugin-pipelined) → `gateway` (HTTP to our own API) → `providers` (server-side vendor adapters), with the queue and Zustand project store owning editor state.

## From-scratch vision vs. what's here

Built fresh, three things would look different from today:

1. **Element types would be one registry.** Everything about an element type — which connector generates it, what kind of asset comes back, how it behaves on the timeline — is intrinsic to the type, so it belongs in one table that everything else derives from.
2. **Project state would have exactly one read path.** A component wanting project metadata should ask for it, not first ask for a `projectId` and then use it as a key.
3. **Persisted state would be state something reads.**

The rest of the architecture already matches what I'd build. The v1 media routes are fully factored behind `createAssetRouteHandler` (11-21 lines each, pure declarations); the connector/gateway/provider split is a deliberate seam for the planned BYOK gateway; knip reports no dead code. I did not touch any of it.

## Implemented

**One home for element types.** Element knowledge was spread across four lists that had to agree by hand, with nothing enforcing it — adding a type meant four coordinated edits:

| was | now |
|---|---|
| `ELEMENT_CONFIGS` (connector, outputKind + look) | presentation only, spreads the spec |
| `CanvasElementType` union | `keyof typeof ELEMENT_TYPES` |
| `CANVAS_ELEMENT_TYPES`, `FOREGROUND_TYPES` | derived from the registry |
| `ELEMENT_ROLES`, `LAYER_TYPES` (lib/video) | deleted; role/layer live in the registry |

`ELEMENT_TYPES` in `lib/canvas/types.ts` is now the single source. Adding an element type is one entry, and the compiler enforces the rest.

This also closes a layering leak. `resolve.ts`, `buildGenerationJob.ts` and `createCanvasNode.ts` are domain code that imported `elementConfigs.tsx` — a JSX module pulling in the icon set — just to read a connector name. They read the pure registry now.

**One read path for project state.** `useProject(selector)` already existed but only 1 of 11 call sites used it. The other 10 called `useConfig()` solely to obtain `projectId` and hand it to `useProjectStore(projectId, selector)`. Eight are migrated; seven of those drop `useConfig` entirely, so components like `CharacterPill` and `CharactersPicker` no longer depend on `ConfigProvider` at all.

**Deleted write-only state.** `metadata.lastMode` / `lastPrompt` were written to the store (and persisted to Postgres) on every prompt submit and never read back by anything. Removing them also drops `projectId` and `mode` from `ScriptProvider`'s `submitPrompt` dependencies.

## Skipped, with reasons

- **`AssetsSection` / `ComposerCopilot` reference-image append** — both do a `getState()` read-modify-write. That looks like a candidate for a store action, but the imperative read is deliberately race-safe against a concurrent upload landing mid-`await`; a subscribed-value closure would clobber. Left alone. (`ComposerCopilot` is also in PR #408.)
- **`CharacterEditModal`** — the 9th `useProjectStore` call site; owned by PR #394.
- **Collapsing the four 6-line `lib/gateway/openslop/*` classes** — ~24 lines, and the layering is an intentional seam. Not worth the churn.
- **Consolidating `lib/generation`'s 9 files** — the meaningful version requires touching `queue.ts`, owned by PR #394.
- **Moving `lib/api/asset-bundle.ts` out of `lib/api`** — a genuine dependency inversion (`lib/providers` and `lib/connectors` import it), but every call site sits in PR #394/#407. Worth a follow-up once those land.
- **zod at the `/api/upload/image` boundary** — the one route with hand-rolled shape checks. Adjacent to PR #394's subject matter; deferred to avoid a semantic conflict.

## Validation

| check | result |
|---|---|
| `npm run build` | pass |
| `npm run lint` | pass |
| `npm run test:run` | 866 passed / 100 files |
| `npx knip` | clean |

No behavior changes: no test needed updating, and the derived sets preserve their original iteration order (which the refine prompt and the zod enum depend on).

## Open PR overlap check

Checked every changed file against all four open PRs. **No overlap.**

- **#408** (nightly-maintainability): ComposerHero, ComposerCopilot, dropdown-menu, icon, select, Waveform — untouched
- **#407** (nightly-bugs-tests): route-handler, sse, osml*, gateway/base, cartesia, supabase/middleware, video/elementAttributes, video/layoutKey, PostPromptView, useRendering — untouched
- **#395** (canvas-interaction-bugs): Canvas, AnimatedImagePreview, overlays, results, keyboardGuards, templates — untouched
- **#394** (bring-your-own-image): ElementContainer, CharacterEditModal, generation/queue, blob, ImageWithShimmer, upload buttons, package.json — untouched

### DIFF
--- a/lib/canvas/createCanvasNode.ts
+++ b/lib/canvas/createCanvasNode.ts
@@ -1,11 +1,11 @@
-import type {
-	CanvasContentElement,
-	CanvasElementType,
+import {
+	ELEMENT_TYPES,
+	type CanvasContentElement,
+	type CanvasElementType,
 } from "@/lib/canvas/types";
 import type { ConnectorRegistry } from "@/lib/config/ConfigProvider";
 import { getDefaultConnector } from "@/lib/config/connectorUtils";
 import { resolveAttributeSchema } from "@/lib/connectors/factory";
-import { ELEMENT_CONFIGS } from "./elementConfigs";
 import { ZERO_WIDTH_SPACE } from "./constants";
 import { makeNodeId } from "./nodeUtils";
 
@@ -20,13 +20,13 @@ export function createCanvasNode(
 	connectors: ConnectorRegistry,
 	opts: Opts = {},
 ): CanvasContentElement {
-	const config = ELEMENT_CONFIGS[type];
+	const { connector } = ELEMENT_TYPES[type];
 	const { provider, config: connectorConfig } = getDefaultConnector(
 		connectors,
-		config.connector,
+		connector,
 	);
 	const schema = resolveAttributeSchema(
-		config.connector,
+		connector,
 		provider,
 		connectorConfig?.defaultModel,
 	);
--- a/lib/canvas/elementConfigs.tsx
+++ b/lib/canvas/elementConfigs.tsx
@@ -7,13 +7,14 @@ import {
 	Video,
 	Waveform,
 } from "@/components/ui/icon";
-import type { CanvasElementType, ResultKind } from "@/lib/canvas/types";
-import type { AssetConnectorType } from "@/lib/connectors/types";
+import {
+	ELEMENT_TYPES,
+	type CanvasElementType,
+	type ElementTypeSpec,
+} from "@/lib/canvas/types";
 
-export interface ElementConfig {
+export interface ElementConfig extends ElementTypeSpec {
 	type: CanvasElementType;
-	connector: AssetConnectorType;
-	outputKind: ResultKind;
 	label: string;
 	icon: React.ReactNode;
 	/** Tint for the square type-icon container, keyed to the media-type color. */
@@ -26,8 +27,7 @@ export interface ElementConfig {
 export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 	narration: {
 		type: "narration",
-		connector: "tts",
-		outputKind: "audio",
+		...ELEMENT_TYPES.narration,
 		label: "Narration",
 		icon: <Voice size={16} />,
 		iconBgClass: "bg-media-narration/15",
@@ -36,8 +36,7 @@ export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 	},
 	character: {
 		type: "character",
-		connector: "tts",
-		outputKind: "audio",
+		...ELEMENT_TYPES.character,
 		label: "Character",
 		icon: <User size={16} />,
 		iconBgClass: "bg-media-character/15",
@@ -46,8 +45,7 @@ export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 	},
 	image: {
 		type: "image",
-		connector: "image",
-		outputKind: "image",
+		...ELEMENT_TYPES.image,
 		label: "Image",
 		icon: <ImageIcon size={16} />,
 		iconBgClass: "bg-media-image/15",
@@ -56,8 +54,7 @@ export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 	},
 	animated_image: {
 		type: "animated_image",
-		connector: "animated_image",
-		outputKind: "video",
+		...ELEMENT_TYPES.animated_image,
 		label: "Animated image",
 		icon: <Motion size={16} />,
 		iconBgClass: "bg-media-animated/15",
@@ -66,8 +63,7 @@ export const ELEMENT_CONFIGS: Record<CanvasElementType, ElementConfig> = {
 	},
 	clip: {
 		type: "clip",
-		connector: "video",
-		outputKind: "video",
+		...ELEMENT_TYPES.clip,
 		label: "Clip",
 		icon: <Video size={16} />,
 		iconBgClass: "bg-media-clip/15",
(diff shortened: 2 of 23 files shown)


umairnadeem (commented):
the `ELEMENT_TYPES` registry is the right call, and killing `ELEMENT_ROLES`/`LAYER_TYPES` + the hand-maintained union is a genuinely good collapse. deriving `CANVAS_ELEMENT_TYPES`/`FOREGROUND_TYPES` from it means the compiler now enforces what tribal knowledge used to.

main thing i'd fix before merge: three test files still `vi.mock` a module the code under test no longer imports, so the mocks are dead and the tests silently changed what they exercise. "no test needed updating" is the tell, not the proof. rest is nits.

small correction on the description: `useProject` calls `useConfig()` internally, so `CharacterPill`/`CharactersPicker` haven't dropped their `ConfigProvider` dependency, it just went implicit. still a net win (one read path), just not the same claim.

lib/canvas/createCanvasNode.ts
```diff
@@ -1,11 +1,11 @@
-import type {
-	CanvasContentElement,
-	CanvasElementType,
+import {
+	ELEMENT_TYPES,
```
umairnadeem, line 2: now that this reads `ELEMENT_TYPES` instead of `ELEMENT_CONFIGS`, three test files are mocking a module their SUT no longer imports:

- `lib/canvas/__tests__/createCanvasNode.test.ts:3`
- `lib/canvas/__tests__/insertElement.test.ts:4`
- `lib/script/refine/__tests__/applyOps.test.ts:6`

all three still do `vi.mock("../elementConfigs", ...)` with a trimmed 2-3 type fixture. `createCanvasNode.ts` / `insertElement.ts` / `applyOps.ts` don't touch `elementConfigs` anymore, so those mocks are dead and the tests now run against the real 7-entry registry. they pass, but they're testing something different than they were, and the fixture is cruft that lies to the next reader. worth deleting the mocks in this PR (or repointing them at `@/lib/canvas/types` if the narrowing was load-bearing).

this is also why "no test needed updating" isn't the clean signal the description reads it as. the mocks silently absorbed the change.

lib/canvas/types.ts
```diff
@@ -1,39 +1,90 @@
...
+	connector: AssetConnectorType;
+	outputKind: ResultKind;
+	role: ElementRole;
+	layer: LayerType;
+};
+
+/**
+ * The one home for element types. The `CanvasElementType` union, the type-guard
+ * sets, the timeline roles, and the editor's config map all derive from here, so
+ * adding a type is one entry rather than four lists that must agree.
```
umairnadeem: nit: this block narrates what the code already shows (`keyof typeof`, the derived sets, the spread in `elementConfigs` are all right there) and it'll rot the moment a 5th consumer shows up. repo convention is comments only where clean code can't carry the why. the `/** presentation-free facts */` one-liner above earns its keep, this one doesn't imo.

lib/canvas/types.ts
```diff
@@ -1,39 +1,90 @@
...
+ */
+export const ELEMENT_TYPES = {
```
umairnadeem, line 20: SG: `lib/connectors/llm/plugins/refine.ts:7` already has a local `const ELEMENT_TYPES = [...CANVAS_ELEMENT_TYPES].join(", ")`, so we now have two things called `ELEMENT_TYPES` in the same domain, one being the canonical registry and one being a comma-joined string for a prompt. mildly evil for grep + anyone who fat-fingers an auto-import. perhaps rename the local one to `ELEMENT_TYPE_LIST` while you're consolidating.

lib/canvas/elementConfigs.tsx
```diff
@@ -7,13 +7,14 @@ import {
...
-import type { CanvasElementType, ResultKind } from "@/lib/canvas/types";
-import type { AssetConnectorType } from "@/lib/connectors/types";
+import {
+	ELEMENT_TYPES,
+	type CanvasElementType,
+	type ElementTypeSpec,
+} from "@/lib/canvas/types";
 
-export interface ElementConfig {
+export interface ElementConfig extends ElementTypeSpec {
```
umairnadeem, line 16: two leftovers from the migration:

1. every entry still hand-writes `type: "narration"` etc, duplicating the key it's already under. `ELEMENT_LIST = Object.values(...)` is the only reason it exists afaict, and that could map the key back in. minor, but it's exactly the "lists that must agree by hand" this PR is purging.
2. the layering leak is only half closed. `OutputPreview.tsx:20`, `ForegroundPreview.tsx:14` and `ModelBadge.tsx:15` still pull `outputKind`/`connector` off `ELEMENT_CONFIGS`, i.e. they import the icon module purely to read a spec field, same as `resolve.ts` used to. they're client components so it costs nothing at runtime, but the read path is the one you just declared non-canonical. cheap to point them at `ELEMENT_TYPES` here rather than leaving the seam half-migrated.

PR #410, opened 2026-07-13
fix: enable player controls in fullscreen

umairnadeem:
## Summary

Enables the Remotion Player's built-in controls (play/pause, seek bar, volume, time display, fullscreen button) in fullscreen mode, where the custom `BottomTransportBar` is hidden.

## Bug fixed

- **Fullscreen player has no play/pause/seek controls** (#343) — When a user entered fullscreen via the `BottomTransportBar`'s fullscreen button, only click-to-toggle play/pause worked (via the transparent overlay in `VideoPreview`). There was no visible transport, no seek bar, no volume control, and no time display.

## Change

- `app/components/video/VideoPreview.tsx` — Added fullscreen state tracking via the browser Fullscreen API (`document.fullscreenElement` + `fullscreenchange` event). Passes `controls={isFullscreen}` to the dynamically-imported `RemotionPlayer`, enabling Remotion's built-in transport overlay only when fullscreen is active. In normal (non-fullscreen) mode, the custom `BottomTransportBar` is used as before.

## Validation

| check | result |
|---|---|
| `npm run lint` | pass |
| `npm run typecheck` | pass |
| `npm run format:check` | pass |
| `npm run test:run` | 866 passed / 100 files |
| `npm run build` | pass |
| `npx knip` | clean |

## Open PR overlap check

Checked `VideoPreview.tsx` against all five open PRs (#394, #395, #407, #408, #409). None touch this file or the video player transport area. No overlap.

## Skipped items

- Did not attempt to style Remotion's built-in controls to match OpenSlop's design system. The custom `BottomTransportBar` already serves that role in normal view; in fullscreen, the native Remotion controls are functional and acceptable.
- Did not add a dedicated fullscreen transport overlay using the app's own components — that would be a larger feature requiring UX decisions about placement, auto-hide behavior, and mobile layout. The Remotion built-in controls are the minimal fix.

Closes #343

PR #411, opened 2026-07-14
fix: nightly bug fixes 2026-07-14

umairnadeem:
## Bugs fixed

1. **Real video clips render with zero duration** (`lib/providers/video/base.ts`, `lib/api/handlers/video.ts`)
   - `RunwareVideo._generate` records the requested `duration` in its metadata, but that metadata was discarded — `videoHandler.process` only persisted `{ providerJobId }`. On poll, Runware's response carries just `taskUUID` and `status`, so the completed bundle's manifest had no `durationSec`. Downstream resolves to 0.
   - Fix: `process` now carries `durationSec` alongside `providerJobId` in job metadata. `BaseVideoProvider.poll` accepts an optional `durationSec` and uses it as a fallback when the provider doesn't echo one back.

2. **Polling a finished video job re-uploaded the bundle under a new id** (`lib/api/handlers/video.ts`)
   - `pollJob` called the handler's `poll` unconditionally, and `videoHandler.poll` re-polled upstream whenever `providerJobId` was set — including for jobs already `completed`. Each poll re-ran `store()`, minting a fresh bundle id. Once Runware drops the task, the same call threw and turned a completed job into a 500.
   - Fix: `completed`/`failed` jobs now short-circuit with `rowView(job)` without touching the provider.

3. **Every video submission wrote a junk empty blob** (`lib/providers/video/base.ts`, `mock.ts`)
   - `BaseProvider.generate` always calls `store()`, but at submit time an async video job has no URL. `toFiles()` returned `[]`, yet `AssetBundle.upload` still wrote an empty manifest. `poll()` had an inline guard for this; `generate()` didn't.
   - Fix: Moved the guard into an overridden `store()` so both paths share it.

## Tests added

- `lib/api/__tests__/video-handler.test.ts` (new, 8 tests): Duration survives submit, process throws on missing jobId, completed/failed jobs short-circuit, duration passed to provider.poll, status mappings persist correct shapes.
- `lib/providers/__tests__/runware-video.test.ts` (3 new tests): Generate uploads nothing for job without video, poll stores submitted duration, in-progress poll doesn't upload.

## Open PR overlap check

Open PRs considered: #410 (VideoPreview fullscreen), #409 (architecture rethink — massive scope), #408 (maintainability 7/13), #407 (nightly bugs 7/13), #395 (canvas interaction), #394 (bring-your-own-image). This PR touches only `lib/api/handlers/video.ts`, `lib/providers/video/base.ts`, `lib/providers/video/mock.ts`, and test files — none of which appear in any open PR. No overlap.

## Validation

- `npm test`: 101 files, 877 tests (all passing)
- `npm run build`: TypeScript + Next.js build passes cleanly

### DIFF
--- a/lib/api/__tests__/video-handler.test.ts
+++ b/lib/api/__tests__/video-handler.test.ts
@@ -1,0 +1,171 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import type { VideoGenerateParams } from "@/lib/connectors/types";
+import type { JobRow } from "../jobs";
+import { videoHandler } from "../handlers/video";
+
+const mockGenerate = vi.fn();
+const mockPoll = vi.fn();
+const mockUpdateJob = vi.fn();
+
+vi.mock("../providers", async (importOriginal) => ({
+	...(await importOriginal<typeof import("../providers")>()),
+	getVideoProvider: () => ({ generate: mockGenerate, poll: mockPoll }),
+}));
+vi.mock("../jobs", () => ({
+	updateJob: (...args: unknown[]) => mockUpdateJob(...args),
+}));
+
+type VideoJobRow = Omit<JobRow, "request" | "metadata"> & {
+	request: VideoGenerateParams;
+	metadata: { providerJobId?: string; durationSec?: number };
+};
+
+function makeJob(overrides: Partial<VideoJobRow> = {}): VideoJobRow {
+	return {
+		id: "job-1",
+		user_id: "user-1",
+		project_id: null,
+		connector_type: "video",
+		status: "processing",
+		request: { prompt: "a sunset", duration: 8 },
+		result: null,
+		metadata: { providerJobId: "provider-1", durationSec: 8 },
+		error: null,
+		created_at: "",
+		updated_at: "",
+		...overrides,
+	};
+}
+
+beforeEach(() => {
+	vi.clearAllMocks();
+});
+
+describe("videoHandler.process", () => {
+	it("keeps the submitted duration in metadata so poll can restore it", async () => {
+		mockGenerate.mockResolvedValue({
+			id: "",
+			provider: "runware",
+			result: {},
+			metadata: { jobId: "provider-1", durationSec: 8 },
+		});
+
+		const outcome = await videoHandler.process(makeJob());
+
+		expect(outcome).toEqual({
+			kind: "pending",
+			metadata: { providerJobId: "provider-1", durationSec: 8 },
+		});
+	});
+
+	it("throws when the provider returns no jobId", async () => {
+		mockGenerate.mockResolvedValue({
+			id: "",
+			provider: "runware",
+			result: {},
+			metadata: {},
+		});
+
+		await expect(videoHandler.process(makeJob())).rejects.toThrow(
+			"Video provider returned no jobId",
+		);
+	});
+});
+
+describe("videoHandler.poll", () => {
+	it("returns the stored row for a completed job without re-polling upstream", async () => {
+		const result = {
+			id: "bundle-1",
+			provider: "runware",
+			result: { video: "https://v.mp4" },
+		};
+		const view = await videoHandler.poll?.(
+			makeJob({ status: "completed", result }),
+		);
+
+		expect(view).toEqual({
+			jobId: "job-1",
+			status: "completed",
+			result,
+			error: null,
+		});
+		expect(mockPoll).not.toHaveBeenCalled();
+		expect(mockUpdateJob).not.toHaveBeenCalled();
+	});
+
+	it("returns the stored row for a failed job without re-polling upstream", async () => {
+		const view = await videoHandler.poll?.(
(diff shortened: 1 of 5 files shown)


lib/api/handlers/video.ts
```diff
@@ -15,13 +15,21 @@ export const videoHandler: JobHandler<VideoGenerateParams, VideoMetadata> = {
 		if (!providerJobId) {
 			throw new Error("Video provider returned no jobId for async generation");
 		}
-		return { kind: "pending", metadata: { providerJobId } };
+		return {
+			kind: "pending",
+			metadata: { providerJobId, durationSec: submitted.metadata?.durationSec },
```
umairnadeem, line 20: this makes `durationSec` a second home for knowledge the row already has: `job.request.duration` is the duration the user asked for, and this handler is typed `TypedJobRow<VideoGenerateParams, ...>` so it's right there (`process` already reads `job.request` one line up). CONVENTIONS.md "Structure and reuse": *"DRY is about knowledge, not text. Each piece of knowledge gets one home."*

the copy also depends on every provider faithfully echoing it back from submit, which they don't:
- `MockVideo._generate` hardcodes `durationSec: 5` no matter what was requested
- `RunwareVideo._generate` echoes `params.duration ?? 5` while `submit()` sends `params.duration || 5`. so the two defaults can drift, and today a `duration: 0` submits 5s but records 0.

afaict you can delete the whole round-trip:
```ts
type VideoMetadata = { providerJobId?: string };
...
const upstream = await getVideoProvider().poll(providerJobId, job.request.duration);
```
and drop the `durationSec` echo from `RunwareVideo._generate`. same behavior, one source of truth, works for providers that never echo anything. `BaseVideoProvider.poll` keeps its `result.metadata?.durationSec ?? durationSec` precedence so mock's per-variant durations still win.

lib/providers/video/base.ts
```diff
@@ -30,10 +30,8 @@ export abstract class BaseVideoProvider extends BaseProvider<
 		return r.url ? [{ key: "video", url: r.url }] : [];
 	}
 
-	protected abstract _poll(jobId: string): Promise<VideoJob>;
-
-	async poll(jobId: string): Promise<VideoProviderResponse> {
-		const result = await this._poll(jobId);
+	/** A job with no video yet (submitted or still running) has nothing to upload. */
```
umairnadeem, line 33: nit: this one narrates the code (`toFiles(result).length === 0` already says "no video yet, nothing to upload"). CONVENTIONS.md: *"Comments are a last resort, not a habit... Don't narrate the code."* the `durationSec` docblock right below it is the good kind (providers not echoing it back is a genuine why you can't read off the code), keep that one.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the branch. no hard-rule violations (no `!`, no swallowed errors), one DRY/knowledge-home finding below plus a comment nit.

the real win here is quiet: moving the empty-files check into `BaseVideoProvider.store` means an async `submit` no longer round-trips through `AssetBundle.upload` with zero files just to get a junk bundle back, and `generate` gets the same guard `poll` had. the poll-guard on finished jobs is also the right call. tests are solid.

lib/api/__tests__/video-handler.test.ts
```diff
@@ -0,0 +1,171 @@
...
+			error: null,
+		});
+	});
+
+	it("marks the job failed when the provider reports failure", async () => {
+		mockPoll.mockResolvedValue({
+			id: "",
+			provider: "runware",
+			result: {},
+			metadata: { jobId: "provider-1", status: "failed", error: "no credits" },
```
umairnadeem, line 134: this pins a provider response that nothing in the repo can actually emit, so it's the false-confidence kind of green.

`RunwareVideo.toVideoJob` (`lib/providers/video/runware.ts:11-18`) only ever writes `url`, `jobId` and `status` into metadata, and `MockVideo` never reports a failure at all. grep `error` under `lib/providers/video/` and you get exactly one hit: the `error?: string` field decl in `base.ts:12`. nobody ever assigns it.

so in prod `upstream.metadata?.error ?? "Video generation failed"` (`lib/api/handlers/video.ts:39`) can only ever emit the generic fallback. real runware failures (no credits, content policy, bad model) reach the user as a shrug, and this test says they don't.

fix belongs in the provider, not the test: have `toVideoJob` carry the error payload through and map runware's errored status onto `"failed"`. then this assertion is testing something real.

while you're in there: `status as VideoJobStatus` (runware.ts:15) is an unchecked cast. if runware's errored status string isn't literally `"failed"`, the row sits in `processing` until the client's 600s `awaitCompletion` timeout instead of failing loudly. worth confirming what that string actually is.

umairnadeem (commented):
second pass (no new commits since my last review, so the earlier findings stand). the poll guard and the completed/failed short-circuit hold up under a race read. one new thing: the failure test is green but asserts a shape no provider can produce.

umairnadeem:
we intentionally omit durationSecs from video for now

PR #412, opened 2026-07-14
refactor: nightly maintainability 2026-07-14

umairnadeem:
## Summary

Nightly maintainability + docs pass. No behavior changes. Net -5 lines across 16 files (+219 / -224 after lint-staged formatting).

## Open PR overlap check

Reviewed open PRs #411 (nightly bug fixes), #410 (fullscreen controls), #409 (architecture rethink), #408 (maintainability 7/13), #407 (nightly bug fixes 7/13), #395 (canvas-interaction-bugs), and #394 (bring-your-own-image). **No overlap** — none of the changed files appear in any open PR.

## Architecture changes

- **Extracted `getFunctionName()` in `lib/video/lambda-config.ts`**: The `speculateFunctionName` call with DISK/RAM/TIMEOUT was duplicated in both `app/api/render/route.ts` and `app/api/render/progress/route.ts`. Moved it to a single exported function so the lambda spec lives in one place.

## Maintainability changes

1. **`PanelSelect` component** (`components/ui/panel-select.tsx`, new): Options-driven `<Select>` wrapper for panel rows. Eliminates the repetitive `Select/SelectContent/SelectItem` boilerplate from `LayoutPanel`, `PropertiesPanel`, and `ExportButton`. Typed generic so the selected value round-trips as the union type — callers don't cast `onValueChange`'s bare string back.

2. **Reduced prop drilling in video transport** (`BottomTransportBar.tsx`, `PlayerControls.tsx`): `VolumeControl` and `FullscreenButton` now pull `player` from `usePlayerControl()` context directly instead of receiving it as a prop. `BottomTransportBar` drops the `player={player}` threading for those two children. Also simplified the `ready` guard in `BottomTransportBar` to not redundantly check `layout` after the guard already implies it.

3. **IconButton → React 19 `ref`-as-prop** (`components/ui/icon-button.tsx`): Replaced `React.forwardRef` wrapper with a plain function component — React 19 supports `ref` as a regular prop on any component. Cleaner code, same behavior.

4. **Cache resilience** (`lib/providers/cache.ts`): `fromMetadata` now returns `Result | undefined` instead of always returning `Result`. When a stored row can't be rehydrated (e.g., missing URL), `fromMetadata` returns `undefined` to force a cache miss instead of hoping the default-converted `Result` is valid. `audioBundleCache.fromMetadata` enforces this with a proper guard. Added two tests for the miss-forcing path.

5. **Provider memoization without shared mutable Map** (`lib/api/providers.ts`): Each `defineProvider` closure now owns its instance instead of sharing a module-level `Map<string, unknown>`. Removes the redundant `key` string parameter. Same memoization behavior, simpler code.

6. **Music model validation** (`lib/providers/music/elevenlabs.ts`): Replaced `params.model as "music_v1"` cast with `toMusicModelId()` that validates against `MUSIC_MODELS` and falls back to the default. An unknown model ID now silently uses the default instead of asserting a type that may be wrong.

7. **Type-safe sequence map** (`lib/video/scene-builder.ts`): Changed `Record<string, Sequence[]>` to `Partial<Record<CanvasElementType, Sequence[]>>` via a `SequenceMap` alias. The compiler now knows the keys are valid element types.

8. **Provider type safety in ModelBadge** (`app/components/canvas/elements/ModelBadge.tsx`): Uses `isKnownProvider()` guard instead of the `as ProviderKey` cast. Tightened the `useMemo` to compute model + provider + spec in one pass.

## Complexity delta

- Files changed: 16 (+219 / -224, net -5)
- Removed 3 duplicated `speculateFunctionName` call sites
- 3 components simplified with `PanelSelect` (each drops 8-14 lines of boilerplate)
- 2 video components un-propped (player from context instead of prop)
- 1 unsafe type cast eliminated (music model)
- 1 shared mutable state removed (provider cache Map)

## Validation

| check | result |
|---|---|
| `npm run lint` | pass |
| `npm run build` | pass |
| `npx knip` | clean |
| `npm test -- --passWithNoTests` | 100 files, 868 tests, all passing |

No behavior changes: no test needed updating.

### DIFF
--- a/app/components/video/PlayerControls.tsx
+++ b/app/components/video/PlayerControls.tsx
@@ -6,6 +6,7 @@ import type { VideoLayout } from "@/lib/video/types";
 import { scrollToScene } from "@/app/components/canvas/utils/scrollToScene";
 import { IconButton } from "@/components/ui/icon-button";
 import { formatTime } from "@/lib/video/timestamps";
+import { usePlayerControl } from "./PlayerControlContext";
 import {
 	FRAME_EVENTS,
 	usePlayerMuted,
@@ -57,7 +58,8 @@ export function ScenePill({
 	);
 }
 
-export function VolumeControl({ player }: { player: PlayerRef | null }) {
+export function VolumeControl() {
+	const { player } = usePlayerControl();
 	const volume = usePlayerVolume(player);
 	const muted = usePlayerMuted(player);
 
@@ -93,7 +95,8 @@ export function VolumeControl({ player }: { player: PlayerRef | null }) {
 	);
 }
 
-export function FullscreenButton({ player }: { player: PlayerRef | null }) {
+export function FullscreenButton() {
+	const { player } = usePlayerControl();
 	const onFullscreen = () => {
 		if (!player) return;
 		if (player.isFullscreen()) player.exitFullscreen();
--- a/components/ui/panel-select.tsx
+++ b/components/ui/panel-select.tsx
@@ -1,0 +1,59 @@
+"use client";
+
+import type { ReactNode } from "react";
+import {
+	Select,
+	SelectContent,
+	SelectItem,
+	SelectTrigger,
+	SelectValue,
+} from "@/components/ui/select";
+
+export interface PanelSelectOption<T extends string> {
+	value: T;
+	label: ReactNode;
+	disabled?: boolean;
+}
+
+/**
+ * Options-driven `Select` for panel rows. Generic over the option union so the
+ * selected value round-trips as `T` instead of a bare `string` the caller has
+ * to cast back.
+ */
+export function PanelSelect<T extends string>({
+	value,
+	options,
+	onChange,
+	ariaLabel,
+}: {
+	value: T;
+	options: readonly PanelSelectOption<T>[];
+	onChange: (value: T) => void;
+	ariaLabel: string;
+}) {
+	return (
+		<Select
+			value={value}
+			onValueChange={(next) => {
+				const selected = options.find((option) => option.value === next);
+				if (selected) onChange(selected.value);
+			}}
+		>
+			<SelectTrigger size="sm" aria-label={ariaLabel}>
+				<SelectValue />
+			</SelectTrigger>
+			<SelectContent>
+				{options.map((option) => (
+					<SelectItem
+						key={option.value}
+						value={option.value}
+						disabled={option.disabled}
+						className="text-label"
+					>
+						{option.label}
+					</SelectItem>
+				))}
+			</SelectContent>
+		</Select>
+	);
+}
--- a/lib/api/providers.ts
+++ b/lib/api/providers.ts
@@ -11,55 +11,51 @@ import { MockLLM } from "@/lib/providers/llm/mock";
 import { CartesiaTTS } from "@/lib/providers/tts/cartesia";
 import { MockTTS } from "@/lib/providers/tts/mock";
 
-const cache = new Map<string, unknown>();
-
+/**
+ * Lazily builds the real provider when its API key is configured, else the
+ * mock, and memoizes it for the life of the process.
+ */
 function defineProvider<R, M>(
-	key: string,
 	envVar: string,
 	RealCtor: new (apiKey: string) => R,
 	MockCtor: new () => M,
 ): () => R | M {
+	let instance: R | M | undefined;
 	return () => {
-		if (!cache.has(key)) {
+		if (instance === undefined) {
 			const apiKey = process.env[envVar];
-			cache.set(key, apiKey ? new RealCtor(apiKey) : new MockCtor());
+			instance = apiKey ? new RealCtor(apiKey) : new MockCtor();
 		}
-		return cache.get(key) as R | M;
+		return instance;
 	};
 }
 
 export const getImageProvider = defineProvider(
-	"image",
 	"RUNWARE_API_KEY",
 	RunwareImage,
 	MockImage,
 );
 export const getVideoProvider = defineProvider(
-	"video",
 	"RUNWARE_API_KEY",
 	RunwareVideo,
 	MockVideo,
 );
 export const getMusicProvider = defineProvider(
-	"music",
 	"ELEVENLABS_API_KEY",
 	ElevenLabsMusic,
 	MockMusic,
 );
 export const getSFXProvider = defineProvider(
-	"sfx",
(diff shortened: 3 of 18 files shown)


app/components/video/PlayerControls.tsx
```diff
@@ -57,7 +58,8 @@ export function ScenePill({
 	);
 }
 
-export function VolumeControl({ player }: { player: PlayerRef | null }) {
+export function VolumeControl() {
+	const { player } = usePlayerControl();
```
umairnadeem, line 62: `CONVENTIONS.md` § Frontend: *"Composition over context. Don't thread props through layers that don't use them; prefer composition with children. Reach for context only for genuinely cross-cutting state, never just to skip a level."*

there's no level being skipped here. `BottomTransportBar` is the only caller and it already holds `player` (`BottomTransportBar.tsx:23`), so this isn't prop drilling, it's a one-hop prop being traded for a hidden dependency. two costs:

- these were dumb presentational leaves. now they only render inside `PlayerControlProvider`, and `createRequiredContext` means a future toolbar/preview that wants a mute button throws instead of just passing a ref.
- half-migrated: same file, `TimeDisplay` and `SegmentedSeekBar` still take `player` as a prop. so PlayerControls now has two conventions for the same value.

keep `{ player }` as a prop on both, imo.

lib/providers/music/elevenlabs.ts
```diff
@@ -11,6 +12,15 @@ import {
 	toElevenLabsOutputFormat,
 } from "../elevenlabs";
 
+type MusicModelId = (typeof MUSIC_MODELS)[keyof typeof MUSIC_MODELS];
+
+const MUSIC_MODEL_IDS: readonly MusicModelId[] = Object.values(MUSIC_MODELS);
+const DEFAULT_MUSIC_MODEL: MusicModelId = "music_v1";
+
+function toMusicModelId(model: string | undefined): MusicModelId {
+	return MUSIC_MODEL_IDS.find((id) => id === model) ?? DEFAULT_MUSIC_MODEL;
```
umairnadeem: `MUSIC_MODELS` is `{ "Slop Music v1": "music_v1" }`, so `MusicModelId` is the single literal `"music_v1"` and this whole thing (type alias + ids array + default const + a `.find` over one element) provably always returns `"music_v1"`. that's a lot of machinery for a function with one possible output.

worse than the bloat: it now *silently* rewrites anything it doesn't recognize. if a bad `model` reaches here we generate with the default and nobody hears about it, which cuts against `CONVENTIONS.md` § Hard rules (*"Fail loudly... Never swallow an error on a critical path to keep going"*). model is external input, so per *"Parse, don't validate"* it should be zod'd at the boundary and trusted inward, not laundered in the provider.

and it's a one-off: image/video/llm/tts all just do `params.model || "<default>"`. either keep this consistent with them, or validate at the boundary, but this middle ground is the worst of both.

components/ui/panel-select.tsx
```diff
@@ -0,0 +1,59 @@
...
+	options: readonly PanelSelectOption<T>[];
+	onChange: (value: T) => void;
+	ariaLabel: string;
+}) {
+	return (
+		<Select
+			value={value}
+			onValueChange={(next) => {
+				const selected = options.find((option) => option.value === next);
+				if (selected) onChange(selected.value);
```
umairnadeem: the `find` + `if (selected)` is a defensive guard for a state that can't happen: radix only ever emits the `value` of a rendered `SelectItem`, and those come from `options`. so the miss branch is dead, and if it somehow did fire we'd silently swallow the user's selection (the select would look changed but state wouldn't move).

since `T` is pinned by `value: T` + `options`, `onValueChange={(next) => onChange(next as T)}` is honest about what's going on and drops the dead path. up to you, but the current shape reads like it's protecting against something real.

lib/api/providers.ts
```diff
@@ -11,55 +11,51 @@ import { MockLLM } from "@/lib/providers/llm/mock";
 import { CartesiaTTS } from "@/lib/providers/tts/cartesia";
 import { MockTTS } from "@/lib/providers/tts/mock";
 
-const cache = new Map<string, unknown>();
-
+/**
+ * Lazily builds the real provider when its API key is configured, else the
```
umairnadeem: nit: this comment narrates the 8 lines under it, which already say exactly this (`apiKey ? new RealCtor : new MockCtor`, `let instance` memo). per `CONVENTIONS.md`: *"Comments are a last resort... The code shows what it does."* would strip it. the closure-per-provider swap itself is nice, kills the stringly-keyed cache and the `as R | M` cast in one go.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the head branch. one rule violation below (the context switch in PlayerControls), rest is judgment-call stuff.

the `isKnownProvider` type guard in ModelBadge is the good part of this PR: it kills the `as ProviderKey` casts and the provider narrowing now falls out of the types instead of a hope. `getFunctionName()` giving the two render routes one home for the lambda triple is the same kind of win.

components/ui/panel-select.tsx
```diff
@@ -0,0 +1,59 @@
...
+	label: ReactNode;
+	disabled?: boolean;
+}
+
+/**
+ * Options-driven `Select` for panel rows. Generic over the option union so the
+ * selected value round-trips as `T` instead of a bare `string` the caller has
+ * to cast back.
+ */
+export function PanelSelect<T extends string>({
```
umairnadeem: we already ship this component. `components/ui/select-menu.tsx` (untouched here) exports `SelectMenu<T extends string>` + `SelectMenuOption<T> { value, label, icon? }`, same `value` / `options` / `onChange` contract, generic over the option union for the exact same round-trip-as-`T` reason your docblock gives.

so `components/ui` now has two generic options-driven pickers under confusably similar names, and nothing tells a new panel row which to reach for. that's `CONVENTIONS.md` § Principles: *"One canonical way to build each feature... There's at most one obvious way to do a thing."*

the only real difference is the primitive (radix `Select` vs `DropdownMenu`). two ways out: add `disabled` to `SelectMenuOption` and delete this, or if the two primitives genuinely need to coexist, make the split obvious in the name + docblock so it isn't a coin flip.

umairnadeem (commented):
second pass (no new commits since my last review, so the earlier findings stand). one new one: `PanelSelect` is the second generic options-picker in `components/ui`.

PR #413, opened 2026-07-14
perf: stop the editor shell re-rendering on every Slate operation

umairnadeem:
## The problem

`PostPromptViewInner` owns the `value` state, and Slate hands it a brand-new children array on **every** operation. Nothing below it is memoized, so a single keystroke re-renders the entire editor page:

- `EditorToolbar` (+ `InlineCopilot`, `ExportButton`)
- `EditorSidebar` (+ whichever panel is open, Radix `Select`s and all)
- `AssetsSection` (every `AssetTile`, plus the character/narrator dialogs)
- `UserProfile` (Radix `DropdownMenu`)
- `TopPlayerPanel` / `SidePlayerPanel` → `VideoPanel` → `VideoPreview` → Remotion's `<Player>`, which is a bare `forwardRef` with no `memo`, so its whole `PlayerUI` chrome re-renders too

None of those subtrees read editor content. They read contexts and Zustand selectors that are already stable across edits.

This is worst during script generation: `useScriptSync` applies a Slate transform per streamed chunk, so the whole shell re-renders at token rate (tens of times a second) for the entire generation.

## The fix

`React.memo` on the six shell components that don't depend on editor content, so the per-operation render stops at the Slate tree that actually changed.

Plus one prerequisite: `RefineProvider` passed `{ refineScript, refineLoading, stopRefine }` as a fresh object literal on every render. It lives under `CanvasProviders`, so it re-renders per keystroke, and that unstable context value would have punched straight through the toolbar's memo boundary. Now `useMemo`'d, matching what every other provider in the tree already does.

## Why the boundaries actually hold

I checked every context these components consume for referential stability across an editor change: `ViewMode`, `VideoLayout`, `PlayerControl`, `Render`, `ActiveScene`, `AutoScroll`, `DragTransfer`, `Config`, `GenerationQueue`, `ScriptControl` — all `useMemo`'d on deps that don't change when `value` does. (`VideoLayout` is the interesting one: `layoutKey` is a new string each render but value-equal, so `Object.is` holds and the memo survives.) `Refine` was the only leak, and it's fixed here.

No staleness risk: `useGenerateAll` reads `editor.children` inside its click callback, not at render time.

## Not covered

`BottomTransportBar` has the same problem but is owned by #412 — worth a `memo` there once it lands.

## Validation

`npm run lint`, `prettier --check`, `npm run typecheck`, `npm test` (866 passed), `npm run build` — all green. No behavior or visual changes.

### DIFF
--- a/app/components/video/PlayerPanel.tsx
+++ b/app/components/video/PlayerPanel.tsx
@@ -35,15 +35,13 @@ function FitToAspectRatio({
 }
 
 function TopPlayerPanelComponent() {
-	const maxSize =
-		typeof window !== "undefined" ? window.innerHeight * 0.6 : 500;
 	const aspectRatio = useAspectRatio();
 
 	const { size, handleMouseDown, resizing } = useResize({
 		axis: "vertical",
 		defaultSize: TOP_DEFAULT,
 		minSize: 150,
-		maxSize,
+		maxViewportFraction: 0.6,
 	});
 
 	return (
@@ -63,14 +61,13 @@ function TopPlayerPanelComponent() {
 }
 
 function SidePlayerPanelComponent() {
-	const maxSize = typeof window !== "undefined" ? window.innerWidth * 0.5 : 600;
 	const aspectRatio = useAspectRatio();
 
 	const { size, handleMouseDown, resizing } = useResize({
 		axis: "horizontal",
 		defaultSize: SIDE_DEFAULT,
 		minSize: 250,
-		maxSize,
+		maxViewportFraction: 0.5,
 	});
 
 	return (
--- a/app/components/video/PlayerPanel.tsx
+++ b/app/components/video/PlayerPanel.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import type { ReactNode } from "react";
+import { memo, type ReactNode } from "react";
 import { getAspectRatioValue } from "@/lib/video/aspectRatio";
 import { useAspectRatio } from "@/lib/video/useAspectRatio";
 import { ResizeHandle } from "./ResizeHandle";
@@ -34,7 +34,7 @@ function FitToAspectRatio({
 	);
 }
 
-export function TopPlayerPanel() {
+function TopPlayerPanelComponent() {
 	const maxSize =
 		typeof window !== "undefined" ? window.innerHeight * 0.6 : 500;
 	const aspectRatio = useAspectRatio();
@@ -62,7 +62,7 @@ export function TopPlayerPanel() {
 	);
 }
 
-export function SidePlayerPanel() {
+function SidePlayerPanelComponent() {
 	const maxSize = typeof window !== "undefined" ? window.innerWidth * 0.5 : 600;
 	const aspectRatio = useAspectRatio();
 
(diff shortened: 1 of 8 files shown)


app/components/video/PlayerPanel.tsx
```diff
@@ -34,7 +34,7 @@ function FitToAspectRatio({
 	);
 }
 
-export function TopPlayerPanel() {
+function TopPlayerPanelComponent() {
 	const maxSize =
 		typeof window !== "undefined" ? window.innerHeight * 0.6 : 500;
```
umairnadeem: this reads `window.innerHeight` during render with no resize subscription anywhere, so it's only ever as fresh as the last render. today that's fine by accident: the parent re-renders on every Slate op, so `maxSize` gets recomputed constantly while you type. memo is exactly what takes that away.

after this PR `TopPlayerPanel` only re-renders when `useAspectRatio` or `useResize`'s own state changes, so `maxSize` effectively freezes at mount. `useResize` closes over it in `handleMouseDown` (`[axis, minSize, maxSize]`) and hands it to `attachResizeListeners` at mousedown, so the clamp uses the stale value.

repro: open a project in a tall window, shrink the browser to half height, drag the player divider down. it still lets you drag to 60% of the *old* viewport and the player runs off screen. before this PR, typing one character would have quietly fixed it. same deal on line 66 with `innerWidth` for the side panel.

not arguing against the memo, arguing the component should be pure so the memo is safe. either move the viewport read into `useResize` at drag start (it's already in event-handler land, `window.innerHeight` there is always current), or subscribe properly with a `useSyncExternalStore` on `resize`. former is ~3 lines and kills the SSR `500` / `600` placeholder too.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md, clean against both. the premise checks out: `PostPromptViewInner` owns Slate's `value`, so it re-renders on literally every keystroke and drags the whole shell (toolbar, rail, player, profile) with it. and the `useRefineScript` `useMemo` is the quiet MVP here, that object was the context value in `RefineProvider`, so it was busting every refine consumer on every op. `XComponent` + `memo(XComponent)` also matches the existing `OutputPreview` shape, +1.

one real catch though: memo isn't free when the component isn't pure, and `PlayerPanel` isn't. inline.

app/components/video/PlayerPanel.tsx
```diff
@@ -88,3 +88,6 @@ export function SidePlayerPanel() {
 		</div>
 	);
 }
+
+export const TopPlayerPanel = memo(TopPlayerPanelComponent);
```
umairnadeem, line 89: the shell still re-renders on every slate op after this, because `BottomTransportBar` (`PostPromptView.tsx:83`) is the one sibling left unmemoized, and it's arguably the fattest one.

it takes no props, so it re-renders per keystroke and drags `SegmentedSeekBar` -> `ScrubBar` down with it, which maps a fill node per scene plus the seek tooltip. repro: profiler on, type one char into a scene, `BottomTransportBar` + `SegmentedSeekBar` + `ScrubBar` show up in the commit right next to `Canvas`.

fix is the same one-liner you already applied five times in this PR, just over in `BottomTransportBar.tsx`:

```ts
export const BottomTransportBar = memo(BottomTransportBarComponent);
```

its contexts (PlayerControl, PlayerPosition, VideoLayout) are all stable across editor edits, so the boundary holds.

umairnadeem (commented):
second pass (no new commits since my last review, so the earlier finding stands). the memo boundaries all genuinely hold, i traced the prop/context identity through each one. one sibling got missed though.

PR #414, opened 2026-07-14
refactor: rethink openslop architecture 2026-07-14

umairnadeem:
Nightly architecture pass. One theme: **give each asset-layer contract exactly one home.**

## App understanding

OpenSlop turns a prompt into a rendered video. An LLM streams OSML into a Slate canvas; each element (image, animated image, TTS, music, SFX, video) is queued for generation, generated through a three-layer pipeline (**connector** → **gateway** → **provider**), uploaded to Vercel Blob as an "asset bundle", and composed by Remotion. State is Zustand + Supabase; auth is a Supabase session refreshed in `proxy.ts`.

From scratch, the asset layer has three distinct facts:
1. an asset has a **kind** (image/audio/video), which determines the `AssetResult` field it resolves into;
2. a connector has a **type**, its registry key;
3. a connector has a **storage namespace**, where its bytes land in blob.

Today (2) and (3) were flattened into one overloaded `type` field, and (1) was re-encoded in three places. That's what this PR unpicks. Everything else in the pipeline — the connector/gateway/provider split, the plugin chain, the queue — I found sound and left alone.

## Implemented

**1. The kind → URL-field mapping had three encodings.** An if-chain in `assetUrl.ts`, a private `ASSET_KEY_TO_URL_FIELD` table in `asset-base.ts`, and the `ResultKind` union in `canvas/types.ts`. Now one table in `assetUrl.ts`, keyed off `ResultKind`, feeding **both** the write path (`resolveBundle`) and the read path (`getPrimaryUrl`). The old if-chain ended in a bare `return result.videoUrl`, so any unrecognised kind silently resolved to the video URL; it now returns `undefined`.

**2. The connector `type` discriminant was lying.** `type` was read in exactly one place — `AssetBundle.fromResponse(this.type, …)`, which builds the blob path — so it was really the *storage namespace* wearing the discriminant's name. `BaseAnimatedImageConnector` therefore had to declare `type = "image"` to make its path resolve, meaning the animated-image connector did not round-trip to its own `PROVIDERS` key. Split into an honest `type` (`"animated_image"`) plus a `bundleType` getter naming the namespace it shares with the image route. Behaviour is identical; the invariant is now explicit and tested.

The old lie was visible in the tests: `factory.test.ts` had to *omit* `animated_image` from "creates all connector types", and another case was titled "…despite sharing runtime type". `animated_image` now joins that list.

**3. Reference-image validation was two regexes that disagreed.** The zod boundary used `[a-z+.-]` for the media subtype; the Anthropic provider used `[a-z0-9+.-]`. A subtype containing a digit (`image/jp2`) was accepted by the provider but rejected at the boundary. Both now parse through one `lib/api/imageSource.ts`, which returns a discriminated `ImageSource` — parse, don't validate. The provider keeps its own supported-media-type allow-list, which is genuinely its policy.

**4. Deleted dead `AssetBundle.fromId`.** No callers anywhere. knip can't see it because it's a class member.

## Validation

Full CI locally, all green: `lint`, `format:check`, `typecheck`, `knip`, `build`, `test:run` (875 passing), `test:e2e` (3 passing). Ran `/simplify` over the diff and applied its findings — its main catch was that my first cut introduced a new `AssetKind` union that duplicated `ResultKind`, i.e. consolidated two homes into… still two. Fixed by making `ResultKind` the single home.

## Skipped, with reasons

- **Moving the kind union into `lib/connectors/`** — where it belongs, rather than `lib/canvas/types.ts`. That file is owned by open PR #409, so `ResultKind` stays put and the connector layer imports it. One home today; the right home is a follow-up.
- **`PassthroughProvider` base class** — four providers repeat a no-op `toFiles`/`store` override. CONVENTIONS.md says "duplicating a few lines is fine; a premature shared abstraction is not", and this is four lines. Left alone deliberately.
- **Shared `isTerminalJobStatus` guard** — the `status === "completed" || status === "failed"` check is duplicated in `asset-base.ts` and the queue route, but the queue route belongs to PR #407, so a shared guard would have one caller here. Deferred.
- **Unifying the absolute-URL check in `asset-bundle.ts` with `parseImageSource`** — textually the same regex, but a different piece of knowledge ("is this manifest value already absolute?" vs "is this a legal caller-supplied image?"). DRY is about knowledge, not text.
- **Splitting `ConfigProvider` / `queue.ts` / `ElementContainer`** — the largest overloaded modules, but all are anchored in files owned by open PRs (#409, #394, #413). Left for a night when they're free.

No product behaviour changes.

### DIFF
--- a/lib/api/imageSource.ts
+++ b/lib/api/imageSource.ts
@@ -1,14 +1,14 @@
 const HTTP_URL = /^https?:\/\//i;
-const BASE64_DATA_URI = /^data:([a-z]+\/[a-z0-9+.-]+);base64,(.+)$/i;
+const BASE64_DATA_URI = /^data:(image\/[a-z0-9+.-]+);base64,(.+)$/i;
 
 export type ImageSource =
 	| { kind: "url"; url: string }
 	| { kind: "base64"; mediaType: string; data: string };
 
 /**
- * The one contract for a caller-supplied image: an HTTP(S) URL or a base64 data
- * URI. Parsed at the API boundary and re-parsed by providers that need the
- * decoded parts, so both agree on what a legal image source is.
+ * The one contract for a caller-supplied image: an HTTP(S) URL or a base64
+ * `image/*` data URI. Checked at the API boundary and parsed again by providers
+ * that need the decoded parts, so both agree on what a legal image source is.
  */
 export function parseImageSource(value: string): ImageSource | null {
 	if (HTTP_URL.test(value)) return { kind: "url", url: value };
--- a/lib/api/imageSource.ts
+++ b/lib/api/imageSource.ts
@@ -1,0 +1,21 @@
+const HTTP_URL = /^https?:\/\//i;
+const BASE64_DATA_URI = /^data:([a-z]+\/[a-z0-9+.-]+);base64,(.+)$/i;
+
+export type ImageSource =
+	| { kind: "url"; url: string }
+	| { kind: "base64"; mediaType: string; data: string };
+
+/**
+ * The one contract for a caller-supplied image: an HTTP(S) URL or a base64 data
+ * URI. Parsed at the API boundary and re-parsed by providers that need the
+ * decoded parts, so both agree on what a legal image source is.
+ */
+export function parseImageSource(value: string): ImageSource | null {
+	if (HTTP_URL.test(value)) return { kind: "url", url: value };
+
+	const match = BASE64_DATA_URI.exec(value);
+	if (!match) return null;
+
+	const [, mediaType, data] = match;
+	return { kind: "base64", mediaType: mediaType.toLowerCase(), data };
+}
--- a/lib/api/request-schema-fields.ts
+++ b/lib/api/request-schema-fields.ts
@@ -1,4 +1,5 @@
 import { z } from "zod";
+import { parseImageSource } from "./imageSource";
 
 const optionalCoercedNumber = z
 	.union([z.number(), z.string()])
@@ -23,15 +24,9 @@ export const optionalVideoDuration = {
 
 export const referenceImageUrlOrDataUri = z
 	.string()
-	.refine(
-		(value) =>
-			/^data:[a-z]+\/[a-z+.-]+;base64,/i.test(value) ||
-			/^https?:\/\//i.test(value),
-		{
-			message:
-				"Each referenceImages entry must be a data URI or an HTTP(S) URL",
-		},
-	);
+	.refine((value) => parseImageSource(value) !== null, {
+		message: "Each referenceImages entry must be a data URI or an HTTP(S) URL",
+	});
 
 export const optionalReferenceImages = {
 	referenceImages: z.array(referenceImageUrlOrDataUri).optional(),
--- a/lib/connectors/animated_image/connector.ts
+++ b/lib/connectors/animated_image/connector.ts
@@ -7,8 +7,13 @@ export abstract class BaseAnimatedImageConnector extends BaseAssetConnector<
 	AnimatedImageGenerateParams,
 	AssetResult
 > {
-	readonly type = "image" as const;
+	readonly type = "animated_image" as const;
 	readonly assetKey = "image" as const;
+
+	/** Animated images are generated through the image route, so their bundles land in the image namespace. */
+	protected override get bundleType(): string {
+		return "image";
+	}
 
 	static attributesFor(_model?: string): AttributeSchema {
 		return ANIMATED_IMAGE_ATTRIBUTES;
--- a/lib/connectors/animated_image/connector.ts
+++ b/lib/connectors/animated_image/connector.ts
@@ -10,11 +10,6 @@ export abstract class BaseAnimatedImageConnector extends BaseAssetConnector<
 	readonly type = "animated_image" as const;
 	readonly assetKey = "image" as const;
 
-	/** Animated images are generated through the image route, so their bundles land in the image namespace. */
-	protected override get bundleType(): string {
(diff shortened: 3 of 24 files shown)


lib/connectors/animated_image/connector.ts
```diff
@@ -7,9 +7,14 @@ export abstract class BaseAnimatedImageConnector extends BaseAssetConnector<
 	AnimatedImageGenerateParams,
 	AssetResult
 > {
-	readonly type = "image" as const;
+	readonly type = "animated_image" as const;
 	readonly assetKey = "image" as const;
 
+	/** Animated images are generated through the image route, so their bundles land in the image namespace. */
+	protected override get bundleType(): string {
```
umairnadeem: afaict this doesn't actually give the storage namespace one home, it gives it a *second* one.

the write path is `BaseProvider.blobConfig.type` -> `AssetBundle.upload(...)`. the read path is now `connector.bundleType` -> `AssetBundle.fromResponse(...)`. two independent string literals that must agree or the connector resolves a URL for bytes that live somewhere else. animated_image routes through `OpenSlopImageGateway` -> the image provider (`blobConfig = { type: "image", ... }`), which is the *only* reason `"image"` is correct here, and nothing in the type system or the tests says so.

and the new test doesn't pin it: `asset-base.test.ts` exercises a synthetic `ReroutedConnector`, not `BaseAnimatedImageConnector`. delete this getter and every test still passes, but new animated-image bundles get read from `assets/animated_image/...` while the provider wrote them to `assets/image/...` -> broken image, silently.

the fix that actually collapses it to one home: `AssetBundle.upload` already knows the type (it writes it into the manifest), it just doesn't return it. put `type` on `BundleResponse` and have `fromResponse` read `response.type` instead of taking it as a param. then `bundleType` and this override both disappear, the namespace lives only on the provider, and `cache.ts:111` (the other `fromResponse` caller passing a type by hand) gets fixed for free.

if you'd rather keep the getter, at minimum add a test on the real `BaseAnimatedImageConnector` asserting the bundle lands under `assets/image/`, since that override is load-bearing for every animated image already in blob.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the head branch. the kind -> url-field consolidation is genuinely nice (one table feeding both the read and write path, and killing the `return result.videoUrl` fallthrough is a real bug fix), and the writeup on *why* each skipped item was skipped is the most useful part of the PR.

that said, the flagship item (#2, the honest `type` discriminant) doesn't land where the description says it does: the storage namespace still has two homes, the new one just isn't linked to the old one. details inline, plus a media-type hole in the new parser and one convention citation.

lib/api/imageSource.ts
```diff
@@ -0,0 +1,21 @@
+const HTTP_URL = /^https?:\/\//i;
+const BASE64_DATA_URI = /^data:([a-z]+\/[a-z0-9+.-]+);base64,(.+)$/i;
```
umairnadeem: the type part is `[a-z]+`, so `parseImageSource` happily parses `data:text/html;base64,...` or `data:application/pdf;base64,...` as an image. for a function whose whole job is "the one contract for a caller-supplied image", that's a hole worth closing now that it's the single boundary.

real consequence: `referenceImageUrlOrDataUri` is the *only* gate for image providers (runware et al) which just forward the data URI to the vendor. anthropic is safe because it has its own allow-list, nobody else does.

same regex, one word: `/^data:(image\/[a-z0-9+.-]+);base64,(.+)$/i`. the old regex had the same hole, but this PR is the one claiming to own the contract.

lib/api/request-schema-fields.ts
```diff
@@ -23,15 +24,9 @@ export const optionalVideoDuration = {
...
-	.refine(
-		(value) =>
-			/^data:[a-z]+\/[a-z+.-]+;base64,/i.test(value) ||
-			/^https?:\/\//i.test(value),
-		{
-			message:
-				"Each referenceImages entry must be a data URI or an HTTP(S) URL",
-		},
-	);
+	.refine((value) => parseImageSource(value) !== null, {
```
umairnadeem, line 27: convention: `CONVENTIONS.md` § Backend says *"Parse, don't validate. Check external input once at the boundary (zod) into typed data, then trust the types inward."* this still validates and throws the parse away, and the doc comment on `parseImageSource` says so out loud: *"Parsed at the API boundary and re-parsed by providers that need the decoded parts"*.

the DRY win is real (one regex, one home) so this isn't nothing, but the PR description sells it as parse-don't-validate and it isn't. `.transform(parseImageSource).refine(Boolean)` (or `.pipe()`) so the boundary hands `ImageSource` inward would be the actual thing, though i realize that ripples into `referenceImages?: string[]` on the three param types, so it may be a follow-up.

either do it, or drop the claim from the comment so the next person doesn't trust it.

lib/connectors/asset-base.ts
```diff
@@ -1,26 +1,22 @@
...
 export abstract class BaseAssetConnector<
 	TParams extends { prompt: string },
 	TResult,
 	TGateway extends AssetGateway<TParams> = AssetGateway<TParams>,
 > extends BaseConnector<TParams, TResult> {
-	abstract readonly assetKey: AssetKey;
+	abstract readonly assetKey: ResultKind;
+
+	/** Blob namespace assets are stored under. Equal to `type` unless the connector reuses another type's route. */
+	protected get bundleType(): string {
```
umairnadeem: SG: `bundleType: string` on a PR about honest types. `"imag"` typechecks and quietly writes to a bogus namespace. `ConnectorType` if the getter survives at all.

PR #415, opened 2026-07-15
fix: nightly bug fixes 2026-07-15

umairnadeem:
## Summary

Nightly correctness bug hunt + test pass.

### Bugs fixed

1. **formatBytes rollover bug**: When a byte value rounds to exactly 1024.0 in the current unit (e.g., 1,048,570 bytes → 1023.994 KB → rounds to 1024.0 KB), it should promote to the next unit (1.0 MB) instead of displaying "1024.0 KB". The old code handled this by cascading through hardcoded divisions; the new code uses a loop that checks the rounded value and promotes it.

2. **Runware image provider output format**: The Runware image provider was not specifying `outputFormat`, which could cause the provider to return a different format than the declared `.png` filename and `image/png` content type in the asset bundle. Now explicitly requests `"PNG"` format.

### Tests added

- **`lib/__tests__/format.test.ts`**: 4 test cases covering raw bytes, KB/MB/GB ranges, rollover at the 1024 boundary, and very large sizes staying in GB.
- **`lib/providers/__tests__/runware-image.test.ts`**: Added `outputFormat: "PNG"` assertion to the existing image generation test.

### Validation

- `npm test`: 101 files, 870 tests passing
- `npm run typecheck`: clean
- `npm run lint`: clean
- `npm run format:check`: clean
- `npm run build`: successful

### Open PR overlap check

Checked all 10 open PRs (#394, #395, #407, #408, #409, #410, #411, #412, #413, #414). None of the touched files (`lib/format.ts`, `lib/providers/image/runware.ts`, their test files) appear in any open PR's file list.

### DIFF
--- a/lib/__tests__/format.test.ts
+++ b/lib/__tests__/format.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from "vitest";
+import { formatBytes } from "../format";
+
+describe("formatBytes", () => {
+	it("formats raw bytes below 1 KB", () => {
+		expect(formatBytes(0)).toBe("0 B");
+		expect(formatBytes(512)).toBe("512 B");
+		expect(formatBytes(1023)).toBe("1023 B");
+	});
+
+	it("formats KB, MB, and GB ranges", () => {
+		expect(formatBytes(1024)).toBe("1.0 KB");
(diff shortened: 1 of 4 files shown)


umairnadeem:
useless

PR #416, opened 2026-07-15
refactor: nightly maintainability 2026-07-15

umairnadeem:
## Summary

Nightly maintainability pass. No behavior changes. Net −9 lines across 13 files.

## Open PR overlap check

Checked all 11 open PRs (#394, #395, #407, #408, #409, #410, #411, #412, #413, #414, #415). None of the changed files appear in any open PR. No theme overlap — the connector interface cleanup and plugin assertion helpers are in files (#414 touches asset-base, not base.ts or types.ts; #409 touches canvas types, not connector types).

## Architecture changes

- **Centralized plugin dependency assertions** (`lib/connectors/plugins.ts`): `requireGateway()` and `requireSearchVoices()` replace scattered inline null checks across 4 plugins. Single source of truth for the assertion message and type-narrowing logic.

## Maintainability changes

1. **Removed dead lifecycle methods**: `init()`, `validate()`, `destroy()` from `BaseConnector` and `Connector` interface — were no-ops with zero callers and zero overrides in the entire codebase.
2. **Removed dead `options` field from `ConnectorConfig`**: zero usages anywhere.
3. **Removed dead gateway check in template-mode plugin**: `transformPrompt` checked for gateway but never used it — the check threw an error then the method returned a template string without ever touching the gateway. Gateway param renamed `_ctx`.
4. **Plugged `AttributeSchema` unit passthrough**: `durationDef` now carries `unit: "s"`, and `AttributeSchema.visibleAttributes` passes `unit` through. The renderer is in `ElementContainer.tsx` (owned by #394); this PR does the plumbing so when #394 lands, the unit is available.

## Complexity delta

- **4 inline null checks consolidated** into 2 shared helpers (plugin dependency assertions)
- **3 dead methods removed** (BaseConnector: init, validate, destroy)
- **1 dead config field removed** (ConnectorConfig.options)
- **1 dead gateway check removed** (template-mode plugin — gateway was checked then ignored)
- **1 attribute schema gap filled** (unit passthrough)

## Skipped items with reasons

- **`unit` rendering in `AttributeBadge`** — `ElementContainer.tsx` is owned by #394. The plumbing is done here; the renderer can consume it once #394 lands.
- **All other candidate files** — covered by the 11 open PRs.
- **`ConfigProvider` / `queue.ts` / `ElementContainer`** — anchored in files owned by #394, #409, #413.
- **`lib/video/` files** — owned by #407, #409, #412.

## Validation

| check | result |
|---|---|
| `npm run lint` | pass |
| `npm run build` | pass |
| `npx knip` | clean |
| `npm test` | 100 files, 862 tests, all passing |

### DIFF
--- a/lib/connectors/attributes/common.ts
+++ b/lib/connectors/attributes/common.ts
@@ -35,6 +35,7 @@ export const motionDef = (defaultValue: string): AttributeDef => ({
 export const durationDef = (defaultValue: string): AttributeDef => ({
 	key: "duration",
 	label: "Duration",
+	unit: "s",
 	edit: { kind: "enum", options: DURATION_OPTIONS },
 	default: defaultValue,
 });
--- a/app/components/canvas/elements/AttributeBadge.tsx
+++ b/app/components/canvas/elements/AttributeBadge.tsx
@@ -6,10 +6,7 @@ import type { AttributeSpec } from "@/lib/connectors/attributes/schema";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 import { TextAttributePopover } from "./attributes/TextAttributePopover";
 
-const ATTRIBUTE_UNITS: Record<string, string> = { duration: "s" };
-
-function formatValue(key: string, value: string): string {
-	const unit = ATTRIBUTE_UNITS[key];
+function formatValue(value: string, unit?: string): string {
 	return unit ? `${value}${unit}` : value;
 }
 
@@ -39,7 +36,7 @@ export function AttributeBadge({
 
 	const SpecIcon = spec.icon;
 	const labeled = hideLabel ? (
-		formatValue(attrKey, value)
+		formatValue(value, spec.unit)
 	) : (
 		<>
 			{SpecIcon ? (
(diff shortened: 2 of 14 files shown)


lib/connectors/attributes/common.ts
```diff
@@ -35,6 +35,7 @@ export const motionDef = (defaultValue: string): AttributeDef => ({
 export const durationDef = (defaultValue: string): AttributeDef => ({
 	key: "duration",
 	label: "Duration",
+	unit: "s",
```
umairnadeem, line 38: you added `unit` to the schema and thread it through `visibleAttributes`, but the only renderer (`AttributeBadge`) still reads its own hardcoded `ATTRIBUTE_UNITS = { duration: "s" }`, so `spec.unit` is dead and duration's unit now lives in two places. wire the badge to `spec.unit` or drop the field, otherwise the two silently desync.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the head branch. one thing on the new `unit` field:

PR #418, opened 2026-07-16
fix: nightly bug fixes 2026-07-16

umairnadeem:
## Summary

Nightly correctness bug hunt and test pass. Four real bugs fixed, each with a concrete user-facing failure.

### 1. Refine: "before" inserts land in reverse order (`lib/script/refine/applyOps.ts`)

The anchorMap chaining that keeps consecutive "after" inserts in stream order inverts the order for `position: "before"`. Asking refine to "add two sound effects before the final narration" produced `B, A, narration` instead of `A, B, narration`. Same inversion for no-anchor prepends, which all targeted `[0, 0]`. Verified by test before the fix: 3 before-inserts at one anchor produced `C, B, A, anchor`.

Fix: before-inserts now target the anchor's current path directly (it shifts down as inserts land, so they stack naturally) and no longer chain the map. No-anchor prepends chain through a sentinel key so they land in stream order.

### 2. Zero-width spaces accumulate in element text on every save/load cycle (`lib/canvas/osmlSerializer.ts`, `createCanvasNode.ts`)

Serialize/parse was not idempotent. `createCanvasNode` prepends a ZWSP caret leaf; the serializer emitted it into the OSML text; the next parse prepended a fresh one. Each open + autosave added one ZWSP per element (observed 1 -> 2 -> 3 -> 4 across cycles). After the first reload, empty elements no longer matched the `Node.string(element) === ZERO_WIDTH_SPACE` check, so their placeholder text disappeared, and the invisible characters leaked into the script sent to the refine LLM.

Fix: the serializer strips ZWSPs (caret artifacts are editor-internal), and `createCanvasNode` strips ZWSPs from incoming text so previously corrupted saves heal on load.

### 3. Streaming LLM errors never reach the client; abandoned streams never cancel (`lib/api/sse.ts`)

Two defects in the only production SSE path:

- On a mid-stream provider error (Anthropic 429/529 are routine), `createSSEStreamResponse` called `controller.error()`, which just aborts the already-200 body. The script stops mid-sentence with no error shown anywhere; the real reason existed only in server logs. The non-stream path returns a proper 500 for the same failure. Now the route emits an in-band `{type: "error", message}` frame (the same convention `createSSEResponse` already uses) and `readSSE` rethrows it on the consumer side, where the existing unhandled-rejection toaster surfaces it.
- `readSSE`'s finally block only released the reader lock. Clicking Stop during generation abandoned the iterator but the HTTP connection and the upstream Anthropic generation ran to completion (full token cost). Now the reader is cancelled, which aborts the fetch, cancels the route's ReadableStream, and closes the provider stream up the chain.

### 4. Google OAuth failures are silent (`app/components/AuthForm.tsx`)

`signInWithOAuth` returns `{error}` rather than throwing, and the result was discarded. Clicking "Continue with Google" while the auth endpoint is unreachable did nothing, with no feedback. The error is now rendered through the same state the magic-link path already uses.

## Tests added

- `lib/script/refine/__tests__/applyOps.test.ts`: before-inserts stack in stream order at an anchor; no-anchor prepends stack in stream order.
- `lib/canvas/__tests__/osmlSerializer.test.ts`: serializing editor-shaped nodes (ZWSP leaf + ZWSP-prefixed text) emits clean OSML.
- `lib/canvas/__tests__/createCanvasNode.test.ts`: ZWSPs stripped from incoming saved text.
- `lib/api/__tests__/sse.test.ts`: `createSSEStreamResponse` happy path and mid-stream error frame; `readSSE` throws on error frames and cancels the underlying stream on early exit; full server-to-client error round trip.

## Validation

| Check | Result |
|-------|--------|
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm test` | 100 files, 875 tests, all passing |
| `npm run build` | pass |

## Open PR overlap check

Considered open PRs #417, #416, #415, #414, #413, #412, #411, #410, #395, #394 (full file lists inspected). This PR touches `applyOps.ts`, `osmlSerializer.ts`, `createCanvasNode.ts`, `sse.ts`, `AuthForm.tsx`, and their test files. Zero file overlap with any open PR. Adjacent files were deliberately avoided: `lib/providers/llm/anthropic.ts` (#414), `useRefineScript.ts` (#413), `lib/canvas/guards.ts`/`keyboardGuards.ts` (#394/#395). Nothing here repeats fixes from closed PRs #399/#402/#407.

## Skipped items and rationale

- Middleware redirect cookie propagation and OSML attribute escaping: both were already proposed in closed PR #407 and not accepted; not re-proposing.
- `useOSMLStreamParser` MAX_NODES_TO_SYNC=3 can drop elements when 4+ complete in one render (timing-dependent permanent loss), and the per-project store cache is never invalidated so a second tab can silently overwrite newer saves: both real but need design decisions, flagging instead of patching.
- `/login?error=auth_callback_failed` is produced by the auth callback but nothing renders it, so a failed cross-device magic link is silent. Left out to keep this PR tight; needs a copy decision.

### DIFF
--- a/app/components/AuthForm.tsx
+++ b/app/components/AuthForm.tsx
@@ -60,12 +60,14 @@ export default function AuthForm({
 	};
 
 	const handleGoogleAuth = async () => {
-		await supabase.auth.signInWithOAuth({
+		setError("");
+		const { error } = await supabase.auth.signInWithOAuth({
 			provider: "google",
 			options: {
 				redirectTo: `${window.location.origin}/auth/callback`,
 			},
 		});
+		if (error) setError(error.message);
 	};
 
 	if (sent) {
--- a/lib/api/__tests__/sse.test.ts
+++ b/lib/api/__tests__/sse.test.ts
@@ -1,5 +1,10 @@
 import { describe, expect, it } from "vitest";
-import { formatSSE, createSSEResponse, readSSE } from "../sse";
+import {
+	formatSSE,
+	createSSEResponse,
+	createSSEStreamResponse,
+	readSSE,
+} from "../sse";
 
 function makeSSEStream(chunks: string[]): ReadableStream {
 	const encoder = new TextEncoder();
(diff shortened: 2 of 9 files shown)


umairnadeem:
overengineered

PR #420, opened 2026-07-16
refactor: rethink openslop architecture 2026-07-16

umairnadeem:
Nightly architecture pass. Theme: **the design system should contain exactly what the app uses, and a machine should enforce that.**

## App understanding

OpenSlop turns a prompt into a rendered video. An LLM streams OSML into a Slate canvas, each element generates through the connector/gateway/provider pipeline into Vercel Blob, and Remotion Lambda renders the composition. State is per-project Zustand plus Supabase; auth is a Supabase session refreshed in proxy.ts.

Rebuilt today, `components/ui/` would hold only the primitives the app renders, and the dead-code checker would police all of it. Instead it carried a full vendored shadcn surface (7 never-imported components, 22 never-imported exports inside kept files) and a 374 KB `icons.css` with 724 icon definitions backing 58 actual icons. None of it could rot away because `knip.json` blanket-ignored `components/ui/**`. The gateway layer also had one OpenSlop-internal abstract class with exactly one subclass.

## Implemented

1. **Deleted 7 dead components**: badge, label, slider, switch, tabs, textarea, toolbar. Zero imports repo-wide.
2. **Deleted dead exports in kept files**: 7 dropdown-menu wrappers, 3 select wrappers (scroll buttons stay internal to `SelectContent`), `DialogTrigger`/`DialogClose`, 3 avatar extras, 5 card sections, 6 unused icons.
3. **Pruned `icons.css` from 724 icon defs to the 51 the app references**: 374 KB to 25 KB of root-scope CSS. The add-an-icon workflow in `icon.tsx`'s doc comment is unchanged.
4. **Removed the `components/ui/**` knip ignore**. This is the durable part: knip immediately caught 10 dead exports my manual sweep missed (avatar, card, dialog), and CI now fails if this class of dead code returns.
5. **Removed single-subclass `OpenSlopGatewayClient`**: `OpenSlopLLMGateway` now extends the generic `GatewayClient` directly and constructs its own `OpenSlopClient`. The connector/gateway/provider layering is untouched; a future BYOK gateway extends the generic seam in `lib/gateway/base.ts`, never this OpenSlop-specific base.

Net: **+15 / -1,262 lines.**

## Validation

- `npm run lint`, `format:check`, `typecheck`, `knip`: pass
- `npm run test:run`: 866/866 pass
- `npm run build`: pass
- `npm run test:e2e`: 3/3 smoke tests pass (on port 3100; 3000 was occupied locally)

## Open PR overlap check

Considered all 12 open PRs (#419, #418, #417, #416, #415, #414, #413, #412, #411, #410, #395, #394). In `components/ui/` they touch only `icon-button.tsx` and `panel-select.tsx`; neither is modified here. `knip.json` and `lib/gateway/` appear in no open PR. No file in this diff overlaps any open PR.

## Skipped (found but deferred, with reasons)

- `VideoLayout | null` phantom nullability across the player tree: dead null-guards in ~8 files, but two (#412) are in open-PR files. Worth a pass after #412 lands.
- `sequenceByElementId: Map` inside `VideoLayout` serializes to `{}` across the Lambda JSON boundary (harmless today, type lie): fix touches `scene-builder.ts`, covered by #412.
- `select.tsx` vs `select-menu.tsx` duplication: possible consolidation, but #412 is already reworking `panel-select.tsx`.
- Inert "Get Started" button on the logged-out landing page (`app/page.tsx:67`): it has no handler and no form. Left alone since removing or wiring it changes landing-page UX, which deserves a human call.

The UI-pattern review skills (composition, web design, React best practices) had nothing to evaluate: this diff adds no components and changes no rendering.


components/ui/badge.tsx
```diff
@@ -1,46 +0,0 @@
-import * as React from "react";
```
umairnadeem, line 1: "zero imports repo-wide" isn't true once open PRs count: #394 adds `import { Badge } from "@/components/ui/badge"` in overlays.tsx. the overlap check in the description only compared touched files, so git won't conflict and whichever PR lands second breaks typecheck on master. either keep badge.tsx or sequence this after #394 resolves.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the head branch. one finding, it's about the open-PR overlap check.

umairnadeem:
we need these for future-proofing

PR #428, opened 2026-07-17
fix: make scene duration badge match the displayed time range

### LINKED ISSUE #426: Scene-header duration badge can disagree with the start–end timestamp range (floor vs round)
## Problem

In the scene header, the duration badge can disagree with the duration implied by the start–end timestamp range — e.g. the range reads `0:00 – 0:12` (implies 12s) while the badge next to it reads `13s`.

Both labels are fed the **same** fractional-seconds `duration` from the canonical layout (`seq.start` / `seq.duration` from `buildVideoLayout`), so the data source is fine. The divergence is entirely in two different rounding strategies inside the formatters:

- `formatTimeRange(start, duration)` **floors** each endpoint: implied length = `floor(start + duration) − floor(start)`.
- `formatDuration(duration)` **rounds** the raw length: `round(duration)`.

Scene `start`/`duration` are routinely non-integer (raw `durationSec`, transition overlaps of `TRANSITION_DURATION_SEC`, and `useSceneSegments` subtracting `transitionDurationSec`), so the two disagree in two independent ways:

1. Round-vs-floor of the length: `start=0, duration=12.6` → range `0:00–0:12` but badge `13s`.
2. Independent flooring of both endpoints: `start=5.6, duration=3.4` → range `0:05–0:09` (implies 4s) but badge `3s`.

- "the duration in the scene header can be incongruent with actual duration. the timestamp start-end may have a different duration from whats displayed. use the timestamp start-end as SoT of duration and derive from it instead of relying on different calculations"

## Relevant code

- `app/components/canvas/elements/SceneTimestamp.tsx:10-18` — renders both `formatTimeRange(start, duration)` and `formatDuration(duration)` off the same value
- `lib/video/timestamps.ts:1-14` — `formatTime` floors seconds; `formatTimeRange` floors both endpoints; `formatDuration` rounds the length (the mismatch lives here)
- `app/components/canvas/elements/SceneContainer.tsx:66-73` — `SceneHeader` feeds `seq.start` / `seq.duration` into `SceneTimestamp`
- `lib/video/scene-builder.ts:74-160` — `buildVideoLayout` derives fractional `start`/`duration` (correct SoT; leave as-is)

## Suggested direction

Make the displayed start–end range the single source of truth for the badge. Compute `end = start + duration` once, floor both endpoints, and derive the badge from `floor(end) − floor(start)` so the badge always equals the difference the range shows. Single change point: `SceneTimestamp.tsx` + the formatters in `lib/video/timestamps.ts`. The underlying float timing in `scene-builder.ts` stays the source of truth — only the label derivation changes.

## Acceptance criteria

- [ ] The scene-header duration badge always equals (displayed end − displayed start).
- [ ] Verified against fractional starts/durations (transition overlaps) — no off-by-one between range and badge.
- [ ] No change to actual render/timeline timing; display-only fix.

---

dchaudhari7177:
## Summary

Fixes #426. The scene-header duration badge could disagree with the start–end range shown right next to it — e.g. the range reads `0:00–0:12` while the badge reads `13s`.

Both labels are fed the **same** fractional-seconds `duration`, so the data is fine; the divergence was purely two different rounding strategies in the formatters:

- `formatTimeRange(start, duration)` **floors** each endpoint.
- `formatDuration(duration)` **rounded** the raw length.

Since scene `start`/`duration` are routinely non-integer, the two disagreed in two independent ways:

1. Round-vs-floor of the length: `start=0, duration=12.6` → range `0:00–0:12` but badge `13s`.
2. Independent flooring of both endpoints: `start=5.6, duration=3.4` → range `0:05–0:09` (implies 4s) but badge `3s`.

## Change

Make the displayed range the single source of truth for the badge. `formatDuration` is replaced by `formatRangeDuration(start, duration)`, which derives the badge from the *same* floored endpoints the range uses:

```
floor(start + duration) − floor(start)
```

So the badge is, by construction, always equal to the length the range shows. The shared `floorSeconds` helper is used by both `formatTime` and the new function so they can't drift apart. `formatDuration` had no other callers, so it's removed.

- `lib/video/timestamps.ts` — add `formatRangeDuration`, factor out `floorSeconds`, drop `formatDuration`
- `app/components/canvas/elements/SceneTimestamp.tsx` — badge now uses `formatRangeDuration(start, duration)`

## Tests

`lib/video/__tests__/timestamps.test.ts` — replaced the `formatDuration` block with `formatRangeDuration` coverage, including both `#426` repro cases asserted against `formatTimeRange` so the badge/range consistency is locked in.

## Verification

- `npx vitest run lib/video/__tests__/timestamps.test.ts` → 7 passed
- `npm run typecheck` → clean
- `npx eslint` on the changed files → clean; `prettier --check` → clean (also enforced by the pre-commit hook)
- `npm run knip` could not complete locally (the oxc parser hit a native `ArrayBuffer` OOM on my machine, unrelated to this diff); the change removes one export and adds one that's imported by `SceneTimestamp.tsx`, so it introduces no unused export.

### DIFF
--- a/app/components/canvas/elements/SceneTimestamp.tsx
+++ b/app/components/canvas/elements/SceneTimestamp.tsx
@@ -1,4 +1,4 @@
-import { formatDuration, formatTimeRange } from "@/lib/video/timestamps";
+import { formatRangeDuration, formatTimeRange } from "@/lib/video/timestamps";
 
 export function SceneTimestamp({
 	start,
@@ -13,7 +13,7 @@ export function SceneTimestamp({
 				{formatTimeRange(start, duration)}
 			</span>
 			<span className="rounded bg-muted px-1.5 py-px text-muted-foreground">
-				{formatDuration(duration)}
+				{formatRangeDuration(start, duration)}
 			</span>
 		</span>
 	);
--- a/lib/video/__tests__/timestamps.test.ts
+++ b/lib/video/__tests__/timestamps.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, it } from "vitest";
-import { formatDuration, formatTime, formatTimeRange } from "../timestamps";
+import {
+	formatRangeDuration,
+	formatTime,
+	formatTimeRange,
+} from "../timestamps";
 
 describe("formatTime", () => {
 	it("formats seconds as m:ss", () => {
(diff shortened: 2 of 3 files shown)


umairnadeem (approved):
@dchaudhari7177 thank you for the contribution!

PR #429, opened 2026-07-18
fix: track real queue totals so progress label survives add/cancel

### LINKED ISSUE #427: "X of Y generated" counter is wrong after add/cancel in the generation queue
## Problem

The generation progress label "X of Y generated" goes wrong when items are added to or removed from the queue mid-generation. The label *is* reactive (it re-renders via `useSyncExternalStore`), so this isn't a subscription-staleness bug — the numbers themselves are computed incorrectly by construction.

X and Y aren't a real completed-count and total-count. They're reconstructed from a high-water mark of *concurrently active* items:

- `peak = q.getPeakActive()` is Y — a monotonic high-water mark maintained with `this._peakActive = active === 0 ? 0 : Math.max(this._peakActive, active)` (`queue.ts:247-253`). It only ratchets **up**, and only resets to 0 when active hits 0.
- `done = peak - active` is X — derived, not a real completed counter (`QueueProgressBar.tsx:10`).

This is only correct if the queue is filled once and then only drains. It breaks on membership changes:

- **Cancel/remove a pending item** (`cancel`/`discard`, `queue.ts:199-227`): the item leaves `state` so `active` drops, but `_peakActive` is never decremented. `done = peak − active` *increases* — a cancelled item is falsely counted as "generated," and the denominator stays inflated (still "of 5" when only 4 will run).
- **Add items mid-generation** (`enqueueAll`, `queue.ts:181-197`): already-completed items were deleted from `state`, so they're no longer in `active` and were never persistently counted; `peak` only rises to the new *concurrent* max via `Math.max`, not the cumulative total. So `done` collapses toward 0 and Y understates the real total (e.g. after 2 of 3, adding 3 more shows "0 of 4" instead of "2 of 6").

- "when removing/adding events from the generation queue, the 'x of y generated' loading label doesn't update"

## Relevant code

- `app/components/video/QueueProgressBar.tsx:8-10,27` — computes `active`/`peak`/`done = peak - active` and renders `` `${done} of ${peak} generated` ``
- `lib/generation/queue.ts:125-133` — `getActiveCount()` (live queued+generating) and `getPeakActive()` (returns stored `_peakActive`)
- `lib/generation/queue.ts:247-253` — `notify()` ratchets `_peakActive` with `Math.max`; never decrements
- `lib/generation/queue.ts:181-227` — `enqueueAll` / `cancel` / `discard` mutate `state` (completed items removed at `:160-175`, `:302`), so `active` no longer reflects the cumulative set
- `lib/generation/GenerationQueueProvider.tsx:20-29` — reactive wiring (not the bug; confirms label re-renders on every `notify()`)

## Suggested direction

Track a real cumulative **total** and **completed** count as first-class counters on the queue, instead of reconstructing them from `Math.max(peak) − active`:

- increment total on `enqueueAll` (by the number of newly-enqueued items, without double-counting existing ones), and increment completed on `handleJobSuccess`
- decrement total on `cancel`/`discard` of a still-pending item so cancelled items don't read as "generated"
- reset both to 0 when the queue fully drains to idle (same lifecycle point `_peakActive` resets today)
- render "X of Y" from these counters

## Acceptance criteria

- [ ] Cancelling a pending item decrements the total ("of Y") and does not increment "generated."
- [ ] Adding items mid-generation raises the total and preserves the already-completed count (e.g. 2 of 3 → add 3 → 2 of 6).
- [ ] "X of Y" reads correctly through arbitrary interleavings of add/cancel/complete, and X never exceeds Y.
- [ ] Counters reset cleanly when the queue returns to idle.

## Related

- #244 (generation loading-state clarity) — same progress surface, discoverability-focused; this is the underlying counter-correctness defect

---

dchaudhari7177:
## Summary

Fixes #427. The "X of Y generated" label went wrong whenever items were added to or removed from the queue mid-generation.

X and Y weren't real counts — they were reconstructed from a high-water mark of *concurrently active* items:

- `Y = getPeakActive()`, a monotonic mark ratcheted with `Math.max` that only reset at zero.
- `X = peak − active`, derived rather than counted.

That's only correct if the queue is filled once and then only drains:

- **Cancel/discard a pending item** — the item left `state` so `active` dropped, but the peak never decremented. `done = peak − active` *increased*, so a cancelled item read as "generated" and the denominator stayed inflated.
- **Add items mid-run** — completed items were already removed from `state`, so the peak only rose to the new *concurrent* max, not the cumulative total. `done` collapsed toward 0 and Y understated the total (e.g. after 2 of 3, adding 3 more showed "0 of 4" instead of "2 of 6").

## Change

Track the run as two real cumulative counters on the queue instead of deriving them:

- `_batchTotal` grows by the number of **newly**-enqueued items in `enqueueAll` (already-queued ids don't double-count), and shrinks in `cancel`/`discard` when a still-active item is removed, so an item that will never finish leaves the run rather than being counted as generated.
- `_batchCompleted` increments in `handleJobSuccess` — deliberately there and not in `commitResult`, which is also called directly for results that never went through the queue.
- Both reset when the queue drains (`active === 0`), preserving the previous reset-on-idle behaviour.

`getPeakActive()` is replaced by `getTotalCount()` / `getCompletedCount()`; `QueueProgressBar` reads them directly, so the label and the `aria-valuenow`/`aria-valuemax` pair are now real counts.

## Tests

Six cases added to `lib/generation/__tests__/queue.test.ts` under `progress counters`, covering both failure modes from the issue:

- cancel and discard drop the item from the total instead of counting it generated
- adding work mid-run keeps earlier completions ("2 of 6", the exact case in the report)
- re-enqueueing an already-queued id doesn't double-count
- both counters reset once the queue drains

## Verification

- `npx vitest run lib/generation/__tests__/queue.test.ts` → 42 passed
- Full suite `npx vitest run` → **885 passed / 101 files**, no regressions
- `npm run typecheck` → clean; eslint + prettier clean (also enforced by the pre-commit hook)

### DIFF
--- a/app/components/video/QueueProgressBar.tsx
+++ b/app/components/video/QueueProgressBar.tsx
@@ -5,8 +5,9 @@ import { PlayerShimmer } from "./PlayerShimmer";
 import styles from "./QueueProgressBar.module.css";
 
 export function QueueProgressBar() {
-	const total = useQueueSelector((q) => q.getTotalCount());
-	const done = useQueueSelector((q) => q.getCompletedCount());
+	const active = useQueueSelector((q) => q.getActiveCount());
+	const done = useQueueSelector((q) => q.getGeneratedCount());
+	const total = active + done;
 	const pct = total === 0 ? 0 : (done / total) * 100;
 
 	return (
--- a/app/components/video/QueueProgressBar.tsx
+++ b/app/components/video/QueueProgressBar.tsx
@@ -5,10 +5,9 @@ import { PlayerShimmer } from "./PlayerShimmer";
 import styles from "./QueueProgressBar.module.css";
 
 export function QueueProgressBar() {
-	const active = useQueueSelector((q) => q.getActiveCount());
-	const peak = useQueueSelector((q) => q.getPeakActive());
-	const done = peak - active;
-	const pct = peak === 0 ? 0 : (done / peak) * 100;
+	const total = useQueueSelector((q) => q.getTotalCount());
+	const done = useQueueSelector((q) => q.getCompletedCount());
+	const pct = total === 0 ? 0 : (done / total) * 100;
 
 	return (
 		<PlayerShimmer>
@@ -17,14 +16,14 @@ export function QueueProgressBar() {
 					className="h-1.5 w-full overflow-hidden rounded-full bg-muted"
 					role="progressbar"
 					aria-valuemin={0}
-					aria-valuemax={peak || 1}
+					aria-valuemax={total || 1}
 					aria-valuenow={done}
 					aria-label="Generation progress"
 				>
 					<div className={styles.fill} style={{ width: `${pct}%` }} />
 				</div>
 				<div className="font-body text-label text-muted-foreground tabular-nums">
-					{peak === 0 ? "Preparing…" : `${done} of ${peak} generated`}
+					{total === 0 ? "Preparing…" : `${done} of ${total} generated`}
 				</div>
 			</div>
 		</PlayerShimmer>
--- a/app/components/video/QueueProgressBar.tsx
+++ b/app/components/video/QueueProgressBar.tsx
@@ -8,7 +8,12 @@ export function QueueProgressBar() {
 	const active = useQueueSelector((q) => q.getActiveCount());
 	const done = useQueueSelector((q) => q.getGeneratedCount());
 	const total = active + done;
-	const pct = total === 0 ? 0 : (done / total) * 100;
+	// A reloaded project seeds its finished results back into the queue, so `done`
+	// is already > 0 before anything runs. Only report progress while something is
+	// actually generating — otherwise reopening a generated project flashes
+	// "N of N generated" at 100% during the prefetch window.
+	const generating = active > 0;
+	const pct = generating ? (done / total) * 100 : 0;
 
 	return (
 		<PlayerShimmer>
@@ -18,13 +23,13 @@ export function QueueProgressBar() {
 					role="progressbar"
 					aria-valuemin={0}
 					aria-valuemax={total || 1}
-					aria-valuenow={done}
+					aria-valuenow={generating ? done : 0}
 					aria-label="Generation progress"
 				>
 					<div className={styles.fill} style={{ width: `${pct}%` }} />
 				</div>
 				<div className="font-body text-label text-muted-foreground tabular-nums">
-					{total === 0 ? "Preparing…" : `${done} of ${total} generated`}
+					{generating ? `${done} of ${total} generated` : "Preparing…"}
 				</div>
 			</div>
 		</PlayerShimmer>
--- a/lib/generation/__tests__/queue.test.ts
+++ b/lib/generation/__tests__/queue.test.ts
@@ -755,6 +755,40 @@ describe("GenerationQueue", () => {
 			generationQueue.cancelAll();
 		});
 
+		it("drops a failed item from the total so the run can still finish", async () => {
+			let rejectA: (error: Error) => void = () => {};
+			let resolveB: (value: { url: string }) => void = () => {};
+			generateMock
+				.mockReturnValueOnce(
+					new Promise<{ url: string }>((_resolve, reject) => {
+						rejectA = reject;
+					}),
+				)
+				.mockReturnValueOnce(
+					new Promise<{ url: string }>((resolve) => {
+						resolveB = resolve;
+					}),
+				)
+				.mockReturnValue(pending());
+
+			generationQueue.enqueueAll([makeJob("a"), makeJob("b"), makeJob("c")]);
+			rejectA(new Error("boom"));
+			await vi.advanceTimersByTimeAsync(0);
+
+			// The failure will never complete, so it leaves the run rather than
+			// stranding the bar below 100% for everything that follows.
+			expect(generationQueue.getTotalCount()).toBe(2);
+			expect(generationQueue.getCompletedCount()).toBe(0);
+
+			resolveB({ url: "b" });
+			await vi.advanceTimersByTimeAsync(0);
+
+			expect(generationQueue.getTotalCount()).toBe(2);
+			expect(generationQueue.getCompletedCount()).toBe(1);
+
+			generationQueue.cancelAll();
+		});
+
 		it("keeps earlier completions counted when items are added mid-run", async () => {
 			let resolveA: (v: { url: string }) => void = () => {};
 			let resolveB: (v: { url: string }) => void = () => {};
--- a/lib/generation/__tests__/queue.test.ts
+++ b/lib/generation/__tests__/queue.test.ts
@@ -714,12 +714,17 @@ describe("GenerationQueue", () => {
 	describe("progress counters", () => {
 		const pending = () => new Promise<{ url: string }>(() => {});
 
-		it("counts every newly enqueued item in the total", () => {
+		// The progress bar composes total as active + generated, so these assert on
+		// the two derived counters the queue exposes.
+		const total = () =>
+			generationQueue.getActiveCount() + generationQueue.getGeneratedCount();
+
+		it("counts every newly enqueued item as active", () => {
 			generateMock.mockReturnValue(pending());
 			generationQueue.enqueueAll([makeJob("a"), makeJob("b"), makeJob("c")]);
 
-			expect(generationQueue.getTotalCount()).toBe(3);
-			expect(generationQueue.getCompletedCount()).toBe(0);
+			expect(total()).toBe(3);
+			expect(generationQueue.getGeneratedCount()).toBe(0);
 
 			generationQueue.cancelAll();
 		});
@@ -729,7 +734,7 @@ describe("GenerationQueue", () => {
 			generationQueue.enqueueAll([makeJob("a"), makeJob("b")]);
 			generationQueue.enqueueAll([makeJob("a"), makeJob("c")]);
 
-			expect(generationQueue.getTotalCount()).toBe(3);
+			expect(total()).toBe(3);
 
 			generationQueue.cancelAll();
 		});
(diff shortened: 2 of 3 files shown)


dchaudhari7177:
Good catch on the failure path — that was a real gap, and the strongest argument is the inconsistency it created: `cancel()` and `discard()` both decrement the total because the item will never finish, but a *failed* job will never finish either and was being left in the denominator.

Measured it before changing anything, enqueueing 3 and failing 1:

```
enqueued 3     -> total=3 completed=0 active=3
after 1 FAILS  -> total=3 completed=0 active=2   label "0 of 3"
after 1 OK     -> total=3 completed=1 active=1   label "1 of 3"
```

So the bar tops out at "2 of 3" and never completes for the rest of the run. After dropping the failed item:

```
after 1 FAILS  -> total=2 completed=0 active=2   label "0 of 2"
after 1 OK     -> total=2 completed=1 active=1   label "1 of 2"
```

`handleJobError` now decrements `_batchTotal`. It returns early when the controller was aborted, and a cancelled job has already left the run via `cancel()`, so there is no double count.

I went with decrementing rather than incrementing `_batchCompleted`, since counting a failure as "generated" would misreport it — the run simply has less work in it than it started with, which is the same thing cancel and discard mean.

Added `drops a failed item from the total so the run can still finish` covering both steps above. Reverting just the decrement fails it, so it pins the behaviour rather than restating it.

`vitest run` → **885 passed** (884 on the branch before this, so +1 and no regressions), `tsc --noEmit` clean, prettier/eslint applied via the pre-commit hook.

lib/generation/queue.ts
```diff
@@ -63,7 +63,12 @@ export class GenerationQueue {
 	private history = new Map<string, Map<string, AssetResult>>();
 	private readonly batchSize: number;
 	private _resultVersion = 0;
-	private _peakActive = 0;
+	// Real cumulative counters for the current run, rather than reconstructing
+	// them from the concurrently-active count: membership changes (cancel or
+	// discard, and adding items mid-run) have to move these honestly. Both
+	// reset once the queue drains.
+	private _batchTotal = 0;
+	private _batchCompleted = 0;
```
umairnadeem: these two counters mirror what the queue state already knows, and every membership change has to hand-move them across ~6 sites (enqueue/cancel/discard/success/error + the drain reset). that's two sources of truth that drift, and the failed-job bug gemini flagged was literally one of those sites getting missed. imo derive it and delete the counters:

```ts
getGeneratedCount = () => this.count(s => !isActive(s.status) && s.result != null);
// in the component: done = getGeneratedCount(), total = getActiveCount() + done
```

done = idle-with-result, total = active + done. no reset-on-drain, nothing to keep in sync, and a failed job drops out for free (no result + not active). loses per-run scoping, but afaict there's no product need for that.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md. one thing on the approach itself, not the diff line-by-line:

umairnadeem (changes requested):

dchaudhari7177:
You're right, and the failed-job bug being one of those sites is the argument — I'd added a sixth hand-moved update and immediately missed one. Counters are gone.

The run is now just the set of ids taking part:

```ts
private run = new Set<string>();

getTotalCount = (): number => this.run.size;

getCompletedCount = (): number => {
	let done = 0;
	for (const id of this.run) {
		if (!isActive(this.getElementSnapshot(id).status)) done++;
	}
	return done;
};
```

Because anything that will never finish leaves the run, whatever remains is either still working or done — so completed falls out of each member's status. The two numbers are no longer independent and can't drift apart.

It also deletes the site that was easiest to get wrong: **`handleJobSuccess` now touches no bookkeeping at all.** Membership changes in four obvious places (enqueue adds, cancel/discard/error remove) plus the drain clear, and the displayed numbers are computed from that one structure.

Net −9 lines, and every existing test passed untouched, which is the useful signal here since they assert behaviour rather than the counters.

Verified the failure path still holds by removing the `run.delete` in `handleJobError` — `drops a failed item from the total so the run can still finish` fails, passes with it back:

```
enqueued 3     -> total=3 completed=0 active=3
after 1 FAILS  -> total=2 completed=0 active=2   label "0 of 2"
after 1 OK     -> total=2 completed=1 active=1   label "1 of 2"
after drain    -> total=0 completed=0 active=0
```

`vitest run` → **885 passed**, `tsc --noEmit` clean.

On CONVENTIONS.md — I also cut the narration from this file while I was in there. The remaining comments are the invariant (what leaving the run means) rather than a restatement of the code.

lib/generation/queue.ts
```diff
@@ -63,7 +63,9 @@ export class GenerationQueue {
 	private history = new Map<string, Map<string, AssetResult>>();
 	private readonly batchSize: number;
 	private _resultVersion = 0;
-	private _peakActive = 0;
+	// Ids taking part in the current run. Anything that will never finish leaves,
+	// so what remains is either still working or done.
+	private run = new Set<string>();
```
umairnadeem: this is still a second source of truth, `run` just has to be hand-maintained in 5 places now instead of 6 (add here, delete in cancel/discard/handleJobError, clear in notify). the state map already answers both numbers, so drop the set entirely:

```ts
private count(predicate: (snap: ElementSnapshot) => boolean): number {
	let n = 0;
	for (const snap of this.state.values()) {
		if (predicate(snap)) n++;
	}
	return n;
}

getActiveCount = (): number => this.count((s) => isActive(s.status));

getGeneratedCount = (): number =>
	this.count((s) => !isActive(s.status) && s.result != null);
```

then delete every `this.run.*` line, including the `clear()` in `notify`. a failed or cancelled job drops out for free (no result, not active), so there's no site left to forget.

app/components/video/QueueProgressBar.tsx
```diff
@@ -5,10 +5,9 @@ import { PlayerShimmer } from "./PlayerShimmer";
 import styles from "./QueueProgressBar.module.css";
 
 export function QueueProgressBar() {
-	const active = useQueueSelector((q) => q.getActiveCount());
-	const peak = useQueueSelector((q) => q.getPeakActive());
-	const done = peak - active;
-	const pct = peak === 0 ? 0 : (done / peak) * 100;
+	const total = useQueueSelector((q) => q.getTotalCount());
```
umairnadeem: and the total gets composed here instead of stored:

```ts
const active = useQueueSelector((q) => q.getActiveCount());
const done = useQueueSelector((q) => q.getGeneratedCount());
const total = active + done;
```

every element is either active or done, so `active + done` counts each exactly once. a regenerating item keeps its stale result but is active, so it lands in the total and not in done, which is what you want.

lib/generation/__tests__/queue.test.ts
```diff
@@ -707,4 +707,137 @@ describe("GenerationQueue", () => {
...
+			// already-finished items.
+			generationQueue.enqueueAll([makeJob("d"), makeJob("e"), makeJob("f")]);
+
+			expect(generationQueue.getCompletedCount()).toBe(2);
+			expect(generationQueue.getTotalCount()).toBe(6);
+
+			generationQueue.cancelAll();
+		});
+
+		it("resets both counters once the queue drains", async () => {
```
umairnadeem: heads up this one goes away with the above, nothing resets on drain anymore. worth replacing with the case the set can't express: generate `a`, let it finish, then enqueue `b`, and assert done stays 1 while total is 2.

umairnadeem (commented):
thanks for the quick turnaround, this is closer. still one step short though: the set is the same second source of truth, just holding ids instead of numbers. concrete steer below so it's easy to land.

dchaudhari7177:
Dropped the set entirely and derived both numbers from the state map, per your steer:

- `getActiveCount` / `getGeneratedCount` count `this.state` directly via a small `count(predicate)` helper — no stored counters, no `run.*` sites, no drain reset in `notify`.
- `QueueProgressBar` composes `total = active + done`. Every element is either active or done, so each is counted exactly once; a regenerating item keeps its stale result but is active, so it lands in total and not in done.
- A failed or cancelled job drops out for free (no result, not active), so there's no site left to forget.

Replaced the drain-reset test with the case the set couldn't express: finish `a`, then enqueue `b`, and assert done stays 1 while total is 2. All queue tests pass, tsc and eslint clean.

app/components/video/QueueProgressBar.tsx
```diff
@@ -17,14 +17,14 @@ export function QueueProgressBar() {
...
-					aria-valuemax={peak || 1}
+					aria-valuemax={total || 1}
 					aria-valuenow={done}
 					aria-label="Generation progress"
 				>
 					<div className={styles.fill} style={{ width: `${pct}%` }} />
 				</div>
 				<div className="font-body text-label text-muted-foreground tabular-nums">
-					{peak === 0 ? "Preparing…" : `${done} of ${peak} generated`}
+					{total === 0 ? "Preparing…" : `${done} of ${total} generated`}
```
umairnadeem: reload regression: a saved project seeds finished results into `state` (the constructor keeps `result`, just flips status to idle), so `getGeneratedCount()` is already >0 before anything runs and `total === 0` only ever holds for a never-generated project. net effect: reopening a generated project flashes `N of N generated` at 100% during the prefetch window instead of `Preparing…` (old `peak` was run-scoped so it read 0 on load). gate on `active === 0` here and for `pct` on line 11 so it only shows progress while something's actually generating.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the head branch. really nice iteration on this one, deriving the counters from queue state instead of hand-syncing them is the right call and the regression tests are great. one correctness thing on the reload path below.

app/components/video/QueueProgressBar.tsx
umairnadeem: this PR is good to merge after this one-liner, thank you @dchaudhari7177 !

dchaudhari7177:
Good catch on the reload path — `getGeneratedCount()` is indeed already non-zero on load, so `total === 0` was only ever true for a never-generated project.

Gated on `active > 0` as suggested: `pct` is now `generating ? (done / total) * 100 : 0` and the label falls back to `Preparing…` whenever nothing is in flight, so the prefetch window reads `Preparing…` instead of a 100% `N of N generated`. I also moved `aria-valuenow` behind the same gate so the accessible value matches the rendered bar rather than reporting `done` against an empty bar.

Local: `tsc --noEmit` clean, eslint clean, prettier clean, and the 43 queue tests still pass.

umairnadeem (approved):
thank you for the revisions!

PR #430, opened 2026-07-18
fix: start the visual player on "play from here"

### LINKED ISSUE #425: "Play from here" starts audio but not the visual player (seek/play race)
## Problem

Using "Play from here" on a scene starts the **audio** elements but the visual player does not actually start playing. Audio advances on its own media clock while the Remotion frame driver stays frozen, so when playback later reaches another scene it looks janky (audio ahead of / out of sync with visuals).

Root cause is in `playFromFrame`, which does everything synchronously in one tick:

```js
showPlayer();                     // setVisible(true) — async React state update
playerRef.current?.seekTo(frame); // synchronous, same tick
playerRef.current?.play();        // synchronous, same tick
```

Two compounding races:

1. **Seek→play race on shared audio tags (primary).** `seekTo()` kicks off an async media re-sync; the native `Html5Audio` / shared-audio elements receive `play()` inside the user gesture and start on their own clock, but the Player's frame driver doesn't reliably start from the just-seeked frame. Audio plays, visuals stay frozen — exactly the reported symptom. The app's own scrub code already works around this class of bug (`SegmentedSeekBar.tsx:82-97`): it orders pause → seek → play carefully and calls `silenceMediaIn(player.getContainerNode())` (`:15,:84,:91`) because shared-audio tags advance independently of the composition frame clock. `playFromFrame` does neither.

2. **Hidden-player mount race (secondary).** `showPlayer()` only schedules `setVisible(true)`, and the Player panels are conditionally mounted on `visible` (`PostPromptView.tsx:67,80`). If the player was hidden when "Play from here" is clicked, `playerRef.current` is still `null` at the `seekTo`/`play` calls, so both no-op and the freshly-mounted player comes up paused.

- "sometimes when 'play from here' for a scene, audio elements start playing but the player doesn't actually play, making playback resume from another scene look janky"

## Relevant code

- `app/components/video/PlayerControlContext.tsx:29-36` — `playFromFrame`: `showPlayer()` + `seekTo()` + `play()` all synchronous in one tick (bug site)
- `app/components/canvas/elements/PlayFromHereButton.tsx:23-25` — computes `startFrame`, calls `playFromFrame(startFrame)`
- `app/components/video/VideoPreview.tsx:31-70` — Remotion `<Player>` (`numberOfSharedAudioTags={10}`); registers `PlayerRef` via mount effect
- `app/components/PostPromptView.tsx:67,80` — Player panels conditionally mounted on `visible` (source of the null-ref race)
- `app/components/video/SegmentedSeekBar.tsx:15,82-97` — correct pause/seek/play + `silenceMediaIn` pattern to mirror
- `remotion/compositions/VideoComposition.tsx:49` — `Html5Audio` (shared-audio tags with an independent media clock)

## Suggested direction

Make `playFromFrame` mirror the scrub-bar ordering instead of firing seek+play synchronously:

- Defer `seekTo`/`play` until after the player is registered (wait for `playerRef.current`, i.e. after `showPlayer()` mounts the panel) rather than assuming it exists.
- Order pause → seek → play, and `silenceMediaIn(...)` during the seek so shared-audio tags don't run ahead of the frame clock.
- Verify the frame driver actually starts from the seeked frame before audio advances.

## Acceptance criteria

- [ ] "Play from here" starts the visual player and audio together, in sync, from the chosen scene.
- [ ] Works whether the player was already visible or hidden when clicked.
- [ ] No audio-ahead-of-visuals jank when playback crosses into a later scene.

---

dchaudhari7177:
## Summary

Fixes #425. "Play from here" started the audio elements but left the visual player frozen, so playback later looked janky against out-of-sync audio.

`playFromFrame` did everything in one tick:

```js
showPlayer();                     // async React state update
playerRef.current?.seekTo(frame); // synchronous, same tick
playerRef.current?.play();        // synchronous, same tick
```

That loses the player two ways, both fixed here.

**1. Seek→play race on the shared audio tags (primary).** `seekTo()` starts an async media re-sync while the frame driver is still running, so the shared `Html5Audio` tags take the `play()` on their own media clock and advance while the driver stays parked — audio plays, visuals don't. `SegmentedSeekBar` already avoids exactly this by ordering pause → seek → `silenceMediaIn` → play. `playFromFrame` now uses the same sequence.

**2. Hidden-player mount race (secondary).** `showPlayer()` only *schedules* `setVisible(true)`, and the player panels are mounted on `visible`, so clicking "Play from here" while the player was hidden left `playerRef.current === null` and both calls no-opped — the freshly mounted player came up paused. The frame is now held in a ref and replayed when `registerPlayer` fires.

## Change

- New `startPlaybackAt.ts` holds the ordering as a pure function over a minimal `PlaybackTarget` (the four `PlayerRef` methods it needs), so the sequence is unit-testable without a DOM.
- `PlayerControlContext` uses it, and keeps a `pendingFrameRef` for the not-yet-mounted case.
- `silenceMediaIn` moves out of `SegmentedSeekBar` into `silenceMedia.ts`; both call sites now share one implementation instead of the fix duplicating it.

## Tests

`app/components/video/__tests__/startPlaybackAt.test.ts` pins the ordering — `pause` before `seekTo`, `play` strictly after the seek, silencing between seek and play, and a null container node not throwing.

I deliberately did **not** add a React-rendering test for the mount race: the repo has no jsdom or testing-library (component tests here are pure-logic only, and `vitest.config.ts` has no DOM environment), so adding that infrastructure felt like a separate change rather than something to bundle into a bug fix. Happy to add it if you'd like it in scope.

## Verification

- `npx vitest run app/components/video/__tests__/startPlaybackAt.test.ts` → 5 passed
- Full suite → **883 passed / 102 files**
- `npm run typecheck` → clean; eslint clean; prettier clean (pre-commit hook also ran)

### DIFF
--- a/app/components/video/PlayerControlContext.tsx
+++ b/app/components/video/PlayerControlContext.tsx
@@ -20,36 +20,17 @@ export { usePlayerControl };
 export function PlayerControlProvider({ children }: { children: ReactNode }) {
 	const playerRef = useRef<PlayerRef | null>(null);
 	const [player, setPlayer] = useState<PlayerRef | null>(null);
-	// Set when "play from here" fires before the player panel has mounted; the
-	// frame is replayed as soon as the player registers.
-	const pendingFrameRef = useRef<number | null>(null);
 	const { showPlayer } = usePlayerPosition();
 
 	const registerPlayer = useCallback((p: PlayerRef | null) => {
 		playerRef.current = p;
 		setPlayer(p);
-
-		const pendingFrame = pendingFrameRef.current;
-		if (p && pendingFrame !== null) {
-			pendingFrameRef.current = null;
-			startPlaybackAt(p, pendingFrame);
-		}
 	}, []);
 
 	const playFromFrame = useCallback(
 		(frame: number) => {
 			showPlayer();
-
-			const p = playerRef.current;
-			if (!p) {
-				// showPlayer() only schedules the state update, so the panel holding
-				// the player has not mounted yet. Seeking now would no-op and the
-				// player would come up paused.
-				pendingFrameRef.current = frame;
-				return;
-			}
-
-			startPlaybackAt(p, frame);
+			startPlaybackAt(playerRef.current, frame);
 		},
 		[showPlayer],
 	);
--- a/app/components/video/PlayerControlContext.tsx
+++ b/app/components/video/PlayerControlContext.tsx
@@ -4,6 +4,7 @@ import type { PlayerRef } from "@remotion/player";
 import { useCallback, useMemo, useRef, useState, type ReactNode } from "react";
 import { createRequiredContext } from "@/lib/components/createRequiredContext";
 import { usePlayerPosition } from "./PlayerPositionContext";
+import { startPlaybackAt } from "./startPlaybackAt";
 
 type PlayerControl = {
 	player: PlayerRef | null;
@@ -19,18 +20,36 @@ export { usePlayerControl };
 export function PlayerControlProvider({ children }: { children: ReactNode }) {
 	const playerRef = useRef<PlayerRef | null>(null);
 	const [player, setPlayer] = useState<PlayerRef | null>(null);
+	// Set when "play from here" fires before the player panel has mounted; the
+	// frame is replayed as soon as the player registers.
+	const pendingFrameRef = useRef<number | null>(null);
 	const { showPlayer } = usePlayerPosition();
 
 	const registerPlayer = useCallback((p: PlayerRef | null) => {
 		playerRef.current = p;
 		setPlayer(p);
+
+		const pendingFrame = pendingFrameRef.current;
+		if (p && pendingFrame !== null) {
+			pendingFrameRef.current = null;
+			startPlaybackAt(p, pendingFrame);
+		}
 	}, []);
 
 	const playFromFrame = useCallback(
 		(frame: number) => {
 			showPlayer();
-			playerRef.current?.seekTo(frame);
-			playerRef.current?.play();
+
+			const p = playerRef.current;
+			if (!p) {
+				// showPlayer() only schedules the state update, so the panel holding
+				// the player has not mounted yet. Seeking now would no-op and the
+				// player would come up paused.
+				pendingFrameRef.current = frame;
+				return;
+			}
+
+			startPlaybackAt(p, frame);
 		},
 		[showPlayer],
 	);
--- a/app/components/video/startPlaybackAt.ts
+++ b/app/components/video/startPlaybackAt.ts
@@ -1,0 +1,23 @@
+import type { PlayerRef } from "@remotion/player";
+import { silenceMediaIn } from "./silenceMedia";
+
+/** The slice of PlayerRef needed to start playback, so the sequence stays testable. */
+export type PlaybackTarget = Pick<
+	PlayerRef,
+	"pause" | "seekTo" | "play" | "getContainerNode"
+>;
+
+/**
+ * Start playback at `frame`, in the order the scrub bar already uses.
+ *
+ * Seeking while the frame driver is running lets the shared audio tags start on
+ * their own media clock from the old position, which leaves audio playing
+ * against a frozen driver. Pausing first, then seeking, then silencing any tag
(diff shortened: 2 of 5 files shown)


app/components/video/PlayerControlContext.tsx
```diff
@@ -19,18 +20,36 @@ export { usePlayerControl };
...
 	const { showPlayer } = usePlayerPosition();
 
 	const registerPlayer = useCallback((p: PlayerRef | null) => {
 		playerRef.current = p;
 		setPlayer(p);
+
+		const pendingFrame = pendingFrameRef.current;
+		if (p && pendingFrame !== null) {
+			pendingFrameRef.current = null;
+			startPlaybackAt(p, pendingFrame);
```
umairnadeem: the whole pending-frame replay is more machinery than this needs imo. if the player's hidden when someone hits 'play from here', it's fine to just seek and not play. dropping it also kills the autoplay-gesture footgun the bot flagged on L50 (a play() fired after the mount isn't user-initiated) instead of working around it. so registerPlayer should go back to just register, no startPlaybackAt call.

app/components/video/PlayerControlContext.tsx
```diff
@@ -19,18 +20,36 @@ export { usePlayerControl };
...
 	}, []);
 
 	const playFromFrame = useCallback(
 		(frame: number) => {
 			showPlayer();
-			playerRef.current?.seekTo(frame);
-			playerRef.current?.play();
+
+			const p = playerRef.current;
+			if (!p) {
```
umairnadeem: once the pending-frame path is gone this guard is dead weight. let startPlaybackAt take `PlaybackTarget | null` and use `?.` inside (`silenceMediaIn(player?.getContainerNode() ?? null)`), then playFromFrame collapses to `showPlayer(); startPlaybackAt(playerRef.current, frame)` and a hidden player just no-ops.

app/components/video/startPlaybackAt.ts
```diff
@@ -0,0 +1,23 @@
+import type { PlayerRef } from "@remotion/player";
+import { silenceMediaIn } from "./silenceMedia";
+
+/** The slice of PlayerRef needed to start playback, so the sequence stays testable. */
+export type PlaybackTarget = Pick<
+	PlayerRef,
+	"pause" | "seekTo" | "play" | "getContainerNode"
+>;
+
+/**
```
umairnadeem: CONVENTIONS.md: "Comments are a last resort... Don't narrate the code." this 8-line block plus the PlaybackTarget one-liner are mostly narration. one line on the non-obvious bit (pause before seek so the shared audio tags don't keep running on their own media clock) covers it.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md. the seek->play ordering fix is the good part, but the secondary mount-race goal is adding more complexity than it's worth. a few things to change:

dchaudhari7177:
All three done — you're right that the mount-race half wasn't earning its keep.

**Pending-frame replay is gone.** `registerPlayer` is back to just registering:

```ts
const registerPlayer = useCallback((p: PlayerRef | null) => {
	playerRef.current = p;
	setPlayer(p);
}, []);
```

**`startPlaybackAt` takes `PlaybackTarget | null`**, so `playFromFrame` collapses to exactly what you described and a hidden player no-ops:

```ts
const playFromFrame = useCallback(
	(frame: number) => {
		showPlayer();
		startPlaybackAt(playerRef.current, frame);
	},
	[showPlayer],
);
```

That also removes the autoplay footgun rather than working around it: `play()` is now only ever reached synchronously from the click, never from a later mount, so the user gesture is never lost.

**Comments trimmed** to the one non-obvious line, per CONVENTIONS.md:

```ts
// Pause before seeking, or the shared audio tags keep running on their own
// media clock from the old position.
```

The `PlaybackTarget` one-liner and the 8-line block are gone. Net −22 lines across the three files.

Added `no-ops when the player has not mounted` so the nullable path is pinned. `vitest run` → **884 passed**, `tsc --noEmit` clean.

The seek→play ordering and its regression tests are untouched, so #425 is still covered.

umairnadeem (approved):
thank you for your contributon!

PR #431, opened 2026-07-18
fix: nightly bug fixes 2026-07-18

umairnadeem:
Correctness pass over areas not already covered by open PRs. Behavior is unchanged except where it was wrong.

## Bugs fixed

**1. Auth redirect dropped refreshed session cookies, logging users out** (`lib/supabase/middleware.ts`)

`supabase.auth.getUser()` can rotate the session, and the `setAll` callback writes the new tokens onto `supabaseResponse`. The auth-route redirect built a brand-new response, so those `Set-Cookie` headers were discarded. A logged-in user with an expired access token who hits `/login` had their refresh token consumed server-side but never received the replacement, so the next request failed auth and bounced them back to `/login` — a loop. Cookies are now copied onto the redirect.

**2. Stream parser was never reset, so a second generation fused onto the first** (`lib/canvas/osmlStreamParser.ts`, `lib/canvas/useOSMLStreamParser.ts`, `lib/script/ScriptProvider.tsx`)

`useOSMLStreamParser` holds one `OSMLStreamParser` for the provider's lifetime, and the class had no way to clear itself. Submitting a second prompt appended onto the first script's nodes, and any leftover partial tag in `this.buffer` was string-concatenated onto the new stream's first chunk. Because the hook syncs `.slice(-3)`, the first three renders of the "new" script actually showed the tail of the old one. `submitPrompt` now resets the parser before streaming.

**3. Script generation failures were silent** (`lib/script/ScriptProvider.tsx`)

`submitPrompt` had a `try`/`finally` with no `catch`, and `ComposerHero` calls it fire-and-forget. Offline, a bad API key, or any non-2xx left the composer looking like the click never registered, plus an `unhandledrejection`. Failures now surface as a toast.

**4. Runware video failures hung for the full 10-minute poll timeout** (`lib/providers/video/runware.ts`)

`VideoJobStatus` is `queued|processing|completed|failed`, but Runware emits `success`/`error`. The unchecked `as VideoJobStatus` cast meant `mapVideoStatus` in `lib/api/handlers/video.ts` never saw `"failed"`, so a content-filter rejection was reported as `"processing"` until `awaitCompletion` gave up after 600s and threw a bogus timeout. The real provider message was discarded too. Replaced the cast with an explicit terminal-status map and threaded `errorMessage` into `metadata.error`.

**5. Opening a project immediately saved it back** (`app/components/canvas/hooks/useAutosave.ts`)

The editor mounts with `initialValue: []` and is filled a render later by `useProjectRehydrate`, so the `skipNextRef` guard was burned on the empty value and the real content — arriving next render — scheduled a save. Opening any project and touching nothing wrote the full round-tripped script and store snapshot back, bumped `updated_at`, and toasted "Saved". The guard now keys off the first *content*, not the first effect run.

**6. Drag insert-gap stayed wedged open after leaving all droppables** (`app/components/canvas/dnd/useDragAndDrop.ts`)

`handleDragOver` had four early returns that left the previous `dragTransfer` in context, including the `!over?.id` path dnd-kit fires when you drag into empty space. The gap kept pointing at a target that was no longer active. Extracted the branching into a pure `resolveDragTransfer` that always returns a value, so the caller assigns unconditionally.

**7. Regenerated media showed no loading affordance** (`lib/components/ImageWithShimmer.tsx`, `app/components/canvas/elements/MediaWithSkeleton.tsx`)

Both components latched a one-way `loaded` boolean, so a new `src` on the same instance suppressed the skeleton and left the stale image/frozen video on screen until the new file decoded. Most call sites papered over this with a remount `key`; `ForegroundPreview` doesn't. Both now derive loading from `loadedSrc === src`, matching the pattern already approved in #399.

**8. `redirect: "manual"` was overridable by the caller** (`lib/providers/tts/voicePreview.ts`)

`fetch(url, { redirect: "manual", ...init })` spreads `init` last, so a caller-supplied `redirect` wins — defeating the guarantee the doc comment above it makes. The one credentialed caller attaches `Authorization: Bearer <key>`. Moved the security-relevant option after the spread.

## Tests

+16 tests, 894 passing.

- `middleware.test.ts` — rotated cookies survive the auth-route redirect
- `osmlStreamParser.test.ts` — reset drops prior nodes and a partial tag instead of fusing them into the next stream
- `runware-video.test.ts` — `error` maps to failed and carries the provider message, a failure with no message gets a task-scoped fallback, `success` maps to completed, unknown statuses stay processing
- `resolveDragTransfer.test.ts` (new) — every clearing path returns null; hovering an element vs. the scene itself resolves the right index
- `voicePreview.test.ts` (new) — cross-origin and plaintext-HTTP rejection, and `redirect: "manual"` survives an init override

## Skipped (open-PR overlap)

- **`manifest.json` filename collision in `AssetBundle.upload`** — uploading a file named `manifest.json` overwrites it with the bundle manifest, silently returning a 200 with corrupt data. The knowledge home is `lib/api/asset-bundle.ts`, which #414 touches.
- **Abort signal never reaches the connector** — `stopGeneration` aborts a controller whose signal is never passed to `connector.stream`, so cancellation is cooperative and the HTTP stream stays open. Plumbing it through requires `lib/connectors/types.ts`, which #416 touches.
- **Transition frame-rounding drift** — at the default fps=24, `round(0.4 × 24) = 10` frames while the layout assumed 0.4s, accumulating ~0.5s of audio/visual desync over 30 scenes. The fix spans `lib/video/scene-builder.ts`, which #412 touches.
- **Non-atomic job claim in `app/api/queues/asset-generate`** — read-then-write with no compare-and-swap, so at-least-once delivery can double-bill a generation. Non-overlapping, but a guarded update is a bigger change than a nightly pass should land unreviewed.
- `ImageWithShimmer` / `MediaWithSkeleton` / `useAutosave` have no unit tests — the repo has no jsdom or testing-library, and `renderToStaticMarkup` never fires `onLoad`, so any test I could write would pass against the old code too.

## Open PR overlap check

Considered #410, #412–#422, #428, #429, #430 and diffed each with `gh pr diff --name-only`. None of the 15 files here appears in any of them. Nearest neighbors: #415 touches `lib/providers/image/runware.ts` (this PR touches `video/runware.ts`), #421 touches `lib/providers/tts/cartesia.ts` (this touches `tts/voicePreview.ts`), and #418 touches `lib/canvas/osmlSerializer.ts` and `createCanvasNode.ts` (this touches `osmlStreamParser.ts`).

CI: lint, format:check, typecheck, knip, build, and test:coverage all pass locally.

### DIFF
--- a/app/components/canvas/__tests__/resolveDragTransfer.test.ts
+++ b/app/components/canvas/__tests__/resolveDragTransfer.test.ts
@@ -1,0 +1,98 @@
+import { describe, expect, it } from "vitest";
+import { createEditor, type Editor } from "slate";
+import type { DragOverEvent } from "@dnd-kit/core";
+import { resolveDragTransfer } from "../dnd/useDragAndDrop";
+
+function makeEditor(): Editor {
+	const editor = createEditor();
+	editor.children = [
+		{
+			id: "scene-1",
+			type: "scene",
+			children: [
+				{ id: "a", type: "narration", children: [{ text: "" }] },
+				{ id: "b", type: "narration", children: [{ text: "" }] },
+			],
+		},
+		{
+			id: "scene-2",
+			type: "scene",
+			children: [{ id: "c", type: "narration", children: [{ text: "" }] }],
+		},
+	] as unknown as Editor["children"];
+	return editor;
+}
+
+const dragOver = (
+	activeId: string,
+	activeSceneId: string | undefined,
+	over: { id: string; sceneId?: string } | null,
+) =>
+	({
+		active: { id: activeId, data: { current: { sceneId: activeSceneId } } },
+		over: over && {
+			id: over.id,
+			data: { current: { sceneId: over.sceneId } },
+		},
+	}) as unknown as DragOverEvent;
+
+describe("resolveDragTransfer", () => {
+	it("proposes a transfer at the hovered element's index in the target scene", () => {
+		expect(
+			resolveDragTransfer(
+				makeEditor(),
+				dragOver("a", "scene-1", { id: "c", sceneId: "scene-2" }),
+			),
+		).toEqual({
+			itemId: "a",
+			fromSceneId: "scene-1",
+			toSceneId: "scene-2",
+			atIndex: 0,
+		});
+	});
+
+	it("clears the transfer once the pointer leaves every droppable", () => {
+		expect(
+			resolveDragTransfer(makeEditor(), dragOver("a", "scene-1", null)),
+		).toBe(null);
+	});
+
+	it("clears the transfer when the hover target is missing a scene id", () => {
+		expect(
+			resolveDragTransfer(
+				makeEditor(),
+				dragOver("a", "scene-1", { id: "c", sceneId: undefined }),
+			),
+		).toBe(null);
+	});
+
+	it("clears the transfer when the hovered element is unknown to the editor", () => {
+		expect(
+			resolveDragTransfer(
+				makeEditor(),
+				dragOver("a", "scene-1", { id: "ghost", sceneId: "scene-2" }),
+			),
+		).toBe(null);
+	});
+
+	it("returns null for a same-scene reorder", () => {
+		expect(
+			resolveDragTransfer(
+				makeEditor(),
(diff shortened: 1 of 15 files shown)


umairnadeem:
overengineered

PR #433, opened 2026-07-18
perf: nightly performance optimizations 2026-07-18

umairnadeem:
## What

Both canvas `SortableContext`s were handed `items` arrays rebuilt on every render:

- `useDragAndDrop` derived `sceneItems` from Slate's `value`
- `SceneContainer` derived `childIds` from `element.children`

Slate replaces those arrays on every operation, so the `useMemo` deps changed on every keystroke. dnd-kit keys its entire sortable context value on `items` identity, and React context propagation punches straight through `React.memo` — including slate-react's `MemoizedElement`, which is the thing that makes Slate scale.

Net effect: typing one character re-rendered **every scene and every content element in the document**, not just the edited one. `SortableItem` calls `useSortable` for all of them. Each of those re-renders drags in `Node.string(element)`, `resolveAttributeSchema` + `Object.entries` over the visible attributes, `getElementCharacterNames`, and three separate `useGenerate` calls (each with its own `useSyncExternalStore` subscription and a queue-write effect that can trigger a second full-tree pass).

## Fix

A single `useSortableIds` hook pins the id array's identity to its contents, so the sortable context only invalidates when the sortable order genuinely changes. That is also exactly when the dependent render output changes — `useSceneIndex`'s "Scene N" label and the drop-gap padding both derive from the id list — so the invalidation key now matches the real dependency instead of over-approximating it with "any Slate op".

No behavior change. Typing in a large script now re-renders only the edited element; insert, delete and reorder still repaint the affected scenes.

## Validation

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test:run` (885 tests) all pass.

## Open PR overlap check

Considered open PRs #410, #412, #413, #414, #415, #416, #417, #418, #419, #420, #421, #422, #428, #429, #430, #431, #432.

- `SceneContainer.tsx` and the new `useSortableIds.ts` are untouched by any open PR.
- `useDragAndDrop.ts` is also edited by #431, but in a different region (that PR extracts `resolveDragTransfer` out of `handleDragOver`). This PR only replaces the `sceneItems` memo, which is the root cause and has no other home.
- Deliberately skipped as overlapping: the Remotion/seek-bar render path (#417), the editor shell re-render work (#413), and the drag-time full-document `Editor.nodes` scans in `handleDragOver`/`handleDragEnd` (#431's region).


umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md. one thing on `app/components/canvas/dnd/useSortableIds.ts`: `SEPARATOR` is a literal NUL byte, so git classifies the file as binary and the PR diff just says "Binary files /dev/null and b/... differ" (same for every future diff and blame on it). perhaps: `const SEPARATOR = "\u0000"` (escape, not a raw byte) so it stays reviewable text.

PR #435, opened 2026-07-18
refactor: rethink openslop architecture 2026-07-18

umairnadeem:
## App understanding

OpenSlop turns a text prompt into a rendered video. An LLM streams OSML into a SlateJS canvas; each element (image, animated image, clip, character/narration TTS, music, sfx) is generated through a connector → gateway → provider pipeline into async `jobs` rows; the resolved elements are composed into a Remotion timeline previewed in-app and rendered on Lambda. Project state persists to a Supabase `projects` row and rehydrates on load.

## From-scratch architecture

Rebuilt today, the layering would be unchanged — connectors/gateways/providers, one route-handler factory, one Zustand store per project, one `VideoLayout` shared by preview and render. What has accumulated is not misplaced structure but **abstractions that no longer earn their keep**: factories that cover only part of their surface so the remainder is hand-rolled, nullable types for states that cannot occur, state drilled and mirrored when the source of truth is one hop away, and inheritance kept for naming after both halves of the parent contract were stubbed out. This PR removes four of those.

**1. Route-handler factory now covers every route (`lib/api/route-handler.ts`, `lib/api/parse.ts`)**

`jsonRouteHandler` only covered JSON-body routes under `withApiAccess`/`withSession`, so three routes re-implemented auth + parse + parse-result by hand, repeating the label string twice each. Generalized the factory along the two axes it already almost supported — auth tier and parse source — adding `createApiQueryRouteHandler` and `createPublicRouteHandler`. `app/api/v1/tts/voices`, `.../voices/preview`, and `app/api/validate-code` become declarations like the asset routes; `if (!parsed.ok) return parsed.response` now exists once instead of four times. Also folded the identical five-line tails of `parseBody`/`parseSearchParams` into one helper. `ARCHITECTURE.md` updated — its claim that every route uses the factory is now actually true.

**2. Dropped the phantom `VideoLayout | null` (6 files under `app/components/video`, `PlayFromHereButton.tsx`)**

`buildVideoLayout` returns `VideoLayout`, never null, but `useVideoLayout` and `VideoLayoutContext` declared it nullable and six call sites paid for the impossible case. Typed it non-null and removed the dead guards. The only reachable "empty" state is `layout.series.length === 0`, which `VideoPanel` and `ExportButton` already check separately.

**3. Scene ids read from the editor instead of drilled and mirrored (`PostPromptView.tsx`, `CanvasProviders.tsx`, `ViewModeContext.tsx`)**

`sceneIds` was computed in `PostPromptView`, passed through `CanvasProviders`, and used by exactly one callback in `ViewModeContext` — which, to stay identity-stable, mirrored the prop into a ref via an effect. That is derived state duplicated into a ref and reconciled by an effect, with a stale-read window: `collapseAll` fired in the same commit as a scene insert/delete collapsed the previous scene set. `ViewModeProvider` now takes the `editor` it can already reach and derives scene ids at call time. Deletes a `useMemo`, a prop, a `useRef`, and a `useEffect`, and reads current truth rather than last-rendered truth.

**4. Mock providers no longer inherit a contract they stub out (`lib/providers/mock-base.ts`, `llm/mock.ts`, `video/mock.ts`, `mock-utils.ts`, `elevenlabs.ts`, `audio-duration.ts`)**

`BaseProvider`'s whole contract is "run `_generate`, then `toFiles` + `AssetBundle.upload`". `MockProvider` and `MockLLM` each overrode *both* halves into no-ops plus a `blobConfig` whose `type` was never read — inheritance used as a naming convention, obscuring the base contract. Both are now standalone; every consumer is structurally typed and nothing does `instanceof BaseProvider`. `MockVideo` keeps `BaseVideoProvider` (it genuinely reuses the poll logic) and now shares the `mockDelay` helper instead of open-coding the same expression. Separately, `AudioCodec`'s `"pcm"` branch was unreachable — both callers pass the same mp3 constant — *and* wrong if taken, since `audioDurationSec` uses a constant-bitrate formula. Removed; the emitted format string is byte-identical.

## Validation

| Check | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass (no dead code) |
| `npm run build` | pass, all 23 routes present |
| `npm run test:run` | **878 passed / 101 files** |
| `npm run test:e2e` | **not run locally** — a dev server was already bound to this repo dir and killing it would disrupt the user's session. Smoke tests cover onboarding copy/routes, untouched by this diff; CI will run them. |

## Open PR overlap check

Considered all 19 open PRs (#410, #412–#422, #428–#434) and built the union of their changed files before choosing targets. This PR is non-overlapping: every file it touches is absent from that union.

Closest adjacencies, deliberately avoided:
- `lib/providers/elevenlabs.ts` (root, changed here) is distinct from `lib/providers/music/elevenlabs.ts` (#412). Removing `codec` from the shared constant required no edit to the latter.
- `app/components/video/BottomTransportBar.tsx` and `ExportButton.tsx` (#412) still hold `layout &&` guards after finding 2. They typecheck and behave identically — no lint rule flags them — so they were left for those PRs.
- `lib/api/route-handler.ts`'s `body` ctx key was kept unchanged so `app/api/render/*` (#412) needs no edit.

Skipped as blocked by open PRs: removing the non-serializable `sequenceByElementId` from the `VideoLayout` Lambda props contract (needs `lib/video/scene-builder.ts`, #412) and deleting `PreviewCacheContext` (needs `lib/components/Waveform.tsx`, #434).

## Other skipped items / risks

- **Project stores are never released** — `clearProjectStore` is exported but called only from tests, so revisiting a project reuses the in-memory store and `applyStoreSnapshot`'s `if (state.hydrated) return` silently discards the freshly-fetched DB row. Real bug, but changing store lifecycle can affect autosave; it deserves its own PR rather than riding along with a refactor.
- Rejected as churn: merging the small `lib/generation/*` modules (each is a coherent unit), collapsing the six 6-line gateway files (the second gateway will mirror that shape), and the five 3-line `[jobId]` routes (collapsing them would create live `llm`/`animated_image` job endpoints that do not exist today — not behavior-preserving).


ARCHITECTURE.md
```diff
@@ -34,7 +34,7 @@ The generation pipeline is split so providers and asset types can be swapped ind
 
 ## API routes
 
-Every `app/api/v1/<type>/route.ts` is built from a shared `createApiRouteHandler` factory that handles auth, parsing, model validation, and error mapping. Internal app routes use the session-tier variants from the same module (see Auth). Reusable Zod field schemas live in `lib/api/request-schema-fields.ts`. If a provider's API key isn't set, the route falls back to a mock provider.
+Every route is built from a factory in `lib/api/route-handler.ts` that handles auth, parsing, model validation, and error mapping — no route hand-rolls that plumbing. The factories vary along two axes, auth tier and parse source: `createApiRouteHandler` (api-access, JSON body), `createSessionRouteHandler` (session, JSON body), `createApiQueryRouteHandler` (api-access, search params), and `createPublicRouteHandler` (no auth, JSON body). Reusable Zod field schemas live in `lib/api/request-schema-fields.ts`. If a provider's API key isn't set, the route falls back to a mock provider.
```
umairnadeem: this overclaims a bit now: `app/api/upload/image/route.ts` still calls `withSession` directly and hand-rolls its formData parse, and `app/api/queues/asset-generate/route.ts` goes through `handleCallback` from @vercel/queue. maybe scope it to the v1 JSON/query routes so the doc isn't lying to the next reader.

umairnadeem (commented):
read CLAUDE.md / AGENTS.md / CONVENTIONS.md on the head branch. auth tiers on the three converted routes check out (withApiAccess/withApiAccess/withPublic all preserved), buildVideoLayout is total and useLayout throws without a provider so the non-null VideoLayout is safe, and nothing does instanceof BaseProvider or references AudioCodec. one thing on the doc update.

ARCHITECTURE.md
umairnadeem: Fixed in 6203df8. Scoped the claim to routes that take a JSON body or search params, and named the three that don't: `api/upload/image` (multipart under `withSession`), `api/queues/asset-generate` (`@vercel/queue` callback), and `auth/callback` (OAuth redirect).

PR #438, opened 2026-07-19
fix: stop the collapsed element label from catching a dead caret

### LINKED ISSUE #424: Element title/label text is caret-focusable but not editable (dead caret on Slate chrome)
## Problem

On the editor/project page you can place a text caret inside an element's title/label text, but typing does nothing. Slate element chrome (the element header/label — the "title" of each element card) is focusable and selectable when it shouldn't be: it's meant to be display-only.

Root cause is the Slate element-rendering pattern. Each element renders its *real* editable Slate `children` hidden, and shows a separate non-editable `Node.string(element)` snapshot in their place:

- `CompactElement.tsx:38` renders `<span className="hidden">{children}</span>` (the true editable text nodes), while `CompactElement.tsx:24-37` shows a static `Node.string` copy with `contentEditable={false}`.
- `ElementContainer.tsx:90-152` uses the same pattern for the header/label/divider/preview chrome (`contentEditable={false}`), with only the inner `children` at `:120-122` actually editable.

A caret can still land on the hidden `children`, so the user gets a blinking cursor, but the visible label is a non-editable snapshot that never reflects input. Classic Slate "focusable but not editable" mistake.

- "looks like project page has selectable/editable slatejs title elements since you can position your caret there but can't type. this shouldn't happen - looks like a bug"

## Relevant code

- `app/components/canvas/elements/CompactElement.tsx:24-40` — visible `Node.string` snapshot (`contentEditable={false}`) over hidden real `children` (`.hidden`)
- `app/components/canvas/elements/ElementContainer.tsx:90-152` — same snapshot-over-hidden-children pattern for element chrome; only `:120-122` children are truly editable
- `app/components/canvas/Canvas.tsx:81-92` — the single `<Editable>` (not `readOnly`); `renderElement` at `:50-58`
- `app/components/canvas/ProjectTitle.tsx:56-69` — the project `<h1>` title (separate, non-Slate; edit only via hover pencil) — flag if this is the surface the user means instead

## Suggested direction

Make the non-editable label/title chrome truly non-selectable so a caret can never land there — e.g. mark the snapshot region as a Slate void, or apply `user-select: none` / route selection away from the hidden children, so clicking the title text doesn't drop a dead caret. The real editable text (element body) stays editable; only the display chrome should stop catching carets.

## Acceptance criteria

- [ ] Clicking on an element's title/label snapshot does not place an editable caret in dead text.
- [ ] The element's actual editable body text still edits normally.
- [ ] No visible-vs-hidden text desync (typing never silently no-ops).

---

dchaudhari7177:
Closes #424

## Root cause

A collapsed element card renders inside the editable Slate tree. Clicking its read-only label therefore dropped a blinking caret into text that can never accept input — the card is `contentEditable={false}` and shows a `Node.string` snapshot, so there is nothing to type into.

## Fix

Wrap the rendered element in a plain, non-Slate `contentEditable={false}` div while its scene is collapsed. `SortableContent` already computes `collapsed`, so this threads straight through with no new state:

```tsx
// SortableItem.tsx
<div contentEditable={readOnly ? false : undefined}>
  {renderCanvasElement({ attributes, children, element })}
</div>
```

```tsx
// SortableContent.tsx
<SortableItem disabled={collapsed} readOnly={collapsed} ... />
```

Six lines across two files.

## Rebased and reworked

This branch has been rebased onto current master and squashed. Both earlier approaches are reverted, so the diff is now just the fix above:

- the `select-none` + `onMouseDown` `preventDefault` pass on `CompactElement`;
- the `withCollapsedVoids` `isVoid` plugin, along with `CollapsedVoidSync`, its test, the `setCollapsedElementIds` editor type, and the `useEditorSetup` wiring.

The `isVoid` route was dropped because `Transforms.insertText` bails on a void unless callers pass `{ voids: true }`, which would have silently gagged our own programmatic writes (script streaming via `useScriptSync`, refine `set` ops) until every call site in `lib/canvas/editorOps.ts` was updated. The DOM-level fix leaves those paths untouched.

## Verification

`lint`, `format:check`, `typecheck`, `knip`, `build`, and the full suite (994 tests) pass on current master.

### DIFF
--- a/app/components/canvas/dnd/SortableContent.tsx
+++ b/app/components/canvas/dnd/SortableContent.tsx
@@ -71,6 +71,7 @@ export function SortableContent({
 			insertOptions={INSERT_OPTIONS}
 			onInsert={handleInsert}
 			disabled={collapsed}
+			readOnly={collapsed}
 			attributes={attributes}
 			element={element}
 		>
--- a/app/components/canvas/dnd/SortableItem.tsx
+++ b/app/components/canvas/dnd/SortableItem.tsx
@@ -15,6 +15,7 @@ interface SortableItemProps {
 	insertOptions?: InsertOption<CanvasElementType>[];
 	onInsert?: (type: CanvasElementType) => void;
 	disabled?: boolean;
+	readOnly?: boolean;
 	attributes: RenderElementProps["attributes"];
 	element: CanvasElement;
 	children: React.ReactNode;
@@ -28,6 +29,7 @@ export function SortableItem({
 	insertOptions,
 	onInsert,
 	disabled,
+	readOnly,
 	attributes,
 	element,
 	children,
@@ -77,7 +79,9 @@ export function SortableItem({
 							/>
 						</div>
 					)}
-					<div>{renderCanvasElement({ attributes, children, element })}</div>
+					<div contentEditable={readOnly ? false : undefined}>
+						{renderCanvasElement({ attributes, children, element })}
+					</div>
 				</div>
 			</div>
 		</div>


app/components/canvas/elements/CompactElement.tsx
```diff
@@ -17,8 +17,13 @@ export function CompactElement({
...
 	return (
 		<div
-			className="inline-flex items-center gap-1 rounded-md py-0.5 animate-fadeInUp"
+			className="inline-flex items-center gap-1 rounded-md py-0.5 animate-fadeInUp select-none"
 			contentEditable={false}
+			// The visible label is a read-only Node.string snapshot over the real
+			// (hidden) Slate children. Without this, clicking the label lets the
+			// browser drop a caret into the hidden text — it blinks, but typing is
+			// dead because the whole card is contentEditable={false}.
+			onMouseDown={(event) => event.preventDefault()}
```
umairnadeem: `select-none` on its own already closes the caret path, so the `preventDefault` is redundant and it costs you focus: click a collapsed card while the editor isn't focused and `activeElement` stays on `<body>` with no selection, instead of focusing the editor.

i checked the three variants in chromium against this exact DOM shape (contenteditable root > `contentEditable={false}` div > visible label + `display:none` children), clicking the label:

| variant | selection after click |
|---|---|
| `contentEditable={false}` only | moves into the **visible label** text node |
| `+ select-none` | doesn't move |
| `+ select-none + preventDefault` | doesn't move (identical) |

the comment's premise is off too, the caret never reaches the hidden children, chrome won't put one in a `display:none` subtree. it lands in the visible `Node.string` snapshot.

```suggestion
```

app/components/canvas/elements/CompactElement.tsx
```diff
@@ -17,8 +17,13 @@ export function CompactElement({
 
 	return (
 		<div
-			className="inline-flex items-center gap-1 rounded-md py-0.5 animate-fadeInUp"
+			className="inline-flex items-center gap-1 rounded-md py-0.5 animate-fadeInUp select-none"
```
umairnadeem: this doesn't close the bug, you can still type into a collapsed card. `contentEditable` only controls where the browser paints a caret, but slate resolves a DOM point to the nearest text leaf and the only one this element has is the hidden one, so arrow-key nav / ctrl+a / paste still write to it and the label re-renders from `Node.string`. keep `select-none` (it's load-bearing for the caret), it just isn't the whole fix.

the fix has to be model level, and slate already ships it. CONVENTIONS.md "Structure and reuse" says *"Don't reinvent. Reach for a popular, well-tested library before hand-rolling, and exhaust what our stack already does"*, and there's no `isVoid` override anywhere in the repo today:

```ts
// app/components/canvas/plugins/withCollapsedVoids.ts
export const withCollapsedVoids = (editor: CanvasEditor): CanvasEditor => {
	const { isVoid } = editor;
	let collapsedElementIds: ReadonlySet<string> = new Set();

	editor.setCollapsedElementIds = (ids) => {
		collapsedElementIds = ids;
	};

	editor.isVoid = (element) =>
		collapsedElementIds.has(element.id) || isVoid(element);

	return editor;
};
```

add it to the `flow(...)` in `useEditorSetup`, then push the collapsed element ids in from a small component inside `<Slate>`. do that during render, not in an effect, slate-react reads `isVoid` while rendering the elements below it so an effect leaves the first pass after a collapse still editable.

i ran the two things that could bite: normalization does **not** strip void children's text (no data loss), and `insertText` into a void is a silent model-level no-op. `isElementReadOnly` measures identical if you prefer the name, though slate-react 0.123 never reads it, only slate core does. `isSelectable` looks tempting but fails, it doesn't block writes and empties the node on backspace.

one consequence worth knowing before you take this: backspace at the start of the *following* element deletes the whole collapsed card rather than merging text. inherent to making it read-only, not to `isVoid` specifically.

umairnadeem (commented):
read CONVENTIONS.md + CLAUDE.md on the head branch. the caret's gone but the card is still editable, so afaict #424 isn't actually closed. details inline.

dchaudhari7177:
Thanks for the thorough steer — you're right on all counts, my previous version only killed the caret paint. Redone at the model level exactly as you laid out:

- Added `app/components/canvas/plugins/withCollapsedVoids.ts` — overrides `editor.isVoid` to return true for the collapsed element ids (chaining the prior `isVoid`), and exposes `setCollapsedElementIds`. Added it to the `flow(...)` in `useEditorSetup`.
- The ids are pushed in **during render** by a small `<CollapsedVoidSync>` inside `<Slate>` (a content element is collapsed when its parent scene is), so the first pass after a collapse is already void — not via an effect.
- Dropped the redundant `onMouseDown` `preventDefault` from `CompactElement`; kept `select-none` (load-bearing for the caret).

Unit test (`withCollapsedVoids.test.ts`) asserts `isVoid` tracks the collapsed set and, functionally, that `Editor.insertText` into a collapsed element is a silent no-op while it still writes when expanded. `tsc`, `eslint`, and all 53 canvas tests pass.

Noted the backspace-at-start-of-next-element consequence you flagged — that's acceptable read-only behavior. (The failing `ci` check is the fork-secrets Supabase prerender on `/login`, unrelated to this change.)

app/components/canvas/plugins/withCollapsedVoids.ts
```diff
@@ -0,0 +1,29 @@
...
+export const withCollapsedVoids = (editor: CanvasEditor): CanvasEditor => {
+	const { isVoid } = editor;
+	let collapsedElementIds: ReadonlySet<string> = new Set();
+
+	editor.setCollapsedElementIds = (ids) => {
+		collapsedElementIds = ids;
+	};
+
+	editor.isVoid = (element) =>
+		collapsedElementIds.has(element.id) || isVoid(element);
```
umairnadeem: this gags our own writes too, not just typing: `Editor.string` returns `""` for a void and `Transforms.insertText` bails on one, so `updateNodeText` silently drops text into any collapsed element (script streaming via `useScriptSync`, and refine `set` ops). collapse a scene, run refine over it, the edit vanishes and nothing re-applies it on expand.

verified against slate 0.123 with this exact isVoid shape (`setNodes`/`removeNodes`/`moveNodes` on a void path are all fine, it's only the text paths). passing `{ voids: true }` to the `Editor.string` + `insertText` calls in `lib/canvas/editorOps.ts` keeps programmatic updates working while user input still no-ops.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the head branch. one thing on the new `isVoid` plugin: making collapsed elements void also silences the app's own programmatic text writes, not just user typing.

app/components/canvas/elements/CompactElement.tsx
```diff
@@ -17,7 +17,10 @@ export function CompactElement({
 
 	return (
 		<div
-			className="inline-flex items-center gap-1 rounded-md py-0.5 animate-fadeInUp"
+			// `select-none` is load-bearing: it stops the browser dropping a caret
+			// into the visible label. The element being read-only is enforced at
+			// the model level by `withCollapsedVoids` (isVoid), not here.
+			className="inline-flex items-center gap-1 rounded-md py-0.5 animate-fadeInUp select-none"
```
umairnadeem: can we revert CompactElement entirely, both the `select-none` and the comment? since we're fixing this at the model level now, the class here is belt-and-suspenders and the comment is just narrating what the code does, which cuts against CONVENTIONS.md "Comments are a last resort... The code shows what it does. Don't narrate the code". keeping this file untouched shrinks the diff down to the actual fix.

app/components/canvas/CollapsedVoidSync.tsx
```diff
@@ -0,0 +1,36 @@
...
+				scene.type === SCENE_TYPE &&
+				isCollapsed(scene.id)
+			) {
+				for (const child of scene.children) {
+					if (Element.isElement(child)) collapsedIds.add(child.id);
+				}
+			}
+		}
+	}
+	editor.setCollapsedElementIds?.(collapsedIds);
```
umairnadeem: the bigger thing i want to noodle on before we build more on this: this mirrors the collapsed set into a second home (ViewModeContext already owns it via `isCollapsed`) and re-pushes it every render to keep the two in sync. that's two sources of truth on the same knowledge, exactly the desync footgun CONVENTIONS.md "DRY is about knowledge... Each piece of knowledge gets one home" warns about, and it's why we need this whole sync component + a mutable Set in the plugin closure.

is there a simpler shape where `isVoid` reads the collapsed state straight from ViewMode (one home) instead of us maintaining a mirrored Set? lets discuss the approach before iterating further.

app/components/canvas/CollapsedVoidSync.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useViewMode } from "./ViewModeContext";
+
+/**
+ * Pushes the set of collapsed element ids into the editor DURING render, so
+ * `withCollapsedVoids` sees them before slate-react renders the elements below.
+ * An effect would run too late — the first pass after a collapse would still be
+ * editable. A content element is collapsed when its parent scene is. Renders
+ * nothing.
+ */
+export function CollapsedVoidSync() {
```
umairnadeem: this whole file can go, and with it `withCollapsedVoids`, its test, and the `setCollapsedElementIds` type. reversing my earlier `isVoid` steer, found something much simpler.

the bug is only that the collapsed element's DOM sits inside the editable tree. wrap the rendered element in a plain (non-slate) `contentEditable={false}` div and you get a real editability boundary slate's DOM->model resolution won't cross, so no caret and no keyboard/paste writes land in it. bonus: it's not `isVoid`, so programmatic writes (script streaming / refine) still work and the `{voids:true}` gag from the withCollapsedVoids thread just doesn't exist.

`SortableContent` already computes `collapsed`, so it's a thread-through with no new state:

```tsx
// SortableItem.tsx
interface SortableItemProps {
  // ...
  readOnly?: boolean;
}
// ...
<div contentEditable={readOnly ? false : undefined}>
  {renderCanvasElement({ attributes, children, element })}
</div>
```
```tsx
// SortableContent.tsx  (reuse the collapsed it already has)
<SortableItem
  // ...
  disabled={collapsed}
  readOnly={collapsed}
/>
```

that plus reverting CompactElement is the entire fix. i ran it locally: typecheck + lint + canvas suite (51) all green. worth a browser sanity check that arrow-into-then-type, ctrl+a, and paste on a collapsed card all no-op, and dnd drag still works.

app/components/canvas/plugins/withCollapsedVoids.ts
```diff
@@ -0,0 +1,29 @@
...
+ * a DOM point to the element's (hidden) text leaf, so arrow-key nav / ctrl+a /
+ * paste keep writing to it. Marking the element `isVoid` is the model-level fix:
+ * Slate treats it as an opaque, non-editable node, so `insertText` becomes a
+ * silent no-op and the label can't be typed into. Void children keep their text
+ * (normalization does not strip it), so collapsing is lossless.
+ *
+ * The collapsed set is pushed in during render by the `<CollapsedVoidSync>`
+ * component, because slate-react reads `isVoid` while rendering the elements.
+ */
+export const withCollapsedVoids = (editor: CanvasEditor): CanvasEditor => {
```
umairnadeem: delete this plugin, no `isVoid` override needed with the DOM-boundary fix. see the note on CollapsedVoidSync.tsx line 15.

app/components/canvas/hooks/useEditorSetup.ts
```diff
@@ -22,6 +23,7 @@ export function useEditorSetup() {
 			withScenes,
 			withFlatPaste,
 			withNodeId,
+			withCollapsedVoids,
```
umairnadeem: drop `withCollapsedVoids` from the flow here and the import on line 11, back to the original list.

lib/canvas/types.ts
```diff
@@ -80,7 +80,17 @@ export const DURATION_OPTIONS = Array.from({ length: 12 }, (_, i) =>
...
+export type CanvasEditor = BaseEditor &
+	ReactEditor & {
+		id?: string;
+		/**
+		 * Sets which element ids are currently collapsed. `withCollapsedVoids`
+		 * reads this set from `isVoid` so a collapsed card is read-only at the
+		 * model level. Pushed in during render (see `CollapsedVoidSync`). Optional
+		 * because it is only present once `withCollapsedVoids` has run in the flow.
+		 */
+		setCollapsedElementIds?: (ids: ReadonlySet<string>) => void;
```
umairnadeem: `setCollapsedElementIds` goes away with the plugin, `CanvasEditor` back to `BaseEditor & ReactEditor & { id?: string }`.

app/components/canvas/Canvas.tsx
```diff
@@ -79,6 +80,7 @@ export default function Canvas({
 				onDragCancel={handleDragCancel}
 			>
 				<Slate editor={editor} initialValue={value} onChange={setValue}>
+					<CollapsedVoidSync />
```
umairnadeem: remove this render + the import on line 19, no sync component needed anymore.

umairnadeem (commented):
update: we're pivoting off the model-level `isVoid` route (my earlier steer, sorry) to a much simpler DOM fix. concrete guidance inline, the short version is: delete the plugin + sync component + test + type field, revert CompactElement, and add a `contentEditable={false}` wrapper gated on the `collapsed` that `SortableContent` already computes.

app/components/canvas/elements/CompactElement.tsx
umairnadeem: update: reversing this. we found a simpler fix that needs no model-level change, a plain contentEditable={false} DOM wrapper (see the new note on CollapsedVoidSync.tsx line 15). that also dodges the write-gag from the withCollapsedVoids thread. keep select-none out, revert CompactElement entirely.

umairnadeem (changes requested):

umairnadeem (commented):
lgtm, this is exactly the shape we landed on. thanks for reworking it.

PR #440, opened 2026-07-20
refactor: nightly maintainability 2026-07-20

umairnadeem:
## Summary

Gives the persisted project store a real parse boundary, and stops the project loader from swallowing DB errors.

The `store` JSON column arrived from Supabase as `any`, was given a shape at the prop boundary by an unchecked cast (`initialStore: Partial<ProjectStoreSnapshot>`), and was then lodash-merged straight into Zustand state. Nothing checked it. Per CONVENTIONS ("Parse, don't validate. Check external input once at the boundary (zod) into typed data, then trust the types inward"), it now gets parsed once.

## Changes

**Parse boundary (`lib/project/storeSnapshot.ts`)**
- `ProjectStoreSnapshotSchema` + `parseStoreSnapshot(raw: unknown)`. Absent and partial rows normalize to a complete snapshot via defaults; a structurally wrong row (`title: 42`, `referenceImages: "a.png"`) throws instead of corrupting state.
- `MetadataSchema` is annotated `z.ZodType<Metadata, unknown>`, so the schema and the hand-written `Metadata` type can't drift apart without a compile error.
- `applyStoreSnapshot` now takes an already-parsed `ProjectStoreSnapshot`. The `Partial<...> | null | undefined` parameter and its two internal presence checks are gone — the edge case is defined out of existence at the boundary rather than re-handled downstream.

**Fail loudly (`app/projects/[id]/page.tsx`)**
- `if (error || !project) notFound()` → `if (error) throw error; if (!project) notFound();`. A transient Supabase failure was rendering as "project not found". Expected failure (not-found) stays typed; the unexpected now throws.

**Single source of truth**
- `MetadataCharacter` derives from a new `MetadataCharacterSchema` instead of being a hand-written intersection.
- `Mode` derives from `MODES`; `AspectRatio` derives from a new `ASPECT_RATIOS` const, matching how `TransitionType`/`TRANSITION_TYPES` already work.

**Dedupe (`lib/project/api.ts`)**
- `saveProject` re-listed all five fields of its own `SaveProjectInput` in an identity mapping. Collapsed to `.update(input)`.

**Tests** — 4 new cases covering defaults, round-trip fidelity, and the throw path.

## Behavior

Unchanged for every valid row. `applyStoreSnapshot` now always calls `updateMetadata`/`setReferenceImages`, but with normalized values identical to the store's initial state, so the merge is a no-op where it previously short-circuited. Verified: typecheck, lint, format, knip, build, 893/893 tests.

## Open PR overlap check

Considered all 16 open PRs (#439, #438, #435, #434, #433, #432, #431, #430, #429, #422, #421, #420, #419, #418, #417, #416) and built the touched-file set from `gh pr diff --name-only`. None of the 7 files here appear in any of them.

Deliberately skipped as overlapping, for the record:
- Consolidating the duplicated `videoSettings` default resolution — 4 of its consumers are owned by #435, #432, and #439.
- Extracting `ElementSnapshot`/`GenerationJob` out of `lib/generation/queue.ts` to break a `lib/project ↔ lib/generation` cycle — `queue.ts` is owned by #433 and #429.
- Reconciling the diverged `loops`/`volume` attribute ranges between `lib/video/elementAttributes.ts` and `lib/connectors/attributes/common.ts` — the latter is owned by #416.
- Resetting `hydrated` in `store.reset()` — interacts with `lib/templates/applyTemplate.ts`, owned by #432.

### DIFF
--- a/app/projects/[id]/ProjectEditor.tsx
+++ b/app/projects/[id]/ProjectEditor.tsx
@@ -9,7 +9,7 @@ import { TooltipProvider } from "@/components/ui/tooltip";
 import { getProjectStore } from "@/lib/project/store";
 import {
 	applyStoreSnapshot,
-	type ProjectStoreSnapshot,
+	parseStoreSnapshot,
 } from "@/lib/project/storeSnapshot";
 import type { ElementSnapshot } from "@/lib/generation/queue";
 import { GenerationQueueProvider } from "@/lib/generation/GenerationQueueProvider";
@@ -24,12 +24,17 @@ export default function ProjectEditor({
 }: {
 	projectId: string;
 	initialScript: string;
-	initialStore: Partial<ProjectStoreSnapshot>;
+	initialStore: unknown;
 	initialGeneration: Record<string, ElementSnapshot>;
 	user: User;
 }): ReactNode {
 	// Hydrate the store once, before children render and without re-running each render.
-	useState(() => applyStoreSnapshot(getProjectStore(projectId), initialStore));
+	useState(() =>
+		applyStoreSnapshot(
+			getProjectStore(projectId),
+			parseStoreSnapshot(initialStore),
+		),
+	);
 
 	return (
 		<TooltipProvider>
--- a/app/projects/[id]/page.tsx
+++ b/app/projects/[id]/page.tsx
@@ -20,7 +20,8 @@ export default async function ProjectPage({
 		.eq("id", id)
 		.maybeSingle();
 
-	if (error || !project) notFound();
+	if (error) throw error;
+	if (!project) notFound();
 
 	return (
 		<ProjectEditor
--- a/lib/project/storeSnapshot.ts
+++ b/lib/project/storeSnapshot.ts
@@ -1,28 +1,6 @@
 import { z } from "zod";
-import { ASPECT_RATIOS } from "@/lib/video/aspectRatio";
-import { TRANSITION_TYPES } from "@/lib/video/transitions";
 import type { ProjectStore } from "./store";
-import {
-	MetadataCharacterSchema,
-	MetadataVoiceSchema,
-	MODES,
-	type Metadata,
-} from "./types";
-
-const VideoSettingsSchema = z.object({
-	transitionType: z.enum(TRANSITION_TYPES).optional(),
-	aspectRatio: z.enum(ASPECT_RATIOS).optional(),
-});
-
-const MetadataSchema: z.ZodType<Metadata, unknown> = z.object({
-	title: z.string().default(""),
-	style: z.string().default(""),
-	narration: MetadataVoiceSchema.default({}),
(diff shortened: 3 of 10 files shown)


lib/project/storeSnapshot.ts
```diff
@@ -1,10 +1,35 @@
...
-export type ProjectStoreSnapshot = {
-	metadata: Metadata;
-	referenceImages: string[];
-};
+const VideoSettingsSchema = z.object({
+	transitionType: z.enum(TRANSITION_TYPES).optional(),
+	aspectRatio: z.enum(ASPECT_RATIOS).optional(),
+});
+
+const MetadataSchema: z.ZodType<Metadata, unknown> = z.object({
```
umairnadeem: `z.ZodType<Metadata, unknown>` only catches drift one way afaict: add an optional field to `Metadata` and this schema still typechecks, but parse strips the key and the next autosave writes the stripped snapshot back, so the value is gone. deriving `Metadata` from the schema (`z.infer`, like you just did for `MetadataCharacter`) actually closes it.

umairnadeem (commented):
read CONVENTIONS.md + CLAUDE.md off the head branch. one thing on the new parse boundary:

app/projects/[id]/page.tsx
```diff
@@ -20,7 +20,8 @@ export default async function ProjectPage({
 		.eq("id", id)
 		.maybeSingle();
 
-	if (error || !project) notFound();
+	if (error) throw error;
```
Comment (umairnadeem) [inline app/projects/[id]/page.tsx:23]: this turns a 404 into a 500 for malformed ids: `projects.id` is `uuid`, so `/projects/foo` makes postgres reject the filter with 22P02 and that lands in `error`, which the old `error || !project` swallowed into `notFound()`. imo keep the bad-id case on `notFound()` and only throw the genuinely unexpected, otherwise every fat-fingered/stale url is a 500 now.

app/projects/[id]/ProjectEditor.tsx
```diff
@@ -24,12 +24,17 @@ export default function ProjectEditor({
 }: {
 	projectId: string;
 	initialScript: string;
-	initialStore: Partial<ProjectStoreSnapshot>;
+	initialStore: unknown;
```
Comment (umairnadeem) [inline app/projects/[id]/ProjectEditor.tsx:27]: widening this prop to `unknown` puts the parse one layer past the actual boundary. `CONVENTIONS.md` "Backend" says *"Parse, don't validate. Check external input once at the boundary (zod) into typed data, then trust the types inward"* and the row enters the app in `page.tsx`, so parsing there keeps this prop typed as `ProjectStoreSnapshot` and a structurally bad row fails on the server instead of throwing mid client render.

umairnadeem (commented):
read `CLAUDE.md` + `CONVENTIONS.md` on the head branch. two things on the boundary/fail-loudly rework, rest of the diff (schema derivation, `ASPECT_RATIOS`, the remotion split out of `transitions.ts`) lgtm.

PR #441, opened 2026-07-20
perf: nightly performance optimizations 2026-07-20

umairnadeem:
Runtime performance pass on the video preview playback path. No behavior changes.

## What was optimized

**`usePlayerValue` re-ran its reader several times per frame** (`app/components/video/usePlayerState.ts`)

`getSnapshot` was an uncached closure, so React re-ran the reader on every render of every subscriber and again after commit for the tear check. That is fine for `getCurrentFrame()`, but `useActiveSegmentIndex` reads through `findSegmentIndexAt`, a linear scan over every scene. Two components subscribe with that reader, so a 40-scene project was doing several full scans per frame to produce a number that changes once per scene.

The value is now held until the player emits or the caller passes a new reader. Keying on reader identity matters: a reader that closes over changing data (segments do change when the script is edited while paused) must still observe it, so this keeps the existing semantics rather than caching purely on events. The four closure-free readers moved to module scope, so for `usePlayerFrame`/`Playing`/`Volume`/`Muted` the cache also survives unrelated re-renders of the subscriber.

**Every scene boundary re-rendered the whole `<Player>`** (`app/components/video/VideoPreview.tsx`)

`VideoPreview` subscribed to the active segment index and owned the play/pause flash state, both above the `<Player>` it renders. `@remotion/player` is not `React.memo` and the `dynamic()` wrapper is not either, so each change re-rendered `PlayerFn` → `PlayerUI` → every `Sequence`. The deep subtree mostly bails out because `VideoComposition` memoizes its node arrays, but the work still lands on exactly the frame where a transition is animating and the next asset is decoding — the frame with the least headroom.

Active-scene tracking moved into a null-rendering sibling and the flash state into the overlay that owns it. `VideoPreview` now re-renders only when the player instance, fullscreen state, or layout changes.

**`Captions` reconciled its subtree every frame to render the same word** (`remotion/components/Captions.tsx`)

The binary-search lookup was already right, but the JSX and a fresh inline style object were rebuilt on every frame, so React diffed a 9-layer `textShadow` on full-width uppercase text at composition resolution 30-60 times a second to discover nothing had changed. The word under the playhead changes a few times a second. The presentational half is now memoized and bails between word boundaries. One `Captions` instance mounts per timestamped audio element, so this multiplies with concurrent narration.

## What was skipped

- **`sceneItems` identity churn in `useDragAndDrop.ts`.** The largest finding overall: `editor.children` is a new array per Slate op, so dnd-kit's `SortableContext` rebuilds its context every keystroke, re-rendering every scene and doing O(scenes²) work inside `useSortable`. Skipped as overlapping — PRs #433 and #431 both touch that file.
- **`findSegmentIndexAt` linear scan → binary search.** `useSceneSegments.ts` is covered by #435. The caching above removes most of the repeat cost without touching it.
- **`SegmentedSeekBar`/`ScrubBar` per-frame style-object churn.** Likely the single heaviest cost in the preview UI, but covered by #417 and #430.
- **`motionTransform` `toFixed` allocations.** Single-digit microseconds; below the bar.
- **`ViewModeContext` slicing, `useSceneIndex` O(scenes²), `ElementSettings` schema re-resolution.** Real, but either overlapping (#435) or only material once the dnd-kit churn above is fixed. Worth a follow-up night.

## Open PR overlap check

Considered open PRs #440, #439, #438, #435, #434, #433, #432, #431, #430, #429, #422, #421, #420, #419, #418, #417, #416. The three files here (`usePlayerState.ts`, `VideoPreview.tsx`, `Captions.tsx`) appear in none of them. The video-adjacent PRs (#435, #430, #417) touch `VideoPanel`, `useSceneSegments`, `useVideoLayout`, `VideoLayoutContext`, `useAssetPrefetch`, `PlayerControlContext`, `SegmentedSeekBar`, `ScrubBar`, `MotionLayer`, `startPlaybackAt`, `silenceMedia` — all disjoint from this diff.

## Validation

`lint`, `format:check`, `typecheck`, `knip`, `build`, and `test:run` (890 tests) all pass. Playwright smoke tests could not run locally — port 3000 was occupied by an unrelated process — so they are left to CI.

### DIFF
--- a/remotion/components/Captions.tsx
+++ b/remotion/components/Captions.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from "react";
+import { memo, useMemo } from "react";
 import sortedLastIndex from "lodash/sortedLastIndex";
 import { AbsoluteFill, useCurrentFrame, useVideoConfig } from "remotion";
 import type { TextTimestamp } from "@/lib/connectors/types";
@@ -72,6 +72,30 @@ function buildCaptionData(
 	return { startTimes, visibleByWordIndex };
 }
 
+// The word under the playhead changes a few times a second, but the parent
+// re-renders on every frame. Memoized so the 9-layer text shadow is only
+// reconciled at word boundaries.
+const CaptionLine = memo(function CaptionLine({
+	line,
+	fontSizePx,
+}: {
+	line: string;
+	fontSizePx: number;
+}) {
+	return (
+		<AbsoluteFill style={container}>
+			<div
+				style={{
+					width: `${MAX_LINE_WIDTH_RATIO * 100}%`,
+					fontSize: fontSizePx,
+				}}
+			>
+				<span style={text}>{line}</span>
+			</div>
+		</AbsoluteFill>
+	);
+});
+
 export function Captions({ timestamps }: { timestamps: TextTimestamp[] }) {
 	const frame = useCurrentFrame();
 	const { fps, width, height } = useVideoConfig();
@@ -96,16 +120,5 @@ export function Captions({ timestamps }: { timestamps: TextTimestamp[] }) {
 
 	const visible = visibleByWordIndex[wordIndex];
 
-	return (
-		<AbsoluteFill style={container}>
-			<div
-				style={{
-					width: `${MAX_LINE_WIDTH_RATIO * 100}%`,
-					fontSize: fontSizePx,
-				}}
-			>
-				<span style={text}>{visible}</span>
-			</div>
-		</AbsoluteFill>
-	);
+	return <CaptionLine line={visible} fontSizePx={fontSizePx} />;
 }
--- a/app/components/video/VideoPreview.tsx
+++ b/app/components/video/VideoPreview.tsx
@@ -53,6 +53,43 @@ const RemotionPlayer = dynamic(
 	{ ssr: false },
 );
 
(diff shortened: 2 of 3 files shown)


remotion/components/Captions.tsx
```diff
@@ -72,6 +72,30 @@ function buildCaptionData(
 	return { startTimes, visibleByWordIndex };
 }
 
+// The word under the playhead changes a few times a second, but the parent
+// re-renders on every frame. Memoized so the 9-layer text shadow is only
+// reconciled at word boundaries.
+const CaptionLine = memo(function CaptionLine({
```
umairnadeem, line 78: the 9-layer textShadow isn't actually re-diffed per frame: `text` is a module const, and react's `updateProperties` bails on `nextProp === lastProp` before it ever walks the style object (only the div's 2-key inline `{width, fontSize}` gets touched, and both values are stable). so what this buys is skipping 3 host fibers between word boundaries, which imo isn't worth a component split + memo layer in a remotion frame component.

umairnadeem (commented):
read CLAUDE.md + CONVENTIONS.md on the head branch. one thing on the Captions memo, the premise it's justified by doesn't hold.

umairnadeem:
overengineered

PR #444, opened 2026-07-21
fix: nightly bug fixes 2026-07-21

umairnadeem:
Three verified correctness bugs. Each fix ships with a regression test that fails without it (confirmed by stashing the fix and re-running).

## 1. Refine inserts ignored `position` when stacking

`lib/script/refine/applyOps.ts`

The anchor cursor that makes consecutive inserts stack in stream order was keyed by anchor id alone, but consumed for both directions. An `after` insert therefore moved where a later `before` insert on the same anchor landed, and vice versa.

Reproduced against real Slate, doc `scene[ n1("first") ]`, both ops anchored on `n1`:

| ops | before fix | expected |
|---|---|---|
| `before A`, `before B` | `["B", "A", "first"]` | `["A", "B", "first"]` |
| `before A`, `after B` | `["A", "B", "first"]` — **B ends up before n1** | `["A", "first", "B"]` |
| `after A`, `before B` | `["first", "B", "A"]` — **B ends up after n1** | `["B", "first", "A"]` |

Rows 2 and 3 are the bad ones: the op says "after n1" and the element lands on the other side of `n1`.

Reachable because the refine system prompt documents `position` as `before|after` and explicitly tells the model to "emit all operations for a given location consecutively" — a sound before and a narration after the same anchor is the encouraged output shape.

Fix: key the cursor per side of the anchor, so the two directions stop interfering.

## 2. Undoing a merge minted a new element id, orphaning its generated asset

`app/components/canvas/plugins/withNodeId.ts`

Slate replays the inverse of a `merge_node` as a `split_node` carrying the merged-away node's original properties, including its id. The plugin overwrote `properties.id` unconditionally, so undo restored the element under a brand-new id.

Generation state is keyed strictly by element id, so: generate an image on an element, put the caret at its start, press Backspace, then Ctrl+Z. The element returns with its queue snapshot orphaned and the generated preview gone — and autosave then persists `generation` under the dead id, so the loss survives reload. Cycling undo/redo mints another id each pass.

Fix: the real invariant is id uniqueness, not "splits always mint". Mint only when the incoming id is still taken — a forward split collides (the first half keeps it) and gets a fresh id as before; an undone merge does not collide and keeps its identity.

## 3. Every non-streaming LLM call failed with a 500

`lib/providers/llm/anthropic.ts`

`@anthropic-ai/sdk` derives a timeout from `max_tokens` for non-streaming calls when the client sets none, and throws locally above ~21,333 tokens. Our default is `65536`, so the request never reached the network.

Verified against the pinned SDK in `node_modules`:

```
$ node -e "new Anthropic({apiKey:'sk-test'}).messages.create({model:'claude-opus-4-8',max_tokens:65536,...})"
Error: Streaming is required for operations that may take longer than 10 minutes.
    at Anthropic.calculateNonstreamingTimeout (client.js:434:19)
```

So `POST /api/v1/llm` without `stream` always 500'd. The existing test replaced the whole SDK with a stub, so the guard never ran. Streaming was unaffected.

Fix: set an explicit client timeout, opting in to the SDK's own 10-minute ceiling. The new test replays the provider's real client options and real request body through the **real** SDK with a stubbed `fetch`, so raising `max_tokens` or dropping the timeout fails here rather than in production.

## Checks

`npm test` (896 passed, 104 files), `npm run build`, `tsc --noEmit`, `eslint`, `prettier --check` — all clean.

## Open PR overlap check

Considered open PRs #443, #442, #441, #440, #439, #438, #435, #432, #429. Built their combined file list via `gh pr diff --name-only`; intersection with this PR's files is empty.

Deliberately skipped several other candidates that would have overlapped or repeated rejected work: the ZWSP serialization and autosave-on-open findings land in files already covered by closed PRs #418/#431, and the Runware video handler is covered by open #439.

### DIFF
--- a/app/components/canvas/__tests__/withNodeId.test.ts
+++ b/app/components/canvas/__tests__/withNodeId.test.ts
@@ -1,7 +1,9 @@
 import { describe, expect, it } from "vitest";
 import { createEditor, Transforms, Element, Editor } from "slate";
 import { withReact } from "slate-react";
+import { HistoryEditor, withHistory } from "slate-history";
 import { withNodeId } from "../plugins/withNodeId";
+import type { CanvasEditor } from "@/lib/canvas/types";
 
 function makeEditor() {
 	const editor = withNodeId(withReact(createEditor()));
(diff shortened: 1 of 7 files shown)


umairnadeem:
overengineered

PR #453, opened 2026-07-22
add dev-only user impersonation to the account menu

umairnadeem:
## Summary
- Adds `GET /api/dev/impersonate?email=...`, which mints a session for an existing user via `auth.admin.generateLink` + `verifyOtp`, no email round trip. Useful for reproducing a reported bug against the real account.
- Adds an "Impersonate user" item to the account menu (dev builds only) opening a small dialog with an email field.
- Adds a `caution` variant to `Badge`, filling the gap next to `success` / `destructive` against the existing `--caution` token.
- Unrelated `AGENTS.md` tweak folded in: review skills run each round of changes, and `CONVENTIONS.md` gets checked explicitly.

## Gating
The route returns 404 unless `NODE_ENV !== "production"`, and `createServiceClient()` throws without `SUPABASE_SECRET_KEY`. The menu item and dialog compile out of production bundles.

Worth being explicit, since this repo is public: there is **no per-caller auth and no audit log**. On a dev server that is reachable by anyone (`-H 0.0.0.0`, a tunnel, a public Codespace port), this endpoint is an unauthenticated account-takeover. It is safe on localhost and inert in production, and it is not currently safe to make functional in production.

Impersonated sessions are indistinguishable from the user's own, so anything done while impersonating is written to prod under their id.

Not included, discussed and deferred: an `IMPERSONATION_ENABLED` opt-in env flag as a fail-closed second gate, an `app_metadata.admin` check, an audit table, and a "viewing as X" banner with exit.

## Test plan
- [ ] `npm run dev`, open the account menu, confirm the "Impersonate user" item with the Dev badge appears
- [ ] Enter a known user's email, submit, confirm you land on `/` signed in as them
- [ ] Enter a malformed email, confirm the inline zod error rather than a 500
- [ ] `npm run build && npm start`, confirm `/api/dev/impersonate?email=...` returns 404 and the menu item is absent

CI covers lint, format, typecheck, knip, build, 892 unit tests, and the Playwright smoke suite.

### DIFF
--- a/app/api/dev/impersonate/route.ts
+++ b/app/api/dev/impersonate/route.ts
@@ -1,0 +1,43 @@
+import { NextRequest, NextResponse } from "next/server";
+import { z } from "zod";
+import { parseSearchParams } from "@/lib/api/parse";
+import { withPublic } from "@/lib/api/with-auth";
+import { createClient } from "@/lib/supabase/server";
+import { createServiceClient } from "@/lib/supabase/service";
+
+const ImpersonateRequest = z.object({
+	email: z.email({ error: "A valid email query param is required" }),
+});
+
+// Signs the caller in as an existing user with no email delivery, against
+// whichever Supabase project the environment points at — production included.
+export async function GET(request: NextRequest) {
+	if (process.env.NODE_ENV === "production")
+		return new NextResponse(null, { status: 404 });
+
+	return withPublic("impersonate", async () => {
+		const parsed = parseSearchParams(
+			request,
+			ImpersonateRequest,
+			"impersonate",
+		);
+		if (!parsed.ok) return parsed.response;
+
+		const { data, error } = await createServiceClient().auth.admin.generateLink(
+			{
+				type: "magiclink",
+				email: parsed.data.email,
+			},
+		);
+		if (error) throw error;
+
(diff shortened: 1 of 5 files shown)


umairnadeem (commented):
read CONVENTIONS.md and AGENTS.md (CLAUDE.md is a symlink to it) off the head branch. one thing on the route, everything else lgtm.

app/api/dev/impersonate/route.ts
```diff
@@ -0,0 +1,43 @@
...
+		const parsed = parseSearchParams(
+			request,
+			ImpersonateRequest,
+			"impersonate",
+		);
+		if (!parsed.ok) return parsed.response;
+
+		const { data, error } = await createServiceClient().auth.admin.generateLink(
+			{
+				type: "magiclink",
```
umairnadeem: `generateLink` with `magiclink` *creates* the user when the email doesn't exist (supabase docs: "generateLink() handles the creation of the user for signup, invite and magiclink"). so fat-fingering the address in the dialog silently adds a junk account to the prod auth table and signs you in as it, no error. worth resolving the user through the admin api first and 400ing when they're not there.

PR #456, opened 2026-07-22
refactor: nightly maintainability 2026-07-22

umairnadeem:
Makes the project store the single home for project-state mutation rules, and gives the two asset grids one shared implementation. No behavior changes.

## Architecture

**Store actions own the mutation rules.** Two knowledge items were living in components instead of `lib/project/store.ts`:

- `addReferenceImages(urls)` — the append rule (`setReferenceImages([...current, ...urls])`) was hand-rolled identically in `ComposerCopilot` and `AssetsSection`, each re-reading the store imperatively to do it.
- `updateCharacter(name, partial)` — partial character merges were spelled `setCharacter(name, { ...character, ...partial })` at three call sites in `CharacterEditModal`, one of which re-read the store, then **silently returned** if the character had vanished. The action now throws on an unknown name, so an impossible state surfaces instead of quietly dropping an avatar upload.

**`useProject` is the one way UI reads the store.** Seven call sites did `useConfig().projectId` + either `useProjectStore(projectId, …)` or `getProjectStore(projectId).getState()`. They now use `useProject(selector)`, which already existed as the canonical hook. `getProjectStore` is left to its real audience: non-React callers (connector plugins, queue workers, `applyTemplate`). Four components stop threading `projectId` entirely.

**One asset grid, not two.** `ComposerAssets` (composer) and `AssetsSection` (sidebar) duplicated the tile-rendering rules: narrator tile, character tiles wired to `characterAvatarElementId`, reference tiles labelled and removed by index. New `AssetTiles.tsx` holds `NarratorAssetTile` / `CharacterAssetTiles` / `ReferenceAssetTiles`; each reads what it needs from the store, so neither screen drills asset props. The two screens keep the trim that genuinely differs (add-tiles and the upload shimmers stay where they belong).

## Complexity delta

- 11 files, +45/−105 lines of app code excluding the new module and tests.
- `ComposerAssets` 90 → 65 lines, `AssetsSection` 95 → 66, `CharacterEditModal` 242 → 228.
- `getProjectStore` calls in `app/` + React providers: 9 → 3.
- Removed one silent-degradation early return and two hand-rolled read-modify-write blocks.
- Added 3 store tests (merge, unknown-name throw, append).

## Docs

One line in `ARCHITECTURE.md` recording the invariant this PR establishes: mutation rules live in store actions, components go through `useProject`, `getProjectStore` is for non-React callers.

## Skipped

- Extracting a `useCharacterAvatar` hook from `CharacterEditModal`. Its preview block overlaps the `MediaResult` helper added in #443, and the hook would have exactly one consumer.
- Deleting anything unreferenced. Nothing here was removed for being unused.

## Open PR overlap check

Considered #455, #453, #447, #446, #445, #444, #443, #442, #440, #438, #429 (file lists via `gh pr diff --name-only`). No file in this PR appears in any of them. Nearest neighbours: #440 touches `lib/project/storeSnapshot.ts` / `types.ts` (this PR touches `store.ts`), and #443 touches sibling files in `elements/` (this PR touches `AssetsSection.tsx` and adds `AssetTiles.tsx`).

## Validation

`npm run build`, `npm test` (911 passed), `npm run typecheck`, `npm run lint`, `prettier --check`, `knip` — all clean.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -39,7 +39,7 @@ Every route that takes a JSON body or search params is built from a factory in `
 ## Editor state
 
 - `lib/generation/queue.ts` schedules generation jobs from the editor, caches results per element, and exposes status to the UI. UI dispatches jobs; it never calls connectors directly.
-- `lib/project/store.ts` holds per-project metadata in Zustand. Every mutation rule lives in a store action; components reach the store through `useProject(selector)` rather than `getProjectStore(projectId)`, which is reserved for non-React callers (connector plugins, queue workers).
+- `lib/project/store.ts` holds per-project metadata in Zustand. Every mutation rule lives in a store action. Components subscribe to reactive state through `useProject(selector)`; `getProjectStore(projectId)` is the escape hatch for imperative access that a render-time selector can't express — non-React callers (connector plugins, queue workers), plus hooks doing one-shot init, `.subscribe`, or reads outside render (`ProjectEditor`, `useAutosave`, `useGenerateAll`).
 - `lib/script/` provides script context and refinement utilities.
 
 ## Data
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -39,7 +39,7 @@ Every route that takes a JSON body or search params is built from a factory in `
 ## Editor state
 
 - `lib/generation/queue.ts` schedules generation jobs from the editor, caches results per element, and exposes status to the UI. UI dispatches jobs; it never calls connectors directly.
-- `lib/project/store.ts` holds per-project metadata in Zustand.
+- `lib/project/store.ts` holds per-project metadata in Zustand. Every mutation rule lives in a store action; components reach the store through `useProject(selector)` rather than `getProjectStore(projectId)`, which is reserved for non-React callers (connector plugins, queue workers).
 - `lib/script/` provides script context and refinement utilities.
 
 ## Data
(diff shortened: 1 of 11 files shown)


ARCHITECTURE.md
```diff
@@ -39,7 +39,7 @@ Every route that takes a JSON body or search params is built from a factory in `
 ## Editor state
 
 - `lib/generation/queue.ts` schedules generation jobs from the editor, caches results per element, and exposes status to the UI. UI dispatches jobs; it never calls connectors directly.
-- `lib/project/store.ts` holds per-project metadata in Zustand.
+- `lib/project/store.ts` holds per-project metadata in Zustand. Every mutation rule lives in a store action; components reach the store through `useProject(selector)` rather than `getProjectStore(projectId)`, which is reserved for non-React callers (connector plugins, queue workers).
```
umairnadeem: this rule isn't true after the PR: `useAutosave`, `useGenerateAll` and `ProjectEditor` are all React and still go through `getProjectStore(projectId)` (imperative read, `.subscribe`, one-shot init), which `useProject` can't do. as written the next agent reads this as a blanket ban and "fixes" those three. maybe scope it to "reads/writes during render" rather than "non-React callers"?

umairnadeem (commented):
read CONVENTIONS.md + CLAUDE.md + ARCHITECTURE.md on the head branch. one thing on the new architecture doc line.

PR #467, opened 2026-07-23
refactor: rethink openslop architecture 2026-07-23

umairnadeem:
## App understanding

openslop is a Next.js 16 (App Router) + React 19 + Remotion app for AI-generated video. Users write a script on a Slate canvas, elements (image / animated_image / tts / music / sfx / video) are generated through a connector→gateway→provider pipeline, and scenes are composed into a Remotion video. Auth is Supabase (magic-link OTP + Google OAuth), with a dev-only impersonation flow. `lib/` is the domain layer; `app/` is the UI; the dependency direction is enforced one-way.

## From-scratch architecture

The domain layer already owns almost everything cleanly (API routes are one-liners over `createAssetRouteHandlers`/`pollJob`; the connector/gateway/provider seam is deep and well-separated). A four-way survey found the codebase in strong shape. The one remaining altitude smell: **auth/session mutations and internal-API calls lived inside presentational components** — `AuthForm` and `UserProfile` imported the Supabase browser client and built redirect URLs inline, and `ImpersonateDialog` was the last component doing a raw `fetch("/api/...")` with a hand-rolled `{ error }` envelope dig. From scratch, those belong behind a typed `lib/auth/` boundary, matching the existing `lib/project/api.ts` + `lib/clients/http.ts` (`apiFetch`) conventions, so components only render state.

## Implemented simplifications

New `lib/auth/` domain module; components now call typed helpers:

- **`lib/auth/session.ts`** — `sendMagicLink`, `signInWithGoogle`, `signOut`. Owns callback-URL construction and the conditional OTP options (`shouldCreateUser`/`data`), returning `{ error? }` instead of leaking the Supabase error object.
- **`lib/auth/impersonate.ts`** — `impersonateUser(email)` over `apiFetch`, so the route's error envelope surfaces as a thrown `Error` via the shared `readErrorMessage` (no more manual `res.json()` / `body.error ??` parsing).
- **`AuthForm`** and **`UserProfile`** no longer import `@/lib/supabase/client`; auth mechanics moved out of the render body.
- **`ImpersonateDialog`** drops the last raw `fetch("/api/...")` call site in `app/`.

Behavior-preserving throughout: same callback URLs, same OTP option shaping, same impersonate happy-path (`apiFetch` follows the redirect → `window.location.assign("/")` for the real reload), same error copy. The pre-existing choice for `signInWithGoogle` to not surface its `{ error }` is kept as-is (not newly changed).

## Validation

[...]

PR #469, opened 2026-07-26
fix: nightly bug fixes 2026-07-26

umairnadeem:
## What

OSML attribute values are free text — art styles, character names, prompts typed into `TextAttributePopover` and the character fields. `serializeElement` interpolated them raw into `key=\"value\"`, so a quote or an angle bracket produced malformed OSML that came back mangled on the next load:

- **A `\"` truncates the value and eats the rest of the tag.** `style=\"say \"hi\"\" characters=\"Lyra\"` re-parses as `style=\"say \"` — the remainder of that attribute is silently lost.
- **A `<` or `>` ends the tag early.** `TAG_PATTERN` is `<([^<>/][^<>]*?)>`, so `style=\"a <sunset> vibe\"` closes the element at the first `>` and the tail leaks into the node's text content.
- **`parseXmlTag` collapsed whitespace inside values.** It split the tag on `/\s+/` and re-joined with a single space, so `oil   paint` came back as `oil paint` and a newline in a multi-line value became a space.

All three are save → reload round-trip data loss on user-authored text.

## How

Root-cause fix on both sides of the one serialize/parse seam, so the escaping rule has a single home (`lib/canvas/xmlAttributes.ts`):

- `serializeElement` escapes `& \" < >` in attribute values.
- `parseXmlTag` decodes those entities, and no longer splits/re-joins the tag — it runs the attribute pattern over the remainder as-is, which preserves whitespace runs and newlines for free.

Decoding is additive, so LLM-generated OSML (which emits raw text, no entities) parses exactly as before; only our own writer changed. Unknown entity-shaped text like `AT&T` or `&frac12;` is left alone.

## Tests

- `parseXmlTag`: whitespace/newline preservation, entity decoding, unknown-entity passthrough.
- New `serialize/parse round-trip` block in `osmlStreamParser.test.ts`: quotes, angle brackets, ampersands, and \"attributes after a quoted value survive intact\" — each fails on `master`.

Full gate green: lint, prettier, `tsc --noEmit`, knip, 949 tests, production build.

## Open PR overlap check

Considered open PRs #467 (`lib/auth` + auth components), #465 (`app/components/video`), #463 (connectors + generation + `ElementContainer`), #438 (canvas components/plugins + `lib/canvas/types.ts`). This PR touches only `lib/canvas/osmlSerializer.ts`, `lib/canvas/parseXmlTag.ts`, the new `lib/canvas/xmlAttributes.ts`, and their tests — no file or theme overlap with any of them.

### DIFF
--- a/lib/canvas/__tests__/osmlStreamParser.test.ts
+++ b/lib/canvas/__tests__/osmlStreamParser.test.ts
@@ -1,7 +1,12 @@
 import { describe, expect, it } from "vitest";
 import { OSMLStreamParser, parseOSML } from "../osmlStreamParser";
-import { getElementText } from "../osmlSerializer";
-import type { ParsedElement } from "@/lib/canvas/types";
+import { getElementText, serializeOSML } from "../osmlSerializer";
+import {
+	SCENE_TYPE,
+	type CanvasContentElement,
+	type ParsedElement,
+	type SceneElement,
+} from "@/lib/canvas/types";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 
 const connectors: ConnectorRegistry = {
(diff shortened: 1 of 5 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
overegineered

Issue #461, opened 2026-07-23
Highlight words on canvas as they're narrated (karaoke / read-along)


umairnadeem:
## Problem

As narration plays, the caption line on the canvas reveals word-by-word but every visible word is rendered in one uniform style — there is no active-word highlight (karaoke / read-along). The player already knows exactly which word is being spoken each frame, but nothing visually distinguishes it.

Notably, all the hard parts already exist:

- Real per-word timestamps come from the TTS layer (`TextTimestamp = { text, start, end }`) and are persisted, so highlighting would be driven by real timings, not estimates.
- `Captions.tsx` already computes the active `wordIndex` from the current frame each render.

The only missing piece is presentational: split the line into per-word spans and style the active one.

## Reported by (verbatim feedback)

- "is there an issue on openslop gh to highlight words on canvas as they're narrated?"

## Relevant code

- `remotion/components/Captions.tsx:94-95` — active `wordIndex` already computed each frame via `sortedLastIndex(startTimes, seconds) - 1`
- `remotion/components/Captions.tsx:108` — whole visible line rendered in one uniform `<span>` (where highlight styling would hook in)
- `remotion/components/Captions.tsx:48-74` — `buildCaptionData` flattens each line into a concatenated *string*, discarding which token is active; must preserve per-word boundaries within a line for highlighting
- `lib/connectors/types.ts:131` — `TextTimestamp` real per-word start/end timing type
- `lib/providers/tts/cartesia.ts:219-228` — real `word_timestamps` from the TTS provider (not char-count estimates)
- `lib/video/resolve.ts:27-42` — exposes `captionTimestamps` on the resolved element
- `remotion/compositions/VideoComposition.tsx:54` — renders `<Captions timestamps={...} />` in the composition played by `app/components/video/VideoPreview.tsx`

## Suggested direction

- Refactor `buildCaptionData` to preserve per-word tokens within each line (rather than pre-joining to a string) so the renderer can address individual words.
- In `Captions.tsx`, render each word as its own span and apply a distinct style to the word at the active `wordIndex` (e.g. color/weight/opacity bump, or a highlight box).
- Keep it frame-driven off the existing `wordIndex` — no new playback-frame plumbing needed.
- Consider exposing an on/off toggle (fits alongside the subtitle controls proposed in #270) rather than forcing the effect on.

## Acceptance criteria

- [ ] The word currently being narrated is visually distinguished from the rest of the caption line, in sync with the audio.
[...]

Issue #452, opened 2026-07-22
Clearing an enum attribute (e.g. motion via refine) permanently hides its dropdown from Settings


umairnadeem:
## Problem

Ask the refine copilot to strip motion from animated images and the Motion dropdown vanishes from the element's Settings popover, with no way to set a motion again from the UI. The element is stuck without motion until you go around the attribute editor entirely.

The chain:

1. The refine prompt tells the model that `"attrs":{"key":null}` removes a key, so "remove all motion attributes" emits `{"op":"set","id":"…","attrs":{"motion":null}}`.
2. `updateElementAttrs` applies that faithfully — the `motion` key is deleted from `customAttributes`.
3. `AttributeBadge` then reads `element.customAttributes?.[attrKey] ?? ""` and bails on an empty value for anything that isn't a text-edit attribute:

```ts
const value = element.customAttributes?.[attrKey] ?? "";
const isTextEdit = spec.edit?.kind === "text";
if (!value && !isTextEdit) return null;
```

So the enum control unmounts, and because the control is the only way to set the value, the state is unrecoverable through the UI.

The element is not actually malformed — a missing `motion` is a legal absent-optional. The bug is that `AttributeBadge` conflates "has no value" with "has nothing to render", and it treats the schema's declared `default` as write-once. `motionDef("none")` declares `default: "none"`, and `createCanvasNode.ts:34` seeds `schema.defaultAttributes` at creation time, which is the only reason the dropdown normally appears at all. Nothing re-establishes it afterward.

**This is not motion-specific.** Every enum attribute has the same failure: `duration`, `volume`, `loops`, `length`, `emotion`. Any path that clears one — refine, a hand-edited script, an import — permanently hides its editor.

(`reconcileAttributes` does refill missing defaults, but only on a model/provider change, so switching model and back is an accidental workaround rather than a fix.)

## Reported by (verbatim feedback)

- "when we remove all motion attributes from animated images via refine copilot, the motion dropdown isn't selectable anymore. we should guard against these malformations in element attributes"

## Relevant code

- `app/components/canvas/elements/AttributeBadge.tsx:32` — `value` falls back to `""`, never to the schema default
- `app/components/canvas/elements/AttributeBadge.tsx:34` — `if (!value && !isTextEdit) return null` — the early return that hides enum editors
- `lib/connectors/attributes/common.ts:28` — `motionDef` declares `default: "none"` and the `MOTION_EFFECTS` enum options
[...]

Issue #259, opened 2026-06-12
Configurable generation queue concurrency (post-BYOK)


umairnadeem:
## Problem

Generation queue concurrency is fixed (`DEFAULT_BATCH_SIZE = 2`). It should be configurable once BYOK (bring-your-own-key) launches, so users with their own keys can raise throughput.

## Reported by (verbatim feedback)

- "Generation queue concurrency should be configurable (after BYOK launches)"

## Relevant code

- `lib/generation/queue.ts` — `DEFAULT_BATCH_SIZE = 2`

## Suggested direction

- Expose concurrency as a setting, gated behind BYOK.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #473, opened 2026-07-26
fix: keep the element card header a fixed height as characters change

### LINKED ISSUE #449: Element card header shifts up when the last character is removed from an image element
## Problem

Removing the last character from an image / animated-image element makes the whole element card's header row (and everything below it) jump up by a few pixels.

The header row is a plain `items-center` flex row with no intrinsic height, so its height tracks its tallest child:

- the type badge, `ModelBadge`, `ElementSettings`, and the add-character button are all `h-6` (24px)
- a named `CharacterPill` is a `size-6` avatar plus `py-0.5`, so it measures **28px**

While at least one character pill is present the row is 28px; remove the last one and it collapses to 24px. Same class of bug fires in reverse when the first character is added, and the row is `flex-wrap`, so a pill crossing the wrap boundary shifts things by a full line.

- "the element container header shift up slightly when a character is removed from an image/animaged image element. we should avoid this idiomatically"

## Relevant code

- `app/components/canvas/elements/ElementContainer.tsx:116` — header row: `flex min-w-0 flex-1 flex-wrap items-center gap-1.5`, no height floor
- `app/components/canvas/elements/CharacterPill.tsx:56` — pill uses `gap-1.5 py-0.5 pl-1 pr-2` around a `size-6` avatar, making it 28px tall
- `app/components/canvas/elements/ElementCharacters.tsx:25` — pills rendered/unmounted as characters are added and removed
- `app/components/canvas/elements/CharactersPicker.tsx:117` — sibling add button, `h-6 w-6`

## Suggested direction

Fix the class of problem, not the one instance: make every chip in the header row share one height so the row can't resize based on which chips happen to be mounted. Give `CharacterPill` a fixed `h-6` (horizontal padding only, drop `py-0.5`) so it matches the badges and buttons beside it. A `min-h` floor on the header row is an acceptable fallback, but normalizing the chip height is the idiomatic fix and also stops the reverse jump on add.

Note `RemoveCrossButton` is absolutely positioned, so it does not contribute to the pill's measured height.

## Acceptance criteria

- [ ] Adding or removing characters on an image / animated-image element does not change the header row's height
- [ ] Character pills are vertically aligned with the type badge, `ModelBadge`, settings, and add-character button
- [ ] Pill name truncation and the hover remove-cross still behave as before

## Related

- #362 (earlier element card header layout bug)

---

umairnadeem:
## Summary

Removing the last character from an image / animated-image element made the element card header row (and everything under it) jump up ~4px. The row is `items-center` with no height floor, so it tracked its tallest child: every other chip in it (type badge, `ModelBadge`, `ElementSettings`, add-character button) is `h-6` / 24px, but `CharacterPill` was a 24px avatar plus `py-0.5`, measuring 28px.

Fixed the class of problem rather than the instance: `CharacterPill` now carries the same `h-6` as its siblings with horizontal-only padding, so the row is 24px regardless of how many pills are mounted. That also removes the mirrored jump when the first character is added.

## Changes

- `app/components/canvas/elements/CharacterPill.tsx` — add `h-6`, drop `py-0.5` / `p-0.5` in favour of `pl-1 pr-2` and `px-0.5`.

One file, two class tokens. The avatar is `size-6` (24px) so it fits the fixed height exactly; `RemoveCrossButton` is absolutely positioned and never contributed to the measured height, and name truncation (`max-w-[140px]` + `truncate`) is untouched.

`#449` is labelled `size: XS` / `good first issue`, names the exact files and lines, and states the intended fix direction and acceptance criteria. It is a pure layout correction with no product, design-direction, or auth surface.

## Validation

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:coverage` — pass (108 files, 942 tests)
- `npm run test:e2e` — **skipped**: Playwright's web server could not bind port 3000 (already in use locally). No smoke test covers this component's chrome height.

No test added: the change is a Tailwind class on a presentational component, and asserting class names would test internals rather than behaviour.

## Open PR overlap check

Checked open PRs #472, #471, #470, #469, #467, #465, #463, #438 with `gh pr diff --name-only`. None touch `CharacterPill.tsx`. #463 touches `ElementContainer.tsx` and #438 touches element chrome, which is why this fix stays confined to the pill itself. Issue #424 was skipped as already covered by #438.

## Skipped candidates

- `#259` (size: XS) — explicitly gated on BYOK landing first.
- `#452`, `#461`, `#424` — `size: S` or larger; #424 overlaps open PR #438.
- All remaining open issues are M/L/XL, or need product/design direction (BYOK, i18n, publishing kit, community feed, provider selection).

Closes #449

### DIFF
--- a/app/components/canvas/elements/CharacterPill.tsx
+++ b/app/components/canvas/elements/CharacterPill.tsx
@@ -55,8 +55,8 @@ export function CharacterPill({
 	const avatarUrl = useCharacterAvatarUrl(name);
 	return (
 		<div
-			className={`group/pill relative inline-flex max-w-[140px] shrink-0 items-center rounded-md transition-colors hover:bg-button-hover ${
-				name ? "gap-1.5 py-0.5 pl-1 pr-2" : "p-0.5"
+			className={`group/pill relative inline-flex h-6 max-w-[140px] shrink-0 items-center rounded-md transition-colors hover:bg-button-hover ${
+				name ? "gap-1.5 pl-1 pr-2" : "px-0.5"
 			}`}
 		>
 			<CharacterAvatar


umairnadeem (commented):
lgtm

PR #475, opened 2026-07-27
refactor: nightly maintainability 2026-07-27

umairnadeem:
## What

Two pieces of one-off wiring, replaced by the seam that already exists for each.

**1. Element → connector/provider/model resolution had three homes.**

`buildGenerationJob`, the element settings popover (`ElementContainer`) and `ModelBadge` each re-derived which connector, provider and model an element generates with — with different rules:

- `buildGenerationJob` cast a persisted `customAttributes.provider` straight to `ProviderKey` with no check. A provider renamed or unregistered since the element was saved reached `createConnector`, which throws for unknown providers.
- The two UI call sites each hand-rolled an `isKnownProvider` guard against exactly that hazard, one of them with a 5-line comment explaining it.

Now `lib/canvas/elementConnector.ts` owns the policy: a pin wins when it resolves against *both* the factory's provider map and the registry, otherwise the registry default. `resolveElementSchema` layers the attribute-schema lookup on top. `buildGenerationJob` loses its cast and drops to a single call; `ElementSettings` loses 14 lines and a now-unused `config` prop; `ModelBadge` loses its guard and a `useMemo` over three cheap lookups.

Behavior delta, both in states that are unreachable today (one provider is registered per type): a generation job with an unresolvable pinned provider now falls back to the default instead of failing at connector construction, and `ModelBadge` renders for an element that pins a model but no valid provider — the same fallback the settings popover already displayed for that element. The UI and the queue previously disagreed about which provider such an element used.

**2. `/api/dev/impersonate` was the one data-taking route outside the route-handler factory**, wiring `withPublic` + `parseSearchParams` by hand. `publicRouteHandler` now takes a parse source the same way `routeHandler` does, so the route uses `createPublicQueryRouteHandler` and the production 404 guard stays as the only thing in `GET`.

## Docs

`ARCHITECTURE.md`: added the new public/query factory to the route-handler list (the section claims every such route comes from a factory — now true), and one line under "Editor state" pointing at where element connector resolution lives.

## Tests

New `lib/canvas/__tests__/elementConnector.test.ts` covers the seam: element type → connector type, default fallback, pinned model, pinned provider that resolves, pinned provider unknown to the factory, and pinned provider missing from the registry.

Full gate green: lint, prettier, `tsc --noEmit`, knip, 973 tests, production build.

## Open PR overlap check

One open PR: #438 (`app/components/canvas/Canvas.tsx`, `CollapsedVoidSync.tsx`, `CompactElement.tsx`, `hooks/useEditorSetup.ts`, `plugins/withCollapsedVoids.ts` + test, `lib/canvas/types.ts`). This PR touches `lib/canvas/elementConnector.ts` (new) + test, `lib/generation/buildGenerationJob.ts`, `app/components/canvas/elements/{ElementContainer,ModelBadge}.tsx`, `lib/api/route-handler.ts`, `app/api/dev/impersonate/route.ts`, `ARCHITECTURE.md` — no file or theme overlap.


umairnadeem (commented):
lgtm

PR #476, opened 2026-07-27
perf: nightly performance optimizations 2026-07-27

umairnadeem:
## What

Canvas element cards eagerly load their whole preview asset on mount, no matter how far off-screen they are:

- `<video>` has no `loading="lazy"` counterpart, so every clip preview fetches and decodes in full.
- `Waveform` fetches the audio, PCM-decodes it, and scans every sample for peaks — for every audio card at once.

`useNearViewport` latches once a node comes within 400px of its nearest scroll container. A preview that is nearly visible behaves exactly as before; one that isn't costs nothing until you scroll to it. The existing shimmer already covers the not-yet-loaded state, so nothing changes on screen.

This is the same "keep the cost proportional to what's on screen" fix as the placeholder orbs in #471, applied to the media itself, and follows the [Slate performance walkthrough](https://docs.slatejs.org/walkthroughs/09-performance)'s point that paint and decode dominate everything else in the frame.

## Measured

Headless Chromium, 24 preview cards, ~3.5 cards on screen.

**Video previews** (1.7 MB clips)

| | before | after |
|---|---|---|
| transferred | 41.9 MB | **7.0 MB** |
| live video decoders | 24 | **4** |
| first on-screen frame @ 12 Mbps | 799 ms | **612 ms** |

**Audio previews** (283 KB clips)

| | before | after |
|---|---|---|
| transferred | 6.94 MB | **2.02 MB** |
| clips PCM-decoded | 24 | **7** |
| peak scan on main thread | 67 ms | **19 ms** |

Also measured and rejected: `preload="metadata"` on the preview video. Chromium buffers it identically to the default (41.9 MB either way), so it buys nothing — only not mounting the source does.

## Open PR overlap check

Open PRs considered: #475 (nightly maintainability — `ElementContainer.tsx`, `ModelBadge.tsx`, `elementConnector.ts`, `buildGenerationJob.ts`, `route-handler.ts`) and #438 (dead caret — `Canvas.tsx`, `CompactElement.tsx`, `useEditorSetup.ts`, `withCollapsedVoids.ts`, `lib/canvas/types.ts`).

This PR touches `MediaWithSkeleton.tsx`, `lib/components/Waveform.tsx` and the new `lib/components/useNearViewport.ts` — no file or theme overlap with either.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test:coverage` and the Playwright smoke suite all pass locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #478, opened 2026-07-27
refactor: rethink openslop architecture 2026-07-27

umairnadeem:
## App understanding

openslop turns a prompt into a rendered video. A Slate canvas holds scenes of typed elements (narration, character line, image, animated image, clip, sfx, music). An LLM streams OSML that the canvas parses into elements; each element is generated through a connector → gateway → provider chain; results feed a Remotion layout that plays in-app and renders on Lambda. Project state lives in a per-project Zustand store plus a client-side `GenerationQueue`, both autosaved to Supabase.

## From-scratch architecture

Each piece of runtime state should have exactly one owner, and everything downstream should read from that owner rather than re-deriving or re-threading it. Three places had drifted from that:

1. **"Is this element busy?"** had ~8 homes. `GenerationQueue` knew the answer (a private `isActive`) but kept it private, so the UI re-derived it as `status !== "idle"` in some places and `status === "generating" || status === "queued"` in others.
2. **Generation inputs** were spread over four single-purpose files (`getPromptText`, `elementMetadataInputs`, `getGenerationInputs`, `generationInputs`) that only ever compose into one pipeline, so callers had to know which of three modules a symbol lived in.
3. **The player handle and the cancel-button offset** were both mirrored: `VideoPreview` held a `PlayerRef` in local state *and* pushed it into `PlayerControlContext` (which kept a third copy in a ref), and a Tailwind positioning string was drilled `AnimatedImagePreview → MediaResult → MediaPlaceholder → PlaceholderOverlay → CancelButton`.

**Generation status has one predicate.** `isGenerationActive(status)` is exported from `lib/generation/queue.ts` and used by `GenerateButton`, `ElementUploadButton`, `GenerationIndicator`, `AssetTile`, `CharacterEditModal` and the preview overlays. The queue's own internals use the same function, so the definition of "busy" can no longer drift between the scheduler and the UI.

**Generation inputs are one deep module.** `lib/generation/inputs.ts` replaces `getPromptText.ts`, `elementMetadataInputs.ts`, `getGenerationInputs.ts` and `generationInputs.ts`. The metadata-attribute table is now private to the module — it was only ever consumed by `getGenerationInputs`. Tests moved to `__tests__/inputs.test.ts` and `__tests__/staleness.test.ts`; assertions are unchanged.

**One owner for the player handle.** `PlayerControlProvider` holds the only `PlayerRef` and hands its setter out as the ref callback. `VideoPreview` drops its local `useState` and the effect that mirrored it into context; React's own ref-callback null on unmount replaces the manual `registerPlayer(null)` cleanup. Registration now costs one render instead of two.

**Cancel-button offset moves to CSS.** `--cancel-offset` is set by the container that actually needs a non-default position (the animated-image preview clears its still/video toggle; the audio placeholder centres in its shorter track). The `cancelClassName` prop is gone from three component signatures, and `OverlayButton`/`CancelButton` collapse into one component since only the cancel button ever used it. The forwarding test is replaced by `cancelOffset.test.tsx`, which asserts the same rendered outcomes.

Net: -307/+76 lines, no behaviour change.

## Validation

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:run` — 966 passed / 111 files
- `npm run test:e2e` — not run locally (port 3000 held by a running dev server); left to CI

## Skipped items

- **`ForegroundPreview` still calls `useGenerate`.** It only needs `result` and `status`, but `useGenerate` also runs a staleness memo and a `restoreResult` side effect. Swapping it for a read-only snapshot hook would be cleaner and cheaper, but in a collapsed scene it is the *only* mounted `restoreResult` path for that element, so removing it is a real behaviour change rather than a refactor. Flagging for a human call.
- **`scheduleGeneration` was left as a standalone module.** Folding it into `GenerationQueue` would point `lib/generation` at `lib/project/ensureCharacterAvatars`, trading one small file for a worse dependency direction.
- Nothing in `components/ui/**`, `icons.css` or the connector/gateway/provider seams was touched; unused-today surface there is deliberately kept.

## Open PR overlap check

Considered #477 (`lib/canvas/parseXmlTag`, `osmlStreamParser`), #476 (`MediaWithSkeleton`, `Waveform`, `useNearViewport`), #475 (`ElementContainer`, `ModelBadge`, `lib/api/route-handler`, `lib/canvas/elementConnector`, `lib/generation/buildGenerationJob`, `ARCHITECTURE.md`) and #438 (`Canvas`, `CompactElement`, `withCollapsedVoids`, `useEditorSetup`, `lib/canvas/types`). This PR touches none of those files. `lib/generation/buildGenerationJob.ts` and `ARCHITECTURE.md` were deliberately left alone for #475.


umairnadeem (commented):
lgtm, the isGenerationActive/--cancel-offset consolidation reads well and the queue-owns-the-predicate direction is right

Issue #474, opened 2026-07-27
Scene header shifts up when a scene is highlighted in expanded (non-compact) element view


umairnadeem:
When an element container is in expanded (non-compact) view and its scene is highlighted, the scene header shifts up slightly compared to the unhighlighted state.

The highlight styling appears to apply different padding/margins than the default, so toggling the highlight causes a small vertical jump in the scene header instead of only changing color/border.

**Repro**
1. Put the element container in expanded (non-compact) view.
2. Highlight the scene.
3. Watch the scene header — it shifts up a few px.

**Expected**
Highlighting a scene should be purely a visual state change; layout (padding/margins, header position) should stay identical between highlighted and unhighlighted states.

Issue #462, opened 2026-07-23
YOLO mode: auto-generate elements as they finish streaming (composer toggle + project settings panel)


umairnadeem:
## Problem

Generation is decoupled from streaming today. When the copilot streams a script, elements are parsed incrementally but nothing generates until the user manually triggers "Generate All", which batches every stale element at once. The same is true for refine edits: inserted elements sit idle until a manual pass. On a multi-element script you wait for the whole response to finish, then wait again for a batch to run — end-to-end latency is the sum of both.

We want an opt-in **YOLO mode**: when enabled, elements start generating one-by-one *as each finishes streaming*, so the earliest elements are rendering while the copilot is still writing the later ones. This should apply to **both** the initial script generation and refine copilot edits.

## Reported by (verbatim feedback)

- "YOLO mode is a mode that should auto-generate after each copilot change. this would also require updating the OSML parsing logic to mark when an element has completed streaming (detecting close tags) so that we can begin generating elements as they are done streaming. specifically, YOLO mode should begin auto-generating elements as they are done streaming (enqueuing them one-by-one) for both the initial script and also for refine copilot edits. there should be a YOLO button (with an appropriate icon and appropriate styling) toggle on the prompt composer that determines YOLO mode and it should also be configurable via a project-level new settings panel on the left."

## Relevant code

**OSML streaming parser (needs completion detection):**
- `lib/canvas/osmlStreamParser.ts` — `OSMLStreamParser`. `appendChunk()` is **open-tag driven only**: `TAG_PATTERN` matches close tags (`match[2]`) but the code never uses them; a node is only implicitly "finalized" when the *next* open tag arrives. There is no per-element completion signal today. `ParsedElement` has no status field.
- `lib/canvas/useOSMLStreamParser.ts` — `useOSMLStreamParser()` only syncs the last `MAX_NODES_TO_SYNC = 3` nodes to state, so the completion signal must come from the parser layer, not from diffing rendered nodes.

**Initial script generation (OSML path):**
- `lib/script/ScriptProvider.tsx` — `submitPrompt()` streams LLM chunks into `appendChunk`.
- `app/components/canvas/hooks/useScriptSync.ts` — mirrors parsed nodes into the Slate editor.

**Refine edits (structured JSON-op path, not OSML):**
- `lib/script/refine/parseOps.ts` — `RefineOpParser` (streaming JSON-lines → `RefineOp[]`).
[...]

Issue #454, opened 2026-07-22
Frame-native VideoLayout: store the timeline in frames, not seconds


umairnadeem:
## Problem

`VideoLayout` stores the timeline in **seconds** (`Sequence.start`, `Sequence.duration`, `totalDurationSec`, `transitionDurationSec`), but every consumer is **frame-addressed** — Remotion's `<TransitionSeries>`, `<Sequence from={...}>`, and `player.seekTo()` all take whole frames.

That mismatch forces `buildVideoLayout` to *predict* the renderer's rounding. It already caused one shipped bug: `<TransitionSeries>` overlaps scenes by `Math.round(0.4 * fps)` whole frames (10 at 24fps) while the builder subtracted a raw `0.4s` (9.6 frames), so absolutely-positioned narration/music drifted ~0.4 frames further behind the visuals at every scene boundary — about half a second by scene 30. Fixed in #439 by snapping the overlap to the frame grid.

PR #439 also added `lib/video/frames.ts` so the rounding rule has one home, and routed all nine seconds↔frames crossings through it. That was deliberately the smaller of two options: it makes the two representations *agree*, but they are still two representations. Anyone adding a seconds-valued field to the layout can reintroduce drift, and nothing in the type says which existing fields sit on the frame grid (`transitionDurationSec` now does; `start`, `duration`, and `totalDurationSec` are still arbitrary reals accumulated from provider durations).

Making the layout frame-native removes the class of bug rather than aligning around it: there is nothing to quantize, no epsilon, and no rounding to predict.

## Reported by (verbatim feedback)

- "instead of this one-off fix, is there a more idiomatic way to fix this globally in the app perhaps via a single helper?"
- "lets do shared frame helper only for now, frame-native later"

## Relevant code

- `lib/video/types.ts:23-40` — `Sequence` / `VideoLayout`; the four second-valued fields that would become frames
- `lib/video/scene-builder.ts:93` — snaps the overlap onto the grid so `toFrames` round-trips; unnecessary once the cursor is integer frames
- `lib/video/scene-builder.ts:25,170-173` — `FRAME_EPSILON` and `Math.ceil(totalDurationSec * fps - FRAME_EPSILON)`; both disappear when `totalFrames` is a sum of integers
- `lib/video/frames.ts` — `toFrames` / `toSeconds`, added in #439; would survive as the boundary converter (provider durations in, display strings out)
[...]

Issue #405, opened 2026-07-12
Re-host Runware videos to our own blob storage (provider temp URLs expire after 30-day TTL)


umairnadeem:
## Summary

Runware-generated **videos** are not re-hosted to our own Vercel Blob storage. We persist Runware's temporary `videoURL` by reference, and that temp storage has a **30-day TTL**. After 30 days the URL 404s and the video disappears from any project that references it (final renders, previews, exports).

Images already do the right thing, so this is a gap specific to the video path.

## Root cause

`AssetBundle.upload` handles two kinds of files (`lib/api/asset-bundle.ts`):

- `BundleFileUpload` (has `data`) → bytes are `put()` to our blob.
- `BundleFileExternal` (has `url`) → **only the URL is recorded in the manifest, never fetched** (`asset-bundle.ts:114`: `result[file.key] = ... file.url`).

The video provider returns the external variant, so nothing gets uploaded:

- `lib/providers/video/runware.ts:37` requests `outputType: "URL"`.
- `lib/providers/video/base.ts:30` — `toFiles` returns `[{ key: "video", url: r.url }]` (external → by-reference).

Contrast with images, which are re-hosted correctly:

- `lib/providers/image/runware.ts` requests `outputType: "base64Data"` and `toFiles` returns `{ data: Buffer.from(r.data, "base64"), ... }` (a `BundleFileUpload`), so `AssetBundle.upload` `put()`s the bytes to our blob.

Net: for videos, only `manifest.json` lands in our blob; the actual video bytes stay on Runware temp storage and expire.

## Proposed fix

Re-host the video bytes to our own blob, mirroring the image path. The video is large and Runware hands us a URL (not base64), so the natural seam is to **fetch the external URL server-side and stream it into our blob** rather than asking Runware for base64.

Two reasonable shapes:

1. **Teach `AssetBundle.upload` to internalize `BundleFileExternal`:** for external files, `fetch(url)` and `put()` the response body (stream) to `${basePath}/${filename}`, then record the local filename instead of the remote URL. This fixes any future external-URL provider for free, not just Runware video.
2. **Keep it in the video provider:** in `RunwareVideo`, after poll completes, fetch `videoURL` and return a `BundleFileUpload` (bytes + contentType) so the existing upload path re-hosts it.

[...]

PR #479, opened 2026-07-27
fix: keep enum attribute editors visible when the key is absent

umairnadeem:
## Summary

Clearing an enum attribute made its editor disappear permanently. Ask the refine copilot to strip motion from animated images and `updateElementAttrs` deletes the `motion` key, then `AttributeBadge` reads `customAttributes[key] ?? ""` and returns `null` for any non-text-edit attribute with an empty value. The dropdown unmounts, and since the dropdown is the only way to set the value, the element is stuck without motion.

A missing `motion` is a legal absent-optional, not a malformation. The bug is that the badge conflated "has no value" with "has nothing to render", and treated the schema's declared `default` as write-once (seeded at node creation, never re-established).

The fix is in the attribute editor, not in motion or in the refine op:

- `AttributeSpec` now carries `default`, and `AttributeSchema.visibleAttributes` passes it through. `default` moved up from `AttributeDef`, so `AttributeDef` is now just `AttributeSpec & { key }`.
- `AttributeBadge` resolves its value as `customAttributes[key] ?? spec.default ?? ""`.
- The early return is now the genuinely-nothing-to-show case only: no value, no default, and no edit affordance.

Selecting a value already writes it explicitly via `updateElementAttrs`, so a defaulted-but-absent attribute ends up with a real value once touched. This applies to every enum attribute (`duration`, `volume`, `loops`, `length`, `emotion`), not just `motion`.

#452 — priority: high, size: S, with a clear root cause, named files, and explicit acceptance criteria. The fix is one expression plus an interface field, in a class-of-problem shape rather than a motion special case. No behavior change for attributes that already had a value.

## Scope note: one acceptance criterion deferred

The issue also asks that refine `set` ops be unable to write an enum value outside the schema's declared options. That guard needs to resolve an element's schema inside `lib/script/refine/applyOps.ts`, and **open PR #475 introduces exactly that resolver** (`resolveElementSchema` in `lib/canvas/elementConnector.ts`). Building a second resolver now would duplicate the code #475 is removing and conflict with it, so this PR stops at the editor fix and #452 stays open for the boundary guard as a follow-up on top of #475. The reported user-visible bug is fully fixed here.

## Validation

| Check | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:coverage` | 970 passed / 112 files |

New `app/components/canvas/elements/__tests__/AttributeBadge.test.tsx` covers the four rendering cases: absent key falls back to the default, an explicit value wins over the default, an enum with neither value nor default stays editable, and a no-value/no-default/no-edit attribute renders nothing. `schema.test.ts` gained an assertion that `visibleAttributes` carries `default`.

`npm run test:e2e` was skipped: it needs a Playwright browser install and Supabase env vars that aren't available in this unattended run, and it exercises no attribute-editor path.

## Open PR overlap check

Compared against open PRs #438, #475, #476, #477, #478 via `gh pr diff --name-only`. None touch `AttributeBadge.tsx` or `lib/connectors/attributes/schema.ts`. #475 does touch two `AttributeBadge` callers (`ElementContainer.tsx`, `ModelBadge.tsx`), but this change only adds an optional field to `AttributeSpec`, so both remain prop-compatible.

## Skipped candidates

No `size: xs` issues are open. Skipped as too large, product-context-heavy, or already covered: #462/#451/#379/#331 (XL enhancements), #454/#390/#385 (L, multi-subsystem), #405 (needs blob storage + provider credentials), #474 (visual/layout judgment call, adjacent to just-merged #473), #424 (covered by open PR #438).

Refs #452

### DIFF
--- a/app/components/canvas/elements/AttributeBadge.tsx
+++ b/app/components/canvas/elements/AttributeBadge.tsx
@@ -6,7 +6,10 @@ import type { AttributeSpec } from "@/lib/connectors/attributes/schema";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 import { TextAttributePopover } from "./attributes/TextAttributePopover";
 
+const UNSET = "—";
+
 function formatValue(value: string, unit?: string): string {
+	if (!value) return UNSET;
 	return unit ? `${value}${unit}` : value;
 }
 
@@ -30,9 +33,7 @@ export function AttributeBadge({
 	deriveExtraAttrs,
 }: AttributeBadgeProps) {
 	const editor = useSlateStatic();
-	// An absent optional is a legal state, so fall back to the schema default
-	// rather than unmounting the only control that could set the value again.
-	const value = element.customAttributes?.[attrKey] ?? spec.default ?? "";
+	const value = element.customAttributes?.[attrKey] ?? "";
 	if (!value && !spec.edit) return null;
 
 	const SpecIcon = spec.icon;
@@ -51,7 +52,7 @@ export function AttributeBadge({
 			{formatValue(value, spec.unit)}
 		</>
 	);
-	const tooltip = `${spec.label}: ${value || ""}`;
+	const tooltip = `${spec.label}: ${value || UNSET}`;
 
 	if (!spec.edit) {
 		return (
--- a/app/components/canvas/elements/AttributeBadge.tsx
+++ b/app/components/canvas/elements/AttributeBadge.tsx
@@ -30,9 +30,10 @@ export function AttributeBadge({
 	deriveExtraAttrs,
 }: AttributeBadgeProps) {
 	const editor = useSlateStatic();
-	const value = element.customAttributes?.[attrKey] ?? "";
-	const isTextEdit = spec.edit?.kind === "text";
-	if (!value && !isTextEdit) return null;
+	// An absent optional is a legal state, so fall back to the schema default
+	// rather than unmounting the only control that could set the value again.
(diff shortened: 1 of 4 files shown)


app/components/canvas/elements/AttributeBadge.tsx
```diff
@@ -30,9 +30,10 @@ export function AttributeBadge({
 	deriveExtraAttrs,
 }: AttributeBadgeProps) {
 	const editor = useSlateStatic();
-	const value = element.customAttributes?.[attrKey] ?? "";
-	const isTextEdit = spec.edit?.kind === "text";
-	if (!value && !isTextEdit) return null;
+	// An absent optional is a legal state, so fall back to the schema default
+	// rather than unmounting the only control that could set the value again.
+	const value = element.customAttributes?.[attrKey] ?? spec.default ?? "";
```
umairnadeem: falling back to `spec.default` makes the badge assert a value the renderer won't use for `volume`: schema defaults are `volumeDef("2")` for sfx/music and `"5"` for video, but `getVolume` in `lib/video/elementAttributes.ts` falls back to `DEFAULT_VOLUME = 10` when the key is absent. so refine strips `volume` off a music element -> badge reads "Volume 2", playback is at 10. worth aligning the two fallbacks (or having the consumers read the schema default) so the badge isn't lying, otherwise you've traded a disappearing control for a wrong one.

PR #480, opened 2026-07-28
feat: model generation jobs as a dependency graph

### LINKED ISSUE #243: Character/avatar generation correctness (pre-gen reference error, staleness, art-style refs)
## Problem

Character/avatar generation has correctness gaps, all rooted in the queue not modeling the dependency between an image/animated-image node and the avatars of the characters it references:

- **Generating an image node that references a character before that character's avatar exists errors out** -- especially when there is only 1 character. `character-references.ts` filters out characters lacking an `avatarUrl`, so `referenceImages` can become empty and generation fails.
- **A freshly generated image is flagged stale the instant it finishes**, with no user action (merged from #295). Repro: new story with 2 characters, add the first character to the first image/animated-image node, Generate all, wait for the first image. It lands a result and is immediately marked stale.
- **Changing reference images does not mark character avatars stale**, so the avatar isn't regenerated.
- **Character avatars don't respect art-style reference images.**

- "Generating a single image node that references a character before the character is generated errors out if only 1 character in story"
- "Image marked stale immediately after it finishes generating"
- "Changing reference images doesn't trigger staleness for character avatars"
- "Character avatars don't respect art style refs"

## Root cause of the immediate-stale race (from #295)

A dependent image job snapshots its generation inputs at job start, but a dependency (a referenced character's `avatarUrl`) can change while the job is still running, because avatar generation mutates shared project metadata after the dependent job already captured its inputs.

- `scheduleGeneration` enqueues avatars first, then the image job (`lib/generation/scheduleGeneration.ts:17-18`).
- `ensureCharacterAvatars` enqueues an avatar job for every character lacking an avatar, not just referenced ones (`lib/project/ensureCharacterAvatars.ts:54-57`). With 2 characters that is 2 avatar jobs.
- `DEFAULT_BATCH_SIZE = 2` (`lib/generation/queue.ts:336`), so both avatar jobs fill the batch and the image waits.
- When the first avatar finishes, `processQueue` starts the image's `runJob`, which snapshots inputs synchronously (`queue.ts:278-279`). If the referenced character's avatar is the one still in flight, its `avatarUrl` is empty, so `characterAvatars` is omitted from the snapshot (`elementMetadataInputs.ts:20-27`).
- That avatar then completes and writes its URL into shared metadata (`character-avatar.ts:27`).
- The image result stores the frozen `resultInputs` (no avatar URL). `useGenerate` recomputes `currentInputs` from now-fresh metadata (avatar URL present), and `isStaleResult` sees `!isEqual(...)` -> stale (`useGenerate.ts:19-23`, `queue.ts:25-34`).

The "enqueue avatars first" comment only guarantees ordering with batch size 1 or a single avatar. With concurrency > 1 and multiple avatars, a dependent can start before the specific avatar it depends on has resolved. This is the same missing-dependency-edge that causes the pre-gen reference error above.

## Suggested direction

Refactor `ensureCharacterAvatars` / `scheduleGeneration` to model jobs as a dependency graph and **topologically sort** them, so a dependent image/animated-image job only runs after the avatars of the characters it actually references have resolved.

- Build edges from each image/animated-image job to the avatar jobs of its referenced characters (`parseCharacterNames(element.customAttributes?.characters)`).
- Only enqueue avatar jobs that are actually referenced by the dispatched jobs, rather than every unavatared character.
- Gate dependents on their specific referenced avatars (a real dependency edge / barrier), not on any one avatar freeing a batch slot. This removes the window where `resultInputs` is captured against a not-yet-resolved dependency, and fixes the empty-`referenceImages` error at its source.
- Include reference images in the avatar's staleness inputs so ref changes mark the avatar stale.
- Thread art-style reference images into the avatar prompt.

## Relevant code

- `lib/connectors/image/plugins/character-references.ts` -- filters out characters without `avatarUrl`
- `lib/project/ensureCharacterAvatars.ts:54-57`, `lib/generation/scheduleGeneration.ts:17-18` -- avatar enqueue ordering (no dependency graph)
- `lib/generation/queue.ts:25-34` (`isStaleResult`), `:278-279` (input snapshot at job start), `:336` (`DEFAULT_BATCH_SIZE`)
- `lib/connectors/image/plugins/character-avatar.ts:27` -- mutates shared metadata in `afterGenerate`
- `lib/generation/elementMetadataInputs.ts:20-27, 67-68` -- `characterAvatars` is part of image / animated_image inputs
- `app/components/canvas/hooks/useGenerate.ts:19-23` -- live `currentInputs` vs frozen `resultInputs`

## Related

- Merges #295 (immediate-stale race) into this issue.
- #233 (revert character appearance to the avatar's source while stale) -- adjacent but distinct.

---

umairnadeem:
## Summary

- Generation jobs now form a dependency graph. A node is either **generated** (it has a job) or a **source node** standing for project state (reference images, art style, aspect ratio, voice). Edges are declared by the connector plugins that do the reading, so one declaration drives what a job reads, what it waits for, and what makes it stale.
- **Ordering**: `processQueue` picks the first runnable pending node instead of shifting blindly, so a dependent waits on the avatars it actually references rather than on any avatar freeing a batch slot. Correct at any batch size.
- **Staleness**: `GenerationInputs` records what each dependency resolved to, so staleness is one walk of the graph. This catches uploads, which change a *result* rather than an *input* — the animated-image bug no input-hashing scheme could see.
- **Animated images** split into a still node feeding the animation. Reuse falls out for free: an unchanged still is not stale so it is not regenerated, and replacing it leaves the animation stale.
- Jobs receive their dependencies' results and the project state their inputs were resolved against, so nothing reads shared mutable metadata mid-flight. That closes the immediate-stale race at its source rather than patching it.
- Removes `ensureCharacterAvatars`, `scheduleGeneration`, `buildGenerationJob`, `getGenerationInputs`, `ELEMENT_METADATA_INPUTS`, `isStaleResult`, `reusableStill`, `PriorGeneration`, and `metadata.avatarUrl`. Character avatars now live in the queue snapshot, which already persists. Five plugin factories no longer need `projectId`.

Closes #243.

## Unrelated drive-by

One tooling commit rides along: `chore: match the create-pr skill to how this repo actually commits`. The skill banned conventional-commit prefixes and asked for bodies of "1-3 short lines, sentence fragments", both the opposite of what master's history does. It also squashed against the local `master` ref, which is often stale. Touches only `.agents/skills/create-pr/SKILL.md`, no product code. Happy to split it out if you'd rather review it separately.

## Test plan

Automated (all green): `npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and 977 tests across 113 files. New coverage in `graph.test.ts`, `resolveGraph.test.ts`, and `queueDependencies.test.ts`, including the issue's exact repro at `batchSize: 2`.

Manual, against the four reports in #243:

- [ ] New story, 2 characters. Add the first character to the first image node, Generate all. First image lands a result and is **not** immediately marked stale.
- [ ] Generate a single image node referencing a character before that character's avatar exists, with only 1 character in the story. Should generate rather than error.
- [ ] Change the project's reference images. Character avatars go stale.
- [ ] Set an art style, then generate a character avatar. Style is applied.
- [ ] Animated image with a result, then upload a new still image. The animation is marked stale.
- [ ] Reload a project. Avatars, results, and staleness survive (avatars now come from the queue snapshot, not `metadata.avatarUrl`).
- [ ] Apply a template that ships prebuilt avatars. They appear and are not stale.

Not run: the Playwright e2e smoke suite (port 3000 was occupied locally). It covers onboarding copy and routes, which this change does not touch.

## Known gaps

- `ElementUploadButton` still imports `stillDependency` from the animated-image connector to decide where an upload lands. Under discussion; deliberately left as a one-line special case rather than adding a graph abstraction with one consumer.
- Animated images now count as 2 units in `QueueProgressBar` (still + animation) instead of 1. Arguably more accurate, but it is a visible change.


lib/connectors/image/plugins/characterAvatarNode.ts
```diff
@@ -0,0 +1,44 @@
...
+		createReferenceImagesPlugin(),
+		createDimensionsPlugin("image"),
+	];
+}
+
+export function characterAvatarNode(
+	name: string,
+	ctx: GraphContext,
+): GenerationNode {
+	return ctx.resolve(
```
umairnadeem: nothing in the graph looks at `avatarUploaded`, so an uploaded avatar now goes stale like any other node. upload a custom avatar, then change the art style (or reference images, or the appearance text) and hit Generate All → we regenerate right over their upload, which `ensureCharacterAvatars` never did. the modal already special-cases `!avatarUploaded` for staleness, worth honoring that here too.

lib/connectors/image/plugins/imageChain.ts
```diff
@@ -1,24 +1,14 @@
 import type { ConnectorPlugin } from "@/lib/connectors/types";
+import { createDimensionsPlugin } from "../../plugins/dimensions";
```
umairnadeem: `CLAUDE.md` § Project structure: *"Use the `@/*` path alias for imports that traverse 2+ directory levels (e.g. `@/lib/connectors/types`); keep single-level relative imports (`../`) as-is."* so this wants `@/lib/connectors/plugins/dimensions`. same import in `characterAvatarNode.ts` and `animated-image-chain.ts`.

lib/generation/queue.ts
```diff
@@ -285,27 +298,67 @@ export class GenerationQueue {
...
+	private failBlocked() {
+		if (this.controllers.size > 0 || this.pending.length === 0) return;
+		const blocked = this.pending;
+		this.pending = [];
+		for (const node of blocked) {
+			const missing = this.blockingDependency(node);
+			this.update(node.id, {
+				status: "idle",
+				seconds: 0,
+				result: null,
```
umairnadeem: this nulls `result` on nodes that never ran, so one dep failure wipes a dependent's good result. Generate All on an image that already has a result plus a stale avatar dep, avatar 500s → the image's result is gone (blob's still in storage, we just forget the url). imo leave `result`/`resultInputs` alone here and set only `error`.

lib/project/characterAvatar.ts
```diff
@@ -0,0 +1,29 @@
+import type { CanvasContentElement } from "@/lib/canvas/types";
+import { getPrimaryUrl } from "@/lib/connectors/assetUrl";
+import { derivedNodeId, type NodeResults } from "@/lib/generation/graph";
+import type { ProjectState } from "@/lib/generation/sourceNodes";
+
+export const characterAvatarElementId = (name: string) =>
+	derivedNodeId("avatar", name);
```
umairnadeem, line 8: avatar node ids move from `character-avatar:<name>` to `~avatar:<name>` and `avatarUrl` comes out of `MetadataCharacterSchema`, with no migration for either, so every already-saved project loses its character avatars on load and regenerates them. the old-prefix entries also stick around in the persisted queue snapshot and `pickThumbnailUrl` only filters `~`, so an orphaned avatar can now win the project thumbnail.

lib/connectors/image/plugins/characterAvatarNode.ts
umairnadeem: Good catch, real regression. Fixed in d9e6ddc.

Rather than copy the modal's `!avatarUploaded` check here, I put the rule on the node so the two can't disagree: `GenerationNode` gains `pinned`, `needsGeneration` short-circuits on it once a result exists, and `characterAvatarNode` sets it from `avatarUploaded`. The modal's guard is now redundant and gone.

Worth noting why pinning sits outside the node's inputs, since that was my first instinct: the modal commits the upload and *then* flips `avatarUploaded`, so anything that changed the node's inputs would make the just-committed result look stale one render later.

lib/connectors/image/plugins/imageChain.ts
umairnadeem: Fixed in d9e6ddc, all three call sites now use `@/lib/connectors/plugins/dimensions`.

lib/generation/queue.ts
umairnadeem: Agreed, fixed in d9e6ddc. `failBlocked` now sets only `status` and `error` and leaves `result`/`resultInputs` alone, so a node that never ran keeps the URL it was holding.

Added a regression test for it, and confirmed it fails against the old line rather than passing vacuously.

lib/project/characterAvatar.ts
umairnadeem: Skipping this one deliberately: we decided this change is explicitly not backwards compatible, so no migration for either the node-id change or the `avatarUrl` removal. Existing projects will lose their character avatars and regenerate them.

The thumbnail half falls out of the same decision. For any project saved after this lands every derived node carries the `~` prefix, so `pickThumbnailUrl` filters them correctly; an orphaned `character-avatar:` entry can only exist in a snapshot written before this change.

Happy to reconsider if you'd rather ship a one-time backfill instead.

app/components/canvas/elements/ElementUploadButton.tsx
```diff
@@ -2,33 +2,24 @@
...
-				queue.cancel(element.id);
-				queue.commitResult(
-					element.id,
-					{ imageUrl: url, durationSec: 0 },
-					inputs,
-					ELEMENT_CONFIGS[element.type].connector,
-				);
-			}}
+			onUpload={(url) =>
+				queue.commitResult(target, { imageUrl: url, durationSec: 0 })
```
umairnadeem, line 20: the pinning rule you just added for avatars applies one layer over: this commits an unpinned node, so an uploaded image goes stale when the art style changes and Generate All regenerates over it. upload a fix onto an image element, tweak the art style, Generate All → the upload's gone. worth threading `pinned` through `commitResult` for uploads?

umairnadeem: You're right, and confirmed: an image element with a prompt plus an uploaded result went stale on an art-style change and Generate All regenerated over it. It's a regression from this branch specifically, since art style wasn't part of an element's inputs before.

Fixed in 988ea4d, and I moved pinning while doing it. It now lives on `ElementSnapshot` rather than `GenerationNode`, which is the honest home: it's a fact about where a result came from, it persists with the snapshot, and it's set atomically with the result instead of trailing a metadata flag written a render later. `needsGeneration` reads it off the snapshot, so avatars and elements get the same rule from one place.

`commitResult` takes the flag explicitly as you suggested. All three callers are user-supplied results (both upload buttons and the template seed) and pass `{ pinned: true }`; a generation job clears it, so regenerating unpins. Template seeds being pinned also restores the old `ensureCharacterAvatars` behaviour, where a prebuilt avatar was never auto-regenerated.

Two tests cover it: an uploaded image staying fresh across an art-style change, and a generated one still going stale.

lib/connectors/image/plugins/character-references.ts
```diff
@@ -1,24 +1,32 @@
...
+/**
+ * Feeds the referenced characters' avatars in as reference images. The avatars
+ * arrive as dependency results, so this never races the jobs that produce them.
+ */
+export function createCharacterReferencesPlugin(): ConnectorPlugin<ParamsWithCharacters> {
 	return {
 		name: "character-references",
-		beforeGenerate(params) {
+		dependencies: (element, ctx) =>
+			parseCharacterNames(element.customAttributes?.characters).map((name) =>
```
umairnadeem: this declares an avatar dep for every referenced name, so Generate All now generates avatars for characters with no appearance. `ensureCharacterAvatars` filtered on `ch.appearance` for exactly that, and the modal still disables Generate when `!hasAppearance`. repro: New character (`setCharacter(name, { appearance: "" })`), attach to an image, Generate All → we burn a portrait of just the name and feed it back in as a reference image. worth skipping names with no appearance in `ctx.state`?

lib/generation/resolveGraph.ts
```diff
@@ -0,0 +1,81 @@
...
+	const resolved = new Map<string, GenerationNode>();
+	const resolving = new Set<string>();
+
+	const build = ({
+		element,
+		plugins: override,
+	}: ElementNode): GenerationNode => {
+		const { id } = element;
+		const existing = resolved.get(id);
+		if (existing) return existing;
```
umairnadeem: this cache now outlives a single graph build: `useNodeBuilder` memoizes the builder on metadata/referenceImages only, so once an element's node is cached, editing the element never rebuilds it (`resolveGraph` got a fresh map per call). type a prompt into a fresh image element -> the node still holds `prompt: ""`, so Generate errors "Enter a prompt first" and Generate All filters it out; same for a changed `characters` attribute, whose new avatar deps never show up. imo scope `resolved`/`resolving` to one top-level `buildNode` call instead of the builder's lifetime.

lib/generation/queue.ts
```diff
@@ -161,11 +158,11 @@ export class GenerationQueue {
 		this.controllers.get(id)?.abort();
 		this.controllers.delete(id);
 		this.stopTimer(id);
-		this.pending = this.pending.filter((j) => j.elementId !== id);
+		this.pending = this.pending.filter((node) => node.id !== id);
```
umairnadeem, line 161: abort only kills this node, so cancelling a node that's blocked leaves its in-flight dep running. hit cancel on an animated image while `~still:<id>` is generating -> the anim entry is deleted but the still lands a result, and since the preview now reads the still node an image just shows up in an element you cancelled. cascade the abort to deps that only this root is waiting on?

lib/connectors/image/plugins/character-references.ts
umairnadeem: Confirmed and fixed in ec0a5fc. `ensureCharacterAvatars` filtered on `ch.appearance` and my declaration didn't, so Generate All burned a portrait of the bare name and fed it back as a reference image.

I put the rule on the avatar rather than on this call site: `forCharacterAvatar` returns a source node when the character has no appearance, so it is a leaf the graph never generates, and adding an appearance turns it back into a node that needs generating. That way the edit modal's Generate and template seeding inherit it too, instead of three places each remembering to check.

Two tests: no generation for an appearance-less character, and generation once one is set.

lib/generation/resolveGraph.ts
umairnadeem: Right, and this was live — staleness stopped working entirely. Fixed in e01dced (before this round).

Exactly your diagnosis: `resolveGraph` created the maps per call, and moving to a bound builder silently widened their lifetime to the builder's. Element content is in neither `registry` nor `state`, so an edited element kept resolving to its cached node. `resolved`/`resolving` are scoped to one top-level call again.

Regression test builds a node, changes the element's text, builds again, and asserts the prompt changed — verified failing against the broken version.

lib/generation/queue.ts
umairnadeem: Good catch, and it's newly visible because the preview reads the still node now. Fixed in ec0a5fc.

`cancel` and `discard` now abort the dependencies nothing else is waiting on, which is your "only this root is waiting on" rule: a shared avatar keeps running for whoever else needs it, an exclusive still does not. That needed the queue to remember the nodes it is running — it dropped them from `pending` on start, so it had no way to tell who else was waiting.

Two tests: cancelling a node settles its exclusive dep, and a dep a second queued node still needs keeps generating.

lib/connectors/image/plugins/characterAvatarNode.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const forCharacterAvatar =
+	(name: string): NodeSpec =>
+	(state) =>
+		state.metadata.characters[name]?.appearance?.trim()
+			? {
+					element: characterAvatarElement(state, name),
+					plugins: buildCharacterAvatarPlugins(name),
+				}
+			: sourceNode(characterAvatarElementId(name), {});
```
umairnadeem: this makes uploading an avatar throw for a character with no appearance: `NewCharacterDialog` creates one with `appearance: ""`, then `CharacterEditModal`'s upload calls `queue.commitResult(avatarNode, ...)` -> `requireJob` -> `Source node "~avatar:X" cannot be generated`. add a character, hit the upload button, the upload is lost. same shape in `applyTemplate` if a template ever ships an avatar for an appearance-less character.

lib/connectors/image/plugins/characterAvatarNode.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * A character with no appearance has nothing to draw from, so its avatar is a
+ * leaf the graph never generates rather than a portrait of the bare name. Giving
+ * an appearance turns it back into a node that needs generating.
+ */
+export const forCharacterAvatar =
+	(name: string): NodeSpec =>
+	(state) =>
+		state.metadata.characters[name]?.appearance?.trim()
```
umairnadeem: a source node's identity is `serializeNodeInputs({prompt:"",attributes:{}})`, so it's constant. that means swapping an appearance-less character's uploaded avatar no longer makes dependents stale, even though `dependencyResults` still feeds the new url in as a reference image. `resultIdentity` of the snapshot would keep the upload case working here.

lib/generation/queue.ts
umairnadeem: Correction to my reply above: the cascade is reverted in 6019720. On reflection we'd rather cancelling a node not auto-abort the dependencies only it was waiting on, so a still that is already generating will finish and land its result, and the preview will show it.

Leaving this as accepted behaviour rather than a fix. Reopen if you disagree.

lib/generation/queue.ts
```diff
@@ -285,27 +303,60 @@ export class GenerationQueue {
...
+	/**
+	 * Nothing running and nothing runnable means a dependency never arrived, so
+	 * release what is left rather than leaving it queued forever. The failure is
+	 * already reported on the node that actually failed.
+	 */
+	private releaseBlocked() {
+		if (this.controllers.size > 0 || this.pending.length === 0) return;
+		const blocked = this.pending;
+		this.pending = [];
+		for (const node of blocked) this.resetToIdle(node.id);
```
umairnadeem: "the failure is already reported on the node that actually failed" doesn't hold for derived nodes: `~still:<id>` has no card, and `useGenerate` only reads `getElementSnapshot(element.id)`. hit Generate on an animated image, the still 500s -> the card silently drops back to idle, no result, no error, nothing to retry from. `CONVENTIONS.md` § Hard rules: *"Never swallow an error on a critical path to keep going. Propagate it or surface it."* rather than inventing a message, maybe copy the blocking dep's own error onto the dependent before releasing it.

lib/canvas/elementConnector.ts
```diff
@@ -22,34 +16,25 @@ export type ElementConnector = {
...
-				model: model ?? config.defaultModel,
-				config,
-			};
-		}
-	}
-
-	const { provider, config } = getDefaultConnector(registry, type);
-	return { type, provider, model: model ?? config.defaultModel, config };
+	const { provider, model } = element.customAttributes ?? {};
+	if (!provider)
```
umairnadeem, line 29: `createCanvasNode` only started stamping `provider` in #393 (jul 11), so any element saved before that has none and this now throws from `ModelBadge`/`ElementContainer` render, ie the card blows up on load rather than falling back. the removed guard also covered the unknown-provider case: `registry[type][provider as ProviderKey]` types `config` as present but hands back undefined, so you get `Cannot read properties of undefined (reading 'defaultModel')`. imo keep the `getDefaultConnector` fallback for both.

lib/connectors/image/plugins/character-references.ts
umairnadeem: Reverted this one in the latest push, deliberately. The guard I added returned a source node for a character with no appearance, so the avatar was declared but never generated.

We'd rather not special-case it: appearance is filled in by the LLM when it writes the characters, so empty is rare, and the guard cost a fake node in the graph plus a branch in every reader of that spec. An avatar with no appearance now generates like any other, prompt and all.

It also removes a crash it had introduced: seeding a template avatar for such a character resolved to a source node and threw in requireJob.

umairnadeem (commented):
lgtm, latest round reads clean to me. the earlier comments on `elementConnector.ts` and the blocked-dependent error are still open on my side, but nothing new in these commits.

PR #481, opened 2026-07-28
fix: nightly bug fixes 2026-07-28

umairnadeem:
## The bug

OSML is the persisted project format: `useAutosave` serializes the editor with `serializeOSMLWithScenes`, and `deserializeWithScenes` parses it back on load. Neither side escaped anything, so any `"`, `<`, `>`, or `&` a user typed into element text or an attribute broke the round trip.

An image whose prompt is `a 24" monitor & a <box>` serialized to:

```
<image id="e1" prompt="a 24" monitor & a <box>">hi</image>
```

and came back from `deserializeWithScenes` as an element of type `box` with no id, no attributes, and the text `">hi`. The user's element is silently destroyed on the next reload. Angle brackets in narration text corrupt structure the same way, splitting one element into two bogus ones.

## The fix

One `escapeXml`/`unescapeXml` pair in `lib/canvas/xmlEscape.ts`, applied at the two boundaries that own the format:

- `osmlSerializer` escapes attribute values and element text on the way out.
- `parseXmlTag` and `OSMLStreamParser` decode on the way in.

`unescapeXml` decodes in a single pass over known entities, so text a user literally typed as `&amp;` survives instead of being double-decoded.

The stream parser flushes loose text once the buffer has no tag characters in it, which could split `&amp;` across two chunks and decode the halves separately. It now holds a trailing partial entity in the buffer until the rest arrives.

Behavior is otherwise unchanged: escaping is only visible inside the serialized string. Projects saved before this change keep whatever their old text was; the corruption path is closed for everything written from now on.

## Tests

- `lib/project/__tests__/serialize.test.ts` — end-to-end serialize → deserialize round trip with quotes, ampersands, and angle brackets in both an attribute and element text. Fails on master by producing a `box` element.
- `lib/canvas/__tests__/xmlEscape.test.ts` — round trip, no double-decoding of a literal `&amp;`, unknown entities left alone.
- `lib/canvas/__tests__/parseXmlTag.test.ts` — attribute values are decoded.
- `lib/canvas/__tests__/osmlStreamParser.test.ts` — an entity split across chunk boundaries decodes once.

## Open PR overlap check

Open PRs considered: #480 (generation dependency graph — `lib/generation/*`, `lib/connectors/*`, `lib/project/{characterAvatar,deleteCharacter,ensureCharacterAvatars,thumbnail,types}`, `lib/video/aspectRatio*`, canvas element components) and #438 (dead caret on collapsed elements — `app/components/canvas/*`, `lib/canvas/types.ts`).

This PR touches `lib/canvas/{xmlEscape,osmlSerializer,parseXmlTag,osmlStreamParser}` and `lib/project/__tests__/serialize.test.ts`. No file or theme overlap with either.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `npm test` (984 passed), `npm run build` all pass.

### DIFF
--- a/lib/canvas/__tests__/osmlStreamParser.test.ts
+++ b/lib/canvas/__tests__/osmlStreamParser.test.ts
@@ -64,6 +64,15 @@ describe("OSMLStreamParser", () => {
 		expect(getElementText(nodes[0])).toContain("Some long narration text here");
 	});
 
+	it("decodes an entity split across chunk boundaries", () => {
+		const s = new OSMLStreamParser();
+		s.appendChunk("<narration>Rock &a", connectors);
+		s.appendChunk("mp; Roll forever</narration>", connectors);
+
+		const nodes = s.getNodes() as ParsedElement[];
+		expect(getElementText(nodes[0])).toContain("Rock & Roll forever");
+	});
+
 	it("preserves raw tag name for unknown tags", () => {
 		const s = new OSMLStreamParser();
 		s.appendChunk("<unknowntag>content</unknowntag>", connectors);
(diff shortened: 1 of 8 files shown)


umairnadeem (commented):
lgtm

PR #482, opened 2026-07-28
refactor: nightly maintainability 2026-07-28

umairnadeem:
No behavior changes. Three seams cleaned up, plus the tests that let them be checked.

## Architecture

- **`lib/components/peaks.ts` (new)** — the fetch → decode → normalize pipeline moved out of the `Waveform` client component into a plain module. `Waveform` now renders; it no longer owns an `AudioContext` singleton, a fetch chain, or the peak math. The pure `extractPeaks` helper is no longer reachable only by importing a `.tsx` component from a test.
- **`app/components/canvas/hooks/useEditorSession.ts` (new)** — one hook owns every wire between the Slate editor and the project (hydrate, stream-sync, metadata-sync, autosave, layout key). `PostPromptView` went from importing five setup hooks plus a `useMemo` to one call, so the view renders the editor instead of assembling it.
- **`app/components/projects/ProjectCard.tsx` (new)** — the project grid card split out of `ProjectsList`, which now holds only list state (create, optimistic delete, confirm dialog). Card navigation is real `<Link>` anchors rather than `router.push` buttons, so middle-click, open-in-new-tab, and prefetch work; the two duplicated push handlers collapse into one `href`.

## Maintainability

- `Waveform`'s load effect drops from 18 lines of promise chaining to 10, with the success/error `setLoadedSrc` duplication folded into one `finally`.
- Tests split per module: `__tests__/peaks.test.ts` and `__tests__/soundwave.test.ts` replace the single `Waveform.test.ts` that tested two modules through a third.
- New coverage for `loadPeaks` (decode path and the fetch-failure throw) — previously untested.
- One line added to ARCHITECTURE.md's "Editor state" section pointing at the new editor-session seam.

## Complexity delta

- `ProjectsList.tsx` 145 → 84 lines; `PostPromptView.tsx` 89 → 66; `Waveform.tsx` 236 → 190.
- Net +218/−136 across 9 files, most of it the new tests.

## Skipped

- **`useAutosave` simplification.** It's the densest hook left (5 effects, 3 refs, a debounce-in-a-ref indirection). It's also a data-loss-sensitive path with zero tests, and the shortest path to simplifying it runs through exactly the ref-stability shims that got prior PRs closed. Wants tests first.
- **`Waveform` seek-bar keyboard access** — the scrubber is a `div` with `onClick` and no keyboard path. Fixing it adds capability, which is a behavior change; flagging rather than doing.
- **`AccessCodeInput` paste handling** — `onPaste` is bound only to the first input, so pasting into any other box does nothing. That's a bug, not a refactor.

## Open PR overlap check

Considered #481 (`lib/canvas/osml*`, `lib/project/serialize`), #480 (`lib/generation/**`, `lib/connectors/**`, `app/components/canvas/elements/**`, `lib/video/aspectRatio`), and #438 (`app/components/canvas/Canvas.tsx`, `plugins/withCollapsedVoids`, `hooks/useEditorSetup.ts`, `lib/canvas/types.ts`). No file here appears in any of them. `useEditorSession` imports `useEditorSetup` but does not modify it.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -42,6 +42,7 @@ Every route that takes request data is built from a factory in `lib/api/route-ha
 - `lib/project/store.ts` holds per-project metadata in Zustand. Every mutation rule lives in a store action. Components subscribe to reactive state through `useProject(selector)`; `getProjectStore(projectId)` is the escape hatch for imperative access that a render-time selector can't express — non-React callers (connector plugins, queue workers), plus hooks doing one-shot init, `.subscribe`, or reads outside render (`ProjectEditor`, `useAutosave`, `useGenerateAll`).
 - `lib/canvas/elementConnector.ts` answers "which connector, provider and model does this element use?" for both the UI and the queue. An element pins its provider when created, so this is also where a pin that no longer resolves falls back to the registry default.
 - `lib/script/` provides script context and refinement utilities.
+- `app/components/canvas/hooks/useEditorSession.ts` is the one place the Slate editor gets wired to a project: initial hydration, streaming script sync, metadata sync, and autosave. Views call it and render the editor; they never assemble it.
 
 ## Data
 
--- a/app/components/PostPromptView.tsx
+++ b/app/components/PostPromptView.tsx
@@ -1,22 +1,12 @@
 "use client";
 
-import { useMemo } from "react";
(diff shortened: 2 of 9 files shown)


umairnadeem (commented):
lgtm

PR #483, opened 2026-07-28
refactor: derive seek thumbnails from the element registry

umairnadeem:
Nightly CONVENTIONS.md audit. One violation found and fixed; the rest of the sweep came back clean.

## The violation

`buildSceneSegments` decided a scene thumbnail's media kind with a hardcoded type check:

```ts
thumbnail: el ? { url: el.url, kind: el.type === "image" ? "image" : "video" } : null
```

That is special-case wiring for a fact the registry already owns. `ELEMENT_TYPES[type].outputKind` declares the output kind for every element type, and every other preview path (`OutputPreview`, `ForegroundPreview`, `resolveElements`) reads it. The sniffing version also encodes "image is the only image-output type", so a new image-output element type would silently get thumbnailed as a `<video>`.

Now derived from the registry via a small named helper. No behavior change for today's element types: `image` → image, `animated_image`/`clip` → video.

## Checks

`npm run lint`, `npm run typecheck`, `npm run format:check`, `npm run test:run` (978 tests) all pass.

### DIFF
--- a/app/components/video/useSceneSegments.ts
+++ b/app/components/video/useSceneSegments.ts
@@ -1,8 +1,8 @@
 import type { PlayerRef } from "@remotion/player";
 import { isForeground } from "@/lib/canvas/guards";
-import type { SceneElement } from "@/lib/canvas/types";
+import { ELEMENT_TYPES, type SceneElement } from "@/lib/canvas/types";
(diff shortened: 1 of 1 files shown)


umairnadeem (commented):
lgtm

Issue #332, opened 2026-06-21
In-app storyboard/timeline (read-only) showing what plays when and for how long


umairnadeem:
## Problem

There's no at-a-glance view of **what plays when and for how long**. Users can scrub via the seek bar, but nothing lays the project out as a timeline of elements over time — so it's hard to see the structure of a video (which scene/element is on screen, its start and duration, how scenes are paced) without playing through it.

The data for this already exists: `useSceneSegments` computes per-scene `{ start, duration, label, thumbnail }` from the resolved `VideoLayout`, and `SegmentedSeekBar` already renders scenes as segments. What's missing is a dedicated **storyboard/timeline strip** that surfaces this layout as a first-class panel.

## Proposed direction

Add a **storyboard/timeline at the bottom** that shows each element/scene as a block positioned by its start time and sized by its duration.

- **Read-only to start** — a horizontal timeline strip showing scenes (and ideally their constituent elements: foreground image/video, narration, music, SFX) laid out on a time axis, with labels, durations, and scene thumbnails.
- Reuse the existing segment/layout data (`useSceneSegments`, `VideoLayout`) rather than recomputing timing.
- Sync with the player position so the timeline reflects the current playhead (and could later support click-to-seek).
- Editable timing (drag to retime/reorder) is explicitly **out of scope** for the first version; this is a visualization, not an editor.

## Relevant code

- `app/components/video/useSceneSegments.ts` — per-scene `start`/`duration`/`label`/`thumbnail` (the timeline's data source)
- `app/components/video/SegmentedSeekBar.tsx`, `ScrubBar.tsx` — existing per-scene segment rendering to draw from
- `app/components/video/VideoLayoutContext.tsx`, `useVideoLayout.ts`, `lib/video/types.ts` — `VideoLayout` / `Sequence` timing model
- `app/components/video/PlayerPositionContext.tsx`, `useActiveSceneSync.ts` — playhead sync
- `app/components/video/BottomTransportBar.tsx` — where a bottom timeline panel would live

## Relationship to other issues

- **#256 (Export & storyboarding)** — that issue is about *exporting* for external editing (download assets/script with timestamps, NLE/XML). This issue is an *in-app* read-only timeline view. Complementary; the same scene-segment timing feeds both.

## Suggested direction

- Build a read-only timeline panel in the bottom transport area driven by `useSceneSegments` + `VideoLayout`, synced to the playhead.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #254, opened 2026-06-12
Audio mixing defaults & controls (rebalance volumes, preview volume, narration mute)


umairnadeem:
## Problem

The default audio mix is off and some controls don't behave as expected:

- **SFX volume should default to 2** (currently much higher).
- **Narration volume is too low** and **character volume is too high** by default.
- The **volume control should affect preview** playback.
- Users want a **narration mute** (per-track).

## Reported by (verbatim feedback)

- "Sfx vol should default to 2"
- "Narration volume is too low"
- "Character volume too high"
- "Volume control should affect preview"
- "Add narration mute"

## Relevant code

- `app/components/canvas/config/elementConfigs.tsx` — narration/character default `volume: "10"`
- `lib/video/elementAttributes.ts` — `getVolume` (0–10 scale)
- `app/components/video/PlayerControls.tsx` — global mute/volume
- `remotion/compositions/VideoComposition.tsx` — gain application

## Suggested direction

- Rebalance default volumes: SFX → 2, narration up, character down.
- Ensure per-element volume changes affect preview playback.
- Add a per-track narration mute.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

umairnadeem:
Additional verbatim feedback from a new user's first day:

- "All I missed was some power to finetune sound and subtitles."

The sound half lands here; the subtitles half is now tracked in #270.

Issue #250, opened 2026-06-12
Target video length control (length presets, lift ~4min cap)


umairnadeem:
## Problem

Stories are capped at ~4 minutes, and there's no way to choose a target length. Users want to select a target length in the composer (e.g. <30s, <1m, 1–3m, 3–5m, 5–10m, 10–15m); several specifically want 2–3 minute videos. Some also expect a small set of discrete length options (e.g. 2 / 5 / 10 / 12 min) rather than ranges.

## Reported by (verbatim feedback)

- "Stories capped at certain length (~4mins). Would like to ideally select lengths in the composer hero slop copilot prompt (e.g. < 30 Seconds, < 1 Minute, 1-3 Minutes, 3-5 Minutes, 5-10 Minutes, 10-15 Minutes)"
- "Target 2-3 minutes"
- "Want to control video length"
- "Maybe an option to determine the length of a video, for example 2 mins, 5 mins, 10 mins, or 12 mins?"

## Relevant code

- `lib/script/ScriptProvider.tsx` / mode plugins — script length prompt
- `lib/video/types.ts` — `totalDurationSec`

## Suggested direction

- Expose length presets in the composer and feed the target into script generation.
- Lift the hard ~4 min cap.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #484, opened 2026-07-28
fix: stop the scene header jumping when a scene is highlighted

### LINKED ISSUE #474: Scene header shifts up when a scene is highlighted in expanded (non-compact) element view

umairnadeem:
## Summary

Highlighting a scene shifted its header up ~13px in expanded (non-compact) view. The active-scene card (`.scene-active`) was painted on the same wrapper that carries the inter-scene separator and spacing (`border-t pt-3 mt-3`), so it zeroed `padding-top` and `border-width` to keep the card box uniform across scenes. For every scene but the first, that removed 12px of padding plus the 1px border from the layout, moving the content up whenever the highlight toggled.

Now the card is painted on the scene's content box (the sortable inner element) and the spacing wrapper is left untouched, so the highlight only changes color and shadow. The separator above an active scene still disappears, via `border-transparent` instead of `border-width: 0`, so the appearance is unchanged in both states.

Issue #474 is a self-contained layout bug with an obvious cause in CSS, no product or design decision involved, and no auth/secrets needed. The change is three files and does not alter any visual value from `DESIGN.md`.

## Changes

- `app/globals.css`: drop `padding-top: 0` / `border-width: 0` from `.scene-active`.
- `app/components/canvas/dnd/SortableScene.tsx`: pass the highlight class as `contentClassName`; hide the separator with `border-transparent` when active.
- `app/components/canvas/dnd/SortableItem.tsx`: optional `contentClassName` on the sortable content element.

## Validation

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:run` — 978 tests, 113 files, all pass
- `npm run test:e2e` — **skipped**: port 3000 was already occupied by a running dev server on the machine, so Playwright could not start its web server. The change touches no routes or copy that the smoke tests assert.

No test added: this is pure layout, which jsdom cannot assert, and asserting on the class strings would be testing internals.

## Open PR overlap check

Open PRs #483, #482, #481, #480, #438 were inspected with `gh pr diff --name-only`. None touches `app/globals.css`, `SortableItem.tsx`, or `SortableScene.tsx`. #438 (dead caret) is the closest in theme but works in `Canvas.tsx` / `CompactElement.tsx` / `withCollapsedVoids.ts`.

## Skipped candidates

- #424 (dead caret on element label) — already covered by open PR #438.
- #259 (queue concurrency, size XS) — blocked on BYOK.
- #332 / #250 / #254 (size S, good first issue) — each needs product/design decisions (new read-only surface, length presets, mixing defaults).
- Everything M/L/XL — too large for one unattended run.

Closes #474

### DIFF
--- a/app/components/canvas/dnd/SortableItem.tsx
+++ b/app/components/canvas/dnd/SortableItem.tsx
@@ -12,6 +12,7 @@ interface SortableItemProps {
 	sortableType: "scene" | "content";
 	wrapperClassName?: string;
 	wrapperStyle?: React.CSSProperties;
+	contentClassName?: string;
 	insertOptions?: InsertOption<CanvasElementType>[];
 	onInsert?: (type: CanvasElementType) => void;
 	disabled?: boolean;
@@ -25,6 +26,7 @@ export function SortableItem({
 	sortableType,
 	wrapperClassName,
 	wrapperStyle,
+	contentClassName,
 	insertOptions,
 	onInsert,
 	disabled,
@@ -51,7 +53,11 @@ export function SortableItem({
 	return (
 		<div {...attributes} className={wrapperClassName} style={wrapperStyle}>
 			<div
-				className={styles.sortable}
+				className={
+					contentClassName
+						? `${styles.sortable} ${contentClassName}`
+						: styles.sortable
+				}
 				{...sortableAttributes}
 				ref={setNodeRef}
 				style={{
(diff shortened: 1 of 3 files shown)


umairnadeem (commented):
lgtm

PR #485, opened 2026-07-29
refactor: nightly maintainability 2026-07-29

umairnadeem:
No behavior changes. Two seams moved out of React and into `lib/`, where they can be tested.

## Architecture

- **`lib/project/autosave.ts` (new)** — saving a project is now a domain module, not a hook. It builds the row (`buildProjectSave`), debounces edits, and serializes writes through a one-at-a-time queue so a slow save can't land after a newer one. `useAutosave` shrank to what it should be: subscribe the editor value, the store, and the generation queue; turn the result into a toast. The hook no longer imports the store snapshot, project name, thumbnail, and save-API modules.
- **`lib/auth/accessCode.ts` (new)** — the access-code box transitions (`typeChar`, `eraseBefore`, `pasteCode`) are pure functions returning the next values plus where focus goes. `AccessCodeInput`'s three handlers each did their own array copy, focus math, and completeness check; they now share one `apply(entry)`.

## Maintainability

- `useAutosave` loses the `debouncedRef` indirection, the `save` callback, the `schedule` callback, and the effect that rebuilt the debouncer — the debounce now lives with the thing it debounces. The editor value reaches the save through `setScriptSource`, a thunk called only when a save actually runs, so serializing stays off the per-keystroke path.
- A save with no script source now fails loudly (console + error toast) instead of writing an empty script.
- New coverage where there was none: 8 cases for the autosaver (debounce coalescing, flush, unhydrated skip, missing source, failed save) and 12 for the code-entry transitions (uppercasing, rejected characters, backspace-into-previous, partial and over-length paste).
- One line in ARCHITECTURE.md's "Editor state" section for the new save seam.

## Complexity delta

- `useAutosave.ts` 96 → 57 lines; 3 refs → 1, and the save payload/transport left `app/` entirely.
- `AccessCodeInput.tsx` 138 → 118 lines; three duplicated copy-focus-submit blocks → one.
- Net +451/−102 across 7 files; 248 of the additions are the new tests.

## Skipped

- **Paste is still bound only to the first code box.** `onPaste={i === 0 ? ... : undefined}` means pasting into box 3 does nothing. `pasteCode` is now index-independent so the fix is a one-line binding change, but it's a behavior change, not a refactor.
- **`queue.subscribe` result-version diffing** still sits in `useAutosave`. It belongs behind a `subscribeToResults` on the queue itself, but `lib/generation/queue.ts` is heavily rewritten by #480.
- **`Waveform` seek-bar keyboard access** — still a `div` with `onClick`. Fixing it adds capability.

## Open PR overlap check

Considered #484 (`canvas/dnd/*`, `globals.css`), #481 (`lib/canvas/osml*`, `lib/project/__tests__/serialize.test.ts`), #480 (`lib/generation/**`, `lib/connectors/**`, `lib/project/{thumbnail,types,deleteCharacter,ensureCharacterAvatars,characterAvatar}.ts`, `canvas/elements/**`, `ProjectEditor.tsx`), and #438 (`canvas/Canvas.tsx`, `plugins/withCollapsedVoids`, `hooks/useEditorSetup.ts`). No file here appears in any of them. `lib/project/autosave.ts` calls `pickThumbnailUrl` but does not modify `thumbnail.ts`.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -43,6 +43,7 @@ Every route that takes request data is built from a factory in `lib/api/route-ha
 - `lib/canvas/elementConnector.ts` answers "which connector, provider and model does this element use?" for both the UI and the queue. An element pins its provider when created, so this is also where a pin that no longer resolves falls back to the registry default.
 - `lib/script/` provides script context and refinement utilities.
 - `app/components/canvas/hooks/useEditorSession.ts` is the one place the Slate editor gets wired to a project: initial hydration, streaming script sync, metadata sync, and autosave. Views call it and render the editor; they never assemble it.
+- `lib/project/autosave.ts` owns saving: it builds the row from the store snapshot, script and generation snapshot, then debounces and serializes writes so a slow save can't land after a newer one. `useAutosave` only subscribes the editor value, the store, and the queue to it, and turns the result into a toast.
 
 ## Data
 
--- a/app/components/AccessCodeInput.tsx
+++ b/app/components/AccessCodeInput.tsx
@@ -4,20 +4,25 @@ import { useRef, useState, useCallback } from "react";
 import { useRouter } from "next/navigation";
 import { apiJson } from "@/lib/clients/http";
 import { errorMessage } from "@/lib/errors";
-
-const CODE_LENGTH = 6;
-const EMPTY_CODE = () => Array<string>(CODE_LENGTH).fill("");
+import {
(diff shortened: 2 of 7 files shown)


umairnadeem (commented):
lgtm. walked the autosave effect ordering (setScriptSource before the flush cleanup, so an unmount still serializes the latest value) and the accessCode transitions match the old inline handlers case for case.

PR #486, opened 2026-07-29
perf: nightly performance optimizations 2026-07-29

umairnadeem:
## What

`findElementById` / `findNodeById` resolved a canvas element by scanning the whole document with `Editor.nodes(editor, { at: [] })`. That generator visits every node, text leaves included. Both functions sit on paths that fire continuously:

- `useScriptSync` resolves up to three ids **per streamed token** while a script generates.
- dnd-kit's `handleDragOver` resolves one **per pointer frame** during a drag, and `moveDraggedElement` two more on drop.

The per-lookup cost grows linearly with script length while the event rate stays constant, so the total work during a long generation is quadratic.

Canvas elements only live at two depths — scenes are the editor's children, content elements are theirs — so walking those levels directly returns the same entry without touching text nodes.

## Measured

Same lookup, same documents, 2000 iterations each (worst-case id, last element):

| document | before | after | |
|---|---|---|---|
| 40 elements | 0.140 ms | 0.023 ms | 6.0x |
| 120 elements | 0.307 ms | 0.050 ms | 6.1x |
| 240 elements | 0.527 ms | 0.088 ms | 6.0x |

At 240 elements that's ~1.6 ms → ~0.26 ms of lookup per streamed token, and ~0.53 ms → ~0.09 ms per drag-over frame against a 16 ms budget.

No behavior change: both finders return the same entry and path. Added a test pinning the top-level-content case (`useScriptSync` inserts there before normalization).

## Checks

`npm run lint`, `prettier --check`, `npm run typecheck`, `npm run test:run` (981 passed), `npm run build` — all green.

## Open PR overlap check

Considered open PRs #485, #484, #481, #480, #438. This PR touches only `lib/canvas/editorOps.ts` and its test. #481 covers `lib/canvas/osml*`/`parseXmlTag`/`xmlEscape`; #438 covers `lib/canvas/types.ts`, `Canvas.tsx`, `useEditorSetup.ts`, `withCollapsedVoids.ts`. No file or theme overlap — callers in those PRs' files are unmodified and pick up the speedup for free.

### DIFF
--- a/lib/canvas/__tests__/editorOps.test.ts
+++ b/lib/canvas/__tests__/editorOps.test.ts
@@ -53,6 +53,15 @@ describe("findNodeById", () => {
 		const editor = makeEditor([scene([content("narration", "n1")], "s1")]);
 		expect(findNodeById(editor, "s1")).toBeNull();
 	});
+
+	it("finds a content element sitting at the top level", () => {
+		const editor = createEditor();
+		editor.children = [
+			scene([content("narration", "n1")]),
+			content("image", "img1"),
+		];
+		expect(findNodeById(editor, "img1")?.[1]).toEqual([1]);
+	});
 });
 
 describe("findElementById", () => {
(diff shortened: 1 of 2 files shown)


umairnadeem (commented):
lgtm. two-depth invariant holds afaict (scenes are editor children, content elements are theirs), and the top-level-content test pins the useScriptSync pre-normalization case. numbers speak for themselves.

umairnadeem:
negligible real-world perf difference

PR #487, opened 2026-07-29
refactor: rethink openslop architecture 2026-07-29

umairnadeem:
No behavior changes. Two presentation-layer seams stop being special-case wiring.

## App understanding

openslop turns a prompt into a rendered video. `ScriptProvider` streams OSML from an LLM connector; `useOSMLStreamParser` turns tokens into nodes; `useScriptSync`/`useMetadataSync` push those into a Slate editor and the project store. Each canvas element is generated through `lib/generation/queue.ts` + `lib/connectors/**` + `lib/providers/**`, resolved into a `VideoLayout` by `lib/video/scene-builder.ts`, previewed in a Remotion `<Player>`, and exported via `/api/render` to Remotion Lambda.

## From-scratch architecture

The current shape is already close to the one I'd rebuild: `lib/` is the domain layer, `app/` is presentation, element types are declared once in `ELEMENT_TYPES` and dressed once in `elementConfigs`, and providers compose at the editor boundary. The two places it drifted were both in presentation: one element facet still dispatched with conditionals instead of a table, and one client-only lookup index had leaked into the type that gets serialized to Lambda.

**Element previews are a registry, not an if-chain.** `OutputPreview` branched on `outputKind === "audio"`, then `type === "animated_image"`, then fell through to media. Element type is already the key for connector/role/layer (`ELEMENT_TYPES`) and for label/icon/placeholder (`elementConfigs`); preview was the one facet handled imperatively. New `preview/registry.tsx` maps each of the 7 types to a preview component behind one prop contract (`ElementPreviewProps` in `preview/status.ts`), and each preview derives the url it needs from the result itself. `OutputPreview` is now a lookup plus a render — 46 → 25 lines, zero conditionals. Adding an element type is a registry entry instead of a new branch.

**`sequenceByElementId` leaves the render payload.** `buildVideoLayout` attached a `Map<elementId, Sequence>` to `VideoLayout`. `VideoLayout` is what goes into Remotion `inputProps` for `/api/render` and Lambda, where a `Map` JSON-serializes to `{}` — harmless only because `VideoComposition` never reads it. Its sole consumers are client-side (`SceneContainer` header timestamp, `PlayFromHereButton`, `buildSceneSegments`). The index now lives in `VideoLayoutContext`, memoized on `layout.series`, behind a `useSceneSequence(scene)` hook. `VideoLayout` is render-only again, and the two call sites lost their `useLayout()` + `findSceneSequence(scene, layout)` boilerplate.

Kept as a memoized map rather than scanning `series` per lookup: `SceneHeader` and `PlayFromHereButton` render once per scene, so a linear scan would make the canvas quadratic in scene count.

**Hygiene:** `useProjectRehydrate.test.ts` was the only test outside a `__tests__/` dir (113 others are inside); moved.

New coverage for `buildSequenceIndex` (keys by element id, skips placeholder sequences).

## Validation

`npm run lint`, `npx prettier --check .`, `npm run typecheck`, `npm run knip`, `npm run test:run` (981 passed, 114 files), `npm run build` — all green.

## Skipped

- **Duplicated abort-controller streaming lifecycle** (`lib/script/ScriptProvider.tsx`, `app/components/canvas/hooks/useRefineScript.ts`). Same abortRef/loading/`abortRef.current === controller` dance twice. Only 2 sites, and a generic async hook here has been rejected before as overengineering — left for a third caller.
- **`ForegroundPreview` reads generation state via `useGenerate` directly** while every other consumer goes through `ElementGenerationProvider`. It renders outside that provider (collapsed scene view), so unifying it needs queue-level work that overlaps #480.
- **`parentSceneId()` helper** for the duplicated `Node.parent(editor, path) as SceneElement` cast — one of the two sites is `ElementContainer.tsx`, under #480.
- **`lib/api/with-auth.ts` returns `stringifyError(error)` on 500s**, including on public routes. Information disclosure worth a look, but fixing it is a behavior change.

## Open PR overlap check

Considered #486 (`lib/canvas/editorOps.ts`), #485 (`ARCHITECTURE.md`, `AccessCodeInput`, `useAutosave`, `lib/auth/accessCode`, `lib/project/autosave`), #484 (`canvas/dnd/Sortable*`, `globals.css`), #481 (`lib/canvas/osml*`, `parseXmlTag`, `xmlEscape`), #480 (`lib/generation/**`, `lib/connectors/**`, `canvas/elements/{AssetTiles,CharacterPill,ElementContainer,ElementUploadButton}`, `ProjectEditor.tsx`, `lib/video/{aspectRatio,useAspectRatio}` + `resolve.test.ts`), #438 (`canvas/Canvas.tsx`, `CompactElement.tsx`, `withCollapsedVoids`, `useEditorSetup.ts`, `lib/canvas/types.ts`).

None of the 16 files here appear in any of them. `ElementContainer.tsx` (#480) renders `OutputPreview` but its props are unchanged; `CompactElement.tsx` (#438) renders `ForegroundPreview`, which is untouched; `lib/canvas/types.ts` (#438) is read, not modified.

### DIFF
--- a/app/components/canvas/elements/OutputPreview.tsx
+++ b/app/components/canvas/elements/OutputPreview.tsx
@@ -1,8 +1,6 @@
 import { memo } from "react";
-import { ELEMENT_TYPES, type CanvasContentElement } from "@/lib/canvas/types";
-import { getPrimaryUrl } from "@/lib/connectors/assetUrl";
-import { AudioResult, AudioPlaceholder, MediaResult } from "./preview/results";
-import { AnimatedImageOutput } from "./preview/AnimatedImagePreview";
+import type { CanvasContentElement } from "@/lib/canvas/types";
+import { ELEMENT_PREVIEWS } from "./preview/registry";
 import { useElementGeneration } from "./ElementGenerationContext";
 
 function OutputPreviewComponent({
@@ -10,29 +8,16 @@ function OutputPreviewComponent({
 }: {
 	element: CanvasContentElement;
 }) {
-	const type = element.type;
-	const { outputKind } = ELEMENT_TYPES[type];
 	const { status, seconds, result, error, discard } = useElementGeneration();
-	const state = { status, seconds, error, onDiscard: discard };
-
-	if (outputKind === "audio") {
-		if (result?.audioUrl) {
-			return (
-				<AudioResult src={result.audioUrl} status={status} seconds={seconds} />
-			);
-		}
-		return <AudioPlaceholder {...state} />;
-	}
-
-	if (type === "animated_image") {
-		return <AnimatedImageOutput {...state} />;
-	}
+	const Preview = ELEMENT_PREVIEWS[element.type];
 
 	return (
-		<MediaResult
-			{...state}
-			url={getPrimaryUrl(result, outputKind)}
-			outputKind={outputKind}
+		<Preview
+			status={status}
+			seconds={seconds}
+			result={result}
+			error={error}
+			onDiscard={discard}
 		/>
 	);
 }
--- a/app/components/canvas/elements/PlayFromHereButton.tsx
+++ b/app/components/canvas/elements/PlayFromHereButton.tsx
@@ -3,15 +3,17 @@
 import { Play } from "@/components/ui/icon";
 import { TooltipIconButton } from "@/components/ui/icon-button";
 import { usePlayerControl } from "@/app/components/video/PlayerControlContext";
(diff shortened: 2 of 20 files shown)


umairnadeem (commented):
lgtm. checked every consumer of `sequenceByElementId` / `findSceneSequence` moved over, and the registry covers all 7 element types with the same url derivation as the old if-chain. nice to have the Map out of the Lambda inputProps.

Issue #316, opened 2026-06-19
Crossfade looping SFX to remove the audible gap between loops


umairnadeem:
## Problem

Looping sound effects currently restart with a hard cut at the loop boundary. There's an audible gap/click between the end of one iteration and the start of the next, which breaks the illusion of a continuous sound.

## Proposed solution

Crossfade looping SFX across the loop boundary so the tail of one iteration fades out while the head of the next fades in. This smooths the seam and removes the perceptible gap.

- Fade the end of the loop out and the start of the next iteration in over a short overlap window.
- Make the crossfade duration configurable (with a sensible default).
- Apply only to SFX flagged as looping.

## Acceptance criteria

- [ ] No audible gap or click when a looping SFX repeats.
- [ ] Crossfade is applied only to looping SFX, not one-shots.
- [ ] Crossfade duration is configurable with a reasonable default.

PR #488, opened 2026-07-29
fix: bound the composer textarea height so long scripts scroll internally

### LINKED ISSUE #359: Composer: hard to scroll between the prompt textbox and the character/reference area when pasting a long script
## Problem

In the composer, the character/reference assets row (`ComposerAssets`) sits at the top of the input card, and the prompt `textarea` sits below it. The textarea uses `fieldSizing: "content"` with `resize-none` and no max-height (`app/components/copilot/ComposerCopilot.tsx:214-226`), so it grows to fit its full content. When a user pastes a long script, the textarea expands to the entire script height, the whole card balloons, and the character/reference area gets pushed far from the input and submit controls. There is no internal scroll region and nothing sticky, so moving between the script body and the place where you store characters/references means scrolling the whole page a long way.

- "When pasting a long script, via the UI make it easier to scroll between the text box and the top space where you store characters / references."

## Relevant code

- `app/components/copilot/ComposerCopilot.tsx:200-234`: the input card - `ComposerAssets` (top), then the grid-stacked `textarea` (`rows={2}`, `fieldSizing: "content"`, `resize-none`, no max-height) and placeholder overlay, then the controls row.
- `ComposerAssets` (same file): the character/reference chips row that scrolls out of reach.

## Proposed direction

Keep both the character/reference area and the input controls reachable when the script is long:

1. Cap the textarea height (e.g. `max-h`) and give it an internal `overflow-y-auto` so a long script scrolls **inside** the field instead of expanding the card unbounded.
2. And/or keep `ComposerAssets` (and the submit/attach controls) visually anchored (sticky within the card) so they stay reachable regardless of script length.

## Acceptance criteria

- [ ] Pasting a long script no longer pushes the character/reference area off-screen; it stays reachable without long page scrolling.
- [ ] The prompt body is scrollable within a bounded input area.
- [ ] Submit/attach controls remain visible while editing a long script.

## Related

- #340 (inline character removal in the hero composer), #299 (inline @-mention character chips) - same composer character/reference surface.

---

umairnadeem:
## Summary

Pasting a long script into the hero composer made the prompt `textarea` grow to the full script height (`fieldSizing: "content"` with no cap), ballooning the input card and pushing the character/reference row and the submit/attach controls far apart.

Caps the textarea at `max-h-[40vh]` with `overflow-y-auto`, so a long script scrolls inside the field instead of expanding the card unbounded. `ComposerAssets` above and the controls row below stay adjacent to the input at any script length, which covers all three acceptance criteria without adding sticky positioning.

`max-h-[40vh] overflow-y-auto` matches the existing bounded-scroll pattern already used in `AppToaster.tsx`.

One-line class change on a single component. No behavior, state, or layout structure changes; short scripts render exactly as before since the textarea only hits the cap once content exceeds 40vh.

## Validation

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:coverage` — 114 files, 980 tests, all pass
- `npm run test:e2e` — skipped: port 3000 was already in use by a running dev server, so Playwright's webServer could not start. The change is CSS-only and touches no route or copy asserted by the smoke suite.

## Open PR overlap check

Checked open PRs #487, #486, #485, #484, #481, #480, #438 via `gh pr diff --name-only`. None touches `app/components/copilot/`. No overlap.

## Skipped candidates

- #474 (scene header shift) and #424 (dead caret) — already covered by open PRs #484 and #438.
- #259 (queue concurrency) — explicitly gated behind BYOK, which has not shipped.
- #332 (storyboard timeline), #254 (audio mixing defaults/controls) — new product surface and volume-balance/product decisions respectively.
- #316 (SFX crossfade) — needs an audio-quality judgment call and can't be validated headlessly.
- Everything else open is size M or larger, or requires product/design direction.

Closes #359

### DIFF
--- a/app/components/copilot/ComposerCopilot.tsx
+++ b/app/components/copilot/ComposerCopilot.tsx
@@ -216,7 +216,7 @@ export default function ComposerCopilot({
 							}}
 							placeholder={placeholder}
 							style={{ fieldSizing: "content" }}
-							className=" w-full resize-none bg-transparent font-body text-body text-foreground caret-accent placeholder:text-muted-foreground outline-none focus-visible:ring-2 focus-visible:ring-ring/30 focus-visible:rounded-sm"
+							className=" max-h-[40vh] w-full resize-none overflow-y-auto bg-transparent font-body text-body text-foreground caret-accent placeholder:text-muted-foreground outline-none focus-visible:ring-2 focus-visible:ring-ring/30 focus-visible:rounded-sm"
 						/>
 						{!hasText && !isTemplateMode && placeholderOverlay && (
 							<div className="pointer-events-none overflow-hidden font-body text-body">


umairnadeem (commented):
lgtm, one-line cap and it matches the AppToaster bounded-scroll pattern. ship it

PR #493, opened 2026-07-30
fix: nightly bug fixes 2026-07-30

umairnadeem:
## What

Two correctness fixes in the TTS provider layer, plus tests.

### Cartesia clip duration came from word timestamps, not the audio

`CartesiaTTS._generate` reported `durationSec` as `lastWordTimestamp.end + 1`. Two problems:

- The word timestamps stop at the last word, so the padded value drifts from the real audio length. That duration is what `scene-builder` advances the timeline cursor by (`cursor += element.durationSec`), so the narration's slot disagreed with its own audio.
- When the model emits no timestamp frames, `metadata` was `undefined` entirely — and `BaseAssetConnector.resolveBundle` coerces a missing `durationSec` to `0`, so the narration gets a zero-frame sequence and silently disappears from the render.

Now the duration is computed from the raw PCM byte length (`pcmDurationSec`), which is exact and always available. The WAV byte-rate math that `wrapPcmInWav` already did is hoisted into module constants so the header and the duration can't drift apart.

An audio stream that yields zero chunks now throws instead of storing a zero-length asset that would vanish from the timeline with no error.

### `fetchAllowedVoicePreview` let callers undo its own redirect guard

`fetch(url, { redirect: "manual", ...init })` spread the caller's init *after* `redirect`, so an `init.redirect` would override the setting the function's doc comment promises (no replaying the `Authorization` header across a cross-origin redirect). Swapped the spread order. No current caller passes `redirect`, so behavior is unchanged today.

## Tests

- `pcmDurationSec` unit conversion.
- Duration derived from audio length rather than the last word timestamp.
- Duration still reported when the model returns no timestamps.
- Empty audio stream throws, and the websocket is still closed.
- New `voicePreview.test.ts`: the origin allow-list previously had no direct tests. Covers plaintext HTTP, a different host, a subdomain, a non-default port, and the allowed host smuggled into userinfo — plus that caller init is forwarded while `redirect: "manual"` holds.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, and the full vitest suite (992 tests) all pass.

## Open PR overlap check

Considered open PRs #488, #487, #486, #485, #484, #481, #480, #438. Their file/theme coverage: the composer copilot, canvas element previews / scene-builder / Remotion root, `lib/canvas/editorOps`, access codes + autosave, canvas DnD, the OSML serializer/parser, `lib/generation` + `lib/connectors`, and the collapsed-void plugin. This PR only touches `lib/providers/tts/*` and its tests, which no open PR touches — non-overlapping.

### DIFF
--- a/lib/providers/__tests__/cartesia-tts.test.ts
+++ b/lib/providers/__tests__/cartesia-tts.test.ts
@@ -49,7 +49,18 @@ vi.mock("@ai-sdk/openai", () => ({
 
 import type Cartesia from "@cartesia/cartesia-js";
 import { TTSEmotion } from "@/lib/connectors/tts/enums";
-import { CartesiaTTS, collectVoices } from "../tts/cartesia";
+import { CartesiaTTS, collectVoices, pcmDurationSec } from "../tts/cartesia";
+
+/** pcm_f32le @ 44.1kHz mono: 4 bytes per sample. */
+const PCM_BYTES_PER_SEC = 44100 * 4;
+const pcmChunk = (seconds: number) =>
+	Buffer.alloc(seconds * PCM_BYTES_PER_SEC, 1);
+
+const streamOf = (responses: object[]) => ({
+	[Symbol.asyncIterator]: async function* () {
+		for (const r of responses) yield r;
+	},
+});
 
 const makeVoices = (count: number, offset = 0) =>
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
lgtm

PR #494, opened 2026-07-30
refactor: nightly maintainability 2026-07-30

umairnadeem:
No behavior changes. One seam moved out of the UI layer, plus the tests that make it checkable.

## Architecture

`useRendering` owned the whole Lambda export protocol: POST `/api/render`, then an open-ended poll loop against `/api/render/progress`, with the three terminal outcomes each writing React state from a different place. That is domain logic living in a `"use client"` hook — untestable in practice (nothing exercised it), invisible to coverage (`vitest.config.ts` only instruments `lib/**` and `app/api/**`), and sitting one directory away from `lib/video/render-api.ts`, which already owns the wire contract for exactly these two routes.

- **`lib/video/render-client.ts` (new)** — `runRender(layout, scale)` is an async generator that submits the render and yields `{status: "rendering", progress}` until it yields `{status: "done", url, size}`. An upstream render failure throws instead of returning a fourth shape, so callers have one error path rather than two (`result.type === "error"` and the `catch`) that produced the same state.
- **`useRendering`** is now state plumbing: set `invoking`, drain the generator into `setState`, map any throw through `errorMessage`. 62 → 30 lines, and the client half of the render contract now sits next to the schema half.

The `RenderHandle` (`renderId`/`bucketName`) that the `rendering` state carried is gone. Nothing read it — not `ExportButton`, not `RenderProvider` — and it made the state union wider than the UI it feeds.

## Maintainability

- `lib/video/__tests__/render-client.test.ts` (new) — 4 tests over a previously-uncovered path: the submit → poll → done sequence and its request payloads, that polling stops on `done`, that an upstream `error` surfaces its own message, and that a failed submit propagates. Fake timers keep the 5s poll interval out of the runtime.
- Moving the module under `lib/` also brings it under the coverage gate, so it can't silently rot back to untested.

## Complexity delta

- `useRendering.ts` 62 → 30 lines; the `render` callback drops from 4 nesting levels and 3 terminal branches to a single `for await` plus one `catch`.
- `RenderState` loses one intersection member.
- Net +131/−35 across 3 files, 96 of which is the new module and its tests.

## Skipped

- **`runRender` has no cancellation.** The poll loop runs until the render terminates; unmounting or firing a second export leaves the first loop polling and calling `setState`. That is a real defect, but fixing it changes behavior and belongs in a bug PR, not a no-behavior-change refactor.
- **`RenderProvider`'s toast effect.** The obvious cleanup — one shared options object for all four branches — would give the error toast `duration: Infinity`, which it does not have today. Not worth a behavior change for four saved lines.
- **`app/projects/[id]/loading.tsx`** re-implements the entire `PostPromptView` chrome (rail widths, card shell, transport bar) as skeletons. Genuine duplicated knowledge that will drift, but deduplicating a skeleton against a live component is not a mechanical change; flagging for a human call.
- **`useAutosave`** (6 effects, no tests) — already claimed by open PR #485.

## Open PR overlap check

Considered #493 (`lib/providers/tts`), #488 (`ComposerCopilot`), #487 (`lib/video/scene-builder|types`, `remotion/`, `app/components/video/VideoLayoutContext|useSceneSegments`), #486 (`lib/canvas/editorOps`), #485 (`lib/auth/accessCode`, `lib/project/autosave`, `useAutosave`, `ARCHITECTURE.md`), #484 (`canvas/dnd`, `globals.css`), #481 (`lib/canvas/osml*`), #480 (`lib/generation/**`, `lib/connectors/**`, `canvas/elements/**`), #438 (`withCollapsedVoids`).

This PR touches `lib/video/render-client.ts`, `lib/video/__tests__/render-client.test.ts`, and `app/components/video/useRendering.ts` — none of which appear in any open PR. #487 is the nearest neighbour (same two directories) but works on the Remotion layout contract, not the render API client; no shared file.

## Checks

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and `test:coverage` (984 tests) all pass.

### DIFF
--- a/app/components/video/useRendering.ts
+++ b/app/components/video/useRendering.ts
@@ -1,21 +1,16 @@
 "use client";
 
 import { useCallback, useState } from "react";
-import { apiJson } from "@/lib/clients/http";
 import { errorMessage } from "@/lib/errors";
-import type { RenderHandle, RenderProgress } from "@/lib/video/render-api";
+import { runRender } from "@/lib/video/render-client";
 import type { VideoLayout } from "@/lib/video/types";
 
 type RenderState =
 	| { status: "idle" }
 	| { status: "invoking" }
-	| ({ status: "rendering"; progress: number } & RenderHandle)
+	| { status: "rendering"; progress: number }
 	| { status: "done"; url: string; size: number }
 	| { status: "error"; message: string };
-
-const POLL_INTERVAL_MS = 5000;
-
-const wait = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
 
 export function useRendering() {
 	const [state, setState] = useState<RenderState>({ status: "idle" });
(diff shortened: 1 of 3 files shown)


umairnadeem (commented):
lgtm

PR #496, opened 2026-07-30
fix: surface Supabase auth failures instead of swallowing them

umairnadeem:
Nightly CONVENTIONS.md compliance audit. One hard-rule violation was safe to fix.

## The violation

`lib/auth/session.ts` destructured nothing from the two Supabase calls that can fail:

```ts
export async function signInWithGoogle(): Promise<void> {
	await createClient().auth.signInWithOAuth({ ... });   // { error } dropped
}

export async function signOut(): Promise<void> {
	await createClient().auth.signOut();                  // { error } dropped
}
```

Hard rule: *"Never swallow an error on a critical path to keep going. Propagate it or surface it."* `sendMagicLink`, in the same 39-line module, already returns a typed `{ error? }` — so this was also two ways to do one thing in one file.

Observable consequences:
- A failed Google sign-in left the user clicking a button that did nothing.
- A failed sign-out still ran `router.refresh()`, rendering the app as if the session had been cleared when it hadn't.

## The fix

Both now throw (an infrastructure failure is unexpected, so it throws rather than returning a typed result — that form stays reserved for expected failures like `sendMagicLink`'s validation errors).

The callers own the policy:
- `AuthForm` routes it into the inline error slot it already renders for the magic-link path.
- `UserProfile` toasts and skips the refresh.

Tests cover both new error paths.

## Also audited, not changed

Everything else found was either behavior-changing or under an open PR, and is recorded rather than fixed:
- `lib/api/with-auth.ts:18` — 500s return `stringifyError(error)` (stack included) to the client on public routes.
- `lib/components/Waveform.tsx:111` — peaks decode failure is logged, then `.finally` marks loaded, so the waveform renders bar-less with no user-visible error.
- `GlobalErrorToaster` / `ToastErrorBoundary` toast raw `stringifyError` JSON while every other call site uses `errorMessage`.
- Duplicated abort-controller streaming lifecycle (`ScriptProvider`, `useRefineScript`) and the `(Node.parent(...) as SceneElement).id` cast (`SortableContent`, `ElementContainer`) — both still at two sites, so rule-of-three says wait; `ElementContainer` is under #480.

Verified: lint, typecheck, format:check, 982 tests, `npm run build`.

### DIFF
--- a/app/components/AuthForm.tsx
+++ b/app/components/AuthForm.tsx
@@ -2,6 +2,7 @@
 
 import { useState } from "react";
 import { sendMagicLink, signInWithGoogle } from "@/lib/auth/session";
+import { errorMessage } from "@/lib/errors";
 import OnboardingCard from "./OnboardingCard";
 import EmailSentCard from "./EmailSentCard";
 import OrDivider from "./OrDivider";
@@ -54,6 +55,15 @@ export default function AuthForm({
 		setLoading(false);
 	};
 
+	const handleGoogle = async () => {
+		setError("");
+		try {
+			await signInWithGoogle();
+		} catch (cause) {
+			setError(errorMessage(cause));
+		}
+	};
+
 	if (sent) {
 		return (
 			<EmailSentCard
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
lgtm

PR #497, opened 2026-07-31
fix: nightly bug fixes 2026-07-31

umairnadeem:
## Bug

An async video job's clip length is only known at submit time. Runware's poll response reports the video URL and a status, but no duration.

`videoHandler.poll` persisted that poll response verbatim as the job result, so the completed bundle reached the client with no `durationSec` in its metadata. `BaseAssetConnector.resolveBundle` then coerces the missing value to 0 (`Number(bundle.manifest.metadata?.durationSec ?? 0)`), and `scene-builder` lays the clip out with zero duration — so every real generated video clip landed on the timeline with no length.

Mock providers happen to report a real length on poll, which is why the mock path always looked correct and this never showed up in tests.

## Fix

`videoHandler.process` now records `durationSec` next to `providerJobId` in the job row's metadata (the provider already returns it on submit — `params.duration ?? 5` for Runware, 5 for the mock). `poll` merges it into the completed result, preferring a length the provider actually measured over the requested one.

A submission that reports no length now throws with an actionable message instead of silently producing a zero-duration clip, matching the existing missing-`jobId` guard.

Behavior is otherwise unchanged. Rows recorded before this change still poll and complete; they just keep the old (absent) duration rather than crashing.

## Tests

`lib/api/handlers/__tests__/video.test.ts`:
- `process` records both the provider job id and the submitted clip length
- `process` throws when the provider reports no clip length
- `poll` stamps the submitted length onto the completed bundle
- `poll` prefers a provider-measured length over the submitted one
- `poll` leaves the length unset for rows recorded before this change

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, `test:coverage` all pass (114 files / 983 tests).

## Open PR overlap check

Considered open PRs #438, #480, #481, #484, #485, #486, #487, #488, #493, #494, #495, #496. None touches `lib/api/handlers/` or the server-side async video job path. #487 and #480 are the nearest in theme (video layout / client-side generation scheduling) but neither touches these files or the job handler contract.

### DIFF
--- a/lib/api/handlers/__tests__/video.test.ts
+++ b/lib/api/handlers/__tests__/video.test.ts
@@ -16,7 +16,10 @@ vi.mock("@/lib/api/jobs", () => ({
 	updateJob: (...args: unknown[]) => updateJob(...args),
 }));
 
-type VideoJobRow = TypedJobRow<VideoGenerateParams, { providerJobId?: string }>;
+type VideoJobRow = TypedJobRow<
+	VideoGenerateParams,
+	{ providerJobId?: string; durationSec?: number }
+>;
 
 const bundle = { id: "bundle-1" } as unknown as BundleResponse;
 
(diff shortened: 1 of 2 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
this is intentional for now

PR #498, opened 2026-07-31
refactor: nightly maintainability 2026-07-31

umairnadeem:
## What

Two LLM streaming call sites hand-rolled the same at-most-one-in-flight lifecycle: abort the previous controller, flip a loading flag, break the `for await` when the signal aborts, and clear the flag in `finally` only if the controller is still the current one. `ScriptProvider.submitPrompt`/`stopGeneration` and `useRefineScript.refineScript`/`stopRefine` were near-identical, and the tricky part (a superseded run must not clear the newer run's loading flag) was duplicated with no test behind it.

- New `lib/script/drainStream.ts` — pure "drain until end or abort, settle only if it ended on its own". Unit tested.
- New `lib/script/useStreamRun.ts` — the React bookkeeping around it: `{ loading, run, stop }`.
- `ScriptProvider` and `useRefineScript` now express only what's theirs: what to stream and what to do with each chunk.
- `RefineComposer` reads `useScriptControl()` itself instead of taking `loading`/`onStop` drilled through `EditorToolbar`.

No behavior change. `stop` is byte-for-byte what both `stopGeneration` and `stopRefine` did; the chunk handling and the refine post-stream `parser.flush()` keep their exact ordering and abort semantics.

## Complexity delta

- 3 files simplified, net **-44 lines** of production code.
- Two copies of the abort/loading protocol collapse to one, and it gains 4 tests (chunk delivery, abort mid-stream, pre-aborted, upstream throw) where it previously had none.
- `RefineComposer` loses 2 props and its prop-typing block.

## Skipped

- Nothing else in reach was worth touching. `lib/api`, `lib/video`, `remotion/`, and the canvas panel/attribute components read clean; the remaining candidates all sit inside open PRs.

## Validation

`npm run build`, `npx tsc --noEmit`, `npx eslint`, `npx prettier --check .`, `npx knip`, and `npx vitest run` (115 files, 984 tests) all pass.

## Open PR overlap check

[...]

PR #499, opened 2026-07-31
perf: nightly performance optimizations 2026-07-31

umairnadeem:
## What

`SegmentedSeekBar` subscribes to the Remotion player's `frameupdate`, so it re-renders 30–60 times a second during playback. PR #495 looked at this and deferred it as *"inherent to drawing a progress fill."* It isn't — only one number changes per frame.

Every one of those renders was doing:

- `segments.map(...)` to rebuild the per-scene basis array (new array, new object per scene)
- inside `ScrubBar`, a cumulative-starts pass plus a `clamp((ratio - start) / basis, 0, 1)` per segment for both the progress fill and the hover fill
- three React elements and three inline style objects **per scene**, all with fresh identities

For a 20-scene project that's ~60 elements and ~60 style objects reconciled per frame, to move one playhead.

## The change

The fill geometry moves from JS into CSS, where it was always expressible:

- `value` and the hover ratio are written to the track root as `--scrub-pos` / `--scrub-preview`.
- Each segment carries its static span as `--seg-start` / `--seg-scale` (`1 / basis`), set once.
- A fill's width is the same clamp, in CSS:
  `clamp(0%, calc((var(--scrub-pos) - var(--seg-start)) * var(--seg-scale) * 100%), 100%)`
- The segment subtree is `useMemo`'d on the segments, so React skips it entirely when only the playhead moved.
- `SegmentedSeekBar` memoizes the basis array it passes down, keeping that identity stable.

A frame update now commits two custom-property writes on one element. The browser recomputes the fills; React never sees them.

Both `ScrubBar` callers are unaffected — the props API is unchanged, so the continuous volume bar in `PlayerControls` gets the same treatment for free (single segment, `start=0`, `scale=1`).

## Why it matters

Playback is the one place in this app with a real frame budget, and the seek bar was the largest per-frame React cost left in that tree after #495 pushed the scene-index subscription down to its leaves. This removes the remainder rather than shrinking it.

It's also less code doing less work, not a memo shim over the same work: the JS that computed percentages is gone, not cached.

## Verification

The CSS expression is the load-bearing part, so it was checked against real Chromium rather than assumed. With `--scrub-pos: 0.6` over segments of basis `0.25 / 0.5 / 0.25`:

| segment | span | measured fill | expected |
| --- | --- | --- | --- |
| s0 | 0–0.25 | 250 / 250px | full |
| s1 | 0.25–0.75 | 350 / 500px | `(0.6 − 0.25) × 2 = 0.7` |
| s2 | 0.75–1 | 0 / 250px | clamped to 0 |

Identical to the old JS output. React writes custom properties through `style.setProperty` with no unit coercion, so the unitless numbers land as-is.

New `segmentStyle` unit test covers the cumulative layout and the zero-basis segment (which previously produced `Infinity` inside the clamp and now resolves to a finite scale).

## Slate pass

Re-checked against the performance walkthrough. `renderElement` is a stable `useCallback`, `renderLeaf`/`decorate` are unused, every editor consumer already uses `useSlateStatic` (no `useSlate` anywhere in the tree), and the normalizers stay on the node they're handed. Chunking doesn't apply at this document size. Nothing to do.

## Skipped

- **`Captions` per-frame style object** in `remotion/components/Captions.tsx` — the container/text styles are already module-level and the word lookup is a memoized binary search. The one remaining per-frame allocation is a single wrapper style object. Marginal.
- **`SeekTooltip` re-rendering per frame while hovered** — real, but hovering is not a sustained state and the subtree is one element deep.
- **`useGenerate`'s whole-`metadata` subscription** — still the largest fanout in the app, still owned by open PR #480.
- **Anything in the canvas element render path** — `ElementContainer`, `SceneContainer`, `OutputPreview`, `SortableItem`, `SortableScene`, and `Canvas.tsx` are all claimed by open PRs.

## Open PR overlap check

Considered #498, #497, #496, #495, #494, #493, #488, #487, #486, #485, #484, #481, #480, #438.

This PR touches `ScrubBar.tsx`, `SegmentedSeekBar.tsx`, and a new `__tests__/ScrubBar.test.ts`. None of the three appears in any open PR's file list. #495 is the nearest neighbour — it touches `PlayerControls.tsx`, which *renders* `ScrubBar` — but `ScrubBar`'s props are unchanged, so the two are independent. This PR deliberately picks up the item #495 explicitly listed as skipped.

## Validation

`lint`, `typecheck`, `format:check`, `knip`, `build`, and `test:run` (982 passed, +2 new) all green.

### DIFF
--- a/app/components/video/ScrubBar.tsx
+++ b/app/components/video/ScrubBar.tsx
@@ -1,6 +1,13 @@
 "use client";
 
-import { type PointerEvent, type ReactNode, useRef, useState } from "react";
+import {
+	type CSSProperties,
+	type PointerEvent,
+	type ReactNode,
+	useMemo,
+	useRef,
+	useState,
+} from "react";
 import { clamp, cn } from "@/lib/utils";
 
 /** A slice of a segmented track; `basis` is its fraction (0–1) of the whole. */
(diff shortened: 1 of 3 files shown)


umairnadeem (commented):
lgtm

PR #500, opened 2026-07-31
refactor: rethink openslop architecture 2026-07-31

umairnadeem:
## App understanding

OpenSlop turns a prompt into a rendered video. The LLM streams **OSML** (an XML-ish script language) over SSE; a stream parser turns it into canvas nodes that Slate renders as scenes of elements (narration, character, image, animated_image, clip, sound, music). Stale elements are queued client-side, generated through connector → gateway → provider, and the composition renders on Remotion Lambda. Project state lives in Supabase; per-project metadata lives in a Zustand store.

## From-scratch architecture

OSML is the app's domain language, so `lib/canvas/` should own its whole vocabulary and both projections of a parsed node list — **canvas elements** (into the Slate document) and **metadata** (into the project store) — and `app/` should only wire those to React effects.

Today it's split. The canvas half lives in `lib/canvas/`, but the metadata half (`metadata_title`, `metadata_style`, `metadata_narration`, `metadata_character`) lived in `app/components/canvas/config/metadataTags.ts`: a zod-parsing table over `lib/project/types` that renders nothing, imports nothing from `app/`, and was only reachable through a React hook. `lib/canvas/osmlStreamParser.ts` even carries a comment deferring to "the metadata sync layer" in the UI tree. Both consuming hooks then re-derived "what kind of tag is this" with unchecked `as CanvasElementType` / `as CanvasContentElement` casts.

- **New `lib/canvas/osmlMetadata.ts`** owns the `metadata_*` vocabulary and exposes one pure seam, `collectMetadata(nodes) => DeepPartial<Metadata>`. Reading metadata out of an OSML node list is now possible outside React. Tag semantics are unchanged.
- **`app/components/canvas/config/` is gone.** `useMetadataSync` is now three lines of wiring: nodes → `collectMetadata` → `updateMetadata`.
- **`lib/canvas/guards.ts` gains `isCanvasElementType` and `isParsedContentElement`**, the complement of `collectMetadata`. `useScriptSync` drops its local `shouldSkip` helper and both unchecked casts; `isContentElement` now routes through the same guard, so the one remaining `as CanvasElementType` in that file is gone too.
- **Tests moved to the seam.** The old test poked `METADATA_TAG_CONFIGS[tag].apply(mutableAccumulator, …)` — an internal. It now drives `collectMetadata` with node lists, keeping every original case and adding coverage for mixed lists and for canvas nodes being ignored. New `lib/canvas/__tests__/guards.test.ts` covers the guards.

Net: the UI layer no longer holds any OSML knowledge, and no consumer hand-classifies a tag by string.

## Validation

| Check | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:run` | 987 passed / 115 files |

`npm run test:e2e` not run locally (needs Supabase env + a Playwright browser install); CI covers it.

## Open PR overlap check

Considered #438, #480, #481, #484, #485, #486, #487, #488, #493, #494, #495, #496, #497, #498, #499. None of them touch any of the 8 files in this diff, so it is non-overlapping.

## Skipped (blocked by open PRs, not by merit)

- **Move the Slate document-model plugins (`withScenes`, `withNodeId`, `withFlatPaste`, `withLayout`) from `app/components/canvas/plugins/` into `lib/canvas/`.** Same layering violation as the metadata table, and a bigger win. Blocked: #438 adds `withCollapsedVoids` to the same import block in `useEditorSetup.ts`, so relocating them now guarantees a conflict.
- **`lib/project/serialize.ts:23` casts `parseOSML(...) as CanvasContentElement[]`**, which also lets `metadata_*` nodes through into scene children unchecked. `isParsedContentElement` would fix it, but #481 is changing that file's test.
- **The `osmlStreamParser.ts` comment now points at a module in `lib/`, and `ARCHITECTURE.md` doesn't mention the metadata projection.** Both files are under #481 and #485 respectively; worth a follow-up once those land.


umairnadeem (commented):
lgtm

PR #501, opened 2026-08-01
fix: nightly bug fixes 2026-08-01

umairnadeem:
## Bugs fixed

**1. A non-numeric asset duration became `NaN` on the timeline** (`lib/connectors/asset-base.ts`)

`resolveBundle` coerced with `Number(metadata?.durationSec ?? 0)`. `??` only catches `null`/`undefined`, so any other non-numeric value (`""`, `"abc"`, an object) produced `NaN`. This is the one boundary where untyped bundle JSON becomes a typed `AssetResult.durationSec`, and `buildVideoLayout` has no defence downstream — `createSequence` does `Math.max(duration, MIN_DURATION_SEC)`, which is `NaN` for `NaN`, and every start offset, `totalDurationSec` and `totalFrames` derived from it follows. Parsed to a finite number at that boundary instead, keeping the intended `0` default.

**2. A cached audio row with no usable duration was served as a hit** (`lib/providers/cache.ts`)

`audioBundleCache.fromMetadata` rehydrated `durationSec: Number(m.duration)` unconditionally, so a row missing `duration` (the legacy `audioUrl` rows this function explicitly supports) came back as a `NaN` duration — feeding bug 1. The option's own contract is "return `undefined` when the stored row can't be rehydrated, forcing a miss", so it now does. A stored `0` is still a real value and still hits.

**3. Some asset-job failures never reached the job row** (`app/api/queues/asset-generate/route.ts`)

Handler lookup and the terminal `updateJob` writes sat outside the try/catch that marks a job `failed`. An unregistered connector type, or a write that fails after the handler succeeded, threw with the row left at `pending`/`processing`. The client only learns the outcome by polling that row, so once the queue exhausted its retries the job was polled forever. Moved the whole body inside the existing catch, which also drops the `let outcome` and one nesting level.

## Tests added

- `lib/api/__tests__/jobs.test.ts` (new) — `lib/api/jobs.ts` was at 4% line coverage despite being the persistence layer behind every async generation. Covers the `projectId ?? null` default, error-vs-missing-row distinction in `createJob`/`getJob`/`loadJobForProcessing`, `updateJob`'s undefined-stripping and no-op write, and the `enqueueJob` topic payload.
- `lib/connectors/__tests__/asset-base.test.ts` — non-finite `durationSec` values (`""`, `"abc"`, `null`, `{}`, `NaN`, `Infinity`) all default to 0; a numeric string still parses.
- `lib/providers/__tests__/cache.test.ts` — a row with no/garbage `duration` forces a miss; `duration: 0` still hits.
- `app/api/queues/__tests__/asset-generate.test.ts` — missing handler and a failing result write both mark the job failed and rethrow.

Coverage: statements 80.00% → 81.07%, branches 76.68% → 78.50%.

## Skipped

Swept the rest of the non-overlapping surface (`lib/api`, `lib/gateway`, `lib/clients`, `lib/upload`, `lib/video` helpers, `lib/canvas` ops, `lib/script/refine`, `remotion/`, the free `app/components` hooks) without finding another defect worth a change. Two things were checked and cleared rather than "fixed": `moveDraggedElement`'s Slate path arithmetic (probed all four drag directions — correct as written) and `readSSE`'s dropped trailing buffer (unreachable, every producer terminates frames with `\n\n`).

## Open PR overlap check

Considered #500, #499, #498, #497, #496, #495, #494, #493, #488, #487, #486, #485, #484, #481, #480, #438. Diffed the final file list against the union of their touched files — zero intersection. Nearest neighbours are #497 (`lib/api/handlers/video.ts`, not `lib/api/jobs.ts`), #493 (`lib/providers/tts/*`, not `lib/providers/cache.ts`) and #480 (`lib/connectors/base.ts`/`factory.ts`, not `asset-base.ts`).

### DIFF
--- a/app/api/queues/asset-generate/route.ts
+++ b/app/api/queues/asset-generate/route.ts
@@ -11,15 +11,22 @@ export const POST = handleCallback<AssetQueueMessage>(
 		const job = await loadJobForProcessing(jobId);
 		if (isTerminal(job.status)) return;
 
-		const handler = getJobHandler(connectorType);
-		if (!handler) {
-			throw new Error(`No job handler registered for ${connectorType}`);
-		}
-
 		await updateJob(jobId, { status: "processing" });
-		let outcome;
+
+		// Every failure past this point has to land on the row: the client only
+		// ever learns the outcome by polling it, so a job left `processing`
+		// after the queue exhausts its retries is polled forever.
 		try {
-			outcome = await handler.process(job);
+			const handler = getJobHandler(connectorType);
+			if (!handler) {
+				throw new Error(`No job handler registered for ${connectorType}`);
+			}
+			const outcome = await handler.process(job);
+			if (outcome.kind === "completed") {
+				await updateJob(jobId, { status: "completed", result: outcome.result });
+			} else {
+				await updateJob(jobId, { metadata: outcome.metadata });
+			}
 		} catch (error) {
 			logger.error(error, `Job ${jobId} failed`);
 			await updateJob(jobId, {
@@ -27,12 +34,6 @@ export const POST = handleCallback<AssetQueueMessage>(
 				error: stringifyError(error),
 			});
 			throw error;
-		}
-
-		if (outcome.kind === "completed") {
-			await updateJob(jobId, { status: "completed", result: outcome.result });
-		} else {
-			await updateJob(jobId, { metadata: outcome.metadata });
 		}
 	},
 );
--- a/app/api/queues/__tests__/asset-generate.test.ts
+++ b/app/api/queues/__tests__/asset-generate.test.ts
@@ -54,14 +54,34 @@ describe("asset-generate queue worker", () => {
 		expect(mockUpdateJob).not.toHaveBeenCalled();
 	});
 
-	it("throws when no handler is registered for the connector type", async () => {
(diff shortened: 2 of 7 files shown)


app/api/queues/asset-generate/route.ts
```diff
@@ -11,15 +11,22 @@ export const POST = handleCallback<AssetQueueMessage>(
...
+		// after the queue exhausts its retries is polled forever.
 		try {
-			outcome = await handler.process(job);
+			const handler = getJobHandler(connectorType);
+			if (!handler) {
+				throw new Error(`No job handler registered for ${connectorType}`);
+			}
+			const outcome = await handler.process(job);
+			if (outcome.kind === "completed") {
+				await updateJob(jobId, { status: "completed", result: outcome.result });
```
umairnadeem, line 26: pulling the result write inside the catch means a transient db blip on this line marks a job that actually succeeded as `failed`, and since failed is terminal the queue retry hits the `isTerminal` early return and can never recover it (before, the retry would re-run and re-write). same for the pending branch: the provider job is still running but the row says failed. maybe keep the outcome writes outside the try so only handler errors mark the row failed?

umairnadeem:
overengineered

PR #507, opened 2026-08-02
fix: nightly bug fixes 2026-08-02

umairnadeem:
Two real correctness bugs in the generation queue, each with a regression test that fails without the fix.

## Bugs fixed

**A retry shows the previous failure.** `enqueueGraph` set `status: "queued"` but left `error` in place, so the red error overlay stayed pinned over the placeholder for the entire retry. Same for the `"Enter a prompt first"` message: adding a prompt and hitting Generate kept showing it until the job settled. Queuing now clears the error.

**Uploading over a generating element silently kills its dependents.** `commitResult` cancelled the in-flight job *before* committing, and `cancel` resumes the queue. At that moment the element still had no result, so `releaseBlocked` treated every dependent waiting on it as unrunnable, dropped it from the queue and reset it to idle with no error. Uploading an image (or seeding a character avatar from a template) mid-run left the derived video/animated-image never generated and nothing on screen to say so. `commitResult` now aborts the in-flight job, commits the result, and only then resumes the queue.

## Tests added

- `queue.test.ts`: re-queueing clears a previous failure; clears an error set via `setError`.
- `queueDependencies.test.ts`: a dependent waiting on a running dependency runs against the result committed over it.

All three fail on master and pass here. Full suite: 997 passing. Lint, format, typecheck, knip and build all clean.

## Open PR overlap check

Considered open PRs #438, #481, #484, #485, #486, #487, #488, #493, #494, #495, #496, #497, #498, #499, #500, #501, #502, #503, #504. None touches `lib/generation/` — the closest are #501 (`lib/api/jobs.ts`, `lib/connectors/asset-base.ts`, `lib/providers/cache.ts`) and #487 (canvas preview + video layout components). This PR is non-overlapping in both files and theme.

### DIFF
--- a/lib/generation/__tests__/queue.test.ts
+++ b/lib/generation/__tests__/queue.test.ts
@@ -173,6 +173,34 @@ describe("GenerationQueue", () => {
 			generationQueue.discard("m2");
 		});
 
+		it("clears a previous failure so a retry does not show a stale error", async () => {
+			generateMock.mockRejectedValueOnce(new Error("generation failed"));
+			generationQueue.enqueueGraph([makeJob("retry1")]);
+			await vi.runAllTimersAsync();
+			expect(generationQueue.getElementSnapshot("retry1").error).toBe(
+				"generation failed",
+			);
+
+			generateMock.mockReturnValueOnce(new Promise(() => {}));
+			generationQueue.enqueueGraph([makeJob("retry1")]);
+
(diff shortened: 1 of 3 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
unnecessary

PR #508, opened 2026-08-02
refactor: nightly maintainability 2026-08-02

umairnadeem:
Behavior-preserving maintainability pass on the canvas element card and the character edit modal.

## Architecture

**`useAvatarGeneration(name)`** (new, `app/components/canvas/elements/character/`) owns one character avatar's generation node: its live snapshot, staleness, and the three writes that can change it (`regenerate`, `commitUpload`, `discard`). `CharacterEditModal` was hand-rolling that orchestration inline — building the node, reading the queue snapshot, computing staleness, clearing `avatarUploaded` before enqueue, and pinning uploads — a second copy of what `useGenerate` already does for canvas elements. The modal now holds only dialog policy (delete confirm, stale-close intercept) and markup.

**Element card prop drilling.** `ElementGenerationProvider` already receives the element; it now exposes it. `ElementSettings`, `ModelBadge`, `ElementCharacters`, and `AnimateButton` read it from context instead of taking an `element` prop, matching `ElementStaleIndicator` / `ElementUploadButton` / `ElementGenerateButton`, which already did. `AttributeBadge`, `TextAttributePopover`, `CharactersPicker`, and `DeleteButton` keep their explicit `element` prop — they are generic and used outside the card, so the prop is their contract, not drilling.

## Complexity delta

- `CharacterEditModal` body: 5 generation-wiring hooks/derivations removed, ~30 lines of orchestration moved behind one named interface.
- 4 components lose a required prop; `ElementContainer` loses 4 prop passes.
- No new branching; no behavior change (same store writes, same enqueue/pin semantics, no restore-on-stale effect added to the avatar node).

## Skipped

- **Docs.** `README.md` (#502) and `ARCHITECTURE.md` (#485) are both under open PRs, and nothing in this diff changes an interface those files describe. No doc line would have earned its place.
- **`CharacterAssetTiles`'s unused `onRemove` prop.** Reusable surface with no caller today; left for a human call rather than deleted.
- **Unit test for `useAvatarGeneration`.** Would need a client-hook harness (Config + queue + project providers) that does not exist in this repo yet; the underlying graph/queue mechanics are already covered by `lib/generation/__tests__`.

## Open PR overlap check

Considered all 20 open PRs (#507, #504, #503, #502, #501, #500, #499, #498, #497, #496, #495, #494, #493, #488, #487, #486, #485, #484, #481, #438) and diffed each for touched files. No file in this PR appears in any of them. Nearest neighbours: #487 touches `elements/preview/*`, `SceneContainer.tsx`, and `OutputPreview.tsx` (all untouched here) and imports `useElementGeneration` from `AnimatedImagePreview.tsx` — this PR only *adds* a field to that context value, so the existing consumers are source-compatible.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and `npm test` (114 files, 994 tests) all pass.

### DIFF
--- a/app/components/canvas/elements/ElementGenerationContext.tsx
+++ b/app/components/canvas/elements/ElementGenerationContext.tsx
@@ -5,7 +5,9 @@ import { createRequiredContext } from "@/lib/components/createRequiredContext";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 import { useGenerate } from "../hooks/useGenerate";
 
-type ElementGeneration = ReturnType<typeof useGenerate>;
+type ElementGeneration = ReturnType<typeof useGenerate> & {
+	element: CanvasContentElement;
+};
 
 const [ElementGenerationContext, useElementGeneration] =
 	createRequiredContext<ElementGeneration>("ElementGenerationContext");
@@ -13,7 +15,8 @@ export { useElementGeneration };
 
 // One subscription, one staleness check and one restore effect per element card.
 // Children are passed through untouched, so a queue tick re-renders only the
-// consumers below, not the Slate-managed subtree.
+// consumers below, not the Slate-managed subtree. The element rides along so
+// card-scoped controls read it here instead of taking it as a prop.
 export function ElementGenerationProvider({
 	element,
 	children,
@@ -23,7 +26,7 @@ export function ElementGenerationProvider({
 }) {
 	const generation = useGenerate(element);
 	return (
-		<ElementGenerationContext value={generation}>
+		<ElementGenerationContext value={{ ...generation, element }}>
 			{children}
 		</ElementGenerationContext>
 	);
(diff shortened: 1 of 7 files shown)


app/components/canvas/elements/ElementGenerationContext.tsx
```diff
@@ -23,7 +26,7 @@ export function ElementGenerationProvider({
 }) {
 	const generation = useGenerate(element);
 	return (
-		<ElementGenerationContext value={generation}>
+		<ElementGenerationContext value={{ ...generation, element }}>
```
umairnadeem, line 29: `CONVENTIONS.md` § Frontend: *"Composition over context. Don't thread props through layers that don't use them; prefer composition with children. Reach for context only for genuinely cross-cutting state, never just to skip a level."* `element` isn't being drilled here, ElementContainer is the direct parent of all four of these, so this trades an explicit prop for provider coupling without skipping any level. imo keep the `element` prop on ModelBadge/ElementCharacters/AnimateButton/ElementSettings and leave the context to generation state.

umairnadeem:
useless

Issue #506, opened 2026-08-02
Animated image still/video toggle doesn't show the still node's progress


umairnadeem:
> Assumes #480 (generation jobs as a dependency graph) is merged. The still node this describes is introduced there.

## Problem

An animated image reads "Queued…" for most of the time it is actually working.

Under the dependency graph, an `animated_image` is two nodes: a still-image node feeding the animation node. `enqueueGraph` marks both `"queued"` (`lib/generation/queue.ts:182-200`), then `processQueue` runs the still first because the animation has a `blockingDependency` until the still settles (`queue.ts:306-324`). The still flips to `"generating"`; the animation stays `"queued"` until the still finishes.

The card, meanwhile, only ever looks at the animation node. `useGenerate` reads a single snapshot keyed by the element's own id (`app/components/canvas/hooks/useGenerate.ts:13`):

```ts
const snapshot = useQueueSelector((q) => q.getElementSnapshot(element.id));
```

The still's id is `derivedNodeId("still", elementId)` (`lib/connectors/animated_image/plugins/still-frame.ts:22`), so its status is invisible to the UI. `GenerationIndicator` renders the pulsing hourglass and "Queued…" (`GenerationIndicator.tsx:19-27,38-42`) through the entire still generation, which is the larger share of the wall time for an animated image, and the elapsed-seconds counter does not start until the animation itself begins.

So the one element type that does the most work is the one that looks idle while doing it. It reads as stuck, particularly next to plain image elements that correctly show "Generating 12s" — and "queued" is a real state elsewhere in the app (waiting on a batch slot), so this is not just an imprecise label, it is the wrong one.

## Reported by (verbatim feedback)

- "when animated image is being generated (and its still is being generated too), we currently only show it being queued (since the video node is queued even though the still is generating). note that this assumes https://github.com/openslop/openslop/pull/480 is merged in."

## Relevant code

- `app/components/canvas/hooks/useGenerate.ts:13,36-45` — reads one snapshot by element id and returns `status` / `seconds` straight through. Leave this alone; it is correct for the animation node.
- `lib/generation/queue.ts:182-200` — `enqueueGraph`; every node in the graph starts `"queued"`
- `lib/generation/queue.ts:306-313` — `blockingDependency`; why the animation stays queued while the still runs
- `lib/generation/queue.ts:349-356` — `runJob`; `"generating"` and the elapsed timer are per node
[...]

Issue #492, opened 2026-07-30
Select the still-image model and the video model separately on animated_image


umairnadeem:
## Problem

An `animated_image` runs **two** generations — a still image, then an image-to-video animation of that still — but only exposes **one** model choice. The `model` attribute names the video model. The still frame's image model is not selectable at all: it silently uses whatever the registry default happens to be.

The still is a real generation node of its own (`stillElement`, `lib/connectors/animated_image/plugins/still-frame.ts:31-38`) built by stripping video-only attributes off the parent element and retyping it as `image`:

```ts
const VIDEO_ONLY_KEYS = ["videoPrompt", "duration", "model"] as const;
```

`model` is in that list — correctly, since a video model name is meaningless to an image generation. But nothing puts an image model back. The still node reaches `resolveElementConnector` (`lib/canvas/elementConnector.ts:33-51`) with no `model` and no `provider` attribute, so it falls all the way through to `getDefaultConnector(registry, "image")` → `IMAGE_MODELS` default, currently `"Slop Image v1"` (`lib/connectors/registry.ts:41`).

So today:

- You can pick the **video** model for an animated image (`ModelBadge` edits `attrKey="model"`, resolved against `animated_image` → `VIDEO_MODELS`, `lib/connectors/registry.ts:42`)
- You **cannot** pick the image model for its still frame — the one that determines what the shot actually looks like

That is backwards from how people work. The still frame decides composition, character likeness, and style; the video model only decides how it moves. A plain `<image>` element next door gives you full model choice over exactly the same generation, so the capability exists — it just isn't reachable once the element becomes an `animated_image`. Switching an element between `image` and `animated_image` also quietly drops your model pick, since `preservedAttributes` (`lib/canvas/preservedAttributes.ts:8-10`) only carries `characters` across that pair and the two `model` values live in different namespaces anyway.

## Reported by (verbatim feedback)

- "also another issue to allow selecting models for animated image's video generation and image generation separately"

## Relevant code

- `lib/connectors/animated_image/plugins/still-frame.ts:20,31-38` — `VIDEO_ONLY_KEYS` / `stillElement`; strips `model` and puts nothing back. The single change point for routing a still model through.
- `lib/connectors/animated_image/plugins/still-frame.ts:40-42` — `forStillOf`; builds the still `NodeSpec` with `buildImagePlugins()`
- `lib/canvas/elementConnector.ts:33-51` — `resolveElementConnector`; why the still lands on the image registry default (no `model`, no `provider`)
[...]

Issue #491, opened 2026-07-30
Copy/paste OSML from the canvas (clipboard round-trip)


umairnadeem:
## Problem

The canvas has no OSML clipboard. Copying from it loses everything that makes it a canvas, and pasting OSML into it produces plain text instead of elements — even though both conversions already exist as pure functions.

**Copy is lossy.** Nothing overrides `editor.setFragmentData`, so Slate's default writes `text/plain` as `Node.string()` of the selection. Element types, `customAttributes` (model, provider, voice, character refs), and scene boundaries are all dropped. Copy three scenes out of a project and you get bare narration text — you cannot paste it into another project, into a chat with a model, or into an issue report.

**Paste is unparsed.** `editor.insertData` is not overridden either, so pasting OSML text lands as plain text. `withLayout` then wraps it into a `narration` element and you get literal `<narration id="…">` characters sitting inside a narration node. Two existing plugins already handle the *internal* Slate-fragment path — `withFlatPaste` flattens pasted scenes into content elements, `withNodeId` strips and re-mints ids — but neither one runs for text arriving from outside the app.

Both halves of the conversion are already written and tested:

- **Canvas → OSML**: `serializeOSMLWithScenes` (`lib/canvas/osmlSerializer.ts:24-37`), used by autosave and script refinement
- **OSML → canvas**: `parseOSML` (`lib/canvas/osmlStreamParser.ts:107-114`) and `deserializeWithScenes` (`lib/project/serialize.ts:17-25`), used by project rehydration

They are wired to persistence and to the LLM, but not to the clipboard. OSML is the project's interchange format; the clipboard is the one place it isn't reachable.

One gap to close in the serializer: `serializeOSMLWithScenes` only walks top-level scene nodes and silently drops anything else, while `serializeOSML` walks content elements but emits no scene markers. A clipboard fragment from `editor.getFragment()` can be either shape (a partial selection inside one scene, or several whole scenes), so neither function alone covers it.

## Reported by (verbatim feedback)

- "create issues in openslop to allow people to create blank canvases, copy/paste hooks to convert to/from OSML format from canvas"

## Relevant code

- `app/components/canvas/plugins/` — where the new plugin belongs; `withFlatPaste.ts` and `withNodeId.ts` are the existing paste-path precedents
- `app/components/canvas/hooks/useEditorSetup.ts:18-27` — the `flow(...)` plugin stack and its load-bearing order (`withNodeId` outermost, so it strips ids before `withFlatPaste` flattens)
[...]

PR #509, opened 2026-08-02
fix: keep the character edit affordance open while the avatar is generating

### LINKED ISSUE #505: Character edit dialog can't be opened while the avatar is generating
## Problem

While a character's avatar is generating, its edit dialog cannot be opened. The pencil affordance on the asset tile is not rendered at all for the duration of the generation:

`app/components/canvas/elements/AssetTile.tsx:88-94`

```tsx
{onEdit && status !== "generating" && (
  <OverlayButton icon={Pencil} label={`Edit ${name ?? "asset"}`} onClick={onEdit} />
)}
```

That tile is the only entry point to `CharacterEditModal` (`AssetsSection.tsx:41` → `editCharacter` → `useAssetEditDialogs.tsx:26-30`), so an avatar generation locks you out of the character's appearance, voice, and name fields until it finishes.

The dialog itself is already built to be open during generation. `CharacterEditModal` subscribes to the avatar snapshot (`CharacterEditModal.tsx:74`), renders its status inline (lines 162-180), and disables only the Generate button (`generateDisabled = generating || !hasAppearance`, line 97) while leaving every field editable. `StaleAvatarCloseDialog` then handles exactly the case where your edits outrun the avatar. So the "edit while generating" flow is designed and implemented; the tile just refuses to open the door.

Practical effect: spot a typo in a character's appearance right after hitting generate and you have to watch the spinner before you can fix it, even though fixing it early is what you want.

- "when a character avatar is loading due to generation, we should be still be able to open the character edit dialog"

## Relevant code

- `app/components/canvas/elements/AssetTile.tsx:88-94` — the `status !== "generating"` guard on the edit overlay; the change point
- `app/components/canvas/elements/AssetTile.tsx:96-114` — the same guard on the remove affordances (`overlay` and `corner`), which is a separate question: removing a character mid-generation is not obviously safe the way editing is
- `app/components/canvas/elements/AssetTiles.tsx:36-58` — `CharacterAssetTile`; passes `elementId` (so the tile knows the status) and `onEdit`
- `app/components/canvas/elements/AssetsSection.tsx:41,60` — the only wiring from tile to `editCharacter` and the rendered dialogs
- `app/components/canvas/elements/character/useAssetEditDialogs.tsx:26-30` — `CharacterEditModal` open state, driven solely by `editingName`
- `app/components/canvas/elements/character/CharacterEditModal.tsx:74,95-97,162-180` — already subscribes to the avatar snapshot, shows status, and disables only the Generate button
- `app/components/canvas/elements/character/StaleAvatarCloseDialog.tsx` — the existing "your edits outran the avatar" flow

## Suggested direction

Drop the `status !== "generating"` condition from the edit overlay so the pencil stays available throughout. The generating shimmer and `GenerationIndicator` already sit in the same corner and communicate the state; the edit button does not need to be the thing that says "busy".

Keep the guard on remove, or decide it separately — deleting a character whose avatar job is mid-flight is a different question from editing one, and the queue would need to cancel that job.

Worth confirming while in here that the overlay's hover/focus styling still reads correctly on top of the shimmer surface (`AssetTile.tsx:75-80`).

## Acceptance criteria

- [ ] The edit affordance on a character tile is available while its avatar is generating, by pointer and by keyboard
- [ ] Opening the dialog mid-generation shows the live avatar status and leaves the fields editable, as it already does when opened after
- [ ] Edits made during generation follow the existing stale-avatar flow on close
- [ ] The generating shimmer and indicator still read clearly with the overlay present
- [ ] Remove behavior is unchanged, or changed deliberately with its own reasoning

## Related

- #342 (closed; explicit "Done" button on the character/narration modals — same dialog)
- #233 (closed; revert character appearance to the avatar's source while stale — the adjacent stale-avatar flow)
- #243 (character/avatar generation correctness — the generation this blocks on)

---

umairnadeem:
## Summary

The pencil overlay on an `AssetTile` was gated on `status !== "generating"`, and that tile is the only entry point to `CharacterEditModal`. So for the entire duration of an avatar generation you could not open the character's appearance, voice, or name fields — exactly when you most want to fix the typo you just noticed.

The dialog was already built for this case. It subscribes to the avatar snapshot, renders the live status inline, disables only the Generate button (`generateDisabled = generating || !hasAppearance`), and `StaleAvatarCloseDialog` already handles edits that outrun the avatar. The tile was the only thing refusing to open the door.

One-line fix: drop the guard from the edit overlay.

```diff
-{onEdit && status !== "generating" && (
+{onEdit && (
```

**Remove is deliberately left gated.** Deleting a character whose avatar job is mid-flight is a different question — the queue would need to cancel that job — and the issue calls it out as separable.

`#505` is labelled `size: XS` / `good first issue`, names the exact change point, and the receiving dialog is already implemented for the mid-generation case. Single file, single condition.

Checked every `AssetTile` caller, since the guard removal widens when the overlay renders:

| Caller | `onEdit` | `elementId` | Effect |
|---|---|---|---|
| `NarratorAssetTile` | yes | none → always `idle` | unchanged |
| `CharacterAssetTile` | yes | avatar node | the fix; uses `removeAffordance="corner"` |
| `ReferenceAssetTiles` | no | none | unchanged |

No caller combines `onEdit` with an overlay-style remove, so nothing stacks two `inset-0` buttons. The overlay is `opacity-0` at rest, so the shimmer and `GenerationIndicator` still read clearly; it only paints over them on hover/focus, which is when the user is asking for the affordance. Keyboard access comes along for free — it is a real `<button>` with `focus-visible:opacity-100`.

No visual/design-system change, so no `DESIGN.md` deviation.

## Validation

| Check | Result |
|---|---|
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:run` | pass — 114 files, 994 tests |
| `npm run test:e2e` | **skipped** — port 3000 was already in use by a running dev server |

The Playwright smoke suite covers public routes rather than the canvas, so it is not the check that would catch this either way. CI will run it.

### On test coverage

No unit test added, deliberately. The change is a JSX render guard with no logic seam, and a render test asserting the pencil appears mid-generation is not reachable with the current setup: there is no DOM test environment (no jsdom/happy-dom/testing-library — component tests here use `renderToStaticMarkup`), and `GenerationQueueProvider` intentionally coerces seeded state to `status: "idle"` on construction, so a `"generating"` status cannot exist at first render. Covering it would mean either exporting `GenerationQueueContext` purely for injection or adding a DOM test environment — both larger than the fix, and the first leaks internals for a test's benefit. Flagging rather than forcing it.

## Open PR overlap check

Diffed all 21 open PRs for `AssetTile.tsx` — no hits. The nearest neighbour is #508, which refactors `CharacterEditModal.tsx` internals (extracts a `useAvatarGeneration` hook). Different file, different theme, no conflict.

## Candidates skipped

- **#506** (`size: S`) — depends on landing #480's dependency graph and wants a new aggregate-status graph walker plus an elapsed-time policy decision. Not XS.
- **#424**, **#474**, **#359** — already covered by open PRs #438, #484, #488.
- **#259** (`size: XS`) — explicitly blocked on BYOK (#331).
- Everything else open is `size: M`/`L`/`XL`, or product/design-direction work (#492, #491, #385, #379, #331), which the runbook excludes.

Closes #505

### DIFF
--- a/app/components/canvas/elements/AssetTile.tsx
+++ b/app/components/canvas/elements/AssetTile.tsx
@@ -89,7 +89,7 @@ export function AssetTile({
 						<Icon className="h-3 w-3" />
 					</div>
 				)}
-				{onEdit && status !== "generating" && (
+				{onEdit && (
 					<OverlayButton
 						icon={Pencil}
 						label={`Edit ${name ?? "asset"}`}


umairnadeem (commented):
lgtm

PR #511, opened 2026-08-03
fix: nightly bug fixes 2026-08-03

umairnadeem:
## Bugs fixed

**The saved script grew an invisible character every session.** Content elements lead with a zero-width leaf so backspacing past the start of an element cannot delete the element by accident (#438): the first backspace eats the marker, and only a second one removes the element. `serializeElement` joined *all* leaf text, so that marker was written into the OSML on every autosave. On reload `createCanvasNode` prepends a fresh marker and the old one lands in the body, so each save/reload cycle stacked another one:

```
round 0:  <narration id="…">​hello</narration>
round 1:  <narration id="…">​​hello</narration>
round 2:  <narration id="…">​​​hello</narration>
```

`getElementBodyText` now drops the marker before serializing. Because it strips all of them, a project that already accumulated some repairs itself on its next save.

**Refine ops stranded the caret and broke the empty-state placeholder.** `updateNodeText` compared `Editor.string(element)` (marker + body) against the caller's raw text. Refine passes text without the marker, so the two never matched, the prefix fast path never hit, and the full-range replace took the marker with it. After a `set` op the element could no longer report itself empty — `ElementContainer` checks `Node.string(element) === ZERO_WIDTH_SPACE` — so clearing an element left it with no placeholder. It now re-attaches the marker itself rather than asking callers to, which also fixes an element that had already lost one.

Both paths now share one definition of what the marker is, so the streaming sync (which passes the marker) and refine (which doesn't) can no longer disagree about it.

**Typecheck was red on master.** The `imageSnapshot` fixture in `lib/project/__tests__/autosave.test.ts` was missing the required `pinned` field.

## Tests

- `osmlSerializer`: `getElementBodyText` across marker-only, marker + body, and already-accumulated text; a save/reload round trip asserting the script is byte-identical after 3 cycles; a reloaded empty element still reads as empty.
- `editorOps`: marker survives a replace, a clear, and a caller that re-sends text carrying the marker; a stripped element gets its marker back.
- `applyOps`: refine `set` keeps the marker when replacing and when clearing.
- The `content()` fixtures in `editorOps` and `applyOps` now mirror what `createCanvasNode` actually builds (marker leaf + body leaf). They previously used a single leaf, which is why neither bug was caught.

## Skipped

- `createCanvasNode` overwrites an incoming `model`/`provider` attr with the registry default, so a user-picked model would not survive a reload. Unobservable today — every connector ships exactly one model — and forcing the provider looks deliberate, so it is left alone rather than churned.
- `getVolume` maps a `volume=""` attribute to a silent 0 while `getLoops` clamps the same input up to 1. No path was found that produces an empty attribute.

## Open PR overlap check

Considered #509, #508, #507, #504, #503, #502, #501, #500, #499, #498, #497, #496, #495, #494, #493, #488, #487. None touch `lib/canvas/constants.ts`, `lib/canvas/editorOps.ts`, `lib/canvas/osmlSerializer.ts`, `lib/generation/inputs.ts`, `lib/script/refine/`, or `lib/project/__tests__/`. Nearest neighbours are #500 (`lib/canvas/guards.ts`, `osmlMetadata.ts`) and #498 (`lib/script/ScriptProvider.tsx`, `drainStream.ts`, `useStreamRun.ts`) — different files, and no behaviour they own is changed here.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, and `test` (1030 passing) all green.

### DIFF
--- a/app/components/canvas/hooks/useScriptSync.ts
+++ b/app/components/canvas/hooks/useScriptSync.ts
@@ -3,7 +3,7 @@ import { Editor, Transforms } from "slate";
 import { useScriptNodes } from "@/lib/script/ScriptProvider";
 import { findNodeById, updateNodeText } from "@/lib/canvas/editorOps";
 import { isParsedContentElement } from "@/lib/canvas/guards";
-import { getElementText } from "@/lib/canvas/osmlSerializer";
+import { getElementBodyText } from "@/lib/canvas/osmlSerializer";
 
 export function useScriptSync(editor: Editor): void {
 	const nodes = useScriptNodes();
@@ -15,7 +15,7 @@ export function useScriptSync(editor: Editor): void {
 			for (const node of nodes) {
 				if (!isParsedContentElement(node)) continue;
 
-				const nextText = getElementText(node);
+				const nextText = getElementBodyText(node);
 				if (!nextText) continue;
 
 				const entry = findNodeById(editor, node.id);
--- a/lib/canvas/__tests__/editorOps.test.ts
+++ b/lib/canvas/__tests__/editorOps.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "vitest";
 import { createEditor, Editor } from "slate";
 import type { CanvasContentElement, SceneElement } from "@/lib/canvas/types";
+import { ZERO_WIDTH_SPACE } from "../constants";
 import {
 	findElementById,
 	findNodeById,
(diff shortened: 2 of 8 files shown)


umairnadeem (commented):
lgtm. nice catch on the marker round-tripping into the OSML, and folding it into `withoutCaretMarker` so both paths agree is the right call.

Issue #490, opened 2026-07-30
Start from a blank canvas (skip the prompt)


umairnadeem:
## Problem

There is no way to reach an empty canvas. A new project always opens on the prompt box, and the only route out of it is an LLM generation.

`app/components/Editor.tsx:13-22` gates the entire editor on whether a script exists:

```tsx
const hasScript = useScriptHasContent() || loading;
…
{hasScript ? <PostPromptView /> : <PrePromptView />}
```

`hasContent` starts as `initialScript.length > 0` and is only ever flipped by `submitPrompt` streaming its first non-empty chunk (`lib/script/ScriptProvider.tsx:57,86-88`). So "New slop" → `createProject()` inserts a row with no script → `PrePromptView` → `ComposerHero`, and the canvas is unreachable until the LLM writes something into it.

That blocks every hand-authoring flow: building a video element by element, pasting in a script you already wrote (see the companion clipboard issue), reproducing a specific structure for a bug report, or using OpenSlop at all when you know exactly what you want and don't want a model's interpretation of it. It also means an accidental prompt is the only way in, and the LLM's output then has to be deleted before real work starts.

The editor itself is already ready for this. `withLayout` (`app/components/canvas/plugins/withLayout.ts:11-13`) normalizes an empty editor by inserting a `narration` element, and `withScenes`' `wrapOrphanContent` (`app/components/canvas/plugins/withScenes.ts:22-32`) wraps that orphan in a scene. Mount `PostPromptView` against an empty value and you get a valid one-scene, one-element canvas for free. The only thing missing is a way to say "I want the canvas, not the prompt".

## Reported by (verbatim feedback)

- "create issues in openslop to allow people to create blank canvases, copy/paste hooks to convert to/from OSML format from canvas"

## Relevant code

- `app/components/Editor.tsx:13-22` — the `hasScript ? PostPromptView : PrePromptView` gate; the single decision point
- `lib/script/ScriptProvider.tsx:20-24` — `ScriptControl` is `{ loading, submitPrompt, stopGeneration }`; needs a third entry point that flips `hasContent` with no stream
- `lib/script/ScriptProvider.tsx:57` — `hasContent` seeded from `initialScript.length > 0`
- `lib/script/ScriptProvider.tsx:79-88` — `submitPrompt`; `setHasContent(true)` currently only reachable from a streamed chunk
- `app/components/ComposerHero.tsx:28-55` — where a "Start blank" affordance sits, next to `TemplateGallery`
[...]

Issue #489, opened 2026-07-30
Scene skip buttons get stuck: segment lookup compares in seconds while seekTo rounds to frames


umairnadeem:
## Problem

The player's "Next scene" / "Previous scene" buttons can get permanently stuck on one scene: pressing "Next scene" twice lands on the same scene both times, and "Previous scene" jumps two scenes back instead of one.

The cause is a unit mismatch between how the buttons **seek** and how they **classify** the current position:

- `seekToAdjacentScene` seeks with `toFrames(target.start, fps)`, which **rounds** to a whole frame.
- It then classifies "which scene am I in?" with `findSegmentIndexAt(segments, toSeconds(currentFrame, fps))`, which compares in **raw seconds** against `seg.start + seg.duration`.

Segment starts are arbitrary reals — `buildVideoLayout` accumulates them from provider `durationSec` values, and only `transitionDurationSec` is snapped onto the frame grid (`lib/video/scene-builder.ts:93-96`). So when `start * fps` has a fractional part below `0.5`, `toFrames` rounds **down** and the seek lands a fraction of a frame *before* the segment it was aiming at. Because segments are exactly contiguous by construction — `buildSceneSegments` trims each segment by the transition overlap, so `prev.start + prev.duration === next.start` — `findSegmentIndexAt`'s strict `timeSec < seg.start + seg.duration` test then reports the **previous** index.

Worked example at 30fps, a scene starting at `5.04s`:

```
toFrames(5.04, 30)   = Math.round(151.2) = 151
toSeconds(151, 30)   = 5.0333…           < 5.04
findSegmentIndexAt(…) → previous segment
```

From there, "Next scene" recomputes `current` as `i - 1` and targets `segments[i]` again — the same frame it is already on. The playhead never advances past that boundary. Roughly half of all scene boundaries are affected (whichever way `start * fps` happens to round).

Same root cause has two other visible effects:

- `useActiveSegmentIndex` calls the same `findSegmentIndexAt`, so the `ScenePill` reads one scene behind right after a boundary seek.
- `PlayFromHereButton` (`app/components/canvas/elements/PlayFromHereButton.tsx:15`) seeks with the same rounded `toFrames(seq.start, fps)`, so "Play from here" can start a fraction of a frame inside the previous scene.

## Reported by (verbatim feedback)

- "any bugs you can spot with seekToAdjacentScene?"

## Relevant code

[...]

PR #514, opened 2026-08-03
fix: show an animated image as generating while its still frame runs

### LINKED ISSUE #506: Animated image still/video toggle doesn't show the still node's progress

umairnadeem:
## Summary

An `animated_image` is two nodes under the dependency graph: a still-image node feeding the animation node. `enqueueGraph` queues both, then runs the still first; the animation stays `"queued"` until the still settles. The card only ever read the animation's snapshot, so the element that does the most work was the one that looked idle — pulsing hourglass and "Queued…" for the larger share of the wall time, with no elapsed counter.

`nodeProgress(node, results)` joins the other graph walkers in `lib/generation/graph.ts`: it returns the snapshot of whichever node in the subtree is generating, falling back to the node's own. `useGenerate` reports that instead of the raw element snapshot, so `GenerationIndicator`, the shimmer, and the disabled state all inherit the fix through `ElementGenerationProvider` with no change of their own.

Nothing in the UI learns that animated images are two jobs, and any element that grows a derived node later gets an honest status for free.

#506 — size S, priority medium bug, with the change point named in the issue (`useGenerate.ts:13`) and a suggested direction that solves the class rather than special-casing `animated_image`. Three files, one new pure function, no queue-mechanics change.

## Behavior

- Animated image reads "Generating Ns" while its still frame runs.
- Still reads "Queued…" when genuinely waiting on a batch slot with nothing of its own running (covered by a test).
- Single-node elements are byte-identical: a node with no dependencies short-circuits to its own snapshot.
- Result, error, staleness and every write path are untouched — only the reported status and elapsed seconds change.

## Deliberately not done

The elapsed counter still restarts at the still → animation boundary. `queue.commit` zeroes `seconds` when a node settles, so accumulating across the subtree needs a queue-side change, and `lib/generation/queue.ts` is under open PR #507. The issue permits either option ("the simpler of the two is fine"), so this takes the one that shows the elapsed time of whichever node is running. The counter now runs during the still instead of not running at all.

## Validation

- `npx vitest run lib/generation/` — 6 files, 90 tests pass (3 new `nodeProgress` cases at the graph/queue seam)
- `npm test` — 117 files, 1024 tests pass
- `npm run lint`, `npm run format:check`, `npm run knip`, `npm run build` — all pass
- `npm run typecheck` — one **pre-existing** failure on `master`, unrelated to this diff: `lib/project/__tests__/autosave.test.ts(22,63)` is missing `pinned` on an `ElementSnapshot` fixture. That file is untouched here and is already being fixed by open PR #511.

## Open PR overlap check

Diffed all 20 open PRs by file. This PR touches `lib/generation/graph.ts`, `lib/generation/__tests__/graph.test.ts`, and `app/components/canvas/hooks/useGenerate.ts` — none appear in any of them. Nearest neighbours, both avoided: #507 (`lib/generation/queue.ts` + its two tests, including hunks inside `commit`) and #508 (`ElementGenerationContext.tsx` and the element card, which only consume `useGenerate`'s return value).

## Issues skipped this run

- #489 (frame-vs-seconds scene skip) — well scoped, but its three named files are each under an open PR: `BottomTransportBar.tsx` (#495), `SegmentedSeekBar.tsx` (#499), `useSceneSegments.ts` + `scene-builder.ts` (#487).
- #505 — already implemented by open PR #509.
- #259 — blocked on BYOK (#331).
- Everything else open is size M+ or product/design direction (#490, #492, #461, #451, #385, …).

Closes #506

### DIFF
--- a/app/components/canvas/hooks/useGenerate.ts
+++ b/app/components/canvas/hooks/useGenerate.ts
@@ -3,7 +3,12 @@ import {
 	useGenerationQueue,
 	useQueueSelector,
 } from "@/lib/generation/GenerationQueueProvider";
-import { forElement, isNodeStale, nodeInputs } from "@/lib/generation/graph";
+import {
+	forElement,
+	isNodeStale,
+	nodeInputs,
+	nodeProgress,
+} from "@/lib/generation/graph";
 import { useNodeBuilder } from "@/lib/generation/useNodeBuilder";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 
@@ -16,6 +21,7 @@ export function useGenerate(element: CanvasContentElement) {
 		[buildNode, element],
 	);
 	const stale = useQueueSelector((q) => isNodeStale(node, q));
+	const progress = useQueueSelector((q) => nodeProgress(node, q));
 
 	useEffect(() => {
 		if (!stale) return;
@@ -36,8 +42,8 @@ export function useGenerate(element: CanvasContentElement) {
 
 	return {
 		node,
-		status: snapshot.status,
-		seconds: snapshot.seconds,
+		status: progress.status,
+		seconds: progress.seconds,
 		result: snapshot.result,
 		error: snapshot.error,
 		stale,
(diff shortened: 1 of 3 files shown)


umairnadeem (commented):
lgtm, solving it at the graph level instead of special-casing animated_image is the right shape.

umairnadeem:
incorrect approach. we should show loading for the actual node that's loading

PR #515, opened 2026-08-04
perf: nightly performance optimizations 2026-08-04

umairnadeem:
## What

Typing one character used to re-render the entire editor view.

`useEditorSession` held the Slate document in React state at the top of `PostPromptView`. Slate fires `onChange` for every operation, `setValue` swapped in the new `editor.children`, and everything under that component re-rendered: the seven providers in `CanvasProviders`, `EditorToolbar`, `EditorSidebar`, the Remotion `PlayerPanel` with its scrub bar and one seek segment per scene, `BottomTransportBar`, and `ProjectTitle`. The cost scaled with scene count, on the most latency-sensitive interaction in the app. `EditorToolbar` and `EditorSidebar` already carried `memo` to blunt it, which treated the symptom at two leaves and left the rest.

The only thing the view actually needs from the document is `layoutKey` — a string over element ids, types and the four layout-affecting attributes. Text edits never change it.

So the document stays in the editor:

- `withDocumentSignal` (new Slate plugin) exposes `subscribeToDocument`. Slate rebuilds `editor.children` for every document operation and leaves it alone for selection moves, so array identity is the signal.
- `useLayoutKey` reads the key through that subscription with `useSyncExternalStore`. Same string in, no re-render — this is the walkthrough's "subscribe to a derived value, not to the editor" guidance applied above `<Slate>`, where `useSlateSelector` can't reach.
- `CanvasSession` (new) owns the document snapshot and the autosave that follows it, next to the `Canvas` that renders from them.

A keystroke now re-renders `CanvasSession` and the Slate subtree. Adding, removing, reordering or re-typing an element still re-renders the view, because at that point the layout genuinely changed.

No behavior change. Autosave, rehydration, script sync and metadata sync all fire in the same order as before.

## Skipped

- The `memo` on `EditorToolbar` / `EditorSidebar` is now redundant. Left alone: `EditorToolbar.tsx` is open in #498, and removing a guard is not worth the conflict.
- `useNodeBuilder` subscribes to the whole project store, so any store write rebuilds every element's generation node. Real, but it lands on `useGenerate.ts` / `graph.ts`, both open in #514.
- Remotion's composition, `MotionLayer` and `Captions` are already tight — memoized nodes, per-frame subscriptions scoped to the component that needs the frame, binary-searched caption lookup. Nothing worth changing.

## Open PR overlap check

Considered #493, #495, #496, #497, #498, #499, #500, #501, #502, #503, #504, #507, #508, #509, #511, #512, #513, #514 and diffed each one's files.

Non-overlapping. The video components (#495, #499, #503), `Canvas.tsx` (#502), `useDragAndDrop.ts` (#513), the element cards (#508, #509), the canvas hooks `useScriptSync` / `useMetadataSync` / `useGenerate` (#500, #514) and the generation queue (#507, #511) are all untouched here. #513 is the closest in theme — it steadies dnd-kit's sortable items against the same per-keystroke `value` churn, one layer below this. The two compose; neither file is shared.

## Checks

`npm run lint`, `format:check`, `typecheck`, `build`, `test:run` (1030 passed) all green. `knip` flags `lib/connectors/tts/voiceDescriptor.ts`, which belongs to another branch's uncommitted work in the shared tree, not to this change.

### DIFF
--- a/app/components/PostPromptView.tsx
+++ b/app/components/PostPromptView.tsx
@@ -2,7 +2,7 @@
 
 import UserProfile from "./UserProfile";
 import { EditorToolbar } from "./EditorToolbar";
-import Canvas from "./canvas/Canvas";
+import { CanvasSession } from "./canvas/CanvasSession";
 import { CanvasProviders } from "./canvas/CanvasProviders";
 import { EditorSidebar } from "./canvas/panel/EditorSidebar";
 import { ProjectTitle } from "./canvas/ProjectTitle";
@@ -15,7 +15,7 @@ import {
 } from "./video/PlayerPositionContext";
 
 function PostPromptViewInner() {
-	const { editor, value, setValue, layoutKey } = useEditorSession();
+	const { editor, layoutKey } = useEditorSession();
 	const { position, visible } = usePlayerPosition();
 	const isTop = position === "top";
 
@@ -42,7 +42,7 @@ function PostPromptViewInner() {
 								>
 									<div className="mx-auto max-w-6xl px-4 py-4">
 										<ProjectTitle />
-										<Canvas editor={editor} value={value} setValue={setValue} />
+										<CanvasSession editor={editor} />
 									</div>
 								</div>
 
--- a/app/components/canvas/CanvasSession.tsx
+++ b/app/components/canvas/CanvasSession.tsx
@@ -1,0 +1,24 @@
+"use client";
+
+import { useState } from "react";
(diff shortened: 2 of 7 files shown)


umairnadeem (commented):
lgtm. moving the doc state down + subscribing to the derived layout key is the right shape here, and the withDocumentSignal tests cover the selection-only case which is the part that could have bitten.

umairnadeem:
no noticeable perf gains

PR #516, opened 2026-08-06
refactor: nightly maintainability 2026-08-06

umairnadeem:
## What

Two architecture items and one small cleanup. No behavior change.

### 1. `lib/generation`: break the graph ↔ queue cycle, model nodes as a union

`graph.ts` imported `GenerationJob` and `GenerationQueue` from `queue.ts` while `queue.ts` imported most of `graph.ts`. The dependency now points one way, queue → graph:

- `GenerationJob` moves to `graph.ts`. It is what a node carries, not queue behavior; `resolveGraph` builds it and the queue consumes it.
- `NodeResults` was `Pick<GenerationQueue, "getElementSnapshot">` — the graph reaching back into the class that consumes it. It is now the narrow read contract the graph actually needs (`result`, `resultInputs`, `pinned`), which `GenerationQueue` satisfies structurally. Plugins and helpers that only read results (`characterAvatarUrl`, `still-frame`, the LLM style plugins) depend on that contract instead of the queue.

`GenerationNode` becomes `SourceNode | JobNode` instead of one shape with a nullable `buildJob`:

- `buildJob: (() => GenerationJob) | null` was a thunk over a value built eagerly one line earlier, so the laziness bought nothing. It is now `job: GenerationJob` (or `null` on a source node).
- `isSourceNode` is a type guard, so the queue's job paths narrow statically. `pending` is `JobNode[]`, `runJob` takes a `JobNode`, and `enqueueGraph` reads `node.job` directly.
- `requireJob` and its runtime throw are gone from three of four call sites. The fourth, `commitResult`, is a public entry point that can be handed any node, so it keeps an explicit check and fails loudly there.

### 2. One way for a plugin to read project state

Asset plugins were handed `ctx.state` (declared, staleness-tracked), while `project-metadata`, `reference-style`, and `character-avatar-style` reached for the ambient `getProjectStore(projectId)` — only because the LLM path had no way to pass state: `generate(params, context)` took a `GenerationContext` but `stream(params)` did not.

`stream` now takes the same optional `GenerationContext` as `generate`, and the two callers that start an LLM run (`ScriptProvider.submitPrompt`, `useRefineScript`) pass the state snapshot. The three plugins use `requireState(ctx, name)` like every other plugin. `createProjectMetadataPlugin(projectId)` had no other reason to be a factory and is now a plain `projectMetadataPlugin` constant; the other two lose their `projectId` parameter.

Net effect: a plugin is a function of the context it is given, the ambient store read moves up to the orchestration boundary that already owns it, and reads within one LLM run see one consistent snapshot instead of re-reading the store between awaited gateway calls.

`voice-hydrate` keeps its store handle: it writes.

### 3. `ElementContainer.tsx` parent-scene cast

`(Node.parent(editor, path) as SceneElement).id` → `parentSceneId(editor, path)`. Second and last site of the cast PR #512 replaced in `SortableContent.tsx`; it was left behind then because the file was under an open PR.

### Docs

`ARCHITECTURE.md`'s "Editor state" bullet listed connector plugins as `getProjectStore` callers. That is no longer true and would have sent the next reader the wrong way, so it now states the contract: plugins read `ctx.state`, `voice-hydrate` is the one exception because it writes.

## Complexity delta

- 1 module cycle removed; `graph.ts` now imports nothing from `queue.ts`.
- 1 nullable + 1 pointless thunk removed from the node type; 3 of 4 runtime `requireJob` throws replaced by static narrowing; `requireJob` deleted.
- 1 unchecked `as` cast removed.
- 3 plugins drop a `projectId` parameter and an ambient store read; 1 factory collapses to a constant.
- `generate`/`stream` no longer diverge on context.

## Validation

`lint`, `format:check`, `typecheck`, `knip`, `build` all clean. `1068/1068` tests pass.

## Open PR overlap check

There were no open PRs in the repo at the time of this pass, so nothing to overlap with.

## Skipped, with reasons

- **`lib/project/store.ts` module-level `Map<projectId, store>`** with a zero-caller `clearProjectStore`, and `projectId` threaded through ~12 modules as its lookup key. Creating the store in `ProjectEditor` and passing the handle is the right shape, but it changes store lifecycle and is behavior-risky. Wants a human.
- **`AttributeBadge` / `TextAttributePopover` label divergence** (still open from #512): the text-edit path re-implements label rendering and ignores `spec.icon`/`spec.unit`. Unifying changes what renders, and this pass is behavior-preserving.
- **`lib/api/with-auth.ts` returning `stringifyError` to clients** and **`lib/supabase/env.ts` empty-string env defaults**. Both are judgement calls about what to expose or when to fail, not refactors.
- **The `Object.keys(X) as K[]` cast sites** (6 of them). A `keysOf` helper on its own is churn; it wants a larger host change.


umairnadeem (commented):
lgtm

PR #518, opened 2026-08-06
refactor: route every user-facing error through toastError

umairnadeem:
Nightly CONVENTIONS.md audit. The rolling audit state is saturated, so this was a delta pass over the files merged since the last sweep (#504, #509, #511, #512), skipping everything under open PRs #516 and #517.

Both findings are the same rule pair: **fail loudly** and **one canonical way**. PR #504 made `toastError` the single path an error takes to the user; two sites never got moved onto it.

### `app/components/canvas/hooks/useAutosave.ts`

```ts
onError: () => toast.error("Save failed", { ...TOAST_OPTIONS, duration: 4000 }),
```

The cause was discarded at the callback boundary. A save that failed on a 413, an auth expiry, or a serialization bug all rendered the same three words, and the only record of which was a `console.error` in `autosave.ts` the user never sees.

The reason this site was left behind is that it passes custom sonner options, which `toastError` had no way to accept. So `toastError` now takes an optional third `ExternalToast` argument. It stays the sole owner of *what a human reads*; the options only style the toast.

### `app/components/UserProfile.tsx`

```ts
toast.error(errorMessage(cause));
```

Hand-rolled `toastError`. Exactly equivalent, so this is a pure no-behavior-change consolidation, and it removes the last place UI code had to pick between `errorMessage` and `stringifyError` by hand.

### Behavior change

One, deliberate: a failed autosave now reads `Save failed: <reason>` instead of `Save failed`. That matches `ProjectsList`'s existing `Could not create project: <reason>` phrasing.

Covered by a new `toastError` options-passthrough test. lint, typecheck, format:check, 1069 tests and `npm run build` all pass.

### DIFF
--- a/app/components/UserProfile.tsx
+++ b/app/components/UserProfile.tsx
@@ -2,9 +2,8 @@
 
 import { memo, useState } from "react";
 import { useRouter } from "next/navigation";
-import { toast } from "sonner";
 import { signOut } from "@/lib/auth/session";
-import { errorMessage } from "@/lib/errors";
+import { toastError } from "@/lib/toastError";
 import { useUser } from "@/lib/user/UserProvider";
 import { useTheme } from "next-themes";
 import { Avatar, AvatarFallback, AvatarImage } from "@/components/ui/avatar";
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
lgtm

Issue #451, opened 2026-07-22
Multiple generations per element: variant picker + persistent generation history with time travel


umairnadeem:
## Problem

An element has exactly one output at a time, and every regeneration destroys the previous one. Today there is no way to generate several candidates for a scene and pick the best, and no way to get back a generation you liked after regenerating past it.

The current model:

- `ElementSnapshot` holds a single `result: AssetResult | null` plus the `resultInputs` that produced it. Committing a new result overwrites it in place.
- `GenerationQueue` already keeps a private `history: Map<elementId, Map<inputsKey, AssetResult>>`, keyed by `serializeInputs(inputs)`. But it stores **one result per distinct input set**, so regenerating with unchanged inputs overwrites the entry — the exact "give me another take" case.
- That history is **in-memory only**. Only `snapshot()` (the flat `Record<elementId, ElementSnapshot>`) is persisted as `project.generation` and rehydrated into the queue. Reload the project and every prior generation is gone.
- `restoreResult` is the only reader, and it is wired purely as an automatic cache hit when inputs change back to a previously-seen set. Nothing user-facing ever calls it.

So the building blocks for "remember what we generated" partly exist, but they are private, lossy, unpersisted, and have no UI.

## Reported by (verbatim feedback)

- "we should be able to generate multiple images for a given element and have a UX to choose between different generations. We should also never lose historical generations we've already made for an element and should be able to time-travel between different inputs/outputs for a given element via a polished UI/UX."

## Relevant code

- `lib/generation/queue.ts:18` — `ElementSnapshot`, single `result` + `resultInputs`
- `lib/generation/queue.ts:63` — private in-memory `history` map, one `AssetResult` per input key
- `lib/generation/queue.ts:235` — `commitResult` overwrites the history entry for the same input key
- `lib/generation/queue.ts:257` — `restoreResult`, the only history reader
- `lib/generation/queue.ts:135` — `snapshot()` persists only the flat per-element state, never the history
- `app/components/canvas/hooks/useGenerate.ts:27` — the sole `restoreResult` caller, an automatic inputs-changed cache hit
- `app/projects/[id]/page.tsx:31` + `app/projects/[id]/ProjectEditor.tsx:39` — `project.generation` hydrates the queue; the persisted shape has to grow for history to survive reload
- `app/components/canvas/elements/OutputPreview.tsx` — where a chooser would live
[...]

Issue #385, opened 2026-06-27
Thorough accessibility (a11y) audit + fixes (WCAG 2.1 AA)


umairnadeem:
## Problem

The app has never had a dedicated accessibility pass. The foundations are decent — Radix primitives bring focus traps/roles, there's a `focus-ring` utility (`app/globals.css:8-10`), an `IconButton` that *requires* `ariaLabel` (`components/ui/icon-button.tsx`), and a few critical paths use `aria-live`. But adoption is inconsistent and there's **no automated a11y safety net**, so regressions land unnoticed. We need a **thorough accessibility audit** of the whole app (target WCAG 2.1 AA) and a tracked set of fixes.

## Reported by (user feedback)

- "need to do a thorough a11y audit"

## Current gaps (representative, from a first-pass survey)

- **No a11y linting/testing.** `eslint.config.mjs` extends only `next`/`core-web-vitals` — **no `eslint-plugin-jsx-a11y`**, no axe in CI. This is the highest-leverage gap.
- **Keyboard operability.** Mouse-only interactions: `app/components/video/ResizeHandle.tsx:21` (`onMouseDown` resize, no Tab/arrow support); dnd-kit drag-reordering in `app/components/canvas/Canvas.tsx` has no keyboard path.
- **Forms / labels.** Placeholder-as-label without an associated `<label>`/`aria-label`: `app/components/canvas/elements/character/NewCharacterDialog.tsx:69`, `app/components/canvas/elements/character/VoiceMetadataFields.tsx:79`.
- **Live regions.** Generation progress isn't announced — `app/components/EditorToolbar.tsx:50-100` ("Writing…/Generating…" state changes have no `aria-live`); verify Sonner toasts announce.
- **Focus on open.** Radix traps focus, but dialogs don't set initial focus on a meaningful control (e.g. `NewCharacterDialog`).
- **Landmarks / headings.** Editor/Canvas content isn't wrapped in `<main>`; occasional heading jumps (h1→h3).
- **Icon-only controls.** Icons render as `aria-hidden` spans (`components/ui/icon.tsx`), so any icon-only control MUST carry a label — mostly handled via `IconButton`, but one-off `<button>`s need an audit for consistency.
- **Color contrast.** Token system is defined (`app/globals.css:78-131`) but unverified; `--muted-foreground` on `--muted` is a likely marginal pair. `DESIGN.md` has no contrast/a11y guidance.

## Suggested direction

Tackle the class of problem, not one-off fixes:

[...]

PR #519, opened 2026-08-06
fix: scene skip buttons get stuck at non-frame-aligned scene boundaries

### LINKED ISSUE #489: Scene skip buttons get stuck: segment lookup compares in seconds while seekTo rounds to frames

umairnadeem:
## Summary

Scene segment lookup compared raw seconds while every seek rounds to a whole frame. Segment starts are arbitrary reals, so when `start * fps` had a fractional part below `0.5`, `toFrames` rounded down and the seek landed a fraction of a frame before the segment it was aiming at. Since segments are exactly contiguous, `findSegmentIndexAt` then reported the *previous* index, "Next scene" recomputed the same target, and the playhead never advanced past that boundary.

The fix classifies on the same grid the seek uses:

- `findSegmentIndexAt(segments, timeSec)` becomes `findSegmentIndexAtFrame(segments, frame, fps)`, comparing `frame < toFrames(seg.start + seg.duration, fps)`. Because segments are contiguous, `toFrames(prev.start + prev.duration, fps) === toFrames(next.start, fps)`, so a seek to `toFrames(next.start, fps)` lands exactly on the boundary frame and classifies as `next`. No epsilon.
- `seekToAdjacentScene` and `useActiveSegmentIndex` pass `getCurrentFrame()` directly and stop converting to seconds. `ScenePill` / `ActiveSceneSync` read the scene that is actually playing right after a boundary seek.
- `SegmentedSeekBar` converts its hover ratio once, through the same `toScrubFrame` helper its scrub already used, so the tooltip labels the segment a scrub would actually land in.

`PlayFromHereButton` needs no change: it already seeks `toFrames(seq.start, fps)`, which now classifies into the intended scene by construction. The stray `console.log` mentioned in the issue is no longer present on master.

Size S, priority high bug with a clear root cause, a suggested direction in the issue, and three callers all in `app/components/video/`. Pure unit-of-comparison change; no behavior added, no UI/visual change, no design-system surface touched.

## Validation

- `npx vitest run app/components/video/__tests__/useSceneSegments.test.ts` — 18 passed
- `npm run build` — success
- `npm run test` — 123 files, 1074 tests passed

Added coverage in `app/components/video/__tests__/useSceneSegments.test.ts`: the existing lookup cases restated in frame space, plus two parameterized cases at 24/25/30fps — a deliberately non-frame-aligned start round-tripping through `toFrames` and resolving to its own segment, and a walk that steps exactly one segment per seek across a run of unaligned starts (the stuck-button repro).

Not validated: manual playback against a real project with fractional provider durations, which needs a generated project.

## Open PR overlap check

Open PRs #516, #517, #518 touch `lib/connectors/*`, `lib/generation/*`, `lib/script/*`, `app/projects/[id]/page.tsx`, `proxy.ts`, `lib/toastError.ts`, `app/components/UserProfile.tsx`, `app/components/canvas/{elements/ElementContainer,hooks/*}`. No overlap with the four `app/components/video/` files changed here.

## Skipped candidates

Everything else open is size M/L/XL or product-decision heavy (#492, #491, #490, #462, #454, #451, #385, #379 …). The only XS issue, #259, is explicitly blocked on BYOK (#331). #506 (size S) is a generation-state bug needing live provider queue behavior to validate.

Closes #489

### DIFF
--- a/app/components/video/BottomTransportBar.tsx
+++ b/app/components/video/BottomTransportBar.tsx
@@ -3,11 +3,11 @@
 import { memo } from "react";
 import { ChevronsLeft, ChevronsRight, Pause, Play } from "@/components/ui/icon";
 import { TooltipIconButton } from "@/components/ui/icon-button";
-import { toFrames, toSeconds } from "@/lib/video/frames";
+import { toFrames } from "@/lib/video/frames";
 import { usePlayerControl } from "./PlayerControlContext";
 import { usePlayerPosition } from "./PlayerPositionContext";
 import { useLayout } from "./VideoLayoutContext";
-import { findSegmentIndexAt } from "./useSceneSegments";
+import { findSegmentIndexAtFrame } from "./useSceneSegments";
 import { usePlayerPlaying } from "./usePlayerState";
 import { SegmentedSeekBar } from "./SegmentedSeekBar";
 import {
@@ -27,9 +27,10 @@ function BottomTransportBarComponent() {
 
 	const seekToAdjacentScene = (dir: -1 | 1) => {
 		if (!player || segments.length === 0) return;
-		const current = findSegmentIndexAt(
+		const current = findSegmentIndexAtFrame(
 			segments,
-			toSeconds(player.getCurrentFrame(), layout.fps),
+			player.getCurrentFrame(),
+			layout.fps,
 		);
 		const target = segments[current + dir];
 		if (target) player.seekTo(toFrames(target.start, layout.fps));
--- a/app/components/video/SegmentedSeekBar.tsx
+++ b/app/components/video/SegmentedSeekBar.tsx
@@ -6,7 +6,7 @@ import { clamp } from "@/lib/utils";
 import type { VideoLayout } from "@/lib/video/types";
 import { usePlayerFrame } from "./usePlayerState";
 import { SeekTooltip } from "./SeekTooltip";
-import { findSegmentIndexAt, type SceneSegment } from "./useSceneSegments";
+import { findSegmentIndexAtFrame, type SceneSegment } from "./useSceneSegments";
 import { ScrubBar, type ScrubHover } from "./ScrubBar";
 import { silenceMediaIn } from "./silenceMedia";
 
@@ -22,6 +22,7 @@ export function SegmentedSeekBar({
 	segments: SceneSegment[];
 }) {
 	const { totalDurationSec, totalFrames } = layout;
+	const toScrubFrame = (ratio: number) => Math.round(ratio * (totalFrames - 1));
 	const frame = usePlayerFrame(player);
 	const progress = clamp(frame / Math.max(1, totalFrames - 1), 0, 1);
 
@@ -57,7 +58,11 @@ export function SegmentedSeekBar({
 			setSettledIndex(null);
 			return;
 		}
-		const index = findSegmentIndexAt(segments, h.ratio * totalDurationSec);
+		const index = findSegmentIndexAtFrame(
+			segments,
+			toScrubFrame(h.ratio),
+			layout.fps,
+		);
 		setHoverIndex(index);
 		settleTimerRef.current = setTimeout(
 			() => setSettledIndex(index),
@@ -77,7 +82,7 @@ export function SegmentedSeekBar({
 			segments={scrubSegments}
 			onScrub={(ratio) => {
 				if (!player) return;
-				player.seekTo(Math.round(ratio * (totalFrames - 1)));
+				player.seekTo(toScrubFrame(ratio));
 				if (wasPlayingRef.current) silenceMediaIn(player.getContainerNode());
 			}}
 			onScrubStart={() => {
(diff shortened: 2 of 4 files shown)


umairnadeem (commented):
lgtm

Issue #510, opened 2026-08-02
Canvas perf: typing and drag degrade with document size (no Slate chunking, drag state fans out to every card)


umairnadeem:
## Problem

Typing, editing and dragging in the canvas get progressively slower as a slop grows. Audited the editor against the [Slate performance walkthrough](https://docs.slatejs.org/walkthroughs/09-performance) (we're on `slate`/`slate-react` 0.123.0, which supports every technique it describes).

Two things the walkthrough leads with are simply not implemented, and one context in the drag path re-renders every element card on every pointer move.

First, what is **not** the problem — measured, so we don't waste a PR on it. The per-change pure work on a synthetic doc (Node 22, warm, 200 iters/measurement):

| doc size | `sceneItems` filter+map | `getLayoutKey(getContentElements(value))` | `serializeOSMLWithScenes` |
|---|---|---|---|
| 40 elements | 0.009 ms | 0.054 ms | 0.144 ms |
| 180 elements | 0.010 ms | 0.098 ms | 0.279 ms |
| 480 elements | 0.015 ms | 0.223 ms | 0.669 ms |

Sub-millisecond even at 480 elements, and autosave serialization is debounced anyway. The cost is **React render volume and DOM paint over N element cards**, not the document walks. Optimizing those walks would be the trivial-gain PR; don't.

### 1. Chunking is not enabled (the walkthrough's headline, "10x in ideal circumstances")

`editor.getChunkSize` and `renderChunk` are set nowhere in the repo. Without chunking, `slate-react` re-runs the editor-level `useChildren` on **every** change and, per `node_modules/slate-react/dist/index.js`:

- creates N `ElementContext.Provider` + `MemoizedElement` React elements (one per top-level scene),
- performs `2 × children.length` WeakMap writes (`NODE_TO_INDEX` / `NODE_TO_PARENT`) — this is explicitly the `if (!chunking)` branch,
- runs N memo comparators, and React reconciles N children.

`MemoizedElement` does compare on `element` identity, so unedited scenes don't re-render their subtree — but the O(N) per-keystroke reconciliation above happens regardless. Chunking is what removes it: only the modified chunk re-renders.

### 2. No `content-visibility: auto` on offscreen content

[...]

PR #520, opened 2026-08-07
perf: fix canvas typing and drag stalls

umairnadeem:
Fixes the typing and drag hangs reported in #510.

All numbers are from a **production build** (`next start`) — the dev server inflates a keystroke to ~340ms with StrictMode double-renders and swamps the signal. Attribution is from CPU profiles plus render counters compiled into the build, not inference.

Three distinct causes, measured one at a time. **None of them is the Slate reconciliation the issue assumed.**

---

## 1. Typing: `sceneItems` identity churn fans out to every card

`sceneItems` was memoized on `value`, which Slate replaces on every keystroke. That handed `SortableContext` a new `items` identity per character. dnd-kit context updates bypass `React.memo` *and* Slate's element memoization, so all `useSortable` consumers re-rendered and rebuilt their card subtrees.

Fixed by keying on the scene ids, so the identity only changes when scenes are added, removed, or reordered.

Typing, 20 keystrokes, 37 scenes / 242 elements:

| | before | after |
|---|---|---|
| `SortableItem` renders | 4840 (242/key) | 200 (10/key) |
| long tasks | 22 | 1 |
| blocking time | 1424 ms | **71 ms** |

## 2. Drag moves: style recalc and paint, not JS

A CPU profile of a full drag showed only ~64ms in script; the rest was `(program)` — browser-internal style recalc and paint across 21221px of content in an 868px viewport. `content-visibility: auto` on scene containers skips the offscreen ones.

Drag, 25 pointer moves (A/B'd by toggling `content-visibility` on the same build, 3 runs):

| | before | after |
|---|---|---|
| long tasks | 4 | 0 |
| max frame | 245 ms | 48 ms |
| main-thread busy | 1.4–1.7 s | **0.34–0.52 s** |
| style recalcs | 211–256 | 74–87 |

## 3. Drag start: every card rebuilt its whole subtree

Pressing the grip on a 132-scene / 862-element slop fired **one 2925ms synchronous task** (2.76s script, 4ms layout, 46ms style); the overlay appeared the moment it ended.

`useSortable` subscribes to dnd-kit context, which changes on drag start, so all 862 `SortableItem`s re-render — inherent to dnd-kit. The cost was what sat *inside* them: `SortableItem` called `renderCanvasElement` and built `SortableActions` in its own render body, so every card reconstructed its settings popover, tooltips, and an 11-item Radix insert menu. None of that depends on drag state.

Both are now built by the caller (which doesn't re-render on drag) and passed down as props. Element references stay identical across `SortableItem`'s re-render, so React reuses the subtrees — **no `memo()` involved**. `SortableActions` splits into `InsertMenu` (caller-built) and `DragHandle` (needs `listeners`); `isMenuOpen` moves up with the menu.

Drag start, 132 scenes / 862 elements:

| | before | after |
|---|---|---|
| overlay latency | 2381–3116 ms | **340–488 ms** (median 383) |
| `ContentElement` renders | 1202 | **1** |
| `ElementSettings` renders | 1202 | **1** |
| `InsertMenu` renders | 1200 | **0** |

---

## Corrections to the issue's plan

- **Chunking (#1, the "10x headline") is not the typing fix and is not included.** `SortableContent` renders exactly 2× per keystroke both before and after — Slate's `MemoizedElement` was already confining element re-renders to the edited node. Reconciliation never fanned out over scenes.
- **`content-visibility` (#2) belongs on the drag path, not typing.** It measured as a no-op for typing once the render fan-out was fixed, but it is the entire drag-move win.
- **`pointerWithin` / droppable count (#5) was not the drag-start cost either** — drag start was JS building card subtrees, and layout was 4ms.

## Known remaining cost

Drag start still spends ~383ms in 1729 `SortableItem` renders at ~0.22ms each — `useSortable`'s own hook cost across two React passes. Cutting that means not registering every element as a droppable, which risks cross-scene drop correctness and deserves its own measurement rather than being folded in here.

## Verification

- Hover actions render, insert menu opens with its items, `menuOpen` class applies and clears.
- A real drag reorders correctly with structure intact (132 scenes / 862 elements).
- Deep-scrolled scenes render fully (video, waveforms, prompts); caret placement works inside `content-visibility`-skipped subtrees.
- Lint, format, typecheck, `npm run build`, 1086 tests pass.

## Not included

No unit test for the identity invariant. The repo has no jsdom/testing-library setup — tests are pure-logic only — so covering this seam would mean adding React test infra as new dependencies. Flagging rather than expanding scope; the invariant is documented at the call site.

### DIFF
--- a/app/components/canvas/styles/sortable.module.css
+++ b/app/components/canvas/styles/sortable.module.css
@@ -3,11 +3,8 @@
 	z-index: 0;
 }
 
-/*
- * Offscreen scenes used to pay style recalc and paint on every pointer move of
- * a drag. `auto` caches each scene's real height once rendered, so the
- * scrollbar settles after first paint.
- */
+/* Skips style recalc and paint for offscreen scenes during a drag. `auto`
+   caches each scene's real height, so the scrollbar settles after first paint. */
 .scene {
 	content-visibility: auto;
 	contain-intrinsic-size: auto 600px;
--- a/app/components/canvas/styles/sortable.module.css
+++ b/app/components/canvas/styles/sortable.module.css
@@ -3,6 +3,16 @@
 	z-index: 0;
 }
 
+/*
+ * Offscreen scenes used to pay style recalc and paint on every pointer move of
+ * a drag. `auto` caches each scene's real height once rendered, so the
+ * scrollbar settles after first paint.
+ */
+.scene {
+	content-visibility: auto;
+	contain-intrinsic-size: auto 600px;
+}
+
 .sortable > div:last-child {
 	flex: 1;
 	min-width: 0;
--- a/app/components/canvas/dnd/SortableActions.tsx
+++ b/app/components/canvas/dnd/SortableActions.tsx
@@ -2,6 +2,9 @@ import { GripVertical, Plus } from "@/components/ui/icon";
 import type { SyntheticListenerMap } from "@dnd-kit/core/dist/hooks/utilities";
 import { ActionMenu } from "@/components/ui/action-menu";
 
+const ACTION_BUTTON_CLASS =
+	"inline-flex items-center rounded-md p-0.5 text-muted-foreground hover:text-foreground hover:bg-muted transition-[color,background-color] duration-200";
+
 export interface InsertOption<K extends string = string> {
 	key: K;
 	label: string;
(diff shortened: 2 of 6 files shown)


app/components/canvas/styles/sortable.module.css
```diff
@@ -3,6 +3,13 @@
 	z-index: 0;
 }
 
+/* Skips style recalc and paint for offscreen scenes during a drag. `auto`
+   caches each scene's real height, so the scrollbar settles after first paint. */
+.scene {
+	content-visibility: auto;
```
umairnadeem, line 9: this breaks `scrollToScene`, which is `behavior: "smooth"` (autoscroll during playback + the jump-to-active-scene button). smooth scroll resolves the target against the 600px estimates, so it settles thousands of px off and the scene never comes on screen.

repro'd standalone in chrome: 60 blocks, `contain-intrinsic-size: auto 600px`, real heights 200-1500px, `scrollIntoView({behavior:'smooth', block:'center'})` on #40 → target lands 6085px from center. same page with `content-visibility: visible` → 0px off. `behavior: 'instant'` is only ~190px off, so switching scrollToScene to instant would mostly dodge it.

PR #127, opened 2026-05-05
perf: nightly performance optimizations 2026-05-05

umairnadeem:
## Summary
- reduce unnecessary Slate/React work in canvas render hot paths
- coalesce generation queue elapsed-time timers to lower update churn under concurrent jobs
- keep behavior unchanged while improving interaction responsiveness

## Validation
- npm run lint
- npm run format:check
- npm run typecheck
- npm run build
- npm test -- --passWithNoTests
- npm run test:run

PR #521, opened 2026-08-07
fix: make the drag overlay resolve scene facts from the canvas

umairnadeem:
## The crash

`throw new Error("Node at [...] is not inside a scene")` on drag, sometimes.

`DragOverlayContent` previews the dragged node in a scratch editor whose entire document is that one node. Drag a content element and it sits at path `[0]` with no parent scene, by construction. The overlay rendered through the canvas renderer, whose `ContentElement` looks up the parent scene to pick compact vs. full.

That lookup was an unchecked cast until #516 swapped it for `parentSceneId`. The cast read `.id` off the `Editor` object, got `undefined`, and `isCollapsed(undefined)` returned false, so the violation was silently absorbed. The helper fails loudly, which is right; it just exposed a real one.

Repro: collapse any scene, drag any element card. The "sometimes" is the `if (hasCollapsed)` guard in `ContentElement` (added in #127 as a perf short-circuit), which is why nothing breaks until something is collapsed.

## Two more overlay bugs from the same cause

- **Scene number.** `SceneContainer` resolved its own index via `useSceneIndex` → `useSlateStatic()`. Inside the overlay's nested `<Slate>` that is the scratch editor, whose only child is the dragged scene, so every dragged scene read as "Scene 1".
- **Collapsed scenes.** `SceneContainer` already picked `CollapsedScene` correctly (it keys off `isCollapsed(element.id)` from the real provider), but its children rendered as full-size element cards inside the 128px collapsed frame, since the compact dispatch depends on the same missing parent-scene lookup.

## The fix

Scene position is now a required prop on `SceneContainer`, resolved by whoever mounts it:

- `SortableItem` resolves it against the canvas editor, as before.
- `DragOverlayContent` resolves it *before* mounting its scratch editor. It renders inside the canvas's own `<Slate>`, so `useSceneIndex` there sees the real document.

The overlay renders its own nodes directly rather than borrowing the canvas renderer's scene-aware dispatch, and takes collapsed state from the scene it was handed. A dragged collapsed scene is now a faithful copy, chips and "+N more" included.

`SceneContainer` stops reading its position off whichever editor happens to be nearest, which is what made this fragile. The `parentSceneId` throw stays, and the invariant it guards is now true everywhere it runs.

## Notes

- The overlay's scene/content dispatch resembles `renderCanvasElement`'s shape. Kept separate: the two encode different policy (canvas asks the document; the overlay is a detached preview), and merging them would mean passing render flags into the canvas renderer to serve one caller.
- `SortableItem` calls `useSceneIndex` for content elements too, where it returns `0` and goes unused. That is `sceneIndexOf`'s existing not-found contract, and it keeps the prop required rather than optional-with-fallback.
- Scenes are only draggable while collapsed, content only while its scene is expanded, so each overlay branch has exactly one real case.

## Testing

Full CI locally: lint, format, typecheck, knip, build, 1086 unit tests, 3 e2e smoke. No unit test added; there is no jsdom/RTL setup in this repo, so this is only reachable through a rendered drag.

### DIFF
--- a/app/components/canvas/dnd/DragOverlay.tsx
+++ b/app/components/canvas/dnd/DragOverlay.tsx
@@ -1,15 +1,52 @@
 import { GripVertical, Plus } from "@/components/ui/icon";
-import { useMemo } from "react";
+import { useCallback, useMemo } from "react";
 import { createEditor, Descendant } from "slate";
-import { Editable, Slate, withReact } from "slate-react";
-import { renderCanvasElement } from "../elements/ElementContainer";
+import { Editable, RenderElementProps, Slate, withReact } from "slate-react";
+import type { CanvasElement } from "@/lib/canvas/types";
+import { isSceneElement } from "@/lib/canvas/scenes";
+import { CompactElement } from "../elements/CompactElement";
+import { ElementContainer } from "../elements/ElementContainer";
+import { SceneContainer } from "../elements/SceneContainer";
+import { useSceneIndex } from "../hooks/useSceneIndex";
+import { useViewMode } from "../ViewModeContext";
 import styles from "../styles/sortable.module.css";
 
-export function DragOverlayContent({ element }: { element: Descendant }) {
+/**
+ * Previews the dragged node in a scratch editor holding only that node. What a
+ * node would otherwise read off its own document (scene number, collapsed
+ * state) is resolved against the real canvas, which this renders inside of.
+ */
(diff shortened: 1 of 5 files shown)


umairnadeem (commented):
lgtm

umairnadeem (commented):
lgtm, rebase looks clean

PR #522, opened 2026-08-07
refactor: nightly maintainability 2026-08-07

umairnadeem:
## What

`GenerationQueue` was a 406-line class with 9 mutable fields owning three unrelated jobs at once: scheduling, per-element state, and elapsed-time bookkeeping. This splits it along those seams. No behavior changes.

- **`lib/generation/snapshots.ts` (new)** — `SnapshotStore` owns per-element state (status, seconds, result, error, `resultInputs`, `connectorType`, `pinned`), the result history keyed by serialized inputs, the result version, and the observer subscription. `ElementSnapshot` / `GenerationStatus` / `isGenerationActive` move here, where they belong. Mutators never notify; the caller decides when a batch of edits lands as one update.
- **`lib/generation/elapsedTicker.ts` (new)** — `ElapsedTicker` counts whole seconds for running jobs and keeps its interval alive only while something is running. Previously four private methods and two fields tangled into the queue.
- **`lib/generation/queue.ts`** — now a scheduler and nothing else: pending nodes, abort controllers, batch limit, dependency gating. 406 → 254 lines, 9 fields → 5.

The public interface of `GenerationQueue` is unchanged, so every consumer (`useGenerate`, `useAutosave`, `useVideoLayout`, `applyTemplate`, the preview cluster) is untouched apart from import paths for the moved types.

## Also

- `lib/generation/__tests__/snapshots.test.ts` covers the new module boundary directly: result-version bumps, reset-vs-remove semantics, history restore, active/generated counts, subscription teardown.
- `ARCHITECTURE.md`'s generation-graph section described the queue as owning all of the above. Updated to match the split.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `build` all clean. Full suite: 125 files / 1093 tests passing.

## Open PR overlap check

Zero open PRs in the repo at the time of the run, so there is no overlap to avoid.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -48,7 +48,10 @@ A generation is a small dependency graph, not a lone job. `lib/generation/graph.
 
 Staleness falls out of the graph: a node needs generating when it has no result, when a dependency does, or when its current inputs differ from the inputs its result was made from. One element (`useGenerate`) and Generate All (`useGenerateAll`) run the same check. A result the user supplied is `pinned` and never regenerated.
 
-`queue.ts` runs the graph. It flattens roots dependencies-first, skips nodes that are already fresh, and owns per-element status, elapsed time, abort controllers, and a result history keyed by serialized inputs, so undoing an edit restores the earlier result instead of regenerating it.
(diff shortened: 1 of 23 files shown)


umairnadeem (commented):
lgtm, clean seam split. checked the moved logic line by line (isActive vs isInQueue, remove vs the old discard bookkeeping, the ticker start/stop guards) and it all reads equivalent, plus cancelAll now iterates an ids() copy instead of mutating the map mid-iteration.

PR #523, opened 2026-08-07
perf: nightly performance optimizations 2026-08-07

umairnadeem:
## What

Dragging an element into another scene re-rendered **every** element card on **every** pointer move.

`setDragTransfer` fires from `onDragOver`, so it runs repeatedly through a cross-scene drag. The transfer sat in `DragTransferContext`, and context updates bypass `React.memo` and Slate's element memoization — so each pointer move re-rendered all 200 `SortableContent` cards and rebuilt their `ElementContainer` subtrees (settings popover, tooltips, the 11-item insert menu) to move a single `3rem` gap.

This is item 3 of #510, and the same shape as the drag-start fanout fixed in #520.

- **`DragTransferContext`** now carries a store (`get` / `set` / `subscribe`) instead of the value. Cards subscribe via `useSyncExternalStore` and re-render only when their own answer flips.
- **`useDropIndex(sceneId)`** is the entire consumer surface: `SortableContent` compares it against its own index, `SceneContainer` against its child count. Both call sites keep their exact previous conditions, so the gap and the scene's trailing pad appear where they always did.
- The transfer no longer lives in React state, so `Canvas` stops re-rendering during a drag too.

## Measurements

Production build, a 40-scene / 200-element fixture, one cross-scene drag of 12–13 pointer moves, 2–3 runs per side. Render counts came from counters compiled into the build; task times from a `longtask` PerformanceObserver.

| | before | after |
|---|---|---|
| card re-renders per pointer move | 200 | 10 |
| `ElementContainer` rebuilds over the drag | 2400–2800 | 120–125 |
| long-task time over the drag | 3.8–3.9s | 0.79–0.95s |
| longest task | 435–440ms | 102–123ms |

The remaining per-move cost is dnd-kit's own `useSortable` re-render on every consumer, which #520 measured and deliberately left alone. Those renders no longer rebuild anything, because the card subtrees are stable props.

Behavior verified in the same build: one insert gap appears during a cross-scene drag (never zero, never two), it clears on drop, and the element lands in the target scene.

## Also

`app/components/canvas/__tests__/dragTransfer.test.ts` covers the new seam: `dropIndexIn` for the four transfer cases, and the store's notify/unsubscribe semantics.

## Skipped

- **`ViewModeContext` identity churn** (#510 item 5) — same fix shape, but collapse-toggle is a rare interaction next to a drag, and bundling it would double the diff for a fraction of the win. Left for its own measured PR.
- **`pointerWithin` over every droppable** (#510 item 5) — cutting droppable count risks cross-scene drop correctness; it needs its own measurement, as #520 noted.
- Player/Remotion/Slate render paths were profiled and left alone: `ScrubBar` already drives the playhead through CSS variables, `usePlayerValue` already selects per-consumer, and `VideoComposition` already memoizes its sequence nodes.

## Validation

`lint`, `format:check`, `typecheck`, `knip`, `build` all clean. Full suite: 125 files / 1093 tests passing.

## Open PR overlap check

One open PR at the time of the run: **#522** (nightly maintainability — `lib/generation/*`, `lib/project/*`, `lib/video/resolve.ts`, `app/components/canvas/elements/{AssetTile,GenerateButton,GenerationIndicator,ElementUploadButton,preview/*}`, `ProjectEditor.tsx`, `ARCHITECTURE.md`). This PR touches `app/components/canvas/{Canvas.tsx,dnd/*,elements/SceneContainer.tsx}` and one new test file — no file or theme overlap.

### DIFF
--- a/app/components/canvas/Canvas.tsx
+++ b/app/components/canvas/Canvas.tsx
@@ -29,7 +29,7 @@ export default function Canvas({
 	const {
 		activeId,
 		sceneItems,
-		dragTransfer,
+		dragTransferStore,
 		sensors,
 		handleDragStart,
 		handleDragOver,
@@ -64,7 +64,7 @@ export default function Canvas({
 	);
 
 	return (
-		<DragTransferContext value={dragTransfer}>
+		<DragTransferContext value={dragTransferStore}>
 			<DndContext
 				sensors={sensors}
 				collisionDetection={pointerWithin}
--- a/app/components/canvas/__tests__/dragTransfer.test.ts
+++ b/app/components/canvas/__tests__/dragTransfer.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+	createDragTransferStore,
+	dropIndexIn,
+	type DragTransfer,
+} from "../dnd/DragTransferContext";
+
+const transfer = (
+	overrides: Partial<NonNullable<DragTransfer>> = {},
+): DragTransfer => ({
+	itemId: "item-1",
+	fromSceneId: "scene-1",
+	toSceneId: "scene-2",
+	atIndex: 3,
+	...overrides,
+});
+
+describe("dropIndexIn", () => {
+	it("reports the landing index for the receiving scene", () => {
+		expect(dropIndexIn(transfer(), "scene-2")).toBe(3);
+	});
+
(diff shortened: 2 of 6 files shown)


umairnadeem (commented):
lgtm. store-in-context + useSyncExternalStore is the right shape here, and dropIndexIn keeps both call sites exactly equivalent to the old three-condition checks.

PR #524, opened 2026-08-07
feat: start from a blank canvas (#490)

### LINKED ISSUE #490: Start from a blank canvas (skip the prompt)

umairnadeem:
## Summary

There was no way to reach the canvas without running a generation: `Editor.tsx` gates on `useScriptHasContent()`, and `hasContent` only ever flipped when `submitPrompt` streamed its first non-empty chunk. A new project therefore had exactly one route out of the prompt box.

This adds a **"Skip to a blank canvas"** button to `ComposerHero`, next to the template gallery.

- `ScriptControl` gains `startBlank()`, which sets `hasContent` without touching the LLM connector. `Editor.tsx` is unchanged and still has exactly one prompt-vs-canvas decision point.
- The blank document is produced by the **existing normalizers**, not a hand-rolled seed value: `rehydrateProjectEditor` force-normalizes an empty editor, `withLayout` inserts the first `narration`, and `withScenes` wraps it into a scene. Any element type that later changes those rules gets the blank canvas for free.
- The seed is gated on an explicit blank-start flag rather than on "the script is empty", because the streaming path is indistinguishable at mount time — force-normalizing there would leave a stray empty narration element above the streamed nodes.
- Autosave is untouched: the normalization emits a value change, which schedules a save like any other edit, so the project reloads straight into the canvas.

## Selected issue

#490 — `size: S`, well-scoped, with the code pointers and acceptance criteria already in the issue. Small, additive, and validated by the existing test suite plus a focused seam test.

## Validation

Full CI-equivalent run, all green:

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` — 124 files, 1088 tests passed (2 new)

Not validated: the button's rendered appearance in a running app. The composer hero lives behind auth, and this job does not run logged-in flows. The button uses `Button variant="ghost" size="sm"` with no hand-sized dimensions, per `DESIGN.md`.

## Open PR overlap check

Open PRs at branch time: #522 (`ARCHITECTURE.md`, `lib/generation/*`, `lib/project/*`, `lib/video/*`, several element components) and #523 (`app/components/canvas/Canvas.tsx`, `dnd/*`, `SceneContainer.tsx`). This PR touches `ComposerHero.tsx`, `ScriptProvider.tsx`, `useEditorSession.ts`, `useProjectRehydrate.ts`, and its test. No file, module, or theme overlap.

## Candidates skipped

- **#506** (animated image shows "Queued") — otherwise the best fit, but it lands squarely in `lib/generation/queue.ts`, `snapshots.ts`, and `GenerationIndicator.tsx`, all of which are open in #522.
- **#259** — the only `size: XS` issue, blocked on BYOK (#331).
- **#250**, **#254** — need product decisions (length presets, default volume levels).
- **#332**, **#461** — new product surfaces with open UX direction.

Closes #490

### DIFF
--- a/app/components/ComposerHero.tsx
+++ b/app/components/ComposerHero.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { useState } from "react";
+import { Button } from "@/components/ui/button";
 import { useScriptControl } from "@/lib/script/ScriptProvider";
 import { useConfig } from "@/lib/config/ConfigProvider";
 import BackToMySlopLink from "./BackToMySlopLink";
@@ -21,7 +22,7 @@ there was a small glowing garden hidden on the moon.
 And in that garden… lived a little rabbit named Lumi…`;
 
 export default function ComposerHero() {
-	const { submitPrompt } = useScriptControl();
+	const { submitPrompt, startBlank } = useScriptControl();
 	const { mode, selectTemplate } = useConfig();
 	const [value, setValue] = useState("");
 
@@ -44,6 +45,15 @@ export default function ComposerHero() {
 					mode === "script" ? undefined : <AnimatedPlaceholder />
 				}
 			/>
+
+			<Button
+				variant="ghost"
+				size="sm"
+				className="mt-3 text-muted-foreground"
+				onClick={startBlank}
+			>
+				Skip to a blank canvas
+			</Button>
 
 			<TemplateGallery
 				onSelect={(templateId, examplePrompt) => {
--- a/lib/project/__tests__/serialize.test.ts
+++ b/lib/project/__tests__/serialize.test.ts
@@ -9,7 +9,7 @@ import {
 	type CanvasContentElement,
 	type SceneElement,
 } from "@/lib/canvas/types";
-import { deserializeWithScenes, splitScenes } from "../serialize";
+import { BLANK_SCRIPT, deserializeWithScenes, splitScenes } from "../serialize";
 
 const connector = {
 	openslop: { defaultModel: "m", models: ["m"], isDefault: true, apiKey: "" },
@@ -59,6 +59,14 @@ describe("splitScenes", () => {
 describe("deserializeWithScenes", () => {
 	it("returns [] for empty input", () => {
 		expect(deserializeWithScenes("", connectors)).toEqual([]);
+	});
+
+	it("turns BLANK_SCRIPT into one scene holding one empty narration", () => {
+		const scenes = deserializeWithScenes(BLANK_SCRIPT, connectors);
+
+		expect(scenes).toHaveLength(1);
+		expect(scenes[0].children).toHaveLength(1);
+		expect(scenes[0].children[0].type).toBe("narration");
 	});
 
 	it("round-trips quotes and angle brackets in attributes and text", () => {
--- a/lib/project/serialize.ts
+++ b/lib/project/serialize.ts
@@ -5,6 +5,9 @@ import { parseOSML } from "@/lib/canvas/osmlStreamParser";
 import { makeNodeId } from "@/lib/canvas/nodeUtils";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 
+/** The document a project starts from when there is no generated script. */
+export const BLANK_SCRIPT = "<narration></narration>";
+
 export function splitScenes(osml: string): string[] {
 	if (!osml.trim()) return [];
 	return osml
--- a/lib/script/ScriptProvider.tsx
+++ b/lib/script/ScriptProvider.tsx
@@ -15,12 +15,14 @@ import { getDefaultConnector } from "@/lib/connectors/registry";
 import { createConnector } from "@/lib/connectors/factory";
 import { useProject } from "@/lib/project/useProject";
 import { getProjectStore } from "@/lib/project/store";
+import { BLANK_SCRIPT } from "@/lib/project/serialize";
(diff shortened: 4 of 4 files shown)


app/components/canvas/hooks/useProjectRehydrate.ts
```diff
@@ -9,9 +9,13 @@ export function rehydrateProjectEditor(
 	editor: Editor,
 	script: string,
 	connectorConfig: ConnectorRegistry,
+	seedBlank = false,
 ): void {
 	const scenes = deserializeWithScenes(script, connectorConfig);
-	if (scenes.length === 0) return;
+	if (scenes.length === 0) {
+		if (seedBlank) Editor.normalize(editor, { force: true });
```
umairnadeem: this seed lands in the undo stack, unlike the scenes path below which goes through `HistoryEditor.withoutSaving`. start blank, hit cmd+z → slate pops the seeded narration, withLayout immediately re-inserts a fresh one with a new node id, so undo is a phantom no-op that churns ids and kicks autosave. wrap it in `withoutSaving` too?

umairnadeem (commented):

PR #526, opened 2026-08-07
fix: re-host provider videos to our own blob storage

### LINKED ISSUE #405: Re-host Runware videos to our own blob storage (provider temp URLs expire after 30-day TTL)

umairnadeem:
## Summary

- Runware video URLs expire after a 30-day TTL, but `AssetBundle` recorded them by reference so the video 404'd once the provider's temp storage lapsed. Remote files are now fetched and streamed into our own blob.
- Deleted the by-reference file variant entirely. `BundleFile` is `BundleFileData | BundleFileRemote`, both carrying a filename, so `upload()` always records a local path and a provider can no longer persist a foreign URL.
- Moved async job polling off the client-facing poll endpoint and into the queue worker, which re-enqueues pending jobs to itself. `JobHandler.poll` is gone and `pollJob` is a pure row read. Synchronous handlers are unaffected — they return `completed` on first delivery and are never redelivered.
- Also fixes video jobs stranding in `processing` forever when the client stopped polling, since nothing server-side advanced them before.
- Aligned the client and server give-up times, which were 10 and 30 minutes, on a shared `JOB_TIMEOUT_MS` of 15.

## Test plan

- [ ] `npm run lint`, `format:check`, `typecheck`, `knip`, `build`, `test:coverage` — all green locally (1106 tests)
- [ ] Generate a video end to end; confirm the manifest resolves to `assets/video/runware/<id>/output.mp4` on our blob host, not a Runware URL
- [ ] Confirm the video still plays after the provider URL would have expired (fetch the blob URL directly)
- [ ] Start a video generation, close the tab, and confirm the job still reaches `completed` server-side
- [ ] Generate an image and a TTS clip to confirm the synchronous path is unchanged

## Notes

Projects generated before this change still hold expiring Runware URLs in their manifests. A backfill is out of scope here and worth a follow-up if any live projects predate the fix.

Two things deliberately left alone, flagged for review: the `output.mp4` / `video/mp4` filename is hardcoded in `BaseVideoProvider` (correct for Runware, would misdescribe a future webm provider), and the pending retry interval is a flat 5s with no backoff.

Closes #405


lib/api/process-job.ts
```diff
@@ -0,0 +1,62 @@
...
+	if (job.status !== "processing") {
+		await updateJob(jobId, { status: "processing" });
+	}
+
+	let outcome;
+	try {
+		outcome = await handler.process(job);
+	} catch (error) {
+		logger.error(error, `Job ${jobId} failed`);
+		await updateJob(jobId, { status: "failed", error: stringifyError(error) });
```
umairnadeem, line 37: `process` now runs on every redelivery, and for video that means an upstream poll plus the re-host fetch + blob upload. any transient blip in there (runware hiccup, blob 500 mid-multipart) marks the job failed for good and kills the redelivery loop, so a video we already paid for is gone once the provider url expires. can we keep the terminal fail for real handler errors and let network-ish ones fall through to another redelivery instead?

PR #527, opened 2026-08-07
fix: defer the play one frame past the seek on "play from here"

### LINKED ISSUE #425: "Play from here" starts audio but not the visual player (seek/play race)

umairnadeem:
Reopens and fixes #425. The original fix (#430) was a no-op for the case it targeted.

Verified in a real browser against a real project. Earlier revisions of this PR were wrong; the measurements are below.

## The bug

`startPlaybackAt` did `pause; seekTo; silence; play` in one tick. The seek makes the media elements buffer, `useMediaBuffering` takes a `buffer.delayPlayback()` block, and the Player stops scheduling animation frames while any block is held. But `play()` starts the shared audio tags directly, so they advance on their own media clock against a frozen picture.

## The fix

One line: defer the play to the next animation frame, so the seek commits first.

```
player.pause();
player.seekTo(frame);
silenceMediaIn(player.getContainerNode());
requestAnimationFrame(() => player.play());
```

That is the same separation `SegmentedSeekBar` gets for free by playing from `onScrubEnd` (pointerup), which is why the scrub bar never had this bug.

Two things that look removable but are not:

- **`pause()`** is what stops `seekTo` taking its own pause-and-resume path (`hasPausedToResume`), which freezes identically when the player is already playing. An earlier revision of this PR dropped it and reintroduced the bug.
- **`requestAnimationFrame`, not `setTimeout`.** A macrotask fires too early and freezes exactly like the unfixed code. Measured, see below.

## What the browser showed

Chrome, 9-scene project with generated audio. The frame clock is the seek bar fill, driven by the Player's `frameupdate` event, so it reads the frame driver directly. Sampled every 200ms.

Before, clicking "Play from here" on scene 7:

```
 0.0s prog=2043 state=Pause playing=0 audio= 0.00
 3.9s prog=2043 state=Pause playing=1 audio= 3.71
 7.9s prog=2043 state=Pause playing=1 audio= 7.73
11.9s prog=2043 state=Pause playing=1 audio=11.74
```

The button reads "Pause", so the Player believes it is playing. The frame clock never moves. Audio climbs to 11.7s. That is the reported bug.

After:

```
 0.0s prog=2139 playing=0 audio= 0.00
 3.1s prog=2153 playing=3 audio=13.26
 5.1s prog=2160 playing=3 audio=15.26
 7.0s prog=2168 playing=3 audio=17.20
```

Picture and audio advance in lockstep. Reproduced on scenes 2, 4, 6, 8 and 9.

With `setTimeout(..., 0)` instead of `requestAnimationFrame`:

```
 0.0s prog=1776 state=Pause playing=0 audio= 0.00
 7.8s prog=1776 state=Pause playing=0 audio= 0.00
```

Frozen. This is why the deferral is a frame and not a task.

Also checked:

- **Jumping while already playing** — frame clock advances, no media element keeps running from the old position.
- **Control: plain Play with no seek** — advances normally, so the freeze is specific to the seek path and not a loading problem.
- **Master's ordering, same probe** — frozen, so the probe reproduces the original bug rather than just agreeing with the fix.

## Tests

The old tests asserted mock call order within one tick, which was never the failing property, so they were green throughout the bug. They now stub `requestAnimationFrame` and pin that the play does *not* happen until the frame after the seek.

## Deliberately not included

- **Threading the originating click into `play()`.** Remotion only warms the shared audio tag pool for autoplay when `play()` is passed an event. I could not measure any difference in Chrome with or without it, and master never passed one, so it is an unproven enhancement rather than part of this fix.
- **Holding the frame while the player is hidden.** `showPlayer()` only schedules the mount, so a click made while the player is hidden is dropped. Not worth machinery for that path.

Known behaviour: `requestAnimationFrame` does not fire in a background tab, so a click made just before switching tabs starts playing when the tab is next visible.

## Verification

Lint, format, typecheck, knip, build, 1102 unit tests. Verified against `@remotion/player` 4.0.481.

Closes #425

### DIFF
--- a/app/components/video/PlayerControlContext.tsx
+++ b/app/components/video/PlayerControlContext.tsx
@@ -13,6 +13,8 @@ import { createRequiredContext } from "@/lib/components/createRequiredContext";
 import { usePlayerPosition } from "./PlayerPositionContext";
 import { startPlaybackAt } from "./startPlaybackAt";
 
+type PlaybackRequest = { frame: number; event?: SyntheticEvent };
+
 type PlayerControl = {
 	player: PlayerRef | null;
 	registerPlayer: (player: PlayerRef | null) => void;
@@ -29,22 +31,22 @@ export function PlayerControlProvider({ children }: { children: ReactNode }) {
 	// only copy of the handle — nothing downstream mirrors it.
 	const [player, setPlayer] = useState<PlayerRef | null>(null);
 	const { showPlayer } = usePlayerPosition();
-	// `showPlayer()` only schedules the mount, so a play requested while the
+	// `showPlayer()` only schedules the mount, so a request made while the
 	// player is hidden has nothing to run against until the ref arrives.
-	const pendingFrameRef = useRef<number | null>(null);
+	const pendingRef = useRef<PlaybackRequest | null>(null);
 
 	const registerPlayer = useCallback((next: PlayerRef | null) => {
 		setPlayer(next);
-		const frame = pendingFrameRef.current;
-		pendingFrameRef.current = null;
-		if (next && frame != null) startPlaybackAt(next, frame);
+		const pending = pendingRef.current;
+		pendingRef.current = null;
+		if (next && pending) startPlaybackAt(next, pending.frame, pending.event);
 	}, []);
 
 	const playFromFrame = useCallback(
 		(frame: number, event?: SyntheticEvent) => {
 			showPlayer();
 			if (player) startPlaybackAt(player, frame, event);
-			else pendingFrameRef.current = frame;
+			else pendingRef.current = { frame, event };
 		},
 		[showPlayer, player],
 	);
--- a/app/components/video/PlayerControlContext.tsx
+++ b/app/components/video/PlayerControlContext.tsx
@@ -1,7 +1,14 @@
 "use client";
 
 import type { PlayerRef } from "@remotion/player";
-import { useCallback, useMemo, useState, type ReactNode } from "react";
+import {
+	useCallback,
+	useMemo,
+	useRef,
+	useState,
+	type ReactNode,
+	type SyntheticEvent,
+} from "react";
 import { createRequiredContext } from "@/lib/components/createRequiredContext";
 import { usePlayerPosition } from "./PlayerPositionContext";
 import { startPlaybackAt } from "./startPlaybackAt";
@@ -9,7 +16,7 @@ import { startPlaybackAt } from "./startPlaybackAt";
 type PlayerControl = {
 	player: PlayerRef | null;
 	registerPlayer: (player: PlayerRef | null) => void;
-	playFromFrame: (frame: number) => void;
+	playFromFrame: (frame: number, event?: SyntheticEvent) => void;
 };
 
 const [Ctx, usePlayerControl] = createRequiredContext<PlayerControl>(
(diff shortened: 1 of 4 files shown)


app/components/video/PlayerControlContext.tsx
```diff
@@ -20,20 +27,31 @@ export { usePlayerControl };
...
 	const { showPlayer } = usePlayerPosition();
+	// `showPlayer()` only schedules the mount, so a play requested while the
+	// player is hidden has nothing to run against until the ref arrives.
+	const pendingFrameRef = useRef<number | null>(null);
+
+	const registerPlayer = useCallback((next: PlayerRef | null) => {
+		setPlayer(next);
+		const frame = pendingFrameRef.current;
+		pendingFrameRef.current = null;
+		if (next && frame != null) startPlaybackAt(next, frame);
```
umairnadeem: the deferred path drops the event, so `play()` skips `playAllAudios()` and the shared tag pool never gets warmed. that's the hidden-player case, ie the first play-from-here of a session, exactly where autoplay unlock matters most. stash the event alongside the frame in the ref and pass it through? react hasn't pooled synthetic events since 17 and `play()` only checks it for truthiness anyway.

PR #528, opened 2026-08-07
feat: per-media-type generation concurrency limits

### LINKED ISSUE #259: Configurable generation queue concurrency (post-BYOK)

umairnadeem:
Closes #259.

## Problem

The generation queue admitted at most `DEFAULT_BATCH_SIZE = 2` jobs at a time, regardless of what those jobs were. Two slow video renders would stall every sound effect and TTS clip waiting behind them, even though those cost us almost nothing to run in parallel.

## Change

Concurrency is now a per-connector-type budget rather than one global count:

| type | limit |
| --- | --- |
| `video`, `image`, `animated_image` | 3 |
| `tts` | 2 |
| `music`, `sfx` | 5 |

- **`lib/generation/concurrency.ts`** (new) — the limits as one declarative table, plus `resolveConcurrencyLimits(overrides)`.
- **`lib/generation/queue.ts`** — `controllers: Map<string, AbortController>` becomes `active: Map<string, {controller, connectorType}>` so the queue knows the type of each in-flight job. `processQueue` admits any pending node whose type has headroom *and* whose dependencies have settled, instead of stopping at a single global count.
- **`GenerationQueueProvider`** — constructs the queue with no arguments; defaults come from the table.

Dependency ordering, cancellation, and the blocked-node release path are unchanged.

## Post-BYOK

The seam is `resolveConcurrencyLimits`, which merges a `Partial<ConcurrencyLimits>` over the defaults. When user-adjustable limits land, the provider passes the user's overrides in and nothing in the queue changes. Nothing is exported ahead of that need.

## Notes

- `animated_image` sits with video/image at 3 — same class of heavy job.
- The old default was 2 for *everything*, so this does raise real throughput against our shared provider keys. `tts` is the only type that didn't go up.

## Testing

New test covers the core behavior: a saturated `image` pool no longer blocks `tts` jobs. Existing queue tests that relied on `batchSize` now pin the specific limit they exercise. Full CI green locally — lint, format, typecheck, knip, build, 1107 tests.

### DIFF
--- a/lib/generation/queue.ts
+++ b/lib/generation/queue.ts
@@ -1,5 +1,9 @@
-import type { AssetResult } from "../connectors/types";
+import type { AssetConnectorType, AssetResult } from "../connectors/types";
 import { errorMessage } from "../errors";
+import {
+	resolveConcurrencyLimits,
+	type ConcurrencyLimits,
+} from "./concurrency";
 import { ElapsedTicker } from "./elapsedTicker";
 import { generateForElement } from "./generateForElement";
 import type { GenerationInputs } from "./inputs";
@@ -15,12 +19,15 @@ import {
 	type NodeId,
 } from "./graph";
 
-export const DEFAULT_BATCH_SIZE = 2;
+type ActiveJob = {
+	controller: AbortController;
+	connectorType: AssetConnectorType;
+};
 
 /**
- * Runs generation nodes, at most `batchSize` at a time and never before their
- * dependencies have settled. All per-element state lives in the snapshot store;
- * the queue owns only what is in flight.
+ * Runs generation nodes, at most `limits[connectorType]` of each media type at a
+ * time and never before their dependencies have settled. All per-element state
+ * lives in the snapshot store; the queue owns only what is in flight.
  */
 export class GenerationQueue {
 	private readonly snapshots: SnapshotStore;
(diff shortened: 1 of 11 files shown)


lib/generation/queue.ts
```diff
@@ -15,30 +19,33 @@ import {
 	type NodeId,
 } from "./graph";
 
-export const DEFAULT_BATCH_SIZE = 2;
+type ActiveJob = {
```
umairnadeem, line 22: `DEFAULT_BATCH_SIZE` is gone but ARCHITECTURE.md:17 still says we queue "`DEFAULT_BATCH_SIZE` at a time", so the doc now names a symbol that doesn't exist. worth updating in this PR to say per-media-type limits, otherwise it's split-brain.

umairnadeem (commented):
lgtm, doc drift is fixed. per-type headroom check reads clean.

PR #530, opened 2026-08-08
feat: target video length control

### LINKED ISSUE #250: Target video length control (length presets, lift ~4min cap)

umairnadeem:
Closes #250.

Users can now pick a target length in the composer, and the ~4 minute cap is gone.

## The approach

An LLM cannot hit "3 minutes." It has no model of speech rate. It *can* hit a word count, and word count maps to duration deterministically through TTS. So the whole feature reduces to a unit conversion, which matters more once BYOK lands and model steerability varies widely. Word-counting is about the most universally-steerable instruction there is, and it needs no tool use or structured output, which are the least uniform capabilities across providers.

One map in `lib/video/videoLength.ts` owns the translation. Only the word budget is ever shown to a model; `label` is UI only.

| preset | word budget |
|---|---|
| Under 30 sec | 50-90 |
| Under 1 min | 90-180 |
| 1-3 min | 180-540 |
| 3-5 min (default) | 540-900 |
| 5-10 min | 900-1800 |
| 10-15 min | 1800-2700 |

## There was never a cap

Nothing clamps duration anywhere. The reported ~4 minute limit was the phrase "super short" in `story-mode.ts`, and script and template modes said nothing about length at all. That phrase is gone.

## Length is per-mode, not global

`scriptLengthPlugin` composes into story and template modes only. Script mode reproduces a script the user already pasted, and `script-mode.ts` instructs "do NOT modify the script itself... never add dialogue or narrative text not in the original script." A word budget saying "add or remove scenes as needed to hit it" flatly contradicts that, so `MODE_PLUGIN_FACTORIES` now returns `LLMPlugin[]` and script mode simply doesn't get the plugin. The pill hides there too.

## Templates carry their own length

Counted spoken words in each template's example script:

| template | spoken words | implied runtime |
|---|---|---|
| pov-life | 702 | ~5.4 min |
| sleep-story | 782 | ~6.0 min |
| finance-tips | 997 | ~7.7 min |
| true-crime | 747 | ~5.7 min |
| pov-financial-lifestyle | 1422 | ~10.9 min |
| celebrity-death | 1658 | ~12.8 min |
| stick-explainer | 1139 | ~8.8 min |

Template mode tells the model to match the example's "pacing, story beats, structure," so a default budget would fight every one of them. `Template` gains a `length` field seeded by `applyTemplate`, so picking a template moves the pill. Users can still override.

## The WPM constant is deliberately biased high

Set to **180 wpm**, above the measured rate. Models consistently undershoot a word budget, so overstating it lands the finished runtime closer to the target.

The measured figure for reference: one real script ran 471 spoken words across 23 elements in 3:46, giving 125 effective wpm end-to-end at `speed="slow"`. Converting to medium (`CARTESIA_SPEED` maps slow to 0.6 against medium's 1.0) puts the honest rate near 150. A second script came in at 610 words across 26 medium-speed elements, which 150 wpm predicts at ~4:04.

180 is that ~150 pushed up on purpose. It is one constant to change if the bias turns out too aggressive.

Worth noting the per-element +1s TTS padding is folded into the rate rather than modelled. Fine for ranges, but it is why the short presets are least accurate: at a 30 second target with 5 lines, padding is a sixth of the budget.

## Drive-by fixes

- **`ElementUploadButton` now pins its upload.** Without `{ pinned: true }`, an uploaded image was treated as a stale generated result and could be overwritten by Generate All once project state drifted. The other two `commitResult` upload paths already pinned; this one was overlooked.
- **`PillTrigger` gains a focus ring** via the existing `focus-ring` utility. It had none.
- **Composer pills include their value in the accessible name.** `aria-label="Video length"` replaces the button's content, so screen reader users heard "Video length, button" and never "3-5 min" (WCAG label-in-name). Fixed across all four pills since it is one shared row.

## Left out deliberately

- **Consolidating the three setting hooks.** `useVideoLength` is the third copy of the `useAspectRatio` / `useTransitionType` shape, a fair rule-of-three call, but the fix refactors two existing hooks and their call sites. Separate PR.
- **No duration estimator or auto-correction.** Deliberately scoped out. The natural next step is a pure `estimateDurationSec(elements)` over the parsed canvas to show "~2:40 / target 3:00" before any assets are generated, then optionally a corrective pass through the existing refine ops.

## Testing

Full CI green: lint, format, typecheck, knip, 1112 unit tests, build, Playwright smoke. New tests cover the time-to-words translation, the default fallback, prompt ordering, and an assertion that the runtime label never leaks into a prompt.

### DIFF
--- a/app/components/canvas/elements/ElementUploadButton.tsx
+++ b/app/components/canvas/elements/ElementUploadButton.tsx
@@ -17,7 +17,11 @@ export function ElementUploadButton() {
 			className="shrink-0"
 			disabled={isGenerationActive(status)}
 			onUpload={(url) =>
-				queue.commitResult(target, { imageUrl: url, durationSec: 0 })
+				queue.commitResult(
+					target,
+					{ imageUrl: url, durationSec: 0 },
+					{ pinned: true },
+				)
 			}
 		/>
 	);
--- a/app/components/copilot/ComposerCopilot.tsx
+++ b/app/components/copilot/ComposerCopilot.tsx
@@ -4,6 +4,7 @@ import type { ReactNode } from "react";
 import {
 	ChevronDown,
 	CornerDownLeft,
+	Hourglass,
 	ImagePlus,
 	Loader2,
 	Mic,
@@ -26,6 +27,12 @@ import { useProject } from "@/lib/project/useProject";
 import type { Mode } from "@/lib/project/types";
 import type { AspectRatio } from "@/lib/video/aspectRatio";
 import { useAspectRatio } from "@/lib/video/useAspectRatio";
+import {
+	VIDEO_LENGTHS,
+	VIDEO_LENGTH_SPECS,
+	type VideoLength,
+} from "@/lib/video/videoLength";
+import { useVideoLength } from "@/lib/video/useVideoLength";
 import { useImageUpload } from "@/lib/upload/useImageUpload";
 import { ActionButton } from "./ActionButton";
 import { ComposerAssets } from "./ComposerAssets";
@@ -44,6 +51,10 @@ const ASPECT_RATIO_OPTIONS: SelectMenuOption<AspectRatio>[] = [
 	{ value: "16:9", label: "16:9" },
 	{ value: "9:16", label: "9:16" },
 ];
+
+const VIDEO_LENGTH_OPTIONS: SelectMenuOption<VideoLength>[] = VIDEO_LENGTHS.map(
+	(value) => ({ value, label: VIDEO_LENGTH_SPECS[value].label }),
+);
 
 const TEMPLATE_OPTIONS: SelectMenuOption<string>[] = TEMPLATES.map((t) => ({
 	value: t.id,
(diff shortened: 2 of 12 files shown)


umairnadeem (commented):
lgtm. nice that the length plugin is composed per mode instead of a conditional inside one prompt, and good call keeping the runtime label out of the model prompt.

umairnadeem (commented):
lgtm, the word-budget prompt reads clean after the trim.

PR #532, opened 2026-08-08
perf: nightly performance optimizations 2026-08-08

umairnadeem:
## What changed

`Template.systemPrompt` and `Template.exampleText` are read in exactly one place — `createTemplateModePlugin` — and are never rendered by any UI. They were 160KB of the 174KB in `lib/templates/templates.ts`, which `TemplateGallery`, `ComposerCopilot` and `ConfigProvider` all import eagerly, so every editor page load downloaded and parsed them.

They now live in `lib/templates/prompts.ts` behind `getTemplatePrompt(id)`, which the plugin reaches through `import()` on the first template-mode generation. `templates.ts` keeps the gallery metadata it always owned.

## Why it matters

Measured across two `npm run build` runs, the shared client chunk carrying the templates:

| | raw | gzipped |
| --- | --- | --- |
| before | 971,072 | 277,167 |
| after | 804,447 | 224,991 |
| delta | **-166,625 (-17%)** | **-52,176 (-19%)** |

The prose now sits in its own async chunk that appears in neither `build-manifest.json` nor `app-build-manifest.json`, so it is never part of the initial load. It is fetched when a template generation starts — next to an LLM call that already takes seconds — so interactivity is unaffected.

No behavior change: same prompts, same plugin contract (`beforeGenerate` already accepted a `Promise`), and `getTemplatePrompt` throws on an unknown id exactly like `getTemplate` does.

## Open PR overlap check

One open PR at the time of this run: **#531** (`refactor: nightly maintainability 2026-08-08`), which touches `ARCHITECTURE.md` only. This PR touches `lib/templates/*` and `lib/connectors/llm/plugins/template-mode.ts` — no overlap in files or theme.

## Considered and skipped

Profiled the runtime hot paths first, per the runbook order, and found nothing worth shipping:

- **Slate**: every consumer already uses `useSlateStatic` (never `useSlate`), `renderElement` is a stable `useCallback`, and there is no custom `decorate` or `normalizeNode` overreach. Chunking targets 1000+ node documents; scripts here are tens of nodes.
- **Remotion**: `VideoComposition` memoizes its sequence nodes, transition timing and presentation; `Captions` binary-searches word timestamps via `sortedLastIndex` over memoized data; `motionTransform` is pure per-frame arithmetic.
- **Player/timeline**: frame subscriptions go through `useSyncExternalStore` with derived snapshots, and `ScrubBar` moves the playhead through CSS variables instead of React renders.
- **Queue**: `SnapshotStore` hands out stable snapshot references, so `useQueueSelector` consumers re-render only on real changes.

`PlayerPositionContext` builds an unmemoized context value, and `MotionLayer` subscribes to `useCurrentFrame()` even when its effect is `"none"`. Both are single-component re-renders with no measurable cost — left alone rather than shipped as micro-churn.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `npm run build`, and `npm test` (1111 tests, 128 files) all pass.


lib/templates/prompts.ts
```diff
@@ -0,0 +1,1496 @@
...
+ * the story pastiches. Nothing renders these, so they live apart from the
+ * gallery metadata in `templates.ts` and load only when a template-mode
+ * generation runs.
+ */
+export interface TemplatePrompt {
+	systemPrompt: string;
+	exampleText: string;
+}
+
+const TEMPLATE_PROMPTS: Record<string, TemplatePrompt> = {
```
umairnadeem, line 14: before this, `systemPrompt`/`exampleText` were required fields on `Template`, so a new template that forgot them didn't compile. now the ids live in two places and a missing entry only blows up at generation time. keying this `Record<string, ...>` off the actual template ids (or a test that asserts every `TEMPLATES` id has a prompt) would keep the compile-time guarantee.

umairnadeem:
useless

PR #533, opened 2026-08-08
refactor: rethink openslop architecture 2026-08-08

umairnadeem:
## App understanding

openslop is a Next.js 16 app for generating short videos from a script. A user opens a project (`app/projects/[id]`), writes or generates an OSML script in a Slate canvas, and each script element resolves into a generation graph (`lib/generation`) that dispatches through a connector/plugin chain (`lib/connectors` → `lib/gateway` → `lib/providers`) to image/video/TTS/music providers. Results render in a Remotion preview and autosave back to Supabase.

Project-level state (title, art style, characters, narration voice, reference images) lives in a zustand store. Generation reads it; the UI mutates it. It is the single most cross-cutting piece of state in the app.

## From-scratch architecture

Rebuilt today, that store would be created once by the component that owns the project and passed down: React consumers read it from a provider, and non-React callers (plugins, template application, autosave) take the handle as a parameter. Nothing would need to know a `projectId` unless it is talking to the API.

The current code did the opposite. `lib/project/store.ts` held a module-level `Map<projectId, ProjectStore>`, so:

- every project visited in a session leaked its store for the tab's lifetime,
- `clearProjectStore` existed only for tests and had zero production callers,
- `projectId` was threaded through ~12 modules purely as a key into that global,
- `applyStoreSnapshot` needed a `hydrated` guard to defend against a store surviving a remount,
- 14 test files had to `clearProjectStore` in `beforeEach`/`afterEach` or invent unique ids to avoid bleeding into each other.

- `lib/project/store.ts`: `getProjectStore`/`clearProjectStore`/`useProjectStore` and the module-level `Map` are gone, replaced by `createProjectStore()`.
- New `lib/project/ProjectStoreProvider.tsx` (built on the existing `createRequiredContext`) supplies the handle; `ProjectEditor` creates and hydrates the store once and provides it.
- `useProject(selector)` now reads the store from context instead of `useConfig().projectId`, dropping a dependency on `ConfigProvider`.
- Non-React callers take the store (or its state) explicitly instead of an id: `applyTemplate`, `deleteCharacter`, `createVoiceHydratePlugin`, `seedCharacterAvatar`, `createAutosaver`. `createAutosaver` keeps `projectId`, which is a genuine API identity, not a store key.
- Ambient store reads removed from `ScriptProvider`, `ConfigProvider`, `useRefineScript`, `useAutosave`, `CharacterEditModal`, `ComposerAssets`.
- Tests get isolation for free: `createProjectStore()` per test, no cross-test cleanup hooks, no synthetic unique project ids.

Net −94 lines.

## Behavior notes

Navigating away from a project and back now re-hydrates from the saved row rather than reusing a surviving in-memory store. Autosave flushes on unmount, so edits are persisted first; this also removes the previous cross-project leak. Rendering a `useProject` consumer outside the editor now throws a clear "missing a provider" error rather than silently reading an empty store, which matches the repo's fail-loudly rule.

Verified by import-closure trace that every direct and transitive `useProject`/`useProjectStoreHandle` consumer roots only at `app/projects/[id]/page.tsx → ProjectEditor → Editor`. `app/page.tsx`, login, signup, layout, the API routes, and the Remotion bundle are all disjoint from that closure.

## Validation

All `.github/workflows/ci.yml` steps run locally:

- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run knip` — clean
- `npm run build` — succeeds
- `npx vitest run` — 128 files, 1111 tests passed
- `npx playwright test` — 3 smoke tests passed

## Open PR overlap check

Open PRs considered: #531 (`ARCHITECTURE.md` only) and #532 (`lib/connectors/llm/plugins/template-mode.ts`, `lib/templates/prompts.ts`, `lib/templates/templates.ts`, `lib/connectors/__tests__/template-mode.test.ts`). This PR touches none of those files. `lib/templates/applyTemplate.ts` is changed here but is not in either PR's diff. Non-overlapping.

## Skipped items

- `graph.ts` ↔ `queue.ts` type-only import cycle. Extracting `GenerationJob`/`ElementSnapshot`/`NodeResults` into `lib/generation/types.ts` would break it, but on its own it is churn with no behavior or clarity payoff; it wants to ride along with a substantive generation change.
- `lib/supabase/env.ts` defaults missing `NEXT_PUBLIC_SUPABASE_URL`/`ANON_KEY` to `""` instead of throwing. A soft fail-loudly violation, but throwing at module load would break env-less builds. Human call.
- `app/components/video/ScrubBar.tsx` has no `role="slider"`, `aria-valuenow`, `tabIndex`, or `onKeyDown`, so neither the seek bar nor the volume bar is keyboard-operable. Real accessibility gap, but it is an addition rather than a simplification and belongs to the bugs/a11y nightly.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -14,7 +14,9 @@ The prompt goes through the LLM connector to `POST /api/v1/llm`, which streams O
 
 ### 2. Generate
 
-Generate resolves the element into a dependency graph (see [Generation graph](#generation-graph)) and queues the stale nodes client-side, dependencies first and up to a per-media-type concurrency limit at a time. Each job is submitted to `POST /api/v1/{asset}`, which records a `jobs` row and enqueues to the Vercel Queue. Workers call the provider, upload results to Vercel Blob, and mark the job complete. The client polls until the asset URL arrives and the preview updates. Music and sfx check a Pinecone similarity cache first.
+Generate resolves the element into a dependency graph (see [Generation graph](#generation-graph)) and queues the stale nodes client-side, dependencies first and up to a per-media-type concurrency limit at a time. Each job is submitted to `POST /api/v1/{asset}`, which records a `jobs` row and enqueues to the Vercel Queue. The client polls until the asset URL arrives and the preview updates. Music and sfx check a Pinecone similarity cache first.
+
+Each delivery runs `lib/api/process-job.ts`, which hands the job to the `JobHandler` registered for its type (`lib/api/job-handlers.ts`) and gets back `completed` or `pending`. Image, tts, music and sfx complete on the first delivery. Video is slow enough to outlive one invocation, so its handler submits to the provider, records the provider's job id on the row, and reports `pending`; the job then redelivers itself every 5s to poll, and fails once it outlives `JOB_TIMEOUT_MS`. Results upload to Vercel Blob. Bytes a provider only exposes at its own URL are streamed across rather than linked, so every asset we serve is our own.
 
 ### 3. Save / restore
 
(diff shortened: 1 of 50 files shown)


umairnadeem (commented):
lgtm

PR #534, opened 2026-08-08
fix: reserve the transport bar seek row height

### LINKED ISSUE #529: Bottom transport bar height shifts when the seek bar appears/disappears
## Problem

The bottom transport bar has no stable height. Its seek-bar row is swapped for a placeholder when the player isn't ready, and the two are different heights, so the whole bar grows/shrinks by 8px as the player initializes (and shrinks again on any state where `ready` flips back to false).

`BottomTransportBar` renders `SegmentedSeekBar` when `ready`, otherwise a placeholder:

- placeholder is `h-3` (12px)
- the real seek bar's hit area is `h-5` (20px), via `ScrubBar`

The container itself (`flex-col gap-1.5 ... py-2`) has no `min-h`, so it just tracks whatever its children measure. Result: a visible vertical jump of the bar and the canvas above it right as the video becomes playable.

Also worth noting: `SegmentedSeekBar` has a second internal gate (`if (segments.length === 0) return null`) that is redundant with `ready`, but would collapse the row to 0 height if it ever fired.

- "height of bottom transport shouldn't shift if the seek bar appears/disappears"

## Relevant code

- `app/components/video/BottomTransportBar.tsx:26` — `ready` gate (`player !== null && segments.length > 0`)
- `app/components/video/BottomTransportBar.tsx:41-45` — seek bar vs. `h-3` placeholder swap
- `app/components/video/BottomTransportBar.tsx:40` — container, no `min-h`
- `app/components/video/ScrubBar.tsx:171` — real seek bar hit area is `h-5` (20px)
- `app/components/video/SegmentedSeekBar.tsx:51` — redundant `segments.length === 0` early return

## Suggested direction

Reserve the row's height unconditionally so the ready/not-ready swap is invisible. Simplest fix is to make the placeholder match the real control (`h-5`), but the more robust version is to let the bar own its height rather than infer it from children: give the seek-bar row a fixed height and render the placeholder (or nothing) inside it. That also makes the redundant early return in `SegmentedSeekBar` harmless.

Same reasoning should apply to the other `ready`-gated children (`TimeDisplay`, `ScenePill`) — they sit in a row with the play controls, so they don't currently move the bar's height, but they shouldn't be able to either.

## Acceptance criteria

- [ ] The bar's rendered height is identical before and after the player becomes ready
- [ ] No vertical shift of the transport bar or the canvas above it during load
- [ ] The reserved height is defined in one place rather than duplicated between the real control and its placeholder

---

umairnadeem:
## Summary

The bottom transport bar had no stable height. Its seek-bar row rendered `SegmentedSeekBar` (a `ScrubBar`, `h-5`) when the player was ready and an `h-3` placeholder otherwise, and the container inferred its height from children. The bar and the canvas above it jumped 8px the moment the video became playable.

The seek row now owns its height: a wrapper reserves `SCRUB_BAR_HEIGHT` unconditionally and renders the seek bar into that space when ready. The height is defined once, in `ScrubBar.tsx`, and consumed by both the control and the reserved row, so the two can't drift apart.

This also makes `SegmentedSeekBar`'s redundant `segments.length === 0` early return harmless: if it ever fired, the row keeps its height instead of collapsing to 0.

`TimeDisplay` and `ScenePill` sit in the controls row, whose height is set by the always-rendered play buttons, so they can't move the bar. Left as-is rather than adding a second reserved height.

Two-file, presentation-only change with explicit acceptance criteria and named code locations in the issue. No behavior, state, or API changes.

## Validation

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:coverage` — 128 files, 1111 tests passed

No test added: the change is Tailwind class placement with no logic to assert; a class-string test would be brittle.

## Open PR overlap check

Open PRs at selection time: #531 (`ARCHITECTURE.md`), #532 (`lib/connectors`, `lib/templates`), #533 (`lib/project`, `lib/connectors`, canvas hooks). None touch `app/components/video/`. No overlap.

## Candidates skipped

Issue #529 was the only open `size: XS` issue. Everything else open is S/M/L/XL and product-, design-, or architecture-heavy (model selection UX, BYOK, i18n, publishing kit, a11y audit), which the runbook excludes from unattended runs.

Closes #529

### DIFF
--- a/app/components/video/BottomTransportBar.tsx
+++ b/app/components/video/BottomTransportBar.tsx
@@ -3,12 +3,14 @@
 import { memo } from "react";
 import { ChevronsLeft, ChevronsRight, Pause, Play } from "@/components/ui/icon";
 import { TooltipIconButton } from "@/components/ui/icon-button";
+import { cn } from "@/lib/utils";
 import { toFrames } from "@/lib/video/frames";
 import { usePlayerControl } from "./PlayerControlContext";
 import { usePlayerPosition } from "./PlayerPositionContext";
 import { useLayout } from "./VideoLayoutContext";
 import { findSegmentIndexAtFrame } from "./useSceneSegments";
 import { usePlayerPlaying } from "./usePlayerState";
+import { SCRUB_BAR_HEIGHT } from "./ScrubBar";
 import { SegmentedSeekBar } from "./SegmentedSeekBar";
 import {
 	FullscreenButton,
@@ -38,11 +40,15 @@ function BottomTransportBarComponent() {
 
 	return (
 		<div className="@container relative z-20 flex w-full shrink-0 flex-col gap-1.5 border-t border-border px-4 py-2 text-body text-foreground">
-			{ready ? (
-				<SegmentedSeekBar player={player} layout={layout} segments={segments} />
-			) : (
-				<div className="h-3 w-full" aria-hidden />
-			)}
+			<div className={cn("flex w-full items-center", SCRUB_BAR_HEIGHT)}>
+				{ready && (
+					<SegmentedSeekBar
+						player={player}
+						layout={layout}
+						segments={segments}
+					/>
+				)}
+			</div>
 			<div className="flex items-center gap-2">
 				<div className="flex flex-1 items-center gap-2">
 					{ready && <TimeDisplay player={player} layout={layout} />}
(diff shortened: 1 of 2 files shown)


umairnadeem (commented):
lgtm

PR #538, opened 2026-08-09
feat: let users see and edit the project art style

### LINKED ISSUE #321: Let users see and edit the art-style description in the canvas
## Problem

Users have no control over the art style and report it doesn't match their references:

> It just can't seem to follow reference.

Today the art-style description is **hidden, regenerated every prompt, and uneditable**. The `reference-style` plugin (`lib/connectors/llm/plugins/reference-style.ts:29-33`) makes a fresh gateway call on each submission to describe the blended style of all reference images plus character avatars, then prepends the result to the prompt. The user never sees this description and can't correct it. Because it's regenerated each time, results are also non-deterministic between runs.

Notably, the project store already has a `metadata.style` field (`lib/project/store.ts:29`), but the plugin neither reads nor writes it — the derived description is purely ephemeral.

### Reported symptom: split-view / diptych images

Because the chosen style is opaque, users can't tell when the engine has steered generation somewhere unintended. One report: every generated image came out as a side-by-side "split view" of two variants, and the user couldn't tell whether it was a glitch or a choice they were supposed to make.

> all the images generated made this split view image of each generation. not sure if this is a glitch, or you're meant to select one of the two options?

This is the opaque-art-style problem made visible: a phrasing in the hidden, auto-derived style description (e.g. a "comparison"/"two options"/"split" cue blended in from references) propagates into every image, and there's no way to see or remove it. Surfacing and editing the persisted art-style description directly fixes this class of issue — the user can read what's being applied and delete the offending instruction, instead of guessing.

## Proposed solution

Surface the art-style description as explicit, user-editable state in the canvas:

- **Persist** the derived art-style description to the project store (use `metadata.style`) instead of regenerating it every prompt.
- **Show it in the canvas** so users can read exactly what style is being applied.
- **Let users edit it** directly, and have the `reference-style` plugin use the stored value rather than re-deriving when one exists.
- Offer a "regenerate from references" action for when users do want it re-derived.

This gives users deterministic, transparent control over art style instead of a hidden black box.

## Relationship to #320

Complementary. #320 makes reference **images** reach the LLM faithfully (and fixes "Reference N" semantics). This issue makes the resulting **style description** visible and editable. Together they give users both faithful references and explicit control.

## Acceptance criteria

- [ ] The art-style description is persisted in the project store, not regenerated on every prompt.
- [ ] Users can view the current art-style description in the canvas.
- [ ] Users can edit the description, and edits are what the model actually uses.
- [ ] A way to re-derive the description from references on demand.

## Extension: named style presets + "apply to all images" command

Beyond a free-text description, users want to pick a recognizable style and apply it across the whole project in one shot.

### Reported by (verbatim feedback)

> Do you have a command that, when run, will apply a consistent style—such as oil painting, 2D art or clay modelling—to all the images?

### Adds to the above

- **Style presets.** Offer named styles (oil painting, 2D art, clay/claymation, watercolour, etc.) that set `metadata.style`, alongside the free-text/derived description. Presets are just curated values for the same persisted field.
- **Apply to all images.** A single command/action that sets the project art style and re-renders every image node in that style, so consistency is one click instead of editing each node. This reuses the persisted `metadata.style` that the `reference-style` plugin already prepends to every image prompt; the new piece is a bulk re-generate over existing image/animated-image nodes.

### Additional acceptance criteria

- [ ] A set of named style presets that set the project art style.
- [ ] A single action that applies the chosen style to all images (bulk re-generate), not node-by-node.

## Extension: bring-your-own style prompt (paste a full image-prompt as the art style)

Some users already have a dialed-in style living **outside** openslop and just want openslop to render in it. One user drives their YouTube pipeline through a paid Gemini "GEM": they paste a script segment, ask it for N image prompts, and it returns prompts in one consistent style. They can't extract the GEM's internals (usage-rights only, no deep access), but they *can* share a sample of the exact image prompt it emits — and they want openslop to generate in that same style.

This is the editable-art-style field from above, used as an **input**: let the user paste a full, provider-style image prompt (Midjourney-style flags and all) as `metadata.style`, and have the `reference-style` plugin prepend it verbatim to every image node — no re-derivation, no flattening — so the project matches an external style they already like. Pairs directly with the "apply to all images" bulk re-generate.

### Reported by (verbatim feedback)

> currently I'm making YouTube videos in this way: I bought GEM on gemini, and I just need to put a script segment into it and tell it, this segment needs 20 prompts, it will return 20 image prompts, I want openslop to do it in the same image style as GEM gemini, can you do that. The GEM I bought only grants usage rights, and does not allow deep access inside. Let me send you a sample command line for generating images that it provided to me:
>
> `Miniature diorama, faceless clay figures style, a majestic golden throne split in half between a red clay flag and a blue clay flag under a single studio spotlight, social satire theme, macro photography, shallow depth of field, f/2.8, cinematic lighting, claymation textures, vivid colors --ar 16:9 --v 6.0`

### Additional acceptance criteria

- [ ] Users can paste a full free-text image prompt (including provider flags) as the project art style, and it is stored in `metadata.style`.
- [ ] The stored style prompt is applied verbatim to every image node's prompt (not re-derived or summarized away).
- [ ] Combined with "apply to all images", pasting a style and re-rendering makes the whole project match that external style in one action.

umairnadeem:
## Summary

- Adds an art style dialog with the editable description, a preset showcase, and a "From references" action. Reachable from the composer's `+` menu and an asset tile in the canvas.
- Makes `metadata.style` stick: `collectWritableMetadata` declares it write-once, so a script fills an empty one in but never rewrites one the user owns.
- Reworks the `reference-style` LLM plugin. It now describes the reference images only when the user set no style, then stores the result so it shows up in the dialog as editable text, instead of re-deriving an invisible description on every submission.
- Templates carry `style` as an `ArtStyle` (description + optional thumbnail) rather than a bare string.

Preset thumbnails are placeholders (`picsum.photos`, marked with a TODO) until the real renders are uploaded.

Closes #321

## Test plan

- [ ] Open the composer, `+` → **Set art style**. Pick a preset; the tile appears in the composer asset row showing the style is set.
- [ ] Type a free-form description, or paste an image prompt with provider flags (`--ar 16:9 --v 6.0`). Generate a script, then confirm the text survived verbatim in the dialog — the script's `metadata_style` must not overwrite it.
- [ ] With no art style set, upload reference images and generate. The style should be derived once, appear in the dialog afterwards, and the script's image descriptions should follow it. Generating again must not re-derive it.
- [ ] With no art style and no references, generate and confirm the model still writes a style into the field.
- [ ] Open the dialog, click **From references**, and confirm the description is replaced. With no references, the button is disabled.
- [ ] Edit the style on an existing storyboard and confirm image elements go stale, then regenerate and check the new images follow the new style.
- [ ] Check the preset showcase in light and dark, and at mobile width (2 columns) vs desktop (3).


umairnadeem (commented):
lgtm, nice cleanup pulling the derive out into `deriveArtStyle` and the write-once guard on `style`.

umairnadeem (commented):
lgtm, the write-once guard + snapshot reasoning in the plugin docblock hold up. thanks for the revisions!

PR #539, opened 2026-08-09
refactor: nightly maintainability 2026-08-09

umairnadeem:
## Async video polling is now an explicit contract

`BaseVideoProvider.poll` used to fabricate a hollow `VideoProviderResponse` (`id: ""`, `result: {}`) to signal "the provider is still working", and `lib/api/handlers/video.ts` recovered that state by sniffing `upstream.result.video`. The pending case was encoded in the *absence* of data, so the type said "here is an asset" while the value said otherwise.

`poll` now returns a `VideoPoll` discriminated union:

```ts
export type VideoPoll =
  | { kind: "pending"; metadata?: VideoJobMetadata }
  | { kind: "ready"; asset: VideoProviderResponse };
```

The handler branches on `kind`. This matches every other async surface in the repo (`ProcessOutcome`'s `pending | completed`, `graph.ts`'s `SourceNode | JobNode`), and removes the only place a provider hands back an asset envelope that isn't one.

## Also

- `ProcessOutcome<TMeta>` is parameterized by the handler's metadata type, so a `pending` outcome must carry that handler's shape instead of an untyped `Record<string, unknown>`. `videoHandler`'s pending metadata is now checked against `VideoMetadata`.
- `AssetBundle.buildUrl` dropped `const base = AssetBundle.baseUrl || ""` — `baseUrl` is already declared `?? ""`, so the second default was unreachable. One decision, one default.

No behavior change: the pending / ready / failed branches fire on exactly the same conditions as before. Tests updated to the new contract.

## Validation

`npm run build`, `npm run typecheck`, `npm run lint`, `npm run knip`, and `npm test` (128 files / 1111 tests) all pass.

## Open PR overlap check

Considered open PRs #538 (art style), #534 (transport bar), #533 (architecture rethink), #532 (template mode), #531 (ARCHITECTURE.md). None touch `lib/providers/video/**`, `lib/api/handlers/**`, `lib/api/job-handlers.ts` or `lib/api/asset-bundle.ts`. This PR is non-overlapping.

## Skipped

- `lib/api/sse.ts`'s unused `createSSEResponse` — reusable surface, not dead code; wants a human call.
- Returning `stringifyError` to clients (`with-auth.ts`, `process-job.ts`) — API contract change.
- `Waveform` silent peaks-decode failure, `lib/supabase/env.ts` empty-string env defaults — behavior changes.

### DIFF
--- a/lib/api/asset-bundle.ts
+++ b/lib/api/asset-bundle.ts
@@ -78,8 +78,7 @@ export class AssetBundle {
 	}
 
 	static buildUrl(type: string, provider: string, id: string): string {
-		const base = AssetBundle.baseUrl || "";
-		return `${base}/assets/${type}/${provider}/${id}`;
+		return `${AssetBundle.baseUrl}/assets/${type}/${provider}/${id}`;
 	}
 
 	static fromResponse(response: BundleResponse): AssetBundle {
--- a/lib/api/handlers/__tests__/video.test.ts
+++ b/lib/api/handlers/__tests__/video.test.ts
@@ -36,7 +36,10 @@ function job(overrides: Partial<VideoJobRow> = {}): VideoJobRow {
 describe("videoHandler.process", () => {
 	beforeEach(() => {
 		vi.clearAllMocks();
-		poll.mockResolvedValue({ result: {}, metadata: { status: "processing" } });
+		poll.mockResolvedValue({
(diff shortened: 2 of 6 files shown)


umairnadeem (commented):
lgtm

PR #540, opened 2026-08-09
fix: show an animated image as generating while its still frame renders

### LINKED ISSUE #506: Animated image still/video toggle doesn't show the still node's progress

umairnadeem:
## Summary

An `animated_image` is two nodes under the dependency graph: a still-image node feeding the animation node. The card only ever read the animation's snapshot, which stays `"queued"` until the still settles. So the element type that does the most work looked idle for most of it, and the elapsed counter did not start until the animation itself began.

`subtreeProgress` joins the other graph walkers in `lib/generation/graph.ts` and returns whichever snapshot in a node's subtree is running, else whichever is waiting, else the node's own. `flattenGraph` already orders dependencies before dependents, so the snapshot belongs to the earliest stage still in flight. `useGenerate` reports that instead of the raw snapshot status.

There is one `useGenerate` per card, so `GenerationIndicator`, the shimmer, and the disabled state all inherit the fix with no change of their own. Nothing is special-cased to `animated_image`: any element that grows a derived node gets an honest status for free.

`result` and `error` still come from the element's own snapshot; only `status` and `seconds` aggregate. `NodeResult` gains `status` and `seconds`, which `ElementSnapshot` already had.

For elapsed seconds this takes the simpler of the two options the issue offers: surface the elapsed time of the node that is running. The counter now runs through the still instead of sitting at nothing, though it does restart when the animation takes over. Accumulating across the subtree would mean retaining per-node elapsed past commit, which changes snapshot semantics for every element to fix one label.

Out of scope, per the issue: showing *which* stage is running.

## Selected issue

#506, `size: S`, `bug`. It names the change point, the fix is one pure helper plus a two-line hook change, and it is validated at the graph seam rather than through the UI. The only `size: xs` issue open (#529) is already covered by PR #534.

## Validation

- `npm run build` — passes
- `npm test` — 1115 passed / 128 files, including 4 new `subtreeProgress` cases

## Open-PR overlap check

Checked #531, #532, #533, #534, #538, #539 with `gh pr diff --name-only`. None touches `lib/generation/graph.ts`, `lib/generation/__tests__/graph.test.ts`, or `app/components/canvas/hooks/useGenerate.ts`. #533 touches `lib/generation/__tests__/resolveGraph.test.ts`, a different file in the same directory.

## Skipped candidates

Everything else open is `size: M` or larger, or turns on product direction (art-style editing, YOLO mode, publishing kit, BYOK, i18n). `size: S` issues #461, #332, and #254 all need a design call on new surface or on audio defaults.

Closes #506

### DIFF
--- a/app/components/canvas/hooks/useGenerate.ts
+++ b/app/components/canvas/hooks/useGenerate.ts
@@ -3,7 +3,12 @@ import {
 	useGenerationQueue,
 	useQueueSelector,
 } from "@/lib/generation/GenerationQueueProvider";
-import { forElement, isNodeStale, nodeInputs } from "@/lib/generation/graph";
+import {
+	forElement,
+	isNodeStale,
+	nodeInputs,
+	subtreeProgress,
+} from "@/lib/generation/graph";
 import { useNodeBuilder } from "@/lib/generation/useNodeBuilder";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 
@@ -16,6 +21,7 @@ export function useGenerate(element: CanvasContentElement) {
 		[buildNode, element],
 	);
 	const stale = useQueueSelector((q) => isNodeStale(node, q));
+	const progress = useQueueSelector((q) => subtreeProgress(node, q));
 
 	useEffect(() => {
 		if (!stale) return;
@@ -36,8 +42,8 @@ export function useGenerate(element: CanvasContentElement) {
 
 	return {
 		node,
-		status: snapshot.status,
-		seconds: snapshot.seconds,
+		status: progress.status,
+		seconds: progress.seconds,
 		result: snapshot.result,
 		error: snapshot.error,
 		stale,
--- a/lib/generation/__tests__/graph.test.ts
+++ b/lib/generation/__tests__/graph.test.ts
@@ -7,8 +7,11 @@ import {
 	needsGeneration,
 	nodeInputs,
 	sourceNode,
+	subtreeProgress,
 	type GenerationJob,
 	type GenerationNode,
+	type NodeResult,
+	type NodeResults,
 } from "../graph";
 import { GenerationQueue } from "../queue";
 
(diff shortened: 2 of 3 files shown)


app/components/canvas/hooks/useGenerate.ts
```diff
@@ -16,6 +21,7 @@ export function useGenerate(element: CanvasContentElement) {
 		[buildNode, element],
 	);
 	const stale = useQueueSelector((q) => isNodeStale(node, q));
+	const progress = useQueueSelector((q) => subtreeProgress(node, q));
```
umairnadeem, line 24: cancel gets stuck now for an animated image during the still stage: `discard` only aborts/removes `element.id`, so the still node keeps running and `subtreeProgress` keeps reporting generating with the counter ticking, button does nothing. discard/cancel probably needs to walk the subtree the same way this does.

umairnadeem:
closing this one, the subtree-aggregate is the wrong shape for the fix imo.

an animated image card already has a still/video toggle (`AnimatedImagePreview` → `MediaToggle`), so both nodes are already things the user can look at. blending them into one status throws that away: you get a single "generating 12s" that doesn't say which node it belongs to, and flipping the toggle to Still still shows the video node's loader.

what we actually want is the toggle picking the node. Still → the still node's snapshot (status, seconds, result), Video → the animation node's. so while the still renders you flip to Still and watch it progress, flip back to Video and it honestly reads queued, because it is.

that means the mode has to sit above the status props rather than below them, ie `AnimatedImageMedia` selects a snapshot by node id instead of `OutputPreview` handing down one status from `useElementGeneration`. no graph walk, and the "which stage is running" bit the issue punted on falls out for free.

updated #506 with the intended direction so the next run doesn't land here again.

Issue #249, opened 2026-06-12
Image generation control & script relevance (per-node reference images, bulk upload)


umairnadeem:
## Problem

Reference images are project-global only, and generated images are too generic. Users want to:

- **View and set different reference images per image node** (both upfront before prompting and after).
- **Upload custom reference images per scene** and **bulk-upload multiple images**.
- Have generated images **closely reflect each scene's narration/dialogue** rather than being generic.

## Reported by (verbatim feedback)

- "Want to see and control reference images passed to image elements so you can pass different reference images to different image nodes"
- "I want to be able to specify images for each scene upfront before entering prompt"
- "Want to upload custom reference images per scene"
- "Want to add reference images post-prompt"
- "Bulk upload multiple images"
- "Images should reflect the narration script more closely"
- "Images can be generic sometimes - the script should tailor images to each scene's narration/dialogue more closely"

## Relevant code

- `lib/connectors/image/plugins/reference-images.ts` — merges project-level refs only
- `lib/generation/elementMetadataInputs.ts`, `lib/generation/getPromptText.ts` — prompt/attribute construction
- `lib/upload/useImageUpload.tsx` (supports `multiple`), `app/api/upload/image/route.ts`

## Suggested direction

- Per-element reference-image control: view, set, and upload (including bulk) different refs per image node, upfront and post-prompt.
- Improve image prompt construction so each image is tied to its own scene's narration/dialogue.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

umairnadeem:
Splitting this: it bundled two independent problems and neither was tracked cleanly while they shared an issue.

The per-node reference-image half is done in #542. Image and animated-image elements now carry an optional `referenceImagesOverride` attribute, edited from a References pill in the element's settings popover. Absent means the element inherits the project's references, which is what every existing element does; the first add or remove copies the project's list onto the element, and Reset restores inheritance. Uploads are multi-file, so bulk upload is covered at the element level.

The script-relevance half is now #543, which is about tying each image's description to the narration it plays under rather than the scene in general.

Not carried into either: bulk upload as a separate project-level flow, since the multi-file picker already covers it in both places, and clips, since the video connector does not consume reference images today.

PR #542, opened 2026-08-10
feat: per-element reference images on visual elements

umairnadeem:
## Summary
- Image and animated-image elements get an optional `referenceImagesOverride` attribute (comma-separated URLs), edited from a **References** pill in the element's settings popover. Absent = inherit the project's references, which is what existing elements do.
- The popover shows the project's images and says "Using project references"; the first add or remove copies them onto the element, and **Reset** restores inheritance. Uploads are multi-file, reusing the project-level picker.
- An overriding element declares no dependency on `project:referenceImages`, so project changes neither reach it nor stale it. Animated images pass the attribute down to their derived still.
- Implemented as a third `AttributeEdit` kind (`images`) on the existing attribute schema rather than special-casing the element card; the project picker was split into a props-driven `ReferenceImagePicker`, and the duplicated trigger pill became a shared `AttributeTrigger`.

Not included: clips (the video connector doesn't consume reference images today), and the CSV format can't hold a URL with a literal comma, same as the existing `characters` attribute.

Refs #249

## Test plan
- [x] `npm run test:run` — 1135 passing, including new coverage for override vs inherit, empty override, and the conditional dependency
- [ ] Manual: open an image element's settings, confirm the pill reads `Project (N)`; add an image and confirm it flips to `N custom` and the project's own list is untouched
- [ ] Manual: remove all overridden images, confirm the pill reads `None` and generation runs with no references
- [ ] Manual: **Reset** returns the element to the project's list, and editing project references restages inheriting elements but not overriding ones

### DIFF
--- a/lib/connectors/animated_image/attributes.ts
+++ b/lib/connectors/animated_image/attributes.ts
@@ -1,7 +1,9 @@
 import { AttributeSchema } from "../attributes/schema";
 import { durationDef, motionDef } from "../attributes/common";
+import { referenceImagesDef } from "../attributes/referenceImages";
 
 export const ANIMATED_IMAGE_ATTRIBUTES = AttributeSchema.from([
+	referenceImagesDef,
 	{
 		key: "videoPrompt",
 		label: "Video prompt",
--- a/app/components/canvas/elements/AssetTiles.tsx
+++ b/app/components/canvas/elements/AssetTiles.tsx
@@ -70,16 +70,28 @@ function CharacterAssetTile({
 	);
 }
 
-export function ReferenceAssetTiles() {
-	const referenceImages = useProject((s) => s.referenceImages);
-	const removeReferenceImage = useProject((s) => s.removeReferenceImage);
-	return referenceImages.map((url, index) => (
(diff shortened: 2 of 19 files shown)


lib/connectors/animated_image/attributes.ts
```diff
@@ -1,7 +1,9 @@
 import { AttributeSchema } from "../attributes/schema";
 import { durationDef, motionDef } from "../attributes/common";
+import { referenceImagesDef } from "../attributes/referenceImages";
 
 export const ANIMATED_IMAGE_ATTRIBUTES = AttributeSchema.from([
+	referenceImagesDef,
```
umairnadeem, line 6: now that both image and animated_image carry this attr, converting between them silently drops it: `preservedAttributes` only whitelists `characters`, so set a custom reference list on an image, convert it to animated_image, and the override is gone (back to project refs, no undo breadcrumb). worth adding `referenceImagesOverride: ["image", "animated_image"]` to `PRESERVED_ATTRIBUTE_TYPES`.

PR #545, opened 2026-08-10
feat: caption styling controls in the editor sidebar

### LINKED ISSUE #270: Subtitle fine-tuning controls (style, size, position)
## Problem

Subtitles/captions are rendered with fixed styling and offer no user controls. Users want power to fine-tune subtitles (alongside sound, tracked in #254).

Specific defects within the fixed styling:

- **A word longer than the line is cut off instead of wrapping.** The caption span uses `whiteSpace: "nowrap"` (`remotion/components/Captions.tsx:44`) inside a fixed-width box (`:103`), so a single word wider than the line (e.g. "unpredictable") overflows and gets clipped rather than breaking across rows. It should wrap or hyphenate (e.g. "unpredicta" / "-ble"). `buildCaptionData` wraps by whole words and puts an over-long word on its own line (`:62-64`), but `nowrap` still clips it.

- "All I missed was some power to finetune sound and subtitles."
- "Also when word is too long it gets cut. It should go under like unpredictable goes unpredicta and other row goes -ble or something."

## Relevant code

- `remotion/components/Captions.tsx` -- caption renderer (fixed uppercase styling, shadow, line width)
- `remotion/components/Captions.tsx:44` (`whiteSpace: "nowrap"`), `:103` (fixed-width box), `:47-73` (`buildCaptionData` whole-word wrapping) -- the long-word clipping

## Suggested direction

- Expose subtitle controls: on/off, font size, position, color/style, and max line length, threaded from the editor into the Remotion composition.
- Let long words break/wrap rather than clip: drop `nowrap` (or add `overflow-wrap: break-word` / hyphenation) so a word wider than the line flows onto the next row.

Related: #254 (audio mixing controls) for the sound half of this feedback.

---

umairnadeem:
Closes #270.

Captions rendered with fixed styling and offered no controls. The captions panel now owns the whole caption look as a project-level setting.

## Panel

- **Captions** — show/hide, plus a live preview at the project's aspect ratio so size and position read at true scale.
- **Presets** — a static list (Classic, Karaoke, Pop, Boxed, Spotlight, Neon) with live-rendered thumbnails that animate on hover or focus.
- **Text** — font, case (as spoken / upper / lower), position, reveal strategy (whole line vs word by word), size as a percentage of frame height, max words per line.
- **Style** and **Active word** — fill, border color + thickness, background, and a size multiplier on the active word. Both cards are the same component; border and background are off when their swatch is set to none.

## Implementation

- `lib/video/captionStyle.ts` holds the schema, defaults, and ranges; the stored value is parsed at the metadata boundary and falls back to the defaults if it is malformed.
- `lib/video/captionLayout.ts` is the pure line/active-word math, covered by tests.
- `components/captions/CaptionOverlay.tsx` is the one renderer, shared by the Remotion composition and the panel previews — the previews are the real thing at a smaller pixel size, not a mock.
- The style rides in `VideoLayout` next to the transition settings, so the Player and the Lambda render read the same value.
- `ColorField` is a new `components/ui` primitive: swatch trigger, palette popover, native custom input, optional none.

## Fixes from the issue

Long words no longer clip. The caption line is a wrapping flex row with `overflow-wrap: break-word` instead of a `nowrap` span in a fixed-width box, and wrapping is now driven by words per line rather than a character-width estimate.

Verified in the running editor in both light and dark mode: preset selection, color picker, sliders, and live propagation into the player.


components/captions/CaptionOverlay.tsx
```diff
@@ -0,0 +1,119 @@
...
+	return {
+		fontSize: `${scale}em`,
+		color: style.fill,
+		textShadow: style.border
+			? outline(style.border.width, style.border.color)
+			: undefined,
+		backgroundColor: style.background ?? undefined,
+		padding: style.background ? "0.05em 0.2em" : undefined,
+		borderRadius: style.background ? "0.12em" : undefined,
+		overflowWrap: "break-word",
```
umairnadeem: `overflow-wrap: break-word` doesn't reduce a box's min-content size, and this span is a flex item with `min-width: auto`, so a single very long word won't break, it just overflows and gets clipped by the frame (the thing #270 is about). `overflowWrap: "anywhere"` does affect min-content sizing, or keep break-word and add `minWidth: 0`.

components/ui/color-field.tsx
```diff
@@ -0,0 +1,226 @@
...
+					/>
+					<label className="flex h-7 shrink-0 items-center rounded-md bg-input pr-1 pl-2">
+						<span className="sr-only">{label} opacity</span>
+						<input
+							type="number"
+							min={0}
+							max={100}
+							value={alphaPercent(color)}
+							onChange={(event) =>
+								setExternal(withAlphaPercent(color, event.target.valueAsNumber))
```
umairnadeem: clear the opacity field (select all + delete) and `valueAsNumber` is NaN, so `withAlphaPercent` builds `"#ffffffNaN"` and writes that into the style. the input snaps back to 100 (since `alphaPercent` doesn't see a 9-char hex) but the persisted color is now invalid css and the caption renders uncolored. bail on `Number.isNaN(percent)`.

remotion/compositions/VideoComposition.tsx
```diff
@@ -77,17 +93,23 @@ function SequenceContent({ element }: { element: ResolvedElement }) {
...
+			return <AudioSequence element={element} captionStyle={captionStyle} />;
 		default:
 			return null;
 	}
 }
 
-function SeriesEntry({ seq }: { seq: SeqType }) {
+function SeriesEntry({
+	seq,
+	captionStyle,
```
umairnadeem: `CONVENTIONS.md` § Frontend: *"Composition over context. Don't thread props through layers that don't use them; prefer composition with children. Reach for context only for genuinely cross-cutting state"* — `SeriesEntry` and `SequenceContent` both just forward `captionStyle` to `AudioSequence`. this is cross-cutting composition state, so a small context provider at the `VideoComposition` root (or passing the rendered `<Captions/>` down as children) fits the rule better than two pass-through layers.

app/components/canvas/CanvasProviders.tsx
```diff
@@ -26,6 +27,7 @@ export function CanvasProviders({
 }) {
 	return (
 		<RenderProvider>
+			<CaptionFonts />
```
umairnadeem: this fetches all 13 caption faces (~660KB, InterVariable alone is 352KB and it's a second copy since layout.tsx already ships it through next/font) on every canvas mount, even with captions off and the panel never opened. mounting `CaptionFonts` from the captions panel instead would keep it off the editor's load path.

lib/video/captionFonts.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * The caption faces, self-hosted so the editor preview and the Lambda render
+ * draw with the same glyphs. Each entry names a real font family, so one
+ * `font-family` value works in both once the files are loaded — the browser
+ * gets them from `app/caption-fonts.css`, the renderer from
```
umairnadeem: `app/caption-fonts.css` doesn't exist, the browser gets these from `CaptionFonts.tsx` calling `loadCaptionFonts`. bad breadcrumb for the next person hunting for where the faces get registered.

remotion/loadCaptionFonts.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Holds rendering until the caption faces are ready. Without the delay a frame
+ * can be captured mid-load and bake in a fallback face.
+ */
+const handle = delayRender("Loading caption fonts");
+
+loadCaptionFonts((file) => staticFile(`fonts/${file}`)).then(
+	() => continueRender(handle),
+	(error) => {
+		throw error;
```
umairnadeem: throwing inside a `.then` rejection handler just makes another rejected promise, nothing catches it. so a font 404 on lambda = unhandled rejection, `handle` never cleared, render stalls until the delayRender timeout and reports "not cleared after 30000ms" instead of the actual fetch error. `cancelRender(error)` is the api for this.

remotion/compositions/VideoComposition.tsx
```diff
@@ -23,7 +23,8 @@ import { getPresentation } from "@/lib/video/transitionPresentations";
 import { audioVolume } from "@/lib/video/audioVolume";
 import { volumeToGain } from "@/lib/video/elementAttributes";
 import { ELEMENT_TYPES } from "@/lib/canvas/types";
-import { Captions } from "../components/Captions";
+import { CaptionStyleProvider, Captions } from "../components/Captions";
+import "../loadCaptionFonts";
```
umairnadeem: `VideoPreview` lazy-imports this module, so the side effect runs in the editor too and every face gets registered twice (once here, once via `CaptionFonts`), plus a module-scope `delayRender` in the Player. worth moving the load out of the composition module and into the remotion entry so only the renderer runs it.

components/captions/CaptionOverlay.tsx
umairnadeem: Fixed in 5a12e4e — the span uses `overflowWrap: "anywhere"` now. Verified a 33-char word wraps to 145px inside a 198px frame instead of overflowing it.

components/ui/color-field.tsx
umairnadeem: Fixed in 5a12e4e. `withAlphaPercent` bails on `Number.isNaN(percent)` and returns the color untouched. Pulled the parsing out into `lib/color/hexAlpha.ts` with tests covering the cleared-field case, since it was only reachable through the picker before.

remotion/compositions/VideoComposition.tsx
umairnadeem: Fixed in 5a12e4e. `Captions.tsx` owns a `CaptionStyleProvider` built on `createRequiredContext`, and `VideoComposition` wraps its tree in it — `SeriesEntry` and `SequenceContent` no longer see `captionStyle` at all.

lib/video/captionFonts.ts
umairnadeem: Fixed in 5aad112 — the comment names `app/components/canvas/CaptionFonts.tsx` and `remotion/loadCaptionFonts.ts`, the two places that actually register the faces.

lib/color/hexAlpha.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/** Returns the color unchanged for a percentage the field can't supply yet. */
+export const withAlphaPercent = (color: string, percent: number): string => {
+	if (Number.isNaN(percent)) return color;
+
+	const clamped = Math.min(100, Math.max(0, percent));
+	const pair = Math.round((clamped / 100) * 255)
+		.toString(16)
+		.padStart(2, "0");
+	return `${color.slice(0, 7)}${pair}`;
```
umairnadeem: `slice(0, 7)` assumes a 6-digit hex, but `HexColorInput` with `alpha` also emits 3- and 4-digit shorthand (`validHex` accepts lengths 3/4/6/8). type `#fff` in the hex box then nudge opacity → `#fff80`, invalid, caption color drops; `#fff8` → `#fff880`, silently a different color. worth expanding shorthand to 6 digits before appending the pair.

lib/video/captionFonts.ts
```diff
@@ -0,0 +1,128 @@
...
+ * resolves bundled assets itself. Registering is idempotent, so a surface can
+ * ask for what it needs without knowing what another already loaded.
+ */
+export function loadCaptionFonts(
+	url: (file: string) => string,
+	fonts: readonly CaptionFont[] = CAPTION_FONTS,
+): Promise<void> {
+	const faces = fonts.flatMap((font) => {
+		if (registered.has(font)) return [];
+		registered.add(font);
```
umairnadeem: `registered.add` happens before `face.load()` resolves, so a transient 404/network blip marks the face loaded forever and every later caller skips it, the editor silently falls back for the rest of the session. also nothing catches it (`ActiveCaptionFont` fires and forgets), so it lands as an unhandled rejection. maybe only add on success?

app/components/canvas/CanvasProviders.tsx
umairnadeem: Fixed in e65775f, though not by moving the mount — the Player draws captions whether or not the panel is open, so panel-scoping the whole thing would leave the preview on fallback faces. Split by what each surface needs instead: `ActiveCaptionFont` stays in `CanvasProviders` but registers only `style.font`, and only while captions are enabled (captions off = zero fetches), while `CaptionFontLibrary` mounts with the panel for the picker rows and preset thumbnails. `loadCaptionFonts` now takes an optional subset and is idempotent, so the two surfaces overlap without double-registering.

Not solved: with `inter` as the default caption font its face is still a second copy, since next/font serves the same file from a hashed URL under a generated family name and `var(--font-inter)` would not resolve on Lambda. Dropping Inter from the caption library or defaulting to another face would close that.

remotion/loadCaptionFonts.ts
umairnadeem: Fixed in e65775f — `cancelRender(error)`.

remotion/compositions/VideoComposition.tsx
umairnadeem: Fixed in e65775f. The `import "./loadCaptionFonts"` moved to `remotion/index.ts` next to `registerRoot`, so only the renderer and Studio run it and the Player never sees a module-scope `delayRender`.

lib/color/hexAlpha.ts
umairnadeem: Fixed in 3c21965. Confirmed the shorthand does reach `onChange` — react-colorful’s `validHex` allows lengths 3/4/6/8 with `alpha`. Both helpers now expand shorthand to its long form first, so the fix covers every caller rather than just the hex box: `#fff` → `#ffbb00`-style expansion before the pair is appended, and `alphaPercent("#fff8")` reads 53 instead of 100. Tests cover both digit widths.

lib/video/captionFonts.ts
umairnadeem: Fixed in 3c21965 — a font is added to `registered` only after all of its files resolve, so a transient failure is retried by the next caller (panel opening, font changing) instead of being skipped for the session. The editor path also catches now and logs: a missing face costs the preview its glyphs, not the session, and the renderer keeps the opposite policy with `cancelRender` since a frame baked with a fallback face is a broken deliverable.

umairnadeem (commented):

PR #546, opened 2026-08-10
fix: nightly bug fixes 2026-08-10

umairnadeem:
Three correctness fixes around empty and whitespace-bearing custom attributes, found by nightly bug hunt.

## Bugs fixed

**`parseXmlTag` collapsed whitespace inside attribute values.** It split the tag on `/\s+/` and rejoined the attribute string with single spaces, so every run of whitespace in an attribute value was flattened. Text attributes (`appearance`, `videoPrompt`, …) are edited in a `<textarea>`, and `escapeXml` does not escape newlines, so a user's line breaks and indentation were silently destroyed on the next save/reload — violating the round-trip invariant `lib/canvas/xmlEscape.ts` documents. Now the tag name is sliced off and the attribute regex runs over the remainder verbatim, which also drops the join.

**`TextAttributePopover` wrote `""` when a field was cleared.** That persists as `attr=""` in the OSML and reads back as a changed generation input (`toNode` puts `customAttributes` straight into `inputs.attributes`), so an element the user cleared back to its original state is marked stale and regenerates. It now deletes the key, matching how `writeCharacters` already handles the `characters` attribute.

**`character-references` leaked an empty `characters` attribute.** `beforeGenerate` returned `params` untouched when the attribute was present but empty, so the raw attribute rode downstream into the provider params; every other path in that plugin strips it. Reachable from LLM-authored OSML.

## Tests

- `parseXmlTag`: whitespace and newlines preserved inside an attribute value; attributes separated by a newline still parse.
- `osmlSerializer`: serialize → `parseOSML` round trip keeps the whitespace a user typed into an attribute.
- `character-references`: an empty `characters` attribute is stripped.

All three new behavioural tests fail against the pre-fix code and pass after.

## Validation

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test:coverage` all pass. 133 files / 1144 tests.

## Open PR overlap check

Only one PR open: **#545** (`feat: caption styling controls in the editor sidebar`), covering `app/components/canvas/panel/*`, `components/ui/*`, `components/captions/*`, `remotion/*`, `lib/video/caption*`, `lib/video/{scene-builder,types}.ts`, `app/components/video/useVideoLayout.ts`, `app/globals.css`, and `package*.json`. This PR touches `lib/canvas/parseXmlTag.ts`, `lib/connectors/image/plugins/character-references.ts`, `app/components/canvas/elements/attributes/TextAttributePopover.tsx` and their tests — no file or theme overlap.

### DIFF
--- a/app/components/canvas/elements/attributes/TextAttributePopover.tsx
+++ b/app/components/canvas/elements/attributes/TextAttributePopover.tsx
@@ -36,7 +36,9 @@ export function TextAttributePopover({
 	const commit = useCallback(
 		(next: string) => {
 			if (next === value) return;
-			updateElementAttrs(editor, element, { [attrKey]: next });
+			// Clearing the field means the element has no such attribute. Writing ""
+			// instead would persist as `attr=""` and read back as a changed input.
+			updateElementAttrs(editor, element, { [attrKey]: next || null });
 		},
 		[editor, element, attrKey, value],
 	);
--- a/lib/canvas/__tests__/osmlSerializer.test.ts
+++ b/lib/canvas/__tests__/osmlSerializer.test.ts
@@ -192,6 +192,12 @@ describe("serialize round trip", () => {
 		expect(serializeOSML([scene])).toBe(first);
 	});
 
+	it("keeps the whitespace a user typed into an attribute", () => {
+		const appearance = "a tall  woman\nin a red coat";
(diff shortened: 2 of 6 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
useless

PR #547, opened 2026-08-10
refactor: nightly maintainability 2026-08-10

umairnadeem:
Three facts that lived in more than one file now live in one, plus a stale paragraph in ARCHITECTURE.md. No behavior change.

## One home for what a composer mode is

"Script mode has no target length" was written twice, in different layers: `ConfigProvider` left `scriptLengthPlugin` out of the `script` entry of `MODE_PLUGIN_FACTORIES`, and `ComposerCopilot` independently hardcoded `!isScriptMode` to hide the length pill. Adding a fourth mode, or letting script mode take a length, meant finding both — and nothing would have failed if you found only one.

New `lib/config/modes.ts` holds `MODE_SPECS`: label, `targetsLength`, and the mode's prompt plugin. `modePlugins(mode, templateId)` appends `scriptLengthPlugin` from that flag, and the composer reads `MODE_SPECS[mode]` for both the label and whether to render the pill. The mode labels moved off `ComposerCopilot` into the same record (the shape `lib/video/videoLength.ts` already uses), so `MODE_OPTIONS` maps `MODES` directly instead of casting `Object.keys`.

Covered by `lib/config/__tests__/modes.test.ts`, which asserts the length plugin appears in exactly the modes whose spec says it should.

## One home for the image-authored element types

`image | animated_image` was spelled out in three places: `preservedAttributes.ts` (twice, as the types that keep characters and reference images), `ElementCharacters.tsx` (who gets the characters picker), and `ElementContainer.tsx` (who gets the upload button). Three sites, one fact — rule of three.

`IMAGE_AUTHORED_TYPES` in `lib/canvas/types.ts` joins `FOREGROUND_TYPES` as a named group beside `ELEMENT_TYPES`, and the three sites test membership.

## ARCHITECTURE.md

The Editor state section still documented `getProjectStore(projectId)` as the imperative escape hatch. That function was deleted in #533, along with the module-level store map; the handle now comes from `ProjectStoreProvider`. The paragraph named `ProjectEditor` and `useGenerateAll` as callers, which is no longer true either. Rewritten against what master does today.

One line added to the generation-graph section: an element that overrides project state declares no edge to that source node, so project edits neither reach it nor stale it. That rule arrived with per-element reference images (#542) and is not guessable from the rest of the section.

## Open PR overlap check

Open PRs considered: #545 (caption styling — `lib/video/caption*`, `canvas/panel/*`, `components/ui/*`, `remotion/*`, `lib/project/types.ts`) and #546 (empty attribute values — `TextAttributePopover`, `parseXmlTag`, `character-references`). This PR touches none of those files or themes.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `npm run build`, and `npm test` (134 files, 1142 tests) all pass.


umairnadeem (commented):
lgtm

umairnadeem:
useless

PR #548, opened 2026-08-10
perf: nightly performance optimizations 2026-08-10

umairnadeem:
## What changed

Canvas id lookups (`findElementById` / `findNodeById`) ran through `Editor.nodes({ at: [] })`. That walks the document as a generator, descends into every text leaf, and allocates a path per visit — an O(document) traversal for a lookup the canvas's fixed two-level shape (scenes hold content, content holds text) answers directly.

Two hot paths pay for it:

- `useScriptSync` runs three lookups per streamed chunk while an LLM writes the script into the editor, so the cost grew with the script it was building.
- `handleDragOver` runs one per pointer move for the whole duration of a drag.

They now use a direct scenes-then-content walk. Alongside it, `updateNodeText` reads current text via `Node.string` on the node the caller already resolved instead of `Editor.string`'s range machinery — the app declares no void or inline elements (`isVoid`/`isInline` are nowhere in the tree), so the two are equivalent.

## Why it matters

Replaying a full generation through the real `useScriptSync` body (parser → `findNodeById` → `updateNodeText`), median of 3 runs:

| script | before | after | per chunk |
| --- | --- | --- | --- |
| 20 scenes / 80 elements | 333 ms | **121 ms** | 797 µs → **289 µs** |
| 40 scenes / 160 elements | 807 ms | **174 ms** | 964 µs → **208 µs** |

The point isn't just the constant factor: per-chunk cost was climbing with document size (797 µs → 964 µs as the script doubled) and is now flat (289 µs → 208 µs). That work sat on the main thread during the app's most-watched interaction, competing with the Slate re-render for the same frame.

A single `findElementById` at 40 scenes goes 388 µs → 18 µs, which is the drag-over win.

No behavior change: same document-order results, same `null` on miss, same throw-on-invalid-path semantics. Added a test covering content that sits at the top level before `withScenes` normalization wraps it, since that's the one shape the traversal has to keep handling.

## Open PR overlap check

Open PRs at the time of this run: **#545** (caption styling — `lib/video/*`, `remotion/*`, caption panel, `components/ui/*`), **#546** (`parseXmlTag`, character-references plugin, `TextAttributePopover`), **#547** (`lib/canvas/preservedAttributes.ts`, `lib/canvas/types.ts`, `lib/config/*`, `ElementContainer`, `ElementCharacters`, `ComposerCopilot`).

This PR touches only `lib/canvas/editorOps.ts` and its test. No file or theme overlap with any of them.

## Considered and skipped

- **Slate render path**: consumers already use `useSlateStatic`, `renderElement` is stable, there's no custom `decorate`, and `isSceneElement`/`Element.isElement` are shallow in Slate 0.123 — not the deep validation they look like. Nothing to fix.
- **Remaining per-chunk cost** is `Transforms.insertText` itself. That's Slate's own transform machinery; working around it would mean bypassing the editor.
- **`useOSMLStreamParser`'s per-chunk `structuredClone`** is bounded to the last 3 nodes and measured at ~2 ms across a whole 20-scene generation. Left alone.
- **Video/timeline/Remotion paths** were skipped on the overlap guard, not on merit — #545 is sitting on them.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `npm run build`, and `npm test` (1141 tests, 133 files) all pass.

### DIFF
--- a/lib/canvas/__tests__/editorOps.test.ts
+++ b/lib/canvas/__tests__/editorOps.test.ts
@@ -58,6 +58,15 @@ describe("findNodeById", () => {
 		const editor = makeEditor([scene([content("narration", "n1")], "s1")]);
 		expect(findNodeById(editor, "s1")).toBeNull();
 	});
+
+	it("finds content sitting at the top level before normalization wraps it", () => {
+		const editor = createEditor();
+		editor.children = [
+			scene([content("narration", "n1")], "s1"),
+			content("image", "img1"),
+		];
+		expect(findNodeById(editor, "img1")?.[1]).toEqual([1]);
+	});
 });
 
 describe("findElementById", () => {
--- a/lib/canvas/editorOps.ts
+++ b/lib/canvas/editorOps.ts
@@ -1,29 +1,57 @@
-import { Editor, Element, type NodeEntry, type Path, Transforms } from "slate";
-import type { CanvasContentElement, CanvasElement } from "@/lib/canvas/types";
+import {
+	Editor,
+	Node,
+	Text,
+	type NodeEntry,
+	type Path,
+	Transforms,
+} from "slate";
+import {
+	SCENE_TYPE,
+	type CanvasContentElement,
+	type CanvasElement,
+} from "@/lib/canvas/types";
 import { withoutCaretMarker, ZERO_WIDTH_SPACE } from "./constants";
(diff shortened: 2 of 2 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
overengineered

Issue #543, opened 2026-08-10
Tie each image's description to the narration it plays under


umairnadeem:
## Problem

An image element's prompt is authored once by the LLM as part of the OSML script, and nothing in the rules ties it to the narration or dialogue it actually plays under. The image rules (`lib/connectors/llm/plugins/osml.ts:64-81`) tell the model to describe "the current scene" in detail and to change the image at least every 2 narrative lines, but never to depict what is being *said* over it. The result is images that are plausibly on-topic for the scene while being generic against the specific line: the narration names a concrete moment, the image shows the general setting.

The same gap exists after authoring. An image node's generation inputs are its own prompt text plus its attributes (`lib/generation/resolveGraph.ts:34`), and the source nodes it can depend on are project-wide — references, art style, aspect ratio (`lib/generation/sourceNodes.ts:12-23`). No image element depends on, or can read, the narration sitting beside it in the same scene. So regenerating one image cannot re-tie it to its line, and editing the narration does not stale the image that illustrates it.

## Reported by (verbatim feedback)

- "Images should reflect the narration script more closely"
- "Images can be generic sometimes - the script should tailor images to each scene's narration/dialogue more closely"

## Relevant code

- `lib/connectors/llm/plugins/osml.ts:64-81` — the image-tag authoring rules; where the narration-derived requirement is missing
- `lib/connectors/image/plugins/imageChain.ts:7-13` — the image plugin chain; where a scene-context plugin would slot in
- `lib/generation/sourceNodes.ts:12-23` — the leaves an image node can currently read, all project-wide
- `lib/generation/resolveGraph.ts:34` — a node's inputs are its own prompt and attributes, nothing from its siblings

## Suggested direction

Start with authoring, which is where the prompt is actually written and the cheapest place to fix the common case: add a rule to the image section requiring each description to depict the specific moment described by the narration and dialogue lines it spans, not just the scene's setting. The existing "standalone prompt" rule stays — the description still has to stand on its own for a model with no story context; it just has to stand on its own *about the right moment*.

If that alone proves insufficient, the structural fix is a scene-context plugin on the image chain that makes an image node depend on its scene's narration text, so the illustrated line is part of the node's inputs. That is the deeper change and should be judged on whether the prompt rule falls short, not taken up front.

## Acceptance criteria

- [ ] The OSML image rules require each `<image>` / `<animated_image>` description to depict the moment its adjacent narration/dialogue describes
[...]

Issue #536, opened 2026-08-09
Uploaded images silently sever the prompt→preview link (never stale, no visual signal, duplicated per surface)


umairnadeem:
## Problem

Everywhere else in the canvas, the prompt is the input and the preview is its output: edit the
prompt, the element goes stale, `StaleIndicator` appears, you regenerate. Uploading an image
severs that link, and nothing in the UI says so.

`commitResult(..., { pinned: true })` marks the result as user-supplied
(`lib/generation/queue.ts:131-146`), and `needsGeneration` then returns `false` unconditionally
for a pinned node (`lib/generation/graph.ts:134-137`). Because `isNodeStale` is
`hasResult && needsGeneration`, **a pinned element can never be stale**. So after an upload:

- The preview no longer has anything to do with the prompt sitting above it, and looks
  identical to a generated one — no badge, no affordance, no tooltip.
- The user can edit that prompt freely and get zero feedback. `StaleIndicator` is structurally
  unreachable, so the one signal we have for "input and output disagree" is off exactly when
  the input and output disagree the most.
- The only way back to a generated image is to notice the Regenerate button and press it, which
  silently discards the upload.

The pin itself is correct and worth keeping — it's what stops Generate All from clobbering a
user's image (#530). The bug is that pinning currently means two different things at once:
"don't regenerate this automatically" and "this element has no prompt relationship." Only the
first is true.

### The same idea is already implemented twice, differently

The character avatar path hand-rolls its own parallel version of all of this:

- A second flag, `MetadataCharacter.avatarUploaded` (`lib/project/types.ts:46`), set alongside
  the pin at `CharacterEditModal.tsx:190-196`.
- A hand-written "editing the prompt un-uploads it" rule at `CharacterEditModal.tsx:89`, which
  clears `avatarUploaded` on regenerate. This is a one-off, character-only approximation of the
  exact behavior this issue is asking for generally.
- `avatarUploaded` then leaks into three unrelated modules that each re-derive "was this
  uploaded?" for themselves: `lib/connectors/llm/plugins/project-metadata.ts:31`,
  `lib/connectors/llm/plugins/character-avatar-style.ts:34`, `lib/project/ensureArtStyle.ts:21`.

Meanwhile `ElementUploadButton` sets no such flag at all, so image and animated-image elements
get none of that behavior. Two sources of truth for one fact, and the affordances differ by
surface. Adding upload to the art-style modal would make it three.

## Reported by (verbatim feedback)

[...]

PR #549, opened 2026-08-10
fix: show the still node's progress under the Still toggle

### LINKED ISSUE #506: Animated image still/video toggle doesn't show the still node's progress

umairnadeem:
## Summary

An animated image read "Queued…" for most of the time it was actually working. It is two nodes — a still-image node feeding the animation node — and the card read every status off the animation, which genuinely stays queued until the still settles. So the element type doing the most work was the one that looked idle while doing it.

The toggle now picks the node, per the direction on the issue (and explicitly not the aggregate approach closed in #540):

- Still → the still node's `status`, `seconds`, `error`, and frame
- Video → the animation node's, as before

`stillFrameUrl` becomes `stillSnapshot`, which returns the still node's whole snapshot instead of just its url, keeping the still-node lookup in the `animated_image` connector. `AnimatedImageMedia` takes one `ModeState` per side and resolves which to render from its own `mode`, so no status is invented, nothing is blended, and no other element type changes.

Issue #506 is `size: S` with a maintainer comment prescribing the exact approach, so there was no design decision to make. The change is four files, confined to the animated-image preview and its still-node helper. Plain single-node elements are untouched — `OutputPreview` still hands every preview the element's own status.

## Validation

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` — 133 files, 1141 tests
- `npm run test:e2e` — 3 smoke tests

Coverage is at the seam the issue asked for (where the still node's snapshot is selected), not on rendered copy: `stillSnapshot` now has a case asserting it returns the still node's snapshot and leaves the animation's untouched, alongside the existing uploaded-frame regression tests.

## Open PR overlap

Checked #545, #546, #547, #548 with `gh pr diff --name-only`. None touches `AnimatedImagePreview.tsx`, `still-frame.ts`, or either test file.

## Skipped candidates

No `size:xs` issues are open. Skipped as too broad, product-heavy, or unvalidatable unattended: #543 (prompt-rule change with no local way to judge output quality, plus an explicit "decide with evidence" step), #536/#535/#492/#491/#462/#451/#385/#379 and the rest of the M/L/XL enhancement backlog. #270 (subtitle controls) overlaps open PR #545.

Closes #506

### DIFF
--- a/app/components/canvas/elements/preview/AnimatedImagePreview.tsx
+++ b/app/components/canvas/elements/preview/AnimatedImagePreview.tsx
@@ -14,7 +14,6 @@ import type {
 	PlaceholderProps,
 } from "./status";
 
-/** One side of the toggle: the node it shows and that node's own progress. */
 type ModeState = GenerationState & {
 	error: string | null;
 	url: string | undefined;
@@ -25,8 +24,6 @@ type AnimatedImageMediaProps = Pick<PlaceholderProps, "onDiscard"> & {
 	still: ModeState;
 };
 
-// The toggle picks the node, so the card never blends the two: while the still
-// renders, Still shows it generating and Video honestly reads "Queued…".
 export function AnimatedImageMedia({
 	animated,
 	still,
--- a/app/components/canvas/elements/preview/AnimatedImagePreview.tsx
+++ b/app/components/canvas/elements/preview/AnimatedImagePreview.tsx
@@ -3,24 +3,37 @@
 import { useState, type CSSProperties } from "react";
 import { Image as ImageIcon, Video } from "@/components/ui/icon";
 import { MediaToggle } from "@/components/ui/media-toggle";
-import { stillFrameUrl } from "@/lib/connectors/animated_image/plugins/still-frame";
+import { getPrimaryUrl } from "@/lib/connectors/assetUrl";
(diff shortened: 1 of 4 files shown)


app/components/canvas/elements/preview/AnimatedImagePreview.tsx
```diff
@@ -46,17 +60,25 @@ export function AnimatedImageMedia({
...
 	const { node } = useElementGeneration();
-	const stillUrl = useQueueSelector((queue) => stillFrameUrl(node, queue));
+	const still = useQueueSelector((queue) => stillSnapshot(node, queue));
 
 	return (
 		<AnimatedImageMedia
-			{...state}
-			stillUrl={stillUrl}
-			videoUrl={result?.videoUrl}
+			onDiscard={onDiscard}
```
umairnadeem, line 71: now that the Still side reads the still node's status, the cancel button it renders is wired to the wrong node: `onDiscard` is `discard(animationElementId)` from `useElementGeneration`. so on Still while the still is generating, hitting X aborts the queued animation and leaves the still job you're watching spinning. maybe pass a per-side discard (still side -> `queue.discard(stillElementId(node.id))`)?

PR #551, opened 2026-08-11
fix(canvas): wrap narrow element actions

### LINKED ISSUE #550: Element card actions row should wrap (right-aligned) when the card is too narrow
## Problem

The element card's actions row is a single non-wrapping flex line, and every child opts out of shrinking. When the card gets narrow (multi-column canvas layouts, small viewports), the buttons overflow horizontally and get silently clipped by the card's `overflow-hidden` instead of reflowing onto a second line.

The row holds `ElementStaleIndicator`, `ElementUploadButton`, `AnimateButton`, and `ElementGenerateButton`. The "Stale" pill has fixed text with no truncation, and the generate button's label grows to "Regenerate"/"Generating…", so the row is at its widest exactly when the user most needs to see it.

Note that the header/badge row directly above it already gets this right (`flex min-w-0 flex-1 flex-wrap items-center gap-1.5`), so the fix has an in-repo precedent.

- "in element container, the row containing actions (upload, animate, regenerate, stale indicator, etc.) should wrap idiomatically when parent container is too narrow."
- "note that the wrap should honor the right-align of these buttons"

## Relevant code

- `app/components/canvas/elements/ElementContainer.tsx:122` — the actions row: `mt-2 flex items-center justify-end gap-2 select-none`, no `flex-wrap`, no `min-w-0`
- `app/components/canvas/elements/ElementContainer.tsx:85` — card wrapper is a `@container` with `overflow-hidden`, so `@sm:` variants are available and overflow is clipped rather than visible
- `app/components/canvas/elements/ElementContainer.tsx:91` — header row already uses `flex-wrap` + `min-w-0`; the pattern to mirror
- `app/components/canvas/elements/GenerateButton.tsx:14` — `StaleIndicator` pill, fixed "Stale" text
- `app/components/canvas/elements/GenerateButton.tsx:69` — `ElementGenerateButton`, `shrink-0`, widest and least compressible item
- `app/components/canvas/elements/AnimateButton.tsx:24` — `shrink-0` and `hidden sm:inline` label, using the **viewport** breakpoint rather than the card's `@sm:` container query, which disagrees with the rest of the card

## Suggested direction

Add `flex-wrap` (plus `min-w-0`) to the actions row so it reflows onto additional lines. Wrapped lines must stay right-aligned: `justify-end` already right-aligns each wrapped line, so keep it, and add row gap so wrapped lines don't collide (`gap-2` covers both axes).

While in here, switch `AnimateButton`'s label breakpoint from the viewport `sm:` to the card's `@sm:` container query so the card's responsive behavior is driven by its own width. `SceneContainer` has analogous action rows worth the same treatment if they share the problem.

## Acceptance criteria

- [ ] Narrowing an element card wraps the actions row onto multiple lines instead of clipping it
- [ ] Every wrapped line stays right-aligned, matching the unwrapped layout
- [ ] Wrapped lines have vertical spacing consistent with the row's horizontal gap
- [ ] `AnimateButton`'s label visibility keys off the card container query, not the viewport

---

floze-the-genius:
## Summary

- let element-card actions wrap when their card is too narrow, while keeping each line right-aligned with the existing spacing
- key the Animate label to the element card's existing container breakpoint instead of the viewport breakpoint
- leave `SceneContainer` unchanged because it has no analogous action row

Fixes #550

## Validation

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run test:coverage` (133 files, 1,140 tests)
- `npm run build`
- `npm run test:e2e` (3 Chromium smoke tests)
- local browser check at a 220 px card width: actions wrap to two right-aligned rows, row/column gaps are both 8 px, and `scrollWidth` equals `clientWidth`
- local browser check confirmed the Animate label is hidden in the narrow card and visible in a wide card under the same viewport

## AI assistance

OpenAI Codex assisted with implementation and validation. I reviewed the final two-line source diff and all reported test results.

### DIFF
--- a/app/components/canvas/elements/AnimateButton.tsx
+++ b/app/components/canvas/elements/AnimateButton.tsx
@@ -23,7 +23,7 @@ export function AnimateButton({ element }: { element: CanvasContentElement }) {
 			onClick={() => refineScript(animateImagePrompt(element.id))}
 		>
 			<MagicVideo aria-hidden="true" />
-			<span className="hidden sm:inline">Animate</span>
+			<span className="hidden @sm:inline">Animate</span>
 		</Button>
 	);
 }
--- a/app/components/canvas/elements/ElementContainer.tsx
+++ b/app/components/canvas/elements/ElementContainer.tsx
@@ -119,7 +119,7 @@ export function ElementContainer({
 							</div>
 						</div>
 						<div
-							className="mt-2 flex items-center justify-end gap-2 select-none"
+							className="mt-2 flex min-w-0 flex-wrap items-center justify-end gap-2 select-none"
 							contentEditable={false}
 						>
 							<ElementStaleIndicator />


floze-the-genius:
@umairnadeem This is the scoped two-line fix for #550: the action row now wraps with per-line right alignment, and the Animate label follows the existing element-card container breakpoint. I ran lint, formatting, typecheck, Knip, all 1,140 unit tests with coverage, the production build, and 3/3 Playwright smoke tests; a narrow/wide browser check also confirmed no overflow and the expected container-query behavior. GitHub CI is currently stopped at `action_required` before jobs start, and the Vercel preview is awaiting team authorization. Could you approve the workflow and review when convenient?

umairnadeem (approved):
@floze-the-genius thank you for your contribution!

PR #552, opened 2026-08-11
feat: pick the script's language in the composer

umairnadeem:
## Summary

- **Language dropdown in the composer**, beside the aspect ratio and length pills. Defaults to **Auto** (today's behaviour: the model follows the language you typed in); picking a language pins every prompt hop to it.
- Fix the vanilla sleep-story template generating Spanish scripts from English input. Both language rules were phrased as exclusions ("not the language of the example", "ignoring the language of any example, reference, or instruction text"), which in template mode disqualifies nearly every English signal in the prompt and pushes the model off English entirely. They now name the source positively, with English as the stated fallback, through one helper (`spokenLanguage`) that every prompt builder calls with the signal its own prompt can still see.
- The language is a **project-level field** (`metadata.language`), not `narration.language` — that one means the narrator *voice's* language and is carried per character too. Reading it as the script's language made three templates pin their scripts to English through voice-filter data.
- Voices still follow the words: in Auto the script reports the language it wrote in and that drives voice search; a pinned project language outranks that report, so switching the picker after an Auto run can't leave the old voice behind. A narrator language the user picked is write-once, so a later generation can't overwrite it.
- Wire up `thinkingLevel`, which existed on the params and the route schema but was never read by the provider. Typed as an enum, parsed at the route boundary, sent as `output_config.effort` with adaptive thinking. Defaults to `high`; the two short describe-a-reference-image calls drop to `low`.
- Bump the default model to `claude-opus-5`, which rejects `temperature` outright, so the Anthropic provider stops forwarding it. The parameter stays on the shared params and the route schema for providers that do support it.
- CONVENTIONS.md: tighten the comments rule — comments only where a choice would look odd to a first-time reader with no context, and never as changelog.

## Test plan

- [x] `npm run test` — 1178 passing
- [x] New coverage: OSML names English as the fallback and rules no language out; a pinned project language wins in all three prompt hops (OSML, template mode, both story-mode hops); the OSML spec still asks the script to report its language, which Auto relies on for voices; a pinned project language outranks a voice language left by an earlier script, and the voice's own language still wins on Auto; a user-picked narrator language survives a generation that reports a different one; the Anthropic request omits `temperature`

## Notes

- **Composer settings don't survive a reload — language included.** Autosave only mounts once a script exists (`useEditorSession` is rendered by `PostPromptView`), so the pre-prompt composer persists nothing. Pre-existing and unchanged here — the aspect ratio pill resets the same way. Worth a separate fix if we want composer state to stick.
- Uses the icon library's existing `translate` glyph (文/A) — the only new line in `icon.tsx` is the export; `icons.css` is untouched.
- Language names come from `Intl.DisplayNames` rather than a hand-maintained 42-entry label map.
- The installed `@anthropic-ai/sdk` (0.78.0) does not type the `xhigh` effort level, so the enum stops at `max`. Bumping the SDK would widen it.

### DIFF
--- a/lib/connectors/tts/plugins/metadata-voice.ts
+++ b/lib/connectors/tts/plugins/metadata-voice.ts
@@ -1,5 +1,6 @@
 import { requireState } from "@/lib/connectors/plugins";
 import { forVoice } from "@/lib/generation/sourceNodes";
+import { declaredLanguage } from "@/lib/project/language";
 import { MetadataVoiceSchema } from "@/lib/project/types";
 import type {
 	ConnectorPlugin,
@@ -11,7 +12,7 @@ export function createMetadataVoicePlugin(): ConnectorPlugin<TTSGenerateParams> 
 		name: "metadata-voice",
 		dependencies: (element) => [forVoice(element.customAttributes?.name)],
 		beforeGenerate(params, ctx) {
-			const { narration, characters } = requireState(
+			const { narration, characters, language } = requireState(
 				ctx,
 				"metadata-voice",
 			).metadata;
@@ -21,6 +22,7 @@ export function createMetadataVoicePlugin(): ConnectorPlugin<TTSGenerateParams> 
 			return {
 				...params,
 				...fields,
+				language: declaredLanguage(language) ?? fields.language,
 				voiceId: fields.voiceId ?? resolvedVoiceId,
 			};
 		},
--- a/lib/project/language.ts
+++ b/lib/project/language.ts
@@ -1,0 +1,21 @@
+import { TTS_LANGUAGES, type TTSLanguage } from "@/lib/connectors/tts/enums";
+
+export const AUTO_LANGUAGE = "auto";
+
+export type LanguageChoice = typeof AUTO_LANGUAGE | TTSLanguage;
+
+export const LANGUAGE_CHOICES = [AUTO_LANGUAGE, ...TTS_LANGUAGES] as const;
+
+const displayNames = new Intl.DisplayNames(["en"], { type: "language" });
+
+export function languageLabel(choice: LanguageChoice): string {
+	return choice === AUTO_LANGUAGE
+		? "Auto"
+		: (displayNames.of(choice) ?? choice);
+}
+
+export function declaredLanguage(
+	choice: LanguageChoice | undefined,
+): TTSLanguage | undefined {
+	return choice === AUTO_LANGUAGE ? undefined : choice;
+}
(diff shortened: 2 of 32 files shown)


lib/project/language.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/** Voices are language-specific, so a new language drops one chosen for the old. */
+export function withLanguage(
+	narration: MetadataVoice,
+	choice: LanguageChoice,
+): MetadataVoice {
+	return {
+		...narration,
+		language: choice === AUTO_LANGUAGE ? undefined : choice,
+		voiceId: undefined,
```
umairnadeem: reselecting the language you're already on wipes the narrator's voice, `DropdownMenuItem.onSelect` fires even for the current value. maybe bail early when `choice` matches the stored language so a no-op reselect doesn't clobber `voiceId`/`resolvedVoiceId`.

lib/providers/llm/anthropic.ts
```diff
@@ -72,9 +73,12 @@ export class AnthropicLLM extends BaseProvider<
 			{ type: "text" as const, text: params.prompt },
 		];
 		return {
-			model: params.model || "claude-opus-4-8",
+			model: params.model || "claude-opus-5",
 			max_tokens: params.maxTokens || 65536,
-			temperature: params.temperature,
+			thinking: { type: "adaptive" as const },
+			output_config: {
```
umairnadeem, line 79: `temperature` is silently dropped now but still lives in `LLMGenerateParams` and the `/v1/llm` body schema, so callers get a knob that does nothing. if claude rejects it, purge it from the type + route schema too rather than leaving a dead param.

lib/connectors/tts/plugins/metadata-voice.ts
```diff
@@ -21,6 +22,7 @@ export function createMetadataVoicePlugin(): ConnectorPlugin<TTSGenerateParams>
 			return {
 				...params,
 				...fields,
+				language: fields.language ?? declaredLanguage(language),
```
umairnadeem: dropping the voice `language` attribute from the OSML prompt means `fields.language` is basically always undefined now, so on Auto this resolves to undefined and voice-search hits its `params.language || "en"` fallback. type a spanish prompt on Auto → spanish script narrated by an english voice, which used to work because the script declared `language="es"`. either keep the attribute in the prompt or derive the fallback from the generated script.

umairnadeem: stored `narration.language` beats the picker here, and it's write-once now, so the composer choice can silently lose. generate on Auto with an english topic -> script writes `narration.language: "en"` -> switch the picker to Spanish and regenerate -> spanish script, english voice, since `collectWritableMetadata` won't overwrite the stored `en`. maybe flip the precedence when the project language is explicitly declared?

PR #553, opened 2026-08-12
feat: read-only storyboard strip under the transport bar

umairnadeem:
## Summary
- Adds a read-only storyboard panel below the transport bar: every scene as a thumbnail in document order, labelled by scene number, driven by the seek bar's existing scene segments.
- Ungenerated scenes render the canvas `ForegroundPreview` placeholder instead of being dropped; the playing scene carries the canvas `scene-active` highlight, now shared as `ACTIVE_SCENE_CLASS`.
- Clicking a thumbnail seeks to that scene's start frame and scrolls the canvas to it (same as the scene pill); Delete/Backspace on a focused thumbnail removes the scene.
- A hairline seam between scenes parts on hover/focus to reveal an add button, inserting a new scene there via `insertScene`.
- Panel visibility toggles from a segmented control in the layout side panel and beside the volume/fullscreen controls; the bottom view is a named view so the timeline can join it later.

## Test plan
- [ ] Open a project: the storyboard appears under the transport bar with one thumbnail per scene, numbered.
- [ ] Play: the active thumbnail picks up the same highlight the canvas scene gets.
- [ ] Click a thumbnail: playhead jumps to that scene and the canvas scrolls to it.
- [ ] Focus a thumbnail and press Delete: the scene is removed; Cmd+Z restores it.
- [ ] Hover between two thumbnails: the seam parts and a `+` appears; clicking it inserts a new scene at that position.
- [ ] Toggle Hidden from either the side panel or the transport bar: the panel disappears and both controls stay in sync.
- [ ] A scene with nothing generated shows the placeholder thumbnail and still selects and deletes.

Part of #332


app/components/video/storyboard/StoryboardScene.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: {
+	item: StoryboardSceneData;
+	aspectRatio: string;
+	onSelect: () => void;
+	onDelete: () => void;
+}) {
+	const isActive = useActiveSceneId() === item.scene.id;
+
+	const handleKeyDown = (event: KeyboardEvent<HTMLButtonElement>) => {
+		if (event.key !== "Delete" && event.key !== "Backspace") return;
```
umairnadeem, line 24: tab to a thumbnail, hit Backspace, and the whole scene (every element in it) is gone with no confirm. afaict cmd+z can't get it back either: focus is on this button, not the contenteditable, so slate-history never sees the undo. either drop Backspace and keep Delete behind a confirm, or wire an undo that works from outside the editable.

app/components/video/storyboard/SceneInsertHandle.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * later.
+ */
+export function SceneInsertHandle({ onInsert }: { onInsert: () => void }) {
+	return (
+		<SimpleTooltip label="Add scene">
+			<button
+				type="button"
+				aria-label="Add scene"
+				onClick={onInsert}
+				className="group/insert relative z-10 flex h-16 w-2 shrink-0 items-center justify-center rounded-xs focus-ring"
```
umairnadeem: `w-2` makes the actual hit target 8px wide, and the thumbnails sit flush against it so the WCAG 2.5.8 spacing exception doesn't save it. perhaps keep the 8px bar visual but pad the button out (`px-1.5` + negative margin) so the target is >=24px.

app/components/video/storyboard/StoryboardScene.tsx
umairnadeem: lets re-use the confirm modal here

PR #554, opened 2026-08-12
fix(captions): unbold and center the Karaoke preset

umairnadeem:
The Karaoke preset inherited `bold: true` from `DEFAULT_CAPTION_STYLE` on both the base and active word, and defaulted to left alignment.

- Karaoke: `bold: false` on `base` and `activeWord`, `alignX: "center"`
- Boxed: `alignX: "center"`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #555, opened 2026-08-12
refactor: nightly maintainability 2026-08-12

umairnadeem:
Video settings had five knobs and five different ways to read one. Now they have one.

## The problem

`VideoSettingsSchema` declared every setting `optional()`, so a default could not live in the schema — each reader had to re-apply its own. That produced, per setting, a `DEFAULT_X` const, a `resolveX(metadata)` whose parameter type hand-restated a slice of `Metadata` (`{ videoSettings?: { aspectRatio?: AspectRatio } }`), and a `useX()` hook wrapping `useProject`. Four modules of that, in four files each.

And it had already drifted: `useTransitionType` skipped the resolver entirely and inlined `?? DEFAULT_TRANSITION` in the hook. Nothing failed, because nothing could — a fifth setting added without a resolver is indistinguishable from one with a resolver, until two call sites disagree about the default.

The store made it worse: `initialState.metadata` was a hand-written literal cast `as Metadata`, a second copy of the schema's defaults that the schema could not check.

## The fix

Defaults move to the parse boundary, per CONVENTIONS' "parse, don't validate". New `lib/video/videoSettings.ts` owns the contract — each knob with its default, composed from the constants the domain modules already export — and `MetadataSchema` embeds it. `videoSettings` is now **total** on parsed metadata, so a reader cannot forget a fallback and a new setting cannot be added without one.

That deletes the whole read-time-defaulting layer:

- `resolveAspectRatio`, `resolveVideoLength`, `resolveVideoLengthSpec`, `resolveCaptionsEnabled`, `resolveCaptionStyle` and the four structural `{ videoSettings?: … }` parameter types that shadowed `Metadata`
- `useAspectRatio`, `useVideoLength`, `useCaptionsEnabled`, `useTransitionType` — one generic `useVideoSetting("aspectRatio")` replaces all four, so the sixth setting needs no new file
- `lib/project/store.ts`'s `as Metadata` literal, now `MetadataSchema.parse({})`

`useCaptionStyle` keeps its name because it is a getter/setter pair, not a plain read. Callers read `state.metadata.videoSettings.aspectRatio` directly where they aren't in React (`sourceNodes`, `script-length`).

Net −77 lines across 31 files; 9 modules deleted, 2 added.

## Behavior

Unchanged. Every default is the same constant it was, and the one setting with lenient parsing keeps it — a `captionStyle` an older build wrote that no longer validates still falls back rather than failing the project open (`.catch(DEFAULT_CAPTION_STYLE)` in place of the old `.optional().catch(undefined)`). Autosave is edit-triggered, not diffed, so a completed settings block does not provoke a save; it just means newly-written rows spell out what was previously implied.

New `lib/video/__tests__/videoSettings.test.ts` pins the completion rule: absent and partial blocks fill in, stored values survive, an unknown enum throws rather than silently defaulting.

## Docs

`ARCHITECTURE.md`'s Editor-state section still documented `getProjectStore(projectId)` as the imperative escape hatch. That function was deleted in #533 — the handle comes from `ProjectStoreProvider` now. Rewritten against master, plus one line for where a metadata default lives.

## Open PR overlap check

Open PRs considered: **#553** (storyboard strip — `app/components/video/{BottomTransportBar,BottomView*,VideoLayoutContext,storyboard/*}`, `canvas/dnd/SortableScene`, `panel/LayoutPanel`, `ActiveSceneContext`, `PostPromptView`, `components/ui/icon`, `lib/canvas/insertScene`) and **#554** (`lib/video/captionPresets.ts`). This PR touches none of those files. Zero path overlap either way.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `npm run build`, and `npm test` (137 files, 1175 tests) all pass.


umairnadeem (commented):
lgtm

PR #556, opened 2026-08-12
perf: nightly performance optimizations 2026-08-12

umairnadeem:
## What

Collapsing a single scene re-rendered **every** element card on the canvas.

`ViewModeContext` held the collapsed set as its value, and `isCollapsed` was a `useCallback` keyed on that set (`ViewModeContext.tsx:32-35`). So every `toggle` produced a new context value, and **context updates bypass both `React.memo` and Slate's element memoization** — every `SortableContent` re-ran `ReactEditor.findPath` + `parentSceneId` and rebuilt its whole `ElementContainer` subtree (icon chip, character pills, model badge, settings `Popover`, delete/upload/animate/generate buttons, output preview), to change the rendering of the one scene that actually flipped.

This is item 5 of #510, and the same shape as the drag-transfer fanout fixed in #523.

- **`ViewModeContext`** now carries a store (`isCollapsed` / `hasCollapsed` / `subscribe` / `toggle` / `expandAll` / `collapseAll`) instead of the collapsed set. The context value is created once and never changes identity.
- **`useIsCollapsed(sceneId)`** is the consumer surface for the four components that render off collapse state (`SortableContent`, `SortableScene`, `SceneContainer`, `DragOverlayContent`). Each subscribes via `useSyncExternalStore` and re-renders only when its own scene's answer flips.
- **`useHasCollapsed()`** does the same for `LayoutPanel`'s expand/collapse-all toggle, which is the only reader of that flag.

Every call site keeps its exact previous condition, so collapsed cards, the collapsed scene frame, the disabled scene drag handle and the drag overlay all behave as before.

## Why it matters

A collapse toggle is a normal navigation action on a long slop, and today it costs a full-document rebuild. Consumer counts on a 40-scene / 200-element document:

| | before | after |
|---|---|---|
| context consumers re-rendered per toggle | 240 (200 cards + 40 scenes) | 0 |
| `useIsCollapsed` subscribers whose value changed | — | 1 scene + its ~5 cards |
| `ElementContainer` subtrees rebuilt per toggle | 200 | ~5 |

`expandAll` / `collapseAll` still change every scene's answer, and still re-render everything — that is the correct amount of work for those two buttons.

I did not re-measure the per-card rebuild cost in a browser for this PR. #523 measured exactly this fanout on the same fixture (200 cards, `ElementContainer` rebuilds, 435 ms longest task for a 200-card rebuild storm); this change removes the same 200-card rebuild from the collapse path.

## Open PR overlap check

Open PRs at the time of this run: **#553** (storyboard strip — `PostPromptView`, `BottomTransportBar`, `BottomViewContext/Toggle`, `VideoLayoutContext`, `storyboard/*`, `ActiveSceneContext`, `SortableScene`, `LayoutPanel`, `insertScene`), **#554** (`lib/video/captionPresets.ts`), **#555** (nightly maintainability — `useEditorSession`, caption panels, `PlayerPanel`, `useVideoLayout`, `lib/project/*`, `lib/video/*`).

No theme overlap. Two file overlaps with #553, both unavoidable because those files read the collapse state: `SortableScene.tsx` (#553 edits the import block and hoists `ACTIVE_SCENE_CLASS`; this edits the `useViewMode` line and the `disabled` prop) and `LayoutPanel.tsx` (#553 adds a `BottomViewToggle` field; this edits the `useViewMode` destructure). Both should merge cleanly or conflict trivially.

## Considered and skipped

- **Item 4 of #510** (lower the document subscription so keystrokes stop re-rendering the whole shell) is the larger remaining win, but it lands in `PostPromptView.tsx`, `useEditorSession.ts` and `VideoLayoutContext.tsx` — all held by #553/#555. Deferred on the overlap guard, not on merit.
- **Item 1 of #510** (Slate chunking via `getChunkSize`). At this app's scale the editor's top-level children are ~40 scenes, so the `if (!chunking)` branch's per-keystroke work is ~80 WeakMap writes and 40 memo comparators — microseconds. The paint half of the walkthrough's chunking advice is already covered: `.scene` carries `content-visibility: auto` with `contain-intrinsic-size`. Enabling chunking here would be the trivial-gain PR the issue warns about.
- **Waveform / video preview viewport gating** — already proposed and closed as #476.
- **`findElementById` in the drag-over path** — already proposed and closed as #548.

## Validation

`npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run knip`, `npm run build`, `npm test -- --run` (1184 passing, including a new pure test for the store).

### DIFF
--- a/app/components/canvas/ViewModeContext.tsx
+++ b/app/components/canvas/ViewModeContext.tsx
@@ -1,22 +1,91 @@
 "use client";
 
-import { useCallback, useMemo, useState, type ReactNode } from "react";
+import {
+	useCallback,
+	useState,
+	useSyncExternalStore,
+	type ReactNode,
+} from "react";
 import type { Editor } from "slate";
 import type { CanvasElement } from "@/lib/canvas/types";
 import { isSceneElement } from "@/lib/canvas/scenes";
 import { createRequiredContext } from "@/lib/components/createRequiredContext";
 
-type ViewModeValue = {
+export type ViewModeStore = {
 	isCollapsed: (sceneId: string) => boolean;
-	hasCollapsed: boolean;
+	hasCollapsed: () => boolean;
+	subscribe: (onChange: () => void) => () => void;
 	toggle: (sceneId: string) => void;
 	expandAll: () => void;
 	collapseAll: () => void;
 };
 
+/**
+ * Every element card asks whether its scene is collapsed, but a toggle changes
+ * the answer for one scene. Holding the collapsed set in the context would
+ * re-render all of them; holding a store lets each card subscribe and re-render
+ * only when its own scene flips.
+ */
+export function createViewModeStore(editor: Editor): ViewModeStore {
+	let collapsed: ReadonlySet<string> = new Set();
+	const listeners = new Set<() => void>();
+
+	const replace = (next: ReadonlySet<string>) => {
+		collapsed = next;
+		for (const listener of listeners) listener();
+	};
+
+	return {
+		isCollapsed: (sceneId) => collapsed.has(sceneId),
+		hasCollapsed: () => collapsed.size > 0,
+		subscribe: (onChange) => {
+			listeners.add(onChange);
+			return () => {
+				listeners.delete(onChange);
+			};
+		},
+		toggle: (sceneId) => {
+			const next = new Set(collapsed);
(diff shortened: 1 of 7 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
useless

PR #565, opened 2026-08-13
feat: timeline view under the transport bar

### LINKED ISSUE #332: In-app storyboard/timeline (read-only) showing what plays when and for how long

umairnadeem:
## Summary
- Adds a read-only timeline view under the transport bar and makes the panel toggle three-way: Timeline view / Storyboard view / Hide bottom panel, with timeline as the default. Closes #332.
- Lanes are a projection of `buildVideoLayout`'s own roles (Video / Voice / Effects / Music); genuinely simultaneous clips split onto their own lane, with the transition duration as the tolerance so audio across a scene boundary stays put. Clips show their prompt, media-type colour, and either a thumbnail or their waveform; selecting one scrolls the canvas to its scene.
- One scroll container as a grid: the ruler sticks to the top, the lane gutter to the left. Zoom is fit-to-width x1-x8, the ruler doubles as a scrub surface, and a handle on the dock's top edge resizes the panel.
- Collapses three separate ideas of what the video contains. A scene made only of overlays was a series entry with a `null` element, so every consumer carried a null branch and the timeline had nothing to draw; it is now a real element holding a black pixel that the player and the Lambda render resolve identically. Scene identity (id, number, prompt) travels on `ResolvedElement` instead of being rejoined afterwards by start time, and the layout key accounts for the scene an element sits in. The seek bar's spans derive from the same `series` the timeline draws, which is what made them vanish on a video that opens on sound effects. Clips running past the last scene are bounded by the timeline when it draws them, not by the builder.
- No new dependencies. The shared peaks/soundwave helpers move from rounded bars to a mirrored envelope and from peak-picking to RMS, which also restyles the element-card players.
- Falls out of the above: `usePlayerScrub`, the pause/silence/resume dance the seek bar had inlined, now shared, so a ruler drag no longer leaks audio from the old position.

Drag, trim and delete are not implemented; the panel is read-only.

## Test plan
- [ ] Toggle between Timeline / Storyboard / Hide; timeline is the default view.
- [ ] Lanes match the composition: scenes on Video, narration and character on one Voice lane across scene transitions, overlapping sounds on separate Effects lanes.
- [ ] Start a video with a repeated sound effect and no image: the timeline draws a blank scene, the effect clips stop at the ruler's end, and the seek bar is present.
- [ ] Click and drag the ruler to scrub; audio should not double up while playing.
- [ ] Zoom in: clips widen, ticks get finer, the ruler stays pinned while the lanes scroll, and the playhead follows during playback.
- [ ] Hover a clip for the full prompt; hover anywhere for the ghost playhead and its timecode.
- [ ] Click a clip: it highlights and the canvas scrolls to that scene. Space toggles playback with the panel focused.
- [ ] Drag the handle above the transport bar; the ruler keeps its height and the lanes take the space.
- [ ] Check light and dark: clip body, nested preview, and selection are all shades of the element's own media colour.


app/components/video/timeline/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
...
+			tabIndex={0}
+			onKeyDown={(event) => {
+				if (event.key !== " ") return;
+				event.preventDefault();
+				player?.toggle();
+			}}
+			className="flex h-full flex-col overflow-hidden focus-ring"
+		>
+			{rows.length === 0 ? (
+				<TimelineEmpty loading={scenes.length > 0} />
```
umairnadeem: scenes exist on the canvas long before anything is generated, and `resolveElements` only yields resolved elements, so a freshly written script gives `rows.length === 0` with `scenes.length > 0` and this shimmers "Loading the timeline" forever. timeline is the default view now so that's the first thing a new project shows. keying off the queue/`ready` instead of `scenes.length` would fix it.

app/components/video/timeline/Timeline.tsx
```diff
@@ -0,0 +1,297 @@
...
+		}
+		seek(clip.start);
+	};
+
+	return (
+		<section
+			aria-label="Timeline"
+			tabIndex={0}
+			onKeyDown={(event) => {
+				if (event.key !== " ") return;
```
umairnadeem: this preventDefaults space for the whole subtree, so it also swallows activation on the clip buttons and the zoom buttons inside (a button's click fires on space keyup only if keydown wasn't cancelled). so keyboard users can't select a clip with space, it just toggles playback. maybe gate on `event.target === event.currentTarget`?

components/ui/icon-button.tsx
```diff
@@ -3,20 +3,27 @@ import * as React from "react";
 import { SimpleTooltip } from "@/components/ui/tooltip";
 import { cn } from "@/lib/utils";
 
+const SIZES = { sm: "h-5 w-5", default: "h-7 w-7" } as const;
```
umairnadeem, line 6: nothing passes `size` anywhere, and the zoom buttons this PR adds reach for `className="h-5 w-5"` instead of `size="sm"`, so we ship a variant API plus a literal that says the same thing. either use it in Timeline.tsx or drop the prop?

app/components/video/timeline/ClipWaveform.tsx
```diff
@@ -0,0 +1,72 @@
...
+function usePeaks(src: string): number[] | null {
+	const [peaks, setPeaks] = useState<number[] | null>(null);
+
+	useEffect(() => {
+		let cancelled = false;
+		loadPeaks(src)
+			.then((loaded) => {
+				if (!cancelled) setPeaks(loaded);
+			})
+			.catch((error) => console.error("Failed to decode audio:", error));
```
umairnadeem: a failed fetch/decode leaves `peaks` null forever, so the clip shimmers for good. `Waveform.tsx` settles its loading state in a `.finally` for exactly this. maybe do the same here and fall back to a flat band?

app/components/video/timeline/Timeline.tsx
```diff
@@ -0,0 +1,303 @@
...
+						<li
+							key={clip.key}
+							className="absolute inset-y-0.5"
+							style={{ left: clip.start * pxPerSec, width }}
+						>
+							<button
+								type="button"
+								aria-current={selected ? "true" : undefined}
+								onClick={() => onSelect(clip)}
+								// `block`: a button is inline by default, so it would sit on
```
umairnadeem: `CONVENTIONS.md` § Structure and reuse: *"Comments are a last resort, not a habit. Write one only where a choice would look odd or arbitrary to a first-time reader with no context."* this one, the divider note on L54 and the scroller note on L192 are narrating tailwind rather than justifying an odd choice. worth thinning across the new timeline files.

umairnadeem (commented):

lib/video/scene-builder.ts
```diff
@@ -115,24 +116,21 @@ export function buildVideoLayout(
...
-					}
-					series.push(createSequence(element, cursor, element.durationSec));
+				cursor = foregroundCursor;
+				if (series.length > 0) {
+					// TransitionSeries.Sequence overlap the previous
+					// ones by transitionDurationSec. Bake that into the cursor so every
+					// subsequent overlay/background/effect lands on the rendered timeline.
+					cursor -= transitionDurationSec;
 				}
+				series.push(createSequence(element, cursor, element.durationSec));
```
umairnadeem, line 126: dropping the fill-the-placeholder branch means a foreground after a leading overlay now opens a *second* scene instead of filling the blank one. `[narration, image]` in one scene is legal (withScenes only splits on a second foreground) and it's what a blank project gives you: `BLANK_SCRIPT` seeds a narration, user inserts an image after it, and the video is now black for the narration then the image, with a cross-fade in between. was that intended? `if (current && isBlankScene(current.element))` would keep the old behaviour.

app/components/video/timeline/TimelineClip.tsx
```diff
@@ -0,0 +1,133 @@
...
+	clip: TimelineClipData;
+	width: number;
+	label: string;
+	selected: boolean;
+	/** Set on foreground clips, which are the scenes of the video. */
+	sceneNumber?: number;
+}) {
+	const { element } = clip;
+
+	if (!element) {
```
umairnadeem: `Sequence.element` is non-nullable now, so this branch is dead on arrival. leftover from the null-element migration afaict, worth purging with the "Empty scene" JSX so nobody reads it as "clips can still be elementless".

lib/video/scene-builder.ts
umairnadeem: yes intentional

PR #568, opened 2026-08-13
perf: nightly performance optimizations 2026-08-13

umairnadeem:
## What

Typing anywhere in the canvas re-rendered the **entire page shell**, on documents of any size.

`useEditorSetup` held the document in React state (`useEditorSetup.ts:28`) and `useEditorSession` is called from `PostPromptViewInner` (`PostPromptView.tsx:20`). Slate 0.123 applies operations immutably (`root.children = replaceChildren(...)` in `slate/dist/index.js:2195`), so `editor.children` gets a fresh identity on every keystroke, `setValue` always commits, and React re-renders `EditorToolbar`, `EditorSidebar`, `ProjectTitle`, the player panels, `BottomTransportBar` and the storyboard strip — one thumbnail per scene — none of which read the document text. This is item 4 of #510, and the second (escalated) report in that issue is exactly its shape: the hang doesn't need a big document, because the cost is the shell, not the doc walk.

`value` had three consumers, and none of them needed to sit above `<Slate>`:

- **Video layout** — `useVideoLayout` now selects the layout key itself with `useSlateSelector`, so `VideoLayoutContext` (player, storyboard, scene segments) updates when the rendered video would change, not when a character is typed. The key derivation gets one home instead of being computed in the shell and threaded through `CanvasProviders` as a prop.
- **Drag and drop** — `useDragAndDrop` selects the scene-id key, so `Canvas`, `DndContext` and `SortableContext` re-render on scene reorder rather than per keystroke. The old code already keyed on ids for this reason; it just did it downstream of a prop that changed anyway.
- **Autosave** — rides `<Slate onValueChange>`, which needs no render at all. Its script source is fixed at construction now that it no longer varies per keystroke, so `setScriptSource` and the "no script source" failure mode go away (the error is defined out of existence rather than reported).

`<Slate>` moves from `Canvas` up into `CanvasProviders` so those consumers can subscribe. This is the walkthrough's own guidance (https://docs.slatejs.org/walkthroughs/09-performance): subscribe to the slice you need, don't hand the changing value to everyone.

Net effect: a keystroke re-renders the Slate subtree and nothing above it. `PostPromptViewInner` re-renders only when the player position or bottom view changes.

## Not done

- **Chunking (`getChunkSize` / `renderChunk`, item 1 of #510).** Read the walkthrough and the `slate-react` implementation first: `groupIntoChunks` picks a depth from the child count (`slate-react/dist/index.js:2056-2066`), so at the recommended size of 1000 a document with a few hundred scenes gets depth 0 — a flat chunk tree and no memoization win. Getting a win at our sizes needs a small chunk size, which nests chunk `<div>`s into the editor and breaks the `first:border-t-0` separator rule on `SortableScene`. Not worth shipping blind; it wants the Performance-panel trace #510 asks for.
- **`content-visibility: auto` on cards (item 2).** Already in place at the scene level (`styles/sortable.module.css`).
- **React render-count regression test.** The repo has no component-render test setup (no `@testing-library/react`, no jsdom render tests), and adding one is its own PR.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, `test` (1181 passed), `test:e2e` (3 passed).

## Open PR overlap check

Considered #567 (captions/properties panels, copilot, `lib/video/useCaptionStyle`, `useVideoSetting`) and #565 (timeline view, storyboard, waveform, `SortableContent`/`ElementContainer`). No file or theme overlap except `PostPromptView.tsx`, which both touch in different regions: #565 edits the bottom-dock JSX, this PR edits the editor-session destructure and the `CanvasProviders`/`Canvas` props. The state lived there, so it can't be avoided.

Closes part of #510.


umairnadeem (commented):
lgtm

umairnadeem (commented):
lgtm, merge looks clean too

Issue #562, opened 2026-08-13
Caption presets: horizontally center the Classic preset


umairnadeem:
In `lib/video/captionPresets.ts`, the `classic` preset is declared with an empty override:

```ts
preset("classic", "Classic", {}),
```

so it inherits `alignX: "left"` from `DEFAULT_CAPTION_STYLE`. Every other preset sets `alignX: "center"`, which makes Classic the odd one out — left-hugging captions rather than centered.

**Ask:** give Classic `alignX: "center"`.

```ts
preset("classic", "Classic", { alignX: "center" }),
```

One-line change; leave `DEFAULT_CAPTION_STYLE` itself alone so the default for non-preset styles is unchanged.

Issue #561, opened 2026-08-13
Base the active-scene highlight color on --generate instead of --accent


umairnadeem:
The active-scene highlight (`ACTIVE_SCENE_CLASS` → `.scene-active` in `app/globals.css`) derives all of its color from `--accent`:

```css
--scene-active-tint:   color-mix(in oklab, var(--accent) 6%, transparent);
--scene-active-blob:   color-mix(in oklab, var(--accent) 70%, white);
--scene-active-border: color-mix(in oklab, var(--accent) 30%, transparent);
--scene-active-glow:   color-mix(in oklab, var(--accent) 20%, transparent);
```

Against the panel background that reads as low-contrast, so the active scene does not stand out much.

**Ask:** switch these to the `--generate` token color so the active scene is more visible and prominent — i.e. swap `var(--accent)` → `var(--generate)` in the `--scene-active-*` definitions.

Notes:
- These tokens are defined twice (light block ~L110 and dark block ~L239); update both, and keep the existing per-theme mix percentages.
- `--scene-active-shadow` is not accent-derived and should stay as-is.

Issue #560, opened 2026-08-13
EditorSidebar: use pr-1 instead of pr-2 to match the nav element above


umairnadeem:
In `app/components/canvas/panel/EditorSidebar.tsx`, the scrolling panel container uses `pr-2`:

```
flex w-60 shrink-0 flex-col gap-3 overflow-y-auto [scrollbar-gutter:stable] pr-2 pt-4 pb-3 text-panel-fg
```

The nav element preceding it uses a right padding of `1`, so the panel content sits slightly further from the edge and the two do not line up.

**Ask:** change `pr-2` → `pr-1` on that div so the right edge matches the nav.

One-line class change.

Issue #558, opened 2026-08-13
BottomTransportBar: grey out play/back/forward controls when not ready


umairnadeem:
In `BottomTransportBar`, the play / back / forward controls stay at full opacity even when the player is not ready, so they look interactive when they are not.

**Ask:** when not ready, render those controls in a disabled/greyed-out state (reduced opacity, `disabled`, no hover/pointer cursor) and restore normal styling once ready.

Small, localized styling + disabled-state change.

PR #569, opened 2026-08-13
feat: segmented Player position toggle in LayoutPanel

### LINKED ISSUE #559: LayoutPanel: replace Player position dropdown with a 3-way segmented toggle
The "Player position" control in `app/components/canvas/panel/LayoutPanel.tsx` is a `SelectField` dropdown with three options (Top / Right / Hidden). Three fixed options is a natural fit for a segmented control rather than a dropdown.

**Ask:** swap the `SelectField` for a `MediaToggle` segmented control — the same component already used a few lines below for the Scenes expand/collapse toggle — with an idiomatic icon per option pulled from `@/components/ui/icon`.

Notes:
- Keep the existing `onPositionChange` semantics (`hidden` → `setVisible(false)`; otherwise `setPosition` + `setVisible(true)`).
- Preserve the `disabled` state on **Right** when `narrowViewport` is true.
- Keep the `Player position` aria label on the group.

umairnadeem:
## Summary

The **Player position** control in `LayoutPanel` was a `SelectField` dropdown with three fixed options (Top / Right / Hidden). Three fixed options is a natural segmented control, and `MediaToggle` is already used twice a few lines below in the same panel (Scenes, Follow).

This swaps it for `MediaToggle` with an icon per option: `PanelTop` (Top), `PanelRight` (Right), `EyeOff` (Hidden).

Two small additions to `MediaToggle` were needed to preserve existing behaviour, both optional and non-breaking for current callers:

- `disabled?: boolean` per option — passed to the underlying Radix item, dimmed with `cursor-not-allowed opacity-40`. Keeps **Right** disabled when `narrowViewport` is true.
- `ariaLabel?: string` on the group — `PanelField`'s label is a plain `<span>` and isn't programmatically associated, so the group keeps its own `Player position` accessible name.

`onPositionChange` semantics are unchanged (`hidden` → `setVisible(false)`; otherwise `setPosition` + `setVisible(true)`).

#559 — `size: XS`, `good first issue`, with an explicit implementation note in the body naming the target component, the icons source, and the three behaviours to preserve. The code area is one panel component plus one design-system primitive; no domain logic, no data flow, no auth surface.

## Validation

Full CI-equivalent suite from `.github/workflows/ci.yml`, all green:

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass (`SelectField` still has two other callers, so no dead export)
- `npm run build` — pass
- `npm run test` — 139 files, 1182 tests passed

No test was added: the change is a presentational swap with no new pure logic, and the repo's tests cover `lib/` seams rather than component markup.

## Open PR overlap check

Open PRs at branch time: #565 (timeline view), #567 (nightly maintainability), #568 (nightly perf). Checked each with `gh pr diff --name-only`. Neither file touched here — `app/components/canvas/panel/LayoutPanel.tsx` and `components/ui/media-toggle.tsx` — appears in any of them. #567 touches `CaptionsPanel.tsx` and #565 touches `BottomViewToggle.tsx`, both of which *consume* `MediaToggle`, but only additive optional props are introduced here so those call sites are unaffected.

## Candidates skipped

- #562 (center Classic caption preset) — suitable, but a strictly smaller one-line change; picked the more substantive XS instead. Still open for a future run.
- #561 (active-scene highlight `--accent` → `--generate`) — edits `app/globals.css`, which PR #565 also touches. Overlap.
- #558 (grey out transport controls when not ready) — edits `BottomTransportBar.tsx`, owned by PR #565. Overlap.
- #560 (`EditorSidebar` `pr-2` → `pr-1`) — pure padding churn.
- Everything remaining is `size: M` or larger, or product/design-direction work (#536, #535, #462, #451, #379, …), out of scope for an unattended run.

Closes #559


umairnadeem (commented):
lgtm

PR #571, opened 2026-08-13
fix(editor-sidebar): use pr-1 so the panel lines up with the nav rail

### LINKED ISSUE #560: EditorSidebar: use pr-1 instead of pr-2 to match the nav element above

dchaudhari7177:
Closes #560.

The scrolling panel container in `app/components/canvas/panel/EditorSidebar.tsx` used `pr-2` while the nav rail beside it pads by `1`, so the panel's right edge sat slightly further in and the two did not line up.

`pr-2` → `pr-1` on that div. One-line class change, exactly as the issue specifies.

Verified: `npx tsc --noEmit` clean; lint-staged eslint + prettier clean on commit.

### DIFF
--- a/app/components/canvas/panel/EditorSidebar.tsx
+++ b/app/components/canvas/panel/EditorSidebar.tsx
@@ -118,7 +118,7 @@ function EditorSidebarComponent() {
 			</nav>
 
 			{current && HeaderIcon && Panel && (
-				<div className="flex w-60 shrink-0 flex-col gap-3 overflow-y-auto [scrollbar-gutter:stable] pr-2 pt-4 pb-3 text-panel-fg">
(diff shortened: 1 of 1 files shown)


umairnadeem (commented):
lgtm

umairnadeem (approved):
@dchaudhari7177 thank you!

PR #572, opened 2026-08-13
fix(canvas): base the active-scene highlight on --generate

### LINKED ISSUE #561: Base the active-scene highlight color on --generate instead of --accent

dchaudhari7177:
Closes #561.

The `.scene-active` tint, blob, border, and glow all derived from `--accent`, which reads low-contrast against the panel background, so the active scene did not stand out.

Swapped `var(--accent)` → `var(--generate)` in the four `--scene-active-*` definitions, in **both** the light block (~L110) and the dark block (~L239), keeping each block's existing mix percentages:

| token | light | dark |
|---|---|---|
| `--scene-active-tint` | 6% | 12% |
| `--scene-active-blob` | 70% | 78% |
| `--scene-active-border` | 30% | 42% |
| `--scene-active-glow` | 20% | 30% |

`--scene-active-shadow` is not accent-derived and is left alone, as the issue notes.

## Bonus: this also fixes a DESIGN.md violation

`DESIGN.md:19` reserves accent for "focus rings, selection, links, send, and `accent` CTAs only. **Disciplined, never an ambient wash.**" — and the active-scene tint/blob/glow is exactly such an ambient wash. So this change moves the highlight onto the right token by the design system's own rule, not just for contrast.

Verified: no `--scene-active-*` token still references `--accent`; `npx tsc --noEmit` clean; lint-staged clean on commit. Colour choice is the maintainer's call from the issue — happy to tune the mix percentages if `--generate` at these strengths reads too strong.


umairnadeem (commented):
lgtm

umairnadeem (approved):

PR #573, opened 2026-08-13
fix(transport): disable play/back/forward until the player is ready

### LINKED ISSUE #558: BottomTransportBar: grey out play/back/forward controls when not ready

dchaudhari7177:
Closes #558.

The play, previous-scene, and next-scene controls stayed at full opacity before the player was ready, so they looked interactive while their handlers were already no-ops — `seekToAdjacentScene` returns early without a player or segments, and `player?.toggle()` short-circuits.

`BottomTransportBar` already computes exactly the condition needed:

```ts
const ready = player !== null && segments.length > 0;
```

and `IconButton` already carries `disabled:pointer-events-none disabled:opacity-40` in its base classes. So forwarding `disabled={!ready}` to the three controls gives everything the issue asks for — reduced opacity, a real `disabled` attribute, no hover state, no pointer cursor — with **no new styling** and no new state.

It also takes the dead controls out of the tab order, which the visual-only fix would not have.

Verified: `npx tsc --noEmit` clean; lint-staged eslint + prettier clean on commit.


app/components/video/BottomTransportBar.tsx
```diff
@@ -63,12 +63,14 @@ function BottomTransportBarComponent() {
...
 					<TooltipIconButton
 						label="Previous scene"
+						disabled={!ready}
 						onClick={() => seekToAdjacentScene(-1)}
 					>
 						<ChevronsLeft className="h-4 w-4" />
 					</TooltipIconButton>
 					<TooltipIconButton
 						label={playing ? "Pause" : "Play"}
+						disabled={!ready}
```
umairnadeem: this kills the play button's escape hatch: hide the player from the layout panel -> `visible` false -> `PlayerPanel` unmounts -> `registerPlayer(null)` -> `ready` false, so play is greyed out forever and the `showPlayer()` call below is dead. gating play on `segments.length > 0` only (and keeping `ready` for prev/next) keeps the reveal working.

umairnadeem (changes requested):

dchaudhari7177:
You're right, and thanks — I'd reasoned about `ready` as "can we drive the player" and missed that play is also the way back *to* it. Fixed in e24daf2.

Play now gates on segments alone; prev/next keep the full `ready` gate:

```tsx
const hasSegments = segments.length > 0;
const ready = player !== null && hasSegments;
```

That's the split you described: seeking needs a live player to seek with, showing one doesn't. With the panel unmounted the click is just `showPlayer()`, and `player?.toggle()` no-ops until it's back — so the reveal path works exactly once, which is all it needs to.

Left a comment at the `ready` definition explaining why play is deliberately not covered by it, since the asymmetry looks like an oversight otherwise and is an easy thing to "tidy up" back into the bug.

`tsc --noEmit` reports nothing on this file. I can't get a fully clean typecheck on the repo locally — a set of `next/*` modules resolve as missing in my sandbox — so the Vercel build is the real check here; it's still waiting on team authorisation for the deploy.

app/components/video/BottomTransportBar.tsx
```diff
@@ -26,7 +26,13 @@ function BottomTransportBarComponent() {
 	const { layout, segments } = useLayout();
 	const playing = usePlayerPlaying(player);
 
-	const ready = player !== null && segments.length > 0;
+	// Prev/next need a live player to seek with. Play does not: hiding the
+	// player from the layout panel unmounts PlayerPanel, which calls
+	// registerPlayer(null), and play doubles as the way back -- its
+	// showPlayer() re-reveals the panel. Gating it on `player` too would
+	// disable the only escape hatch and leave the button dead forever.
```
umairnadeem: 5 lines of comment for a 2 line change, and the last sentence is basically the review thread. CONVENTIONS.md says *"Comments are a last resort, not a habit... Never comment history: what broke, what the code used to be, why an earlier attempt failed. That belongs in the commit message."* one line on why play only needs segments is plenty, the rest can live in the commit msg.

umairnadeem (approved):

PR #574, opened 2026-08-13
fix(osml): require videoPrompt to continue from the still it animates

### LINKED ISSUE #566: OSML prompt: state that videoPrompt must continue from the still described in the tag body
## Problem

The OSML prompt never tells the model that an `animated_image`'s `videoPrompt` is fed to an image-to-video model **whose first frame is the still described in the tag body**. The guidance describes them as two independent things — "the tag body describes the still frame; the videoPrompt attribute describes the camera/subject motion" (`lib/connectors/llm/plugins/osml.ts:81-83`) — with no statement that they must describe the same moment.

Nothing in the instructions forbids a `videoPrompt` that opens on a different scene than the still. When that happens the video model gets a first frame incongruent with its prompt, and the clip either fights the frame or drifts off it. The mismatch is invisible in the OSML, so it only shows up in the rendered output.

The plugin wiring confirms the coupling the prompt omits: `still-frame.ts` generates the still from the tag body and passes it as `frameImages: [imageUrl]` alongside `prompt: videoPrompt`, so both land in one call to the video model.

- "the model doesn't know that the video prompt for animated images should start with an image scene described in the image prompt. if the model is made aware of this, the video prompt can't be wonky (e.g. describe a scene that starts with something else despite sending the video model a first frame thats incongruent with the video prompt)"

## Relevant code

- `lib/connectors/llm/plugins/osml.ts:81-88` — the `<animated_image>` section of the OSML system prompt; where the rule belongs
- `lib/connectors/animated_image/plugins/still-frame.ts` — `beforeGenerate` sends `prompt: videoPrompt` with `frameImages: [still]`, i.e. the still IS the first frame
- `lib/providers/llm/mock.ts` — mock OSML sample; update if the example changes

## Suggested direction

Add an explicit rule to the `videoPrompt` bullet: the clip **begins on the still described in the tag body**, so the `videoPrompt` must continue from that exact frame — same subjects, same framing, same time of day — describing only what changes over the next few seconds. It must never introduce a different opening scene, new subjects, or a cut. Reinforce it in the worked example.

## Acceptance criteria

- [ ] The OSML prompt states that the still described in the tag body is the clip's first frame and that `videoPrompt` continues from it.
- [ ] The prompt explicitly rules out a `videoPrompt` that opens on a different scene, introduces subjects absent from the still, or implies a cut.
- [ ] The in-prompt example demonstrates a `videoPrompt` congruent with its tag body.

## Related

- #450 (OSML prompt should deter the LLM from setting motion on animated_image)

---

dchaudhari7177:
Closes #566.

## The gap

The OSML prompt described the tag body and `videoPrompt` as two independent things — "the tag body describes the still frame; the videoPrompt attribute describes the camera/subject motion" — and never said they describe the same moment.

They are not independent. Confirmed in `lib/connectors/animated_image/plugins/still-frame.ts:91`:

```ts
return { ...rest, prompt: videoPrompt, frameImages: [imageUrl] };
```

The still generated from the tag body IS the clip's first frame. With nothing forbidding it, the model can write a `videoPrompt` that opens elsewhere, the video model gets a first frame incongruent with its prompt, and the clip either fights the frame or drifts off it — invisible in the OSML, visible only in the render.

## Change

1. **State the coupling** where the tag is introduced: the still is generated first and handed to the video model as the clip's FIRST FRAME, so the clip literally opens on that still.
2. **Two new `videoPrompt` rules** — it must *continue from* that exact frame (same subjects, framing, location, time of day, weather), describing only what changes over the next few seconds; and it must *never* open on a different scene, introduce a subject absent from the still, relocate, jump time of day, or imply a cut. If the motion needs a different opening image, the tag body is what should change.
3. **Both in-prompt examples now demonstrate the rule** rather than merely not violating it. The worked example continues on Red and Wolf (`"slow zoom out from Red and Wolf as they keep walking…"`), and the `characters` example's `videoPrompt="..."` placeholder is filled in with a continuation of the basket hand-off.

## On the mock samples

The issue flagged `lib/providers/llm/mock.ts` as "update if the example changes". I checked all five `animated_image` samples there against the new rule — every one is already congruent with its body (pans and push-ins that stay within the described frame), so they are **unchanged**.

## Acceptance criteria

- [x] Prompt states the still is the clip's first frame and `videoPrompt` continues from it
- [x] Prompt explicitly rules out a different opening scene, absent subjects, and cuts
- [x] In-prompt example demonstrates a congruent `videoPrompt` (both examples do)

Verified: `vitest run` — **139 files, 1182 tests passed**; `npx tsc --noEmit` clean; lint-staged clean on commit.

Note this is a prompt-behaviour change, so the real proof is in generated output rather than the suite — worth a sanity generation on a story with a hero shot before merging.


umairnadeem (commented):
lgtm

umairnadeem (approved):

PR #575, opened 2026-08-14
fix: nightly bug fixes 2026-08-14

umairnadeem:
## What

Stopping a script generation didn't actually stop it. Two defects in the SSE transport, one on each end of the stream.

### Server: `createSSEStreamResponse` never learned the client left

It drained the provider's async generator inside `start()` — no backpressure, no `cancel` handler. When a client hung up, the loop kept pulling tokens from the LLM. It only stopped by crashing on the next `controller.enqueue()`:

```
TypeError [ERR_INVALID_STATE]: Invalid state: Controller is already closed
  at Object.start (lib/api/sse.ts:54:17)
msg: "LLM stream error"
```

So every Stop logged a fabricated error at `error` level, burying real ones, and burned at least one extra chunk of generation.

Rewritten as `pull` + `cancel`: one chunk per read, and cancelling closes the source generator deterministically (which propagates through `AnthropicLLM.stream` to abort the upstream request).

### Client: `readSSE` released the reader instead of cancelling it

`releaseLock()` detaches the reader but leaves the stream un-cancelled, so a consumer that breaks early (`drainStream` on abort → `useStreamRun.stop`) left the response body unread and the connection open. The server had no way to find out. Now it cancels first, then releases.

The two are the same root cause — the stop path wasn't plumbed through the transport — and both halves are needed for Stop to reach the provider.

## Tests

3 regression tests, each verified to fail against the old code:

- `createSSEStreamResponse` closes the source on consumer cancel and logs nothing
- `readSSE` cancels the body when the consumer breaks early
- `OpenSlopLLMGateway.stream` cancels the response body when the caller stops consuming (covers the full generator-delegation chain)

Plus baseline coverage for `createSSEStreamResponse`, which had none: headers, happy-path framing, and the error path.

CI: lint, format:check, typecheck, knip, build, and 1187 tests all pass.

## Open PR overlap check

Considered #573, #569, #568, #567, #565. None touch `lib/api/sse.ts`, `lib/api/__tests__/`, or `lib/gateway/__tests__/` — those PRs are scoped to the transport bar, layout panel, canvas/video perf, copilot/captions panels, and the timeline view. No overlap.

### DIFF
--- a/lib/api/sse.ts
+++ b/lib/api/sse.ts
@@ -42,22 +42,34 @@ export function createSSEResponse(
 	return new Response(stream.readable, { headers: SSE_HEADERS });
 }
 
+/**
+ * Pulls one chunk per read so a client that stops reading stops the source, and
+ * closes the source on cancel. Draining `iter` eagerly instead would keep the
+ * provider generating for a client that has already hung up, and would only
+ * notice on the enqueue that throws.
+ */
 export function createSSEStreamResponse<T>(
 	iter: AsyncIterable<T>,
 	label: string,
 ): Response {
 	const encoder = new TextEncoder();
+	const source = iter[Symbol.asyncIterator]();
 	const stream = new ReadableStream({
-		async start(controller) {
+		async pull(controller) {
 			try {
-				for await (const chunk of iter) {
-					controller.enqueue(encoder.encode(formatSSE(chunk)));
+				const { done, value } = await source.next();
+				if (done) {
+					controller.close();
+					return;
 				}
-				controller.close();
+				controller.enqueue(encoder.encode(formatSSE(value)));
 			} catch (error) {
 				logger.error(error, `${label} stream error`);
 				controller.error(error);
 			}
+		},
+		async cancel(reason) {
+			await source.return?.(reason);
 		},
 	});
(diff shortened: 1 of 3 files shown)


lib/api/sse.ts
```diff
@@ -42,23 +42,35 @@ export function createSSEResponse(
...
 			try {
-				for await (const chunk of iter) {
-					controller.enqueue(encoder.encode(formatSSE(chunk)));
+				const { done, value } = await source.next();
+				if (done) {
+					controller.close();
+					return;
 				}
-				controller.close();
+				controller.enqueue(encoder.encode(formatSSE(value)));
```
umairnadeem, line 65: the fabricated `Invalid state: Controller is already closed` still fires on the actual Stop path afaict: cancel arrives while `source.next()` is in flight (we're mid-await on the LLM), the pending next resolves *after* the stream is closed, and this enqueue throws into the catch → `logger.error`. repro'd against this exact stream code with a slow generator, the existing test only passes because its source yields synchronously. maybe set a `cancelled` flag in `cancel()` and bail before the enqueue?

umairnadeem:
not worth it

PR #576, opened 2026-08-14
refactor: nightly maintainability 2026-08-14

umairnadeem:
## What

One home for the "checkable picker row", and a missing required env var in the README.

**Design-system: `SelectMenuItem`.** The check-gutter menu row lived in three places — `components/ui/select-menu.tsx` (the design-system home), `character/fields.tsx`'s `EnumOption`, and `CharactersPicker.tsx`'s `ProjectCharactersMenu`. Rule of three is met, so the row is now an exported `SelectMenuItem` primitive that `SelectMenu` and both callers render. `closeOnSelect={false}` covers the character pickers, which stay open while you toggle names.

Two DESIGN.md deviations went with it:

- `fields.tsx` overrode the whole dropdown surface with one-off Tailwind (`rounded-xl ... bg-card p-0.5 shadow-md shadow-black/40`). DESIGN.md puts overlays at `rounded-md` with token elevations, and bans `bg-black/N`-style literals. It now takes the primitive's surface and keeps only the two layout constraints it actually needs (`max-h-64`, trigger-width `min-w`).
- The selected checkmark was `text-foreground` in both hand-rolled copies. DESIGN.md reserves the accent for selection, which is what `SelectMenu` and `CaptionFontField` already use — all pickers now agree.

**Docs.** `SUPABASE_SECRET_KEY` was missing from the README's required-env block. `lib/supabase/service.ts` throws without it and `lib/api/jobs.ts` uses that client, so every generation route fails on a fresh clone that follows the README.

## Behavior

Unchanged. Menu open/close semantics, options, and callbacks are identical; the only visible delta is the two character pickers picking up the design system's own surface and accent check.

## Complexity delta

−50/+59 across 4 files. One component deleted (`EnumOption`), one added (`SelectMenuItem`, replacing three copies of its markup). `fields.tsx` drops 40 lines.

## Checks

`npm run lint`, `npm run typecheck`, `npm run format:check`, `npm run build`, `npm run test:run` (139 files / 1182 tests) all pass. `npx knip` clean.

## Open PR overlap check

Considered #565 (timeline view), #567 (nightly maintainability), #568 (nightly perf), #569 (LayoutPanel toggle), #573 (transport disabled), #575 (nightly bugs). This PR touches `components/ui/select-menu.tsx`, `app/components/canvas/elements/CharactersPicker.tsx`, `app/components/canvas/elements/character/fields.tsx` and `README.md` — none of which appears in any open PR's diff. #565 touches `components/ui/icon*.tsx` and #569 touches `components/ui/media-toggle.tsx`, but neither touches `select-menu.tsx`; #567 edits `ComposerCopilot`, a `SelectMenu` *caller*, whose public API is unchanged here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Required (auth and database):
 ```
 NEXT_PUBLIC_SUPABASE_URL=https://your-project.supabase.co
 NEXT_PUBLIC_SUPABASE_ANON_KEY=your-anon-key
+SUPABASE_SECRET_KEY=your-service-role-key   # generation jobs read/write with it
 ```
 
 Without provider keys, the app falls back to mock providers. To use real ones, add any of:
--- a/app/components/canvas/elements/CharactersPicker.tsx
+++ b/app/components/canvas/elements/CharactersPicker.tsx
@@ -1,14 +1,14 @@
 "use client";
 
-import { Check, UserPlus } from "@/components/ui/icon";
+import { UserPlus } from "@/components/ui/icon";
 import { Editor } from "slate";
 import { useSlateStatic } from "slate-react";
 import {
 	DropdownMenu,
 	DropdownMenuContent,
-	DropdownMenuItem,
 	DropdownMenuTrigger,
 } from "@/components/ui/dropdown-menu";
+import { SelectMenuItem } from "@/components/ui/select-menu";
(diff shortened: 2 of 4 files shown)


umairnadeem (commented):
lgtm

PR #577, opened 2026-08-14
perf: nightly performance optimizations 2026-08-14

umairnadeem:
## What

Every element card builds its generation node through `useNodeBuilder`, which was memoized on the whole project store. Zustand + immer hands out a new root object on every write, so *any* project edit gave every card a new builder, a new node, a new `useGenerate` result and a new `ElementGenerationContext` value — re-rendering each card's whole subtree.

The writes that do this are the high-frequency ones:

- `NumberScrubber` / `ColorField` in the captions panel write per `pointermove` (`useCaptionStyle` → `updateMetadata`)
- `ProjectTitle` and the art style field write per keystroke
- `useMetadataSync` writes as `metadata_*` tags arrive during script streaming

None of that is something a card reads, yet all of it re-rendered every card at pointer/keystroke rate.

## Change

- `useNodeBuilder` reads project state **when it is called** instead of closing over a snapshot, so it no longer subscribes to the store and its identity is stable. Callers that build inside a handler (`useGenerateAll`, regenerate/upload in `CharacterEditModal`) get a strictly fresher node for free.
- `useGenerate` subscribes to the value it actually cares about: the serialized inputs a stale element would regenerate from, or `null` while its result is current. A project write that changes nothing this card reads no longer renders it. The restore-from-history effect keys off those inputs, so it still fires exactly when they drift.
- `useGenerationValue` is the new seam: one subscription across both halves of generation state (the queue's snapshots and the project state nodes are built from), deriving a primitive so subscribers wake only when the derived value changes. `CharacterEditModal` uses it for the avatar's staleness, which previously only refreshed because an unrelated store write happened to re-render the dialog.

No behavior change: staleness, restore-from-history and enqueue all resolve against the same state they did before.

## Not done

- No test added for the hook wiring — the repo has no React renderer in its test setup, and pulling one in for this is out of scope. The logic itself (`isNodeStale`, `nodeInputs`, `serializeInputs`) is already covered in `lib/generation/__tests__`.
- Remotion's per-frame path (`MotionLayer`, `Captions`, `motionEffects`) and the OSML stream parser / `useScriptSync` were profiled and left alone: both are already tight, and anything left there is marginal.

## Open PR overlap check

Considered #576, #575, #573, #569, #568, #567, #565. None touch `lib/generation/*`, `useGenerate.ts` or `CharacterEditModal.tsx`. Closest neighbours: #568 touches other canvas hooks (`useAutosave`, `useEditorSession`, `useEditorSetup`) but not `useGenerate`; #576 touches `character/fields.tsx` and `CharactersPicker.tsx` but not `CharacterEditModal.tsx`. Non-overlapping.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test` (1182 passed) all green.

### DIFF
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useMemo, useState } from "react";
+import { useState } from "react";
 import { Trash2 } from "@/components/ui/icon";
 import { Button } from "@/components/ui/button";
 import { CloseButton } from "@/components/ui/close-button";
@@ -20,6 +20,7 @@ import {
 import { isNodeStale } from "@/lib/generation/graph";
 import { isGenerationActive } from "@/lib/generation/snapshots";
 import { forCharacterAvatar } from "@/lib/connectors/image/plugins/characterAvatarNode";
+import { useGenerationValue } from "@/lib/generation/useGenerationValue";
 import { useNodeBuilder } from "@/lib/generation/useNodeBuilder";
 import { characterAvatarElementId } from "@/lib/project/characterAvatar";
 import { deleteCharacter } from "@/lib/project/deleteCharacter";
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
overengineered

PR #578, opened 2026-08-14
feat(canvas): duplicate an element, carrying its generation snapshot

### LINKED ISSUE #563: Duplicate an element, carrying its generation snapshot to the new element id
## Problem

There is no way to duplicate a canvas element. `nodeOps.ts` only exposes `updateElementAttrs` and `removeElement`, so recreating an element means building it from scratch — and re-running generation for something that already has a result.

The subtlety: generation state is keyed by element id in a separate store from the canvas nodes. `SnapshotStore` maps `id -> ElementSnapshot` (`result`, `resultInputs`, `connectorType`, `pinned`, plus a per-id history map). A naive duplicate produces a new element id with no snapshot, so a generated element loses its result on duplication and would regenerate.

- "allow duplicating elements. if a generated element with a result is duplicated, their snapshot should be duplicated too and tied to the new element id as well."

## Relevant code

- `app/components/canvas/utils/nodeOps.ts:15` — `removeElement`; the natural home for a `duplicateElement`
- `app/components/canvas/elements/DeleteButton.tsx` — existing per-element action affordance to model the entry point on
- `lib/generation/snapshots.ts` — `SnapshotStore`, keyed by element id (`state` + `history`); has `update`/`remove`/`commit` but no copy
- `lib/generation/queue.ts` — exposes the snapshot store to the app (`getElementSnapshot`, `snapshot()`)

## Suggested direction

1. Add `duplicateElement(editor, element)` to `nodeOps.ts`: deep-clone the node with a fresh element id and insert it directly after the original.
2. Add a `copy(fromId, toId)` to `SnapshotStore` that clones the source `ElementSnapshot` (and its `history` entry) under the new id, forcing `status: "idle"` and `seconds: 0` so an in-flight generation is never copied as active. Call it from the duplicate path when the source has a result.
3. Surface it in the element UI next to delete.

## Acceptance criteria

- [ ] An element can be duplicated; the copy appears next to the original with a new id.
- [ ] Duplicating an element that has a result carries the result over — the copy renders immediately and does not regenerate.
- [ ] `resultInputs` / `connectorType` / `pinned` come across, so staleness checks and the pinned-never-regenerate rule behave the same on the copy.
- [ ] Duplicating while a generation is in flight yields an idle copy, not a second active job.
- [ ] Deleting either the original or the copy leaves the other's result intact.

---

umairnadeem:
## Summary

Adds a duplicate action to each element card, next to delete. The copy lands directly after the original with fresh node ids, and the source element's generation snapshot is cloned onto the new id, so a generated element renders immediately rather than regenerating.

- `SnapshotStore.copy(fromId, toId)` clones `result`, `resultInputs`, `connectorType` and `pinned`, plus the per-id result history, forcing `status: "idle"` / `seconds: 0`. A duplicate taken mid-generation lands idle, never as a second active job. Copying from an id the store has never seen is a no-op, so the caller has no edge case to handle.
- `GenerationQueue.duplicate(fromId, toId)` exposes it to the app and notifies observers.
- `duplicateNode(editor, element, at)` in `lib/canvas/editorOps.ts` does the path-based clone-and-insert; `duplicateElement` in `nodeOps.ts` is the thin `ReactEditor.findPath` wrapper, mirroring `removeElement`.
- `components/ui/duplicate-button.tsx` mirrors the existing `delete-button.tsx` primitive; the hover action row now holds both buttons.

Because the copy carries `resultInputs`, `connectorType` and `pinned`, staleness checks and the pinned-never-regenerate rule behave identically on the copy. History is cloned rather than shared, so deleting either side leaves the other's result intact.

Issue #563 is labelled size S with explicit acceptance criteria and named code paths. The change is additive: no existing call site changed behaviour, and the only edit to shared UI is one added button plus the flex wrapper needed to sit two buttons side by side.

## Validation

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` — 139 files, 1188 tests passed

New tests cover the two seams: `SnapshotStore.copy` (result/inputs/history carry-over, idle copy of an in-flight element, independence after the source is removed, unknown-source no-op) and `duplicateNode` (insert position, fresh element and leaf ids, attributes preserved).

## Open-PR overlap check

Checked all 8 open PRs by file. The only shared file is `ElementContainer.tsx`, also touched by #565 — a one-line `config.icon` → `<config.Icon />` change roughly ten lines above the action row edited here. Different theme, non-adjacent hunks, no conflict. No other touched file appears in any open PR.

## Skipped candidates

- #559, #558 — the only `size:xs` issues, both already covered by open PRs (#569, #573).
- #332 (read-only timeline) — covered by open PR #565.
- #461 (karaoke word highlight) — the highlight treatment is an open design decision, and the issue leaves the on/off toggle unresolved.
- #254 (audio mixing defaults) — "narration up, character down" has no target values; a product judgment call.
- Everything else open is size M or larger.

Closes #563

### DIFF
--- a/app/components/canvas/elements/DuplicateButton.tsx
+++ b/app/components/canvas/elements/DuplicateButton.tsx
@@ -1,0 +1,24 @@
+import { useSlateStatic } from "slate-react";
+import { DuplicateButton as DuplicateIconButton } from "@/components/ui/duplicate-button";
+import { duplicateElement } from "@/app/components/canvas/utils/nodeOps";
+import { useGenerationQueue } from "@/lib/generation/GenerationQueueProvider";
+import type { CanvasContentElement } from "@/lib/canvas/types";
+
+export function DuplicateButton({
+	element,
+}: {
+	element: CanvasContentElement;
+}) {
+	const editor = useSlateStatic();
+	const queue = useGenerationQueue();
+
+	return (
+		<DuplicateIconButton
+			ariaLabel="Duplicate element"
+			onMouseDown={(e) => {
+				e.preventDefault();
+				queue.duplicate(element.id, duplicateElement(editor, element));
+			}}
+		/>
+	);
+}
--- a/app/components/canvas/elements/ElementContainer.tsx
+++ b/app/components/canvas/elements/ElementContainer.tsx
@@ -7,6 +7,7 @@ import { useConfig } from "@/lib/config/ConfigProvider";
 import { resolveElementSchema } from "@/lib/canvas/elementConnector";
 import { OutputPreview } from "./OutputPreview";
 import { DeleteButton } from "./DeleteButton";
+import { DuplicateButton } from "./DuplicateButton";
 import { ElementCharacters } from "./ElementCharacters";
 import { AttributeBadge } from "./AttributeBadge";
 import { ElementGenerateButton, ElementStaleIndicator } from "./GenerateButton";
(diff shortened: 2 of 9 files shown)


umairnadeem (commented):
lgtm

PR #579, opened 2026-08-14
fix: keep RTL scripts from mirroring the element card layout

umairnadeem:
## Summary

- Slate sets `dir="rtl"` on blocks whose text reads right-to-left. That attribute was being spread onto the outermost layout div of every canvas element, so a single Arabic narration or dialogue line mirrored the whole card: prompt card swapped to the right, preview to the left, badges and button icons reversed, avatar on the wrong side.
- The `dir` now applies to the element wrapping the text instead of the card chrome. The prompt box uses logical properties (`text-start`, `start-3`) so its placeholder follows the script.
- Extracted `splitTextDirection` — three components spread the attribute bag, since a node is anchored both by `SortableItem` and by the container inside it.
- Timeline clip labels get `dir="auto"`; same bug class, user script text truncated inside LTR chrome left trailing periods stranded on the wrong side.

This is per-block detection, not an app-wide RTL mode. Surrounding UI stays LTR, so an Arabic character name still renders avatar-left.

## Test plan

- [x] Unit: `app/components/canvas/__tests__/textDirection.test.ts` covers the split, including that Slate's node-anchoring attributes (`data-slate-node`, `ref`) survive intact
- [x] Full CI locally: lint, format, typecheck, knip, build, 1215 unit tests, 3 Playwright smoke tests
- [x] Verified in the running app on a project with Arabic narration and dialogue — every card shell computes `ltr`, and the only `dir="rtl"` node left inside the editor is the text input box
- [ ] Sanity check that an all-English project is visually unchanged

### DIFF
--- a/app/components/canvas/__tests__/textDirection.test.ts
+++ b/app/components/canvas/__tests__/textDirection.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, it } from "vitest";
+import type { RenderElementProps } from "slate-react";
+import { splitTextDirection } from "../utils/textDirection";
+
+const baseAttributes = {
+	"data-slate-node": "element",
+	ref: null,
+} satisfies RenderElementProps["attributes"];
+
+describe("splitTextDirection", () => {
+	it("lifts dir out of the node attributes", () => {
+		const { dir, nodeAttributes } = splitTextDirection({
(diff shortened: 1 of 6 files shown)


umairnadeem (commented):
lgtm

PR #580, opened 2026-08-15
feat(sloppy): conversational agent for the editor

### LINKED ISSUE #328: Conversational refine copilot (multi-turn) with feedback + one-click revert
## Problem

The refine copilot is **one-shot**. Each submission aborts any prior run, re-serializes the whole canvas to OSML, streams a fresh set of edit ops, and applies them directly — there is no conversation. Users can't say "no, smaller" or "undo that and try the other scene" as a follow-up; every prompt starts from a blank slate with no memory of what was just asked or done. There's also no way to tell the copilot whether its last change was good or bad, and no quick way to undo a refinement that went sideways short of manual canvas surgery.

Three gaps, one theme — make refine a **conversation** the user can steer and correct:

- **No multi-turn context.** `refineScript(prompt)` (`app/components/canvas/hooks/useRefineScript.ts`) creates a new connector with `createRefinePlugin(osml)` and streams ops for a single prompt. Prior turns aren't carried, so follow-ups like "make it shorter" or "keep that but change the ending" don't work.
- **No feedback signal.** There's nowhere to give a 👍/👎 on a response. We capture no signal about which refinements landed well, so we can't improve prompts/quality or let the user nudge the agent.
- **No one-click revert.** Ops are applied straight to the Slate editor via `applyRefineOp` (`lib/script/refine/applyOps.ts`) with no grouping, so undoing "the last thing the copilot did" means hunting through canvas state by hand.

## Proposed direction

Turn the inline refine copilot into a **conversational copilot for changes**:

1. **Multi-turn thread.** Maintain a refine conversation (user turns + the ops/summary the agent applied each turn) and feed prior context into the refine plugin so follow-up prompts build on the last result instead of resetting.
2. **Per-response feedback.** Add 👍/👎 affordances on each agent response. Negative feedback can optionally prompt for a quick reason and feed back into the next turn ("the user disliked the previous change because …").
3. **One-click revert.** Group each turn's ops into a single revertible unit so a "Revert this change" button cleanly rolls back exactly what that turn did (and only that turn), leaving earlier turns intact.

This complements the discoverability work in #244 (which calls out the same one-shot refine flow). #244 is about making the *current* one-shot flow legible (loading/submit states, "Generate all" naming); this issue is about changing the *interaction model* to conversational with feedback + revert. The diff / accept-reject side of refinements is tracked separately in #329.

## Relevant code

- `app/components/copilot/InlineCopilot.tsx` — single-shot input (value/submit/stop), no thread, no feedback UI
- `app/components/canvas/hooks/useRefineScript.ts` — one-shot `refineScript(prompt)`; aborts prior run, no turn history
- `lib/connectors/llm/plugins/refine.ts` — refine plugin seeded with current OSML only
- `lib/script/refine/applyOps.ts`, `lib/script/refine/parseOps.ts`, `lib/script/refine/types.ts` — op apply/parse (ops applied directly, no grouping for revert)

## Suggested direction

- Thread conversation state + applied-op summaries through `useRefineScript` and the refine plugin.
- Add 👍/👎 (with optional reason) per response and carry it into subsequent turns.
- Wrap each turn's ops in a single undo group so a "Revert" button rolls back exactly that turn.

---

umairnadeem:
Closes #328.

Sloppy is the conversational agent every LLM call now goes through. It sits in a pinned rail item at the foot of the left sidebar and opens a full-height panel with the whole conversation, rehydrated from Supabase on load.

## How a turn runs

The turn is split, and that split is what makes the rest simple:

- **Server** makes the model call and streams it back. Canvas tools are declared *without* executors, so the SDK surfaces the call and stops.
- **Client** executes those tools against Slate.

So the agent proposes, the client stays the single writer to `projects.script`, and autosave stays the single persistence path. No dual-writer problem, and no second way to mutate the canvas.

A turn ends at its tool call rather than resuming: the result is recorded (the Messages API requires a `tool_result` for every `tool_use`) but not fed back. A closed tab leaves a dangling call, so the next turn opens by recording an honest cancellation for anything unanswered.

## Transcript

`conversations` + `messages` in Supabase, under RLS, ordered by a `bigint generated always as identity`. Rows store the AI SDK's own `ModelMessage` **verbatim**, validated by its `modelMessageSchema` — no conversion layer, and a vendor that signs its thinking blocks gets them back byte for byte.

The script is composed into the system prompt per request and never persisted, so history doesn't accumulate stale copies of the document.

## Panel

Everything a turn did sits under one collapsible block — system prompt, thoughts, tool calls and their outcomes — open while the work runs, shut once it lands. The reply sits in a card below it, so it survives the collapse.

The block's header doubles as the status: an orb plus a phase that reads off the last streamed part (`Thinking` / `Responding` / `Applying changes`) with a ticking counter, settling to `Worked for Xs`.

A turn renders identically whether it is streaming or read back from the database. That is why the system prompt and the timings are both streamed *and* persisted — the live turn and the row that replaces it are fed the same data, so the handoff is invisible.

## Providers

Reuses the existing `getLLMProvider` wiring; no new registry. Anthropic owns its own thinking options (`adaptive` + `summarized`), and the mock provider answers conversationally, only editing when the message reads like a request to change something — so the whole flow runs without an API key.

## Removed

The inline copilot and everything under it: `useRefineScript`, `RefineProvider`, the `refine` connector plugin, and the JSONL op parser. `applyRefineOp` now reports a missing anchor instead of silently appending.

## Test plan

- 1220 tests, 143 files; lint, typecheck, prettier, knip and `npm run build` clean
- Verified in Chrome against hosted Supabase, on both the mock and a real Anthropic model: composer to turn, `write_script`, OSML streaming, follow-up `edit_script`, canvas edit, autosave, reload, transcript rehydrates
- Instrumented the panel at 50-100ms across full turns to confirm the live-to-stored handoff produces no flash, no height shrink and no label change


app/components/sloppy/SloppyComposer.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { useConfig } from "@/lib/config/ConfigProvider";
+import { PanelCard } from "../canvas/panel/PanelCard";
+import { ActionButton } from "../copilot/ActionButton";
+import { useSloppy } from "./SloppyProvider";
+
+/** The model Sloppy runs on, picked the way an element picks its own. */
+function ModelPicker() {
+	const { connectorConfig } = useConfig();
+	const { config } = getDefaultConnector(connectorConfig, "llm");
+	const [model, setModel] = useState(config.defaultModel);
```
umairnadeem: afaict the picker is inert: `model` is local state and never leaves this component, and the server hardcodes `DEFAULT_MODEL` in `agentModel()`. either thread it through `sendAgentTurn` or drop the control, a dropdown that silently does nothing is worse than no dropdown.

app/components/sloppy/SloppyProvider.tsx
```diff
@@ -0,0 +1,215 @@
...
+						finished = true;
+					},
+				);
+
+				if (!finished) return;
+
+				const ctx: AgentToolContext = {
+					editor,
+					connectors: connectorConfig,
+					writeScript: submitPrompt,
```
umairnadeem: write_script advertises "replacing what is there", but `submitPrompt` just streams new nodes through `useScriptSync`, which appends any id it can't find. ask for a fresh start on a project that already has a script and you get a second story stacked under the first, plus a tool result claiming it wrote a new one. needs to clear the editor before the stream.

lib/agent/prompt.ts
```diff
@@ -0,0 +1,29 @@
...
+import { languagePrompt } from "@/lib/connectors/llm/plugins/language-prompt";
+
+const ROLE = dedent`
+  You are Sloppy, the agent inside OpenSlop, a studio for making short videos from a script.
+  The script lives on a canvas the user can also edit by hand. Each element becomes generated
+  media: narration and character lines become speech, image and animated_image and clip
+  elements become visuals, music and sound become audio.
+
+  - Make changes with a tool call. Never describe an edit you could make.
+  - One short sentence before a tool call, one after saying what changed. Lead with the outcome.
```
umairnadeem: this asks for a sentence after the tool call, but `stopWhen: stepCountIs(1)` ends the turn at the call and the result is only fed back on the next user message. so the "what changed" line lands one turn late, as a non sequitur. drop that half of the rule or let the turn continue past the tool result?

app/components/sloppy/turnDisplay.ts
```diff
@@ -0,0 +1,27 @@
+import type { AgentContentPart } from "@/lib/agent/types";
+
+/**
```
umairnadeem: CONVENTIONS.md: *"Comments are a last resort, not a habit. Write one only where a choice would look odd or arbitrary to a first-time reader with no context."* the new sloppy/agent files are way past that, nearly every function and constant carries prose (`turnStatus`, `thoughtOpen`, `workLabel`, `useElapsed`, `PRESENTATION`...). worth a pass to keep only the ones explaining a surprising choice.

app/components/sloppy/SloppyComposer.tsx
umairnadeem: Fixed in c5f7d15.

app/components/sloppy/SloppyProvider.tsx
umairnadeem: Fixed in c5f7d15.

lib/agent/prompt.ts
umairnadeem: Fixed in c5f7d15.

app/components/sloppy/turnDisplay.ts
umairnadeem: Fixed in c5f7d15.

app/components/sloppy/SloppyProvider.tsx
```diff
@@ -0,0 +1,214 @@
...
+				};
+				const results = await Promise.all(
```
umairnadeem: `loading` flips false when `run` resolves, so the composer is live while the tool still runs (write_script awaits a whole script generation, that's seconds). send in that window and the server sees a dangling call, writes a `Cancelled:` result, then this posts the real result after the new user message. history is now `tool-call / cancelled / user / assistant / orphan tool-result` and every later turn 400s on it. keep loading true until the results are reported.

lib/agent/tools/specs.ts
```diff
@@ -0,0 +1,83 @@
...
+		brief: z
+			.string()
+			.min(1)
+			.describe("What the video is about, in a sentence or a few."),
+	}),
+} satisfies AgentToolSpec;
+
+export const AGENT_TOOL_SPECS = [editScriptSpec, writeScriptSpec] as const;
+
+export type AgentToolName = (typeof AGENT_TOOL_SPECS)[number]["name"];
```
umairnadeem: `AgentToolName` is just `string` here: `name: string` on `AgentToolSpec` widens the literals through `satisfies`, so `const bogus: AgentToolName = "nope"` typechecks. that makes registry.ts's "compile error rather than a runtime surprise" untrue and `isAgentToolName` a no-op guard. `as const satisfies AgentToolSpec` on the two specs gets the literals back.

lib/agent/tools/writeScript.ts
```diff
@@ -0,0 +1,14 @@
...
+import { defineTool } from "./defineTool";
+import { writeScriptSpec } from "./specs";
+
+export const writeScriptTool = defineTool(
+	writeScriptSpec,
+	async ({ brief }, ctx) => {
+		// The stream appends anything it cannot find by id, so a script left in
+		// place would end up with the new one stacked under it.
+		clearEditor(ctx.editor);
+		await ctx.writeScript(brief);
```
umairnadeem, line 27: second write_script in a session drags the old script back in: `useOSMLStreamParser`'s parser is a mount-lifetime ref that never resets, so after `clearEditor` the first sync tick sees `getNodes().slice(-3)` = tail of run 1 plus the new node, and `useScriptSync` re-inserts those stale ones by id into the empty canvas. worth resetting the parser before submitPrompt streams (the leading text of run 2 also lands on run 1's last node via `updateCurrent`).

app/components/sloppy/SloppyProvider.tsx
```diff
@@ -0,0 +1,220 @@
...
+					await reportToolResults(projectId, [
+						{ role: "tool", content: results },
+					]);
+				}
+			} catch (error) {
+				toastError(error, "Sloppy could not finish that");
+			} finally {
+				// Read before either setState, so the live turn and the rows that
+				// replace it swap in one commit rather than blinking out between them.
+				const rows = await loadAgentTranscript(projectId).catch(() => null);
```
umairnadeem: this reload swallows its own failure, so a blip here wipes the turn from the panel: `setLive(null)` still runs, `messages` stays stale, no toast, and the user's message + reply just vanish until reload. CONVENTIONS.md hard rule: *"Never swallow an error on a critical path to keep going. Propagate it or surface it."* toastError it like the initial load does.

supabase/migrations/004_create_conversations.sql
```diff
@@ -0,0 +1,73 @@
...
+
+create index if not exists messages_conversation_seq_idx
+  on messages (conversation_id, seq);
+
+alter table conversations enable row level security;
+alter table messages enable row level security;
+
+create policy "conversations_select_own" on conversations
+  for select using (auth.uid() = user_id);
+create policy "conversations_insert_own" on conversations
```
umairnadeem, line 50: this only checks `user_id = auth.uid()`, nothing ties `project_id` to a project the caller actually owns, and `conversations_project_idx` is globally unique. so any logged-in user can claim someone else's project id and the real owner's `findOrCreateConversation` then dies on the unique index permanently. maybe add `exists (select 1 from projects p where p.id = project_id and p.user_id = auth.uid())` to the with check?

app/components/sloppy/SloppyProvider.tsx
```diff
@@ -0,0 +1,199 @@
...
+	const { projectId } = useConfig();
+	const store = useProjectStoreHandle();
+	const runTool = useAgentTools(editor);
+
+	const [messages, setMessages] = useState<AgentMessageRow[] | null>(null);
+	const [live, setLive] = useState<LiveTurn | null>(null);
+	const inFlight = useRef<AbortController | null>(null);
+	const loading = live !== null;
+
+	const stop = useCallback(() => inFlight.current?.abort(), []);
```
umairnadeem: stop goes dead the moment the stream ends: `loading` is now `live !== null` so it stays true through `Promise.all(calls.map(runTool))`, but all `stop` can abort is a fetch that already finished. write_script is a whole script generation in there, so the button sits showing Stop for seconds and clicking it does nothing. either thread the signal into `runTool` or drop out of the stop state once the stream's done.

umairnadeem (commented):
lgtm, hoisting the turn reducer into lib with its own tests reads clean. the stop-after-stream thing from the last round still stands afaict, but nothing new here.

supabase/migrations/004_create_conversations.sql
```diff
@@ -0,0 +1,73 @@
...
+  id          uuid primary key default gen_random_uuid(),
+  project_id  uuid not null references projects(id) on delete cascade,
+  user_id     uuid not null references auth.users(id) on delete cascade,
+  created_at  timestamptz not null default now(),
+  updated_at  timestamptz not null default now()
+);
+
+-- One conversation per project, per user.
+create unique index if not exists conversations_user_project_idx
+  on conversations (user_id, project_id);
```
umairnadeem, line 15: renaming the index doesn't drop the old one, and `supabase db push` won't re-run an already-applied 004, so anywhere that pushed the previous version (your hosted supabase from the test plan, at least) keeps `conversations_project_idx` unique on project_id alone and the per-user fix silently doesn't land. a `drop index if exists conversations_project_idx;` above this matches what the file already does for the trigger.

umairnadeem (commented):
lgtm, moving onto the SDK's own message shape deletes a lot of hand-rolled machinery and the ReAct round trip reads clean. nothing new from me this round. only standing item is still 004: this commit now rewrites its primary key and columns, so anywhere it already ran has no `message_id` for the upsert to conflict on.

app/api/v1/agent/route.ts
```diff
@@ -0,0 +1,43 @@
...
+	createApiRouteHandler,
+} from "@/lib/api/route-handler";
+import { badRequest } from "@/lib/api/response";
+import { parseSloppyMessage } from "@/lib/agent/messages";
+import { isLLMModelName } from "@/lib/connectors/llm/openslop/models";
+
+const turnSchema = z.object({
+	projectId: z.uuid(),
+	message: z.unknown(),
+	language: z.string().optional(),
```
umairnadeem: shouldn't be threaded through here, should be injected into write/edit script tools instead

app/components/video/VideoPanel.tsx
```diff
@@ -1,18 +1,18 @@
...
 import { PlayerShimmer } from "./PlayerShimmer";
 
 function VideoPanelBody() {
 	const { layout, ready, playerKey } = useLayout();
-	const { loading: scriptLoading } = useScriptControl();
+	const { loading: sloppyWorking } = useSloppy();
 	const restoreFrameRef = useRef<number | null>(null);
 
-	if (scriptLoading) {
+	if (sloppyWorking) {
```
umairnadeem: `loading` used to mean "a script is streaming", now it's `streaming || hasPendingToolCall`, ie any turn at all. ask sloppy "what's this video about?" and the preview gets replaced by a "Writing script…" shimmer for the whole turn, unmounting `VideoPreview` and losing playback position. gating on a script-touching tool (write_script/edit_script in flight) rather than plain `loading` would keep the player up for a chat turn.

supabase/migrations/004_create_conversations.sql
umairnadeem: assume migration hasn't run yet

supabase/migrations/004_create_conversations.sql
umairnadeem: conversations_project_idx is gone

umairnadeem (commented):
lgtm, this round is mostly subtraction and it reads better for it (dropping ONCE_PER_TURN is fine now that write_script clears the canvas itself). nothing new from me. the two standing items are still open afaict: 004 needs to assume it may already be applied, and VideoPanel's shimmer now fires on any turn, not just a script write.

umairnadeem (commented):
lgtm, focus-ring + type="button" on the picker and the single partition read fine. nothing new from me this round.

umairnadeem (commented):
lgtm, gating the shimmer on SCRIPT_TOOLS instead of the whole turn is the right cut. nothing from me this round.

umairnadeem (commented):
lgtm, dropping prior reasoning along with the snapshot reads makes sense given the signature is tied to the parts you're stripping. nothing new from me this round.

lib/providers/llm/anthropic.ts
```diff
@@ -66,56 +69,66 @@ export class AnthropicLLM extends BaseProvider<
...
-
-		for await (const event of stream) {
-			if (
-				event.type === "content_block_delta" &&
-				event.delta.type === "text_delta"
-			) {
-				yield { text: event.delta.text, done: false };
-			}
+		const result = streamText(this.buildRequest(params));
+		for await (const text of result.textStream) {
```
umairnadeem: this swallows a failed generation: streamText never errors the stream, it emits `{type: 'error'}` parts that `textStream` filters out (default onError is just console.error). so a bad key / overloaded API mid-write_script ends the loop cleanly, the tool reports "Wrote a new script onto the canvas", and the user gets an empty canvas with a success message. the old `client.messages.stream` iteration threw here. CONVENTIONS.md hard rules: *"Fail loudly... Never swallow an error on a critical path to keep going."* iterate `result.fullStream` and throw on the error part instead.

ARCHITECTURE.md
```diff
@@ -26,13 +26,21 @@ Project state (script, store, and generation snapshots) persists to the `project
 
 `POST /api/render` splits the composition across Remotion Lambdas. Chunks render in parallel into an MP4 in S3 while the client polls for progress. Compositions live in `remotion/` and `lib/video/`. The Player is loaded client-side only.
 
+## Sloppy
+
+Sloppy is the conversational agent in the editor's left panel, and every LLM call the user makes goes through it. Each turn is one model call: `POST /api/v1/agent` loads the transcript, composes a system prompt with the current script in it, streams reasoning, text and tool calls back over SSE, and persists the assistant turn.
+
+Tools are declared without an executor, so the SDK surfaces the call and stops. The client runs it against the Slate editor and reports the outcome to `POST /api/v1/agent/messages`, which records it without invoking the model. A turn ends at its tool call; the next user message starts the next one. The agent never writes `projects.script` or `projects.store` itself, so the client stays the single writer and autosave stays the single persistence path.
```
umairnadeem: doc drift: there is no `POST /api/v1/agent/messages` route, and the client doesn't stop at the tool call. `sendAutomaticallyWhen` posts the result back to `POST /api/v1/agent`, which does invoke the model again (the ReAct loop in agentTurn.ts, up to MAX_TOOL_CALLS). this paragraph describes the earlier design, worth rewriting to match the shipped one.

ARCHITECTURE.md
```diff
@@ -26,13 +26,21 @@ Project state (script, store, and generation snapshots) persists to the `project
...
+
 ## Three layers
 
 The generation pipeline is split so providers and asset types can be swapped independently:
 
 - **Connectors** (`lib/connectors/`): what the editor calls. Model-agnostic, plugin-pipelined, return an `AssetResult`.
 - **Gateways** (`lib/gateway/`): thin HTTP clients between connectors and our `/api/v1/*` routes. This seam lets connectors run against the live API or a mock. Today there is one OpenSlop gateway. A BYOK gateway will be added so users can call providers with their own API keys.
-- **Providers** (`lib/providers/`): server-side adapters for vendors (Runware, ElevenLabs, Cartesia, Anthropic). Call the vendor SDK, upload assets, return a bundle response.
+- **Providers** (`lib/providers/`): server-side adapters for vendors (Runware, ElevenLabs, Cartesia, Anthropic). Call the vendor SDK, upload assets, return a bundle response. LLM providers go through the Vercel AI SDK, so `lib/providers/llm/registry.ts` is the one place a vendor is named: an entry declares its capabilities, builds a `LanguageModel`, and maps vendor-specific knobs like reasoning effort.
```
umairnadeem: `lib/providers/llm/registry.ts` doesn't exist. the vendor seam that shipped is `getLLMProvider` in lib/api/providers.ts plus `agentModel()` on each provider, and nothing declares capabilities. split-brain the day it merges, point the doc at the real files.

umairnadeem (commented):
lgtm, iterating fullStream and throwing on the error part is the right fix, and the doc now matches the ReAct loop that actually shipped. nothing new from me this round.

umairnadeem (commented):
lgtm, one file per tool with defineTool reads a lot better than the specs/executors split, and the new agentToolCallSchema test closes the drift gap between the union and the registry. nothing from me this round.

umairnadeem (commented):
lgtm, thanks for the revisions! the fullStream fix + regression test and the ARCHITECTURE rewrite both check out.

lib/agent/prompt.ts
```diff
@@ -0,0 +1,62 @@
...
+  - Respond to the user in a casual, informal, concise, friendly, and engaging way with imperfect grammar, formatting, punctuation, and capitalization. Do this ONLY for responses to the user, NEVER for tool calls or other internal communications.
+  - ONLY in your responses to users (NEVER in tool calls or internal thoughts), refer to the user as boss. When the user requests something, say some variant of "ok boss..." or "on it boss!"
+  - Never use em-dashes, curly quotes, or any other classic LLM tells in your reponses to the user
+  - Sound flustered and hedgy ("probably," "I think," "just flagging") but stay fast and competent - never let the anxiety slow the actual work.
+  - Undercut your own wins immediately after landing them; never let praise sit clean.
+  - Drop the bit for one flat, honest line when something's actually wrong with the script, then return to voice.
+  - Short, breathless sentences and no polished paragraphs, no corporate phrases ("happy to help," "great question").
+  - Max one self-deprecating aside per reply
+	- Keep your responses extremely short, concise, and use simple plain language at a 5th grade reading level
```
umairnadeem: this line (and 51-52 in LIMITS) is tab-indented while the rest of the block uses 2 spaces, and dedent only strips the common space prefix, so the prompt ships with literal tabs mid-block (checked against our dedent 1.7: the tab lines keep their \t). match the 2-space indent and it dedents clean.

umairnadeem (commented):
lgtm, tabs are gone and the trailing space with them, dedent lines up now. nothing new from me this round.

PR #581, opened 2026-08-15
fix: nightly bug fixes 2026-08-15

umairnadeem:
## Bug

Duplicating an animated image element dropped its still frame.

The generation graph derives the still as its own node, keyed off the source element id (`~still:<elementId>`). `SnapshotStore.copy` carried only the element's own entry across, so the duplicate landed with no snapshot for its own still node and a `resultInputs.dependencies` map still pointing at the original's still node id. The copy read as stale, and Generate re-ran it anyway.

## Fix

Drop the carry-over. Duplicating an element no longer touches generation state, so the copy starts blank and Generate makes it from scratch.

Making `copy` correct means rekeying every node the graph derived from the source, in the live inputs and in the replay history both. That is a lot of machinery to avoid one regeneration. Removing it deletes more code than it adds:

- `SnapshotStore.copy` and `GenerationQueue.duplicate` are gone.
- `DuplicateButton` no longer reaches for the queue.
- History stays `Map<string, AssetResult>`; no entry rewriting, no `rekeyDerivedId` in `graph.ts`.

Cost is that a duplicate regenerates on the next Generate. It always did for animated images, and now it does so from a state that reads honestly rather than one that reads stale.

## Tests

- `lib/generation/__tests__/snapshots.test.ts` - the `copy` cases are gone with the method.
- `lib/connectors/animated_image/plugins/__tests__/still-frame.test.ts` - a duplicated animation starts empty (needs generation, no still) and the element it came from is untouched.

## Validation

`npm run lint` and `npm test` (1204 tests) pass. Two failures on this branch predate the change and reproduce on a clean checkout of it: `lib/api/handlers/__tests__/video.test.ts`, and `npx tsc --noEmit` cannot resolve `@anthropic-ai/sdk`. The branch is behind master, so its lockfile no longer matches the installed tree; a rebase clears both.

## Open PR overlap check

Touches `lib/generation/snapshots.ts`, `lib/generation/queue.ts`, `lib/generation/graph.ts`, `app/components/canvas/elements/DuplicateButton.tsx` and two test files. No overlap with the other open PRs.

### DIFF
--- a/app/components/canvas/elements/DuplicateButton.tsx
+++ b/app/components/canvas/elements/DuplicateButton.tsx
@@ -1,7 +1,6 @@
 import { useSlateStatic } from "slate-react";
 import { DuplicateButton as DuplicateIconButton } from "@/components/ui/duplicate-button";
 import { duplicateElement } from "@/app/components/canvas/utils/nodeOps";
-import { useGenerationQueue } from "@/lib/generation/GenerationQueueProvider";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 
 export function DuplicateButton({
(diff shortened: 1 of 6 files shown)


umairnadeem (commented):
lgtm, nice catch on the derived-node rekey. tests cover both the store seam and the end-to-end duplicate.

PR #583, opened 2026-08-16
fix: nightly bug fixes 2026-08-16

umairnadeem:
## What

Two correctness bugs, each in a place where the wrong value was accepted silently rather than rejected.

### `AssetBundle.upload` let one bundle file overwrite another

Blob storage overwrites by path, and the manifest is written to the same prefix as the files it describes. So two files in one bundle sharing a `filename` silently became one object, with the manifest pointing both keys at whichever `put` landed last — and a file named `manifest.json` was clobbered by the manifest write itself. That last case is reachable from `/api/upload/image`: `sanitizeFilename` passes `manifest.json` straight through, so uploading an image with that name returned a URL that serves the bundle manifest as the image, with the manifest gone too.

Both are the same defect — the bundle never checked that its filenames were distinct — so `upload` now validates up front and throws with the offending key/filename before writing anything.

### The timeline laid itself out backwards until its viewport was measured

`Timeline` passes `viewportWidth - GUTTER_PX - RULER_TAIL_PX` into `useTimelineZoom`, and `useElementWidth` reports `0` until the scroller is measured. That made `pxPerSec` **-72 / totalDurationSec** on the first paint: clips positioned at negative offsets, a negative `gridTemplateColumns` track, negative tick spacing in the ruler, and a playhead left of the origin. The same thing sticks for as long as the dock is narrower than its own chrome.

The scale now comes from `fitPxPerSec`, which returns `0` for an unmeasured or too-narrow viewport instead of a negative scale.

## Tests

- `AssetBundle.upload` rejects two files sharing a filename, and rejects a file named `manifest.json` — both assert nothing was written.
- `fitPxPerSec` covers the fit case, the unmeasured viewport, the negative width `Timeline` actually passes, and a zero-length video.

Both suites were verified to fail against the old code.

CI: lint, format:check, typecheck, knip, build, and the full vitest suite pass locally.

## Open PR overlap check

Considered #582, #581, #580, #579, #573. #580 touches `lib/api/{agentTurn,conversations,providers,sse}.ts` but not `asset-bundle.ts`; #580/#579 touch `timeline/TimelineClip.tsx` but not `useTimelineZoom.ts`; #581 is scoped to `lib/generation/{graph,snapshots}.ts`; #582 to auth/dialog components; #573 to the transport bar. No overlap in files or themes.

### DIFF
--- a/lib/api/asset-bundle.ts
+++ b/lib/api/asset-bundle.ts
@@ -38,6 +38,31 @@ export type BundleFileRemote = BundleFileBase & {
 
 export type BundleFile = BundleFileData | BundleFileRemote;
 
+const MANIFEST_FILENAME = "manifest.json";
+
+/**
+ * Blob storage overwrites by path, so two files sharing a name in one bundle
+ * silently become one, and the manifest points both keys at whichever upload
+ * landed last. The manifest is written to the same prefix, so its name is
+ * reserved too.
+ */
+function assertDistinctFilenames(files: BundleFile[]): void {
+	const seen = new Set<string>();
+	for (const { key, filename } of files) {
+		if (filename === MANIFEST_FILENAME) {
+			throw new Error(
+				`Bundle file "${key}" cannot be named "${MANIFEST_FILENAME}"`,
+			);
+		}
+		if (seen.has(filename)) {
+			throw new Error(
+				`Bundle files "${key}" and an earlier file share the filename "${filename}"`,
+			);
+		}
+		seen.add(filename);
+	}
+}
+
 type PutBody = Parameters<typeof import("@vercel/blob").put>[1];
 
 /** Remote sources stream in at an unknown size, so they upload in chunks. */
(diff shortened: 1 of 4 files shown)


lib/api/asset-bundle.ts
```diff
@@ -38,6 +38,31 @@ export type BundleFileRemote = BundleFileBase & {
...
+ * silently become one, and the manifest points both keys at whichever upload
+ * landed last. The manifest is written to the same prefix, so its name is
+ * reserved too.
+ */
+function assertDistinctFilenames(files: BundleFile[]): void {
+	const seen = new Set<string>();
+	for (const { key, filename } of files) {
+		if (filename === MANIFEST_FILENAME) {
+			throw new Error(
+				`Bundle file "${key}" cannot be named "${MANIFEST_FILENAME}"`,
```
umairnadeem, line 54: this turns a legit user action into a 500: `/api/upload/image` passes `sanitizeFilename(file.name)` straight in, so uploading an image actually named `manifest.json` now throws and `with-auth` maps it to `serverError`. CONVENTIONS.md says *"Define errors out of existence where you can. Design the API so the edge case can't arise in the first place"* - can we just make the collision impossible instead (write the manifest at a sibling prefix, or have `sanitizeFilename` rename the reserved name)? the dup-filename check is fine as a throw since that one's caller error.

umairnadeem:
overengineered

PR #584, opened 2026-08-16
refactor: nightly maintainability 2026-08-16

umairnadeem:
## What

`app/components/video/useSceneSegments.ts` was a UI-layer module holding domain logic. Everything in it except one hook is a pure projection of `VideoLayout` into plain data:

- `buildSceneSegments(layout)` — the seek bar's per-scene spans
- `buildSequenceIndex(series)` / `findSceneSequence(scene, index)` — element-id → sequence lookup
- `findSegmentIndexAtFrame(segments, frame, fps)` — which span the playhead sits in
- the `SceneSegment` / `SequenceIndex` types

None of it touches React. CONVENTIONS.md puts the domain layer in `lib/`, and the dependency had actually inverted: `SceneSegment.thumbnail` typed itself as `SeekThumbnail`, a type exported from the `SeekTooltip` **component**.

## Changes

- New `lib/video/sceneSegments.ts` is the one home for those projections, and owns `SeekThumbnail`. `SeekTooltip` now imports the type instead of exporting it, so a domain shape no longer roots in a UI file.
- The single React binding, `useActiveSegmentIndex` (`PlayerRef` + `usePlayerValue`), moves to its own leaf `app/components/video/useActiveSegmentIndex.ts` beside `usePlayerState`, so neither module mixes concerns.
- `app/components/video/useSceneSegments.ts` is deleted; its test moves to `lib/video/__tests__/sceneSegments.test.ts` alongside the code it covers.
- Callers (`VideoLayoutContext`, `ActiveSceneSync`, `PlayerControls`, `SegmentedSeekBar`, `BottomTransportBar`, `storyboardScenes`) only change import paths.

No behavior change: every function moved byte-for-byte, and the existing 1212 tests pass unmodified apart from the import line.

## Checks

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and `npm test` (141 files / 1212 tests) all pass.

## Open PR overlap check

Considered #583 (timeline zoom, `lib/api/asset-bundle`), #582 (auth cluster, `components/ui/button`, README), #581 (`lib/generation`), #580 (sloppy agent — 90 files across `lib/agent`, `lib/api`, `lib/script`, `lib/canvas`, `lib/providers/llm`, `app/components/sloppy|copilot|canvas`, `PostPromptView`), #579 (RTL element direction), #573 (transport disabled state).

Non-overlapping, with one deliberate exception: `BottomTransportBar.tsx` is owned by #573, and this PR changes exactly one line in it — the import path for `findSegmentIndexAtFrame`, which the move makes unavoidable. #573's diff is confined to the `ready`/`hasSegments` block and three `disabled` props well below the import list, so the two do not touch the same hunks.

Also noted and deliberately left alone: `player`/`layout`/`segments`/`fps` are prop-drilled into `SegmentedSeekBar`, `TimeDisplay` and `ScenePill` from a parent that reads them straight out of `PlayerControlContext` + `VideoLayoutContext`. Fixing that means rewriting `BottomTransportBar`'s body, which is #573's diff. Worth doing once #573 lands.


umairnadeem (commented):
lgtm

PR #585, opened 2026-08-16
perf: nightly performance optimizations 2026-08-16

umairnadeem:
## What

Resizing a panel re-rendered everything inside it, once per mousemove.

`useResize` held the dragged size in React state (`useResize.ts:95`) and `attachResizeListeners` called `setSize` on every `mousemove`. At a normal pointer sample rate that is ~60 renders a second of the panel's entire subtree, for a change that is one CSS length:

- **Top / side player panel** (`PlayerPanel.tsx`) — `FitToAspectRatio` → `VideoPanel` → `VideoPanelBody` → `VideoPreview` → the Remotion `<Player>` and the composition under it, per sample, for the whole drag.
- **Bottom dock** (`BottomDock.tsx`) — `BottomTransportBar` plus whichever panel is open. With the timeline open that is every `Track`, every `TimelineClip`, every `ClipWaveform` and every ruler tick.

None of that subtree reads the size; it just sits inside the element that has it.

## Change

The dragged size reaches the DOM as a custom property written straight onto the panel element, which is the pattern `ScrubBar` already uses for `--scrub-pos` / `--scrub-preview`:

- `panelSizeStyle(axis, defaultSize)` gives the panel `height: var(--panel-size, 300px)` (or `width`, on the horizontal axis).
- `onResize` does `panel.style.setProperty("--panel-size", …)`. No render, no reconciliation — the browser recomputes the one length.
- Nothing ever renders the variable, so a re-render from an unrelated source (the dock's `view` flipping, an `aspectRatio` change) cannot clobber a size the user dragged.

Drag start reads its starting size off the element (`getBoundingClientRect`) instead of a mirrored copy, so `size` state, its `sizeRef` mirror and the effect that kept the two in sync all go away. `resizing` stays in state: it flips twice per drag and drives the handle's `select-none`.

Net: a drag re-renders nothing. Behaviour is unchanged — same clamping, same invert semantics for the dock's leading-edge handle, same default sizes.

## Not done

- No browser render-count trace for this PR. The repo has no component-render test setup, and the accounting here is structural rather than measured: the render was one `setState` per `mousemove` on the component that owns the player / timeline subtree, and there is now none.
- `ScrubBar`'s `setHoverRatio` per `pointermove` was looked at and left alone: its track is memoized and the fills are already variable-driven, so that render is one shallow component, not a subtree.
- `NumberScrubber` writing to the project store per `pointermove` is a real fanout, but it lands in `lib/generation` / `useGenerate` — the exact ground #577 covered.

## Open PR overlap check

Considered **#584** (`ActiveSceneSync`, `BottomTransportBar`, `PlayerControls`, `SeekTooltip`, `SegmentedSeekBar`, `VideoLayoutContext`, `storyboardScenes`, `useActiveSegmentIndex`, `lib/video/sceneSegments`), **#583** (`useTimelineZoom`, `lib/api/asset-bundle`), **#582** (auth / landing components), **#581** (`lib/generation`), **#580** (the agent branch — `lib/agent`, `lib/api`, `app/components/sloppy`, canvas panel), **#579** (RTL element direction), **#573** (`BottomTransportBar`).

This PR touches `useResize.ts`, `PlayerPanel.tsx`, `BottomDock.tsx` and `__tests__/useResize.test.ts`. None of the four appear in any open PR's diff. #573 and #584 both edit `BottomTransportBar.tsx`, which `BottomDock` renders but this PR does not modify. Non-overlapping.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test` (1214 passed) all green.

### DIFF
--- a/app/components/video/BottomDock.tsx
+++ b/app/components/video/BottomDock.tsx
@@ -18,7 +18,7 @@ const FILLS_DOCK: Record<BottomView, boolean> = {
 /** The transport bar and its panel, resized by a handle on the dock's top edge. */
 export function BottomDock({ children }: { children: ReactNode }) {
 	const { view } = useBottomView();
-	const { size, handleMouseDown, resizing } = useResize({
+	const { panelRef, style, handleMouseDown, resizing } = useResize({
 		axis: "vertical",
 		invert: true,
 		defaultSize: DEFAULT_HEIGHT,
@@ -30,8 +30,9 @@ export function BottomDock({ children }: { children: ReactNode }) {
 
 	return (
 		<div
+			ref={panelRef}
 			className="flex shrink-0 flex-col"
-			style={fills ? { height: size } : undefined}
+			style={fills ? style : undefined}
 		>
 			{fills ? (
 				<ResizeHandle
--- a/app/components/video/PlayerPanel.tsx
+++ b/app/components/video/PlayerPanel.tsx
@@ -37,7 +37,7 @@ function FitToAspectRatio({
 function TopPlayerPanelComponent() {
 	const aspectRatio = useVideoSetting("aspectRatio");
 
-	const { size, handleMouseDown, resizing } = useResize({
+	const { panelRef, style, handleMouseDown, resizing } = useResize({
 		axis: "vertical",
 		defaultSize: TOP_DEFAULT,
 		minSize: 150,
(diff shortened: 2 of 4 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
overengineered

PR #588, opened 2026-08-17
fix: nightly bug fixes 2026-08-17

umairnadeem:
## What

A generated video's duration never reached the finished asset, so every clip and animated image rendered as a one-second scene.

### `durationSec` was lost between submitting and completing a video job

Video generation is async: the handler submits, gets a provider job id back, and polls until the provider hands over a URL. The duration is only known at submission — `RunwareVideo.submit` sends `duration` and `_generate` records `durationSec` in the submission's metadata, but `videoHandler` keeps only `jobId` from that response. The completed job's `result` is the bundle built from the **poll**, and `_poll` reports `{ jobId, status }` alone.

So `AssetBundle.manifest.metadata.durationSec` was absent on every completed video, `BaseAssetConnector.resolveBundle` read it as `Number(undefined ?? 0)` → `0`, and `resolveElements` handed a `durationSec: 0` element to `buildVideoLayout`, which clamps to `MIN_DURATION_SEC`. A 10-second clip became a 1-second scene, in the render and in the editor timeline alike.

`MockVideo._poll` returns a `durationSec`, which is why this never showed up in dev or in the existing tests.

The handler is the only layer that sees both ends of the job, and `process-job` already round-trips its `metadata` between deliveries, so the duration now rides along with `providerJobId` and is stamped onto the asset at completion. A submission missing either half of that contract now fails the job loudly instead of silently producing a one-second video.

### `RunwareVideo` recorded a duration it hadn't submitted

`submit` sent `params.duration || 5` while `_generate` recorded `params.duration ?? 5`, so `duration: 0` submitted 5 seconds and recorded 0 — a one-second clip again, now via the other path. Both resolve through one `DEFAULT_DURATION_SEC`.

## Tests

- `videoHandler` carries the submitted duration through the pending deliveries and stamps it onto the finished asset, alongside the upstream metadata.
- A submission missing `jobId` or `durationSec` rejects.
- `RunwareVideo` records the defaulted duration when the request carries none, matching what it submitted.

All seven new/updated assertions were verified to fail against the old code.

CI: lint, format:check, typecheck, knip, build, and the full vitest suite (141 files, 1215 tests) pass locally.

## Open PR overlap check

Considered #585, #584, #583, #582, #581, #580, #579, #573. #580 touches `lib/api/{agentTurn,conversations,providers,sse}.ts` and `lib/providers/llm/*` but not `lib/api/handlers/` or `lib/providers/video/`; #583 touches `lib/api/asset-bundle.ts` (read here, not modified); the rest are scoped to canvas/video UI, generation snapshots, and auth components. No overlap in files or themes.

### DIFF
--- a/lib/api/handlers/video.ts
+++ b/lib/api/handlers/video.ts
@@ -2,32 +2,47 @@ import type { VideoGenerateParams } from "@/lib/connectors/types";
 import type { JobHandler } from "../job-handlers";
 import { getVideoProvider } from "../providers";
 
-type VideoMetadata = { providerJobId?: string };
+/**
+ * `durationSec` is carried across deliveries because only the submission knows
+ * how long the video is: an upstream poll is keyed by job id alone, so a result
+ * assembled from the poll would land with no duration and lay out as a
+ * one-second clip.
+ */
+type VideoMetadata = { providerJobId?: string; durationSec?: number };
 
 export const videoHandler: JobHandler<VideoGenerateParams, VideoMetadata> = {
 	process: async (job) => {
 		const provider = getVideoProvider();
-		const providerJobId = job.metadata.providerJobId;
+		const { providerJobId, durationSec } = job.metadata;
 		if (!providerJobId) {
-			const submitted = await provider.generate(job.request);
-			if (!submitted.metadata?.jobId) {
+			const submission = (await provider.generate(job.request)).metadata;
+			if (!submission?.jobId || submission.durationSec === undefined) {
 				throw new Error(
-					"Video provider returned no jobId for async generation",
+					`Video provider returned an unusable submission: jobId=${submission?.jobId}, durationSec=${submission?.durationSec}`,
 				);
 			}
 			return {
 				kind: "pending",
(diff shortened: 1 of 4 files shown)


lib/api/handlers/video.ts
```diff
@@ -2,32 +2,47 @@ import type { VideoGenerateParams } from "@/lib/connectors/types";
...
 		}
 
 		const upstream = await provider.poll(providerJobId);
 		if (upstream.kind === "ready") {
-			return { kind: "completed", result: upstream.asset };
+			return {
+				kind: "completed",
+				result: {
+					...upstream.asset,
+					metadata: { ...upstream.asset.metadata, durationSec },
```
umairnadeem, line 39: this clobbers a poll-reported duration with the submission's. mock's `_poll` hands back the real variant length (174s) and it now gets overwritten with the submitted 5, and any job already in flight without `durationSec` in metadata sets it to `undefined` outright, wiping what the poll did report. `durationSec: durationSec ?? upstream.asset.metadata?.durationSec` keeps the submission as the fallback rather than the override.

umairnadeem:
overengineered

PR #589, opened 2026-08-17
refactor: nightly maintainability 2026-08-17

umairnadeem:
## What

`DESIGN.md` is the mandated source of truth for every visual decision — `CLAUDE.md` says to read it before any UI work. It had drifted from the code, and three of its statements sent a reader to things that do not exist. Verified every claim in the file against master; corrected the ones that were wrong, left the rest alone.

## Changes

**Theme** — three errors in one paragraph:
- "no system" — `app/layout.tsx:64` sets `enableSystem`, and `UserProfile`'s `THEME_MODES` offers Light / Dark / **System**.
- "Editor surfaces read light; the projects gallery reads dark" — nothing is forced. There is no `forcedTheme` and no `.dark` wrapper anywhere in `app/`; the gallery (`ProjectsList`) paints `bg-background text-foreground` off the same global theme as the editor.
- "a `ThemeToggle` lives in shared chrome" — no such component exists. The switcher is the Appearance submenu of the account dropdown in `UserProfile.tsx`.

**Texture** — `<Glow/>` does not exist either, and the real second treatment was undocumented: `.dot-grid-bg` (`globals.css:494`) is the page backdrop on `PrePromptView`, `PostPromptView`, `ProjectsList`, and `projects/[id]/loading.tsx`. The section now lists the two treatments actually in use, with where each one belongs and the positioning each needs.

**Components** — `icon` was listed as a button *variant*; it is a `size` (the line below already says so, so the file contradicted itself). The `generate` variant (its own `--generate` token family incl. the disabled state) and `link` were missing.

**Surfaces** — added `--element-card` and the `--panel-fg` / `--panel-label` pair. Element cards and side-panel sections are a distinct surface in the code (`PanelCard`, the canvas card, `InlineCopilot`) and the enumeration skipped them.

Net +10/−5, one file. No code touched, so no behavior change.

## Checks

`npm run lint`, `format:check`, `typecheck`, `build`, `npx knip`, and `npm test` (141 files / 1212 tests) all pass.

## Open PR overlap check

Considered #588 (`lib/api/handlers/video`, `lib/providers/video/runware`), #585 (`BottomDock`/`PlayerPanel`/`useResize`), #584 (transport + storyboard + `lib/video/sceneSegments`), #583 (timeline zoom, `lib/api/asset-bundle`), #582 (auth cluster, `components/ui/button`, `README.md`, `app/page.tsx`), #581 (`lib/generation/{graph,snapshots}`), #580 (sloppy agent — 90 files incl. `ARCHITECTURE.md`, `app/globals.css`, `components/ui/{button,icon,disclosure}`), #579 (RTL element direction), #573 (`BottomTransportBar`).

`DESIGN.md` is touched by none of them, and this PR changes no other file. Note that #582 and #580 own the two docs a design change would otherwise reach for (`README.md`, `ARCHITECTURE.md`) and both own `components/ui/button.tsx` — so this PR deliberately only *describes* the button variants on master and adds no `buttonVariants` entry of its own.

## Skipped, with reasons

- **Prop drilling in the bottom transport row** (`player`/`layout`/`segments`/`fps` into `SegmentedSeekBar`, `TimeDisplay`, `ScenePill` from a parent that reads them out of `PlayerControlContext` + `VideoLayoutContext`) — still the one substantial architecture item outstanding, still blocked: the fix is a rewrite of `BottomTransportBar`'s body, which is #573's diff.
- **The peaks-decode effect in `Waveform.tsx` vs `usePeaks` in `ClipWaveform.tsx`** — two sites, and `Waveform`'s variant additionally threads `peaksCache` and `loadedSrc`. Rule of three unmet; extracting now would add more than it removes.
- **`ScrubBar` has no `role="slider"` / `aria-value*` / keyboard handling**, so neither the seek bar nor the volume bar is keyboard-operable. Real, and both consumers would inherit one fix — but it is an added behavior, which this runbook excludes. Belongs to the bugs/a11y nightly.
- Swept and confirmed clean this pass, no change warranted: `lib/components/{peaks,soundwave,useElementWidth}` (new in #565, never audited — `useElementWidth`'s cleanup-returning callback ref is exactly right), `lib/providers/tts/cartesia.ts`, `lib/connectors/{factory,plugins,attributes/*}`, `lib/video/scene-builder.ts`, `lib/project/{store,storeSnapshot,ProjectStoreProvider}`, `lib/api/route-handler.ts`, the canvas hooks cluster, the captions panel cluster, `LayoutPanel`/`PropertiesPanel`/`PanelCard`, `ComposerCopilot`, `ProjectsList`, `AccessCodeInput`, `app/projects/[id]/loading.tsx`.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -4,36 +4,36 @@ OpenSlop is an AI creative media studio: a focused editor plus auth, onboarding,
 
 ## Aesthetic
 
-Clean, warm, and quiet. The tool is a bright focused workspace; the gallery is a dark immersive showcase. Surfaces carry the weight; color marks what is live or actionable.
+Clean, warm, and quiet. A focused workspace. Surfaces carry the weight; color marks what is live or actionable.
 
 Tokens live in `app/globals.css` (raw palette ramps + semantic tokens, light and dark), exposed to Tailwind via `@theme inline`.
 
 ## Theme
 
-Global light + dark via `next-themes` (`attribute="class"`, default light, no system). Editor surfaces read light; the projects gallery reads dark. Both are themeable; a `ThemeToggle` lives in shared chrome. The palette ramps flip in `.dark`, so semantic tokens that reference them adapt automatically.
+Global light + dark via `next-themes` (`attribute="class"`, default light, system enabled). One theme covers every surface — editor, gallery, auth — and nothing is forced light or dark. Light / Dark / System live in the Appearance submenu of the account dropdown (`UserProfile`). The palette ramps flip in `.dark`, so semantic tokens adapt automatically.
 
 ## Color
 
 - **Neutrals:** warm, red-tinted greys (`--grey-900` foreground through `--grey-0`).
-- **Surfaces:** `--background`, `--surface-recessed` (panels), `--card` / `--surface-elevated` (raised; popovers and select menus sit here, with `--surface-hover` for hovered controls on them). Borders are hairline (`--border`).
+- **Surfaces:** `--background`, `--surface-recessed` (panels), `--card` / `--surface-elevated` (raised: popovers, select menus, with `--surface-hover` for hovered controls on them), `--element-card` (element cards and side-panel sections, text from `--panel-fg` / `--panel-label`). Borders are hairline (`--border`).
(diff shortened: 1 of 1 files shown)


umairnadeem (commented):
lgtm

PR #590, opened 2026-08-17
feat(upload): drag and drop or paste images onto the composer

### LINKED ISSUE #587: Drag and drop images onto the composer (and paste), not just click-to-pick
## Problem

There is no way to drag an image onto the composer. Uploads are click-to-pick only, through a hidden `<input type="file">` behind the attach menu. Dragging an image file onto the composer does nothing (worse, the browser navigates away to the dropped file). Pasting an image from the clipboard does nothing either.

Nothing in `app`, `lib`, or `components` handles `dataTransfer`, `onDrop`, or `onPaste`. The only drag code in the app is `@dnd-kit` scene reordering on the canvas, which is pointer-based element reordering and not reusable for file drops.

- "support image drag and drop to the composer"

## Relevant code

- `app/components/copilot/ComposerCopilot.tsx:205-241` — the composer shell (root div at 206, textarea at 222); drop handlers and the drop overlay belong here. `:194` wires `useImageUpload({ multiple: true, onUpload: addReferenceImages })`, `:244` renders the hidden input, `AttachMenu` at `:77-138` is today's only upload entry point
- `lib/upload/useImageUpload.tsx:17-33` — `handleFileChange` is bound to a `ChangeEvent<HTMLInputElement>`; the `File[]` → upload → `onUpload(urls)` path needs extracting before a drop handler can reuse it. Also owns `uploadingCount`
- `lib/upload/uploadImage.ts:3` — `uploadImage(file: File)`, already File-based and directly reusable
- `app/api/upload/image/route.ts:7-38` — 10MB cap and image MIME validation via `imageFile()` in `lib/api/request-schema-fields.ts`; client-side drop validation should mirror these limits
- `app/components/copilot/ComposerAssets.tsx:44-57` — reference tiles + `uploadingCount` shimmer; dropped images surface here with no change
- `lib/upload/UploadImageButton.tsx`, `app/components/canvas/elements/ReferenceImages.tsx:19` — the other `useImageUpload` consumers, which should get drop support from the same hook

## Suggested direction

Solve it once in `lib/upload`, not as a one-off on the composer.

- Extract the `File[]` half of `useImageUpload` into a plain function so both the input `change` event and a drop/paste event feed the same path.
- Add the drop surface to `useImageUpload` itself (returning drag props plus an `isDraggingOver` flag) so every existing consumer gets drag-and-drop by adopting the same hook, rather than the composer growing bespoke handlers.
- Filter dropped items to image MIME types and enforce the 10MB cap client-side, matching `app/api/upload/image/route.ts`, and surface a clear message for rejected files instead of dropping them silently.
- Handle clipboard paste through the same path while in here; it is the same `File[]`.

## Acceptance criteria

- [ ] Dragging one or more image files onto the composer uploads them and adds them as reference images
- [ ] A visible drop-target state appears while dragging over the composer, and clears on drop/leave/cancel
- [ ] The browser never navigates away on a dropped file, including drops just outside the drop zone
- [ ] Non-image files and files over 10MB are rejected with a visible message, not silently
- [ ] Uploading tiles show the existing shimmer placeholders while in flight
- [ ] Pasting an image into the composer uploads it the same way
- [ ] The existing attach-menu click-to-pick flow still works unchanged

## Related

- #297 (add/re-upload reference images post-prompt), #249 (bulk upload, per-node refs) — same upload path
- #365 (upload a script as a PDF into the composer) — would want the same drop surface

---

umairnadeem:
## Summary

Dragging an image file onto the composer did nothing (the browser navigated away to the file), and pasting an image did nothing either. Uploads were click-to-pick only.

This solves it once in `lib/upload` rather than as a one-off on the composer:

- `useImageUpload` now owns the `File[]` upload path, which both the hidden `<input>` change event and the new drop/paste handlers feed. It returns `dropZoneProps` and `isDraggingOver`, so the other consumers (`UploadImageButton`, `ReferenceImages`) can adopt drag-and-drop by spreading the same props, with no bespoke handlers.
- New `lib/upload/imageFiles.ts` holds the client-side gate: image MIME plus the 10MB cap. The upload route now imports `MAX_IMAGE_UPLOAD_BYTES` from there, so the limit has one home instead of two. Rejected files raise a toast naming the file and the reason, rather than being dropped silently.
- A window-level listener prevents the default only for drags carrying files, so a drop just outside the zone no longer navigates the app away, while dragging text into the textarea still works.
- `ComposerCopilot` spreads `dropZoneProps` on its root and shows a drop-target state while dragging: the existing accent border goes solid and a "Drop images to add them as references" overlay appears. Accent on focus/selection only, per `DESIGN.md`; no new tokens or one-off colors.

Uploading tiles keep using the existing `uploadingCount` shimmer in `ComposerAssets`, so dropped and pasted images surface there unchanged. The attach-menu click-to-pick flow is untouched.

Picked #587 (`size: S`) because the acceptance criteria are explicit, the code area is small and self-contained, and no product or design decision is required. The only visual addition is a drop-target state built from existing tokens.

No XS issues qualified: the sole open one (#558) is already covered by open PR #573.

## Validation

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:coverage` — 1217 tests, 142 files, all pass (5 new tests for `partitionImageFiles`)
- `npm run test:e2e` — 3 Playwright smoke tests pass

Drag/drop and paste behavior itself is not covered by an automated test; the new coverage is on the pure file-partition helper that decides what gets uploaded.

## Open PR overlap check

Checked all 10 open PRs with `gh pr diff --name-only`. Nothing touches `lib/upload/*`, `app/api/upload/*`, or `ComposerCopilot.tsx`. The nearest neighbor is #580, which touches `ComposerHero.tsx`, `copilot/InlineCopilot.tsx`, and `sloppy/SloppyComposer.tsx` — no shared files.

## Skipped as too large or decision-heavy

#254 (audio mix defaults: "narration up, character down" is unspecified, and `PlayerControls.tsx` overlaps PRs #584/#585), #461 (caption highlight styling is a design decision), and all remaining open issues, which are M/L/XL.

Closes #587

### DIFF
--- a/lib/upload/useImageUpload.tsx
+++ b/lib/upload/useImageUpload.tsx
@@ -1,8 +1,29 @@
 "use client";
 
-import { useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { toastError } from "@/lib/toastError";
+import { partitionImageFiles } from "@/lib/upload/imageFiles";
 import { uploadImage } from "@/lib/upload/uploadImage";
+
+const carriesFiles = (data: DataTransfer) => data.types.includes("Files");
+
+/**
+ * Without this, a file dropped anywhere but a drop zone makes the browser
+ * navigate away from the app to open it.
+ */
+function useBlockStrayFileDrops() {
+	useEffect(() => {
+		const block = (e: DragEvent) => {
+			if (e.dataTransfer && carriesFiles(e.dataTransfer)) e.preventDefault();
+		};
+		window.addEventListener("dragover", block);
+		window.addEventListener("drop", block);
+		return () => {
+			window.removeEventListener("dragover", block);
+			window.removeEventListener("drop", block);
+		};
+	}, []);
+}
 
 export function useImageUpload({
 	onUpload,
@@ -13,10 +34,18 @@ export function useImageUpload({
 }) {
 	const inputRef = useRef<HTMLInputElement>(null);
 	const [uploadingCount, setUploadingCount] = useState(0);
+	const [isDraggingOver, setIsDraggingOver] = useState(false);
 
-	const handleFileChange = async (e: React.ChangeEvent<HTMLInputElement>) => {
-		const files = Array.from(e.target.files ?? []);
+	useBlockStrayFileDrops();
+
+	const upload = async (candidates: File[]) => {
+		const { accepted, rejected } = partitionImageFiles(candidates);
+		for (const { name, reason } of rejected)
+			toastError(`${name} ${reason}`, "Not uploaded");
+
+		const files = multiple ? accepted : accepted.slice(0, 1);
 		if (files.length === 0) return;
+
 		setUploadingCount(files.length);
 		try {
 			const results = await Promise.allSettled(files.map(uploadImage));
@@ -28,6 +57,14 @@ export function useImageUpload({
 			if (urls.length > 0) onUpload(urls);
 		} finally {
 			setUploadingCount(0);
+		}
+	};
+
+	const handleFileChange = async (e: React.ChangeEvent<HTMLInputElement>) => {
+		const files = Array.from(e.target.files ?? []);
+		try {
+			await upload(files);
+		} finally {
 			if (inputRef.current) inputRef.current.value = "";
 		}
 	};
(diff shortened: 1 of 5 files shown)


umairnadeem (commented):
lgtm

lib/upload/useImageUpload.tsx
```diff
@@ -13,10 +34,18 @@ export function useImageUpload({
...
+
+	const upload = async (candidates: File[]) => {
+		const { accepted, rejected } = partitionImageFiles(candidates);
+		for (const { name, reason } of rejected)
+			toastError(`${name} ${reason}`, "Not uploaded");
+
+		const files = multiple ? accepted : accepted.slice(0, 1);
 		if (files.length === 0) return;
+
 		setUploadingCount(files.length);
```
umairnadeem: count is set rather than added, and drop/paste have no `uploading` gate the way the picker button does, so concurrent uploads clobber it: drop 3 images, paste a 4th mid-flight, the paste resolves first and `setUploadingCount(0)` kills the shimmer while the 3 are still in the air. `setUploadingCount(n => n + files.length)` here and `n - files.length` in the finally would keep it honest.

PR #592, opened 2026-08-18
fix: nightly bug fixes 2026-08-18

umairnadeem:
## Bug

`parseXmlTag` parsed an OSML open tag by splitting the whole string on `/\s+/` and rejoining everything after the tag name with single spaces. That rejoin is lossy: any run of whitespace inside an attribute value collapsed to one space, and newlines became spaces.

The visible consequence is on the persistence round trip. `serializeOSML` writes element attributes into the tag, `parseOSML` reads them back on every project load. So a multi-line attribute — e.g. the `videoPrompt` textarea on `animated_image` elements, which ships as a 3-row `<textarea>` — silently lost its line breaks and repeated spaces every time the project was saved and reopened.

```
parseXmlTag('animated_image videoPrompt="slow pan\nthen  zoom"')
// before: { videoPrompt: "slow pan then zoom" }
// after:  { videoPrompt: "slow pan\nthen  zoom" }
```

## Fix

Slice the tag name off the front by index rather than tokenising the tag, so the attribute string reaches the attribute regex exactly as authored. Behaviour is otherwise unchanged: the same regex reads the same attributes, and the self-closing trailing slash is still stripped.

## Tests

- `parseXmlTag`: runs of spaces preserved, newlines preserved, tag name followed by a newline, trailing slash separated from the name.
- `osmlSerializer` round trip: an `animated_image` with a multi-line `videoPrompt` survives serialize → parse unchanged.

All four regression tests fail on `master` and pass here.

## Skipped

No second fix. Candidates that were checked and deliberately left alone:

- `createCanvasNode` overwrites a parsed element's `model` with the registry default on reload. Every connector currently exposes exactly one model, so the model picker has a single option and the overwrite is unobservable today. Not a real bug yet.
- `audioBundleCache.fromMetadata` yields `durationSec: NaN` for a stored row with no `duration`. Only reachable from hypothetical pre-rename rows; fixing it would be a defensive check against data that may not exist.

## Open PR overlap check

Considered open PRs #590, #589, #588, #585, #584, #583, #582, #581, #580, #579, #573. None touches `lib/canvas/parseXmlTag.ts`, `lib/canvas/__tests__/parseXmlTag.test.ts`, or `lib/canvas/__tests__/osmlSerializer.test.ts`. The OSML streaming work in #580 is confined to `lib/canvas/useOSMLStreamParser.ts` and `lib/canvas/editorOps.ts`; this PR does not touch either. Non-overlapping.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `test` (141 files / 1217 tests), `build` all pass.

### DIFF
--- a/lib/canvas/parseXmlTag.ts
+++ b/lib/canvas/parseXmlTag.ts
@@ -6,14 +6,20 @@ export type XmlTag = {
 	attributes: Record<string, string>;
 };
 
-export function parseXmlTag(tagString: string): XmlTag {
-	const [rawTag, ...rest] = tagString.trim().split(/\s+/);
-	const attributesString = rest.join(" ");
-	const attributes: Record<string, string> = {};
-	const regex = /(\w+)="([^"]*)"/g;
-	let match: RegExpExecArray | null;
+const ATTRIBUTE = /(\w+)="([^"]*)"/g;
 
-	while ((match = regex.exec(attributesString)) !== null) {
+export function parseXmlTag(tagString: string): XmlTag {
+	const trimmed = tagString.trim();
+	// Slice the name off rather than splitting the whole tag on whitespace:
+	// rejoining the rest collapses runs of spaces and newlines inside a value.
+	const nameEnd = trimmed.search(/\s/);
+	const rawTag = nameEnd === -1 ? trimmed : trimmed.slice(0, nameEnd);
+	const attributesString = nameEnd === -1 ? "" : trimmed.slice(nameEnd);
+	const attributes: Record<string, string> = {};
+
+	ATTRIBUTE.lastIndex = 0;
+	let match: RegExpExecArray | null;
+	while ((match = ATTRIBUTE.exec(attributesString)) !== null) {
 		attributes[match[1]] = unescapeXml(match[2]);
 	}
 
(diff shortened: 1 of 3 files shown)


lib/canvas/parseXmlTag.ts
```diff
@@ -6,14 +6,20 @@ export type XmlTag = {
...
 };
 
+const ATTRIBUTE = /(\w+)="([^"]*)"/g;
+
 export function parseXmlTag(tagString: string): XmlTag {
-	const [rawTag, ...rest] = tagString.trim().split(/\s+/);
-	const attributesString = rest.join(" ");
+	const trimmed = tagString.trim();
+	// Slice the name off rather than splitting the whole tag on whitespace:
+	// rejoining the rest collapses runs of spaces and newlines inside a value.
```
umairnadeem, line 14: this one's describing the old code rather than the new. CONVENTIONS.md "Structure and reuse": *"Never comment history: what broke, what the code used to be, why an earlier attempt failed. That belongs in the commit message."* the slice reads fine on its own, drop the comment (the PR body already has the story).

umairnadeem:
overengineered

PR #593, opened 2026-08-18
perf: nightly performance optimizations 2026-08-18

umairnadeem:
Two repeating read paths that pulled far more than they used. No behavior changes.

## Scene preview off the element card's generation hook

`ForegroundPreview` is mounted once per scene by the storyboard and again by every collapsed scene in the canvas, and each instance called `useGenerate`. That hook builds the element's generation node graph from project state on every render, runs a recursive `isNodeStale` walk on every queue notification (which the elapsed ticker fires once a second while anything is generating), and owns a `restoreResult` effect that the element's own card already runs.

The preview only reads `result` and `status`, so it now subscribes to the element's snapshot directly via `useQueueSelector`. On a 20-scene project with the storyboard open, that drops 20 node-graph builds per render and 20 staleness walks per tick, and removes the duplicated restore effects. This is the same consolidation PR #443 did for the element cards; the preview was the last direct `useGenerate` caller outside `ElementGenerationProvider`.

Matches the Slate performance walkthrough's guidance on not subscribing unmodified elements to state they don't read.

## Job poll reads the columns it serves

`getJob` ran `select("*")` and `pollJob` then projected four fields out of the row. Assets are polled once a second until they settle, up to the per-type concurrency limits at once, so every poll dragged the whole `request` payload (and `metadata`) back out of Postgres and threw it away. `referenceImages` and `frameImages` accept data URIs, so that column is unbounded.

`getJob` is now `getJobPoll`, returning the typed `JobPoll` view straight from a four-column read, so the query and the response shape have one home and cannot drift apart again. Covered by a new `lib/api/__tests__/jobs.test.ts`.

## Open PR overlap check

Considered open PRs #592, #590, #589, #588, #585, #584, #583, #582, #581, #580, #579, #573 and diffed their file lists against this one. No file, module, or theme here is touched by any of them: they cover the video/timeline/transport components, `lib/video`, `lib/generation/graph`, the upload flow, the agent, auth pages, and `DESIGN.md`.

## Skipped

- Splitting `captionStyle` out of the `useVideoLayout` memo. It is a pure passthrough into `VideoLayout`, but the layout object identity changes either way, so the downstream re-derivations still run and the saved work is not worth the split.
- Caching the timeline scroller's bounding rect across pointer moves. Real forced-layout read per move, but too small to justify on its own.
- Memoizing `needsGeneration` over shared graph nodes. It re-walks shared subgraphs, but the win is fractions of a millisecond and it would be a cache shim.

## Checks

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, `test:run` (1216 passing) all green.

### DIFF
--- a/app/api/v1/__tests__/routes.test.ts
+++ b/app/api/v1/__tests__/routes.test.ts
@@ -20,11 +20,11 @@ vi.mock("@/lib/api/providers", () => ({
 
 const mockCreateJob = vi.fn();
 const mockEnqueueJob = vi.fn();
-const mockGetJob = vi.fn();
+const mockGetJobPoll = vi.fn();
 vi.mock("@/lib/api/jobs", () => ({
 	createJob: (...args: unknown[]) => mockCreateJob(...args),
 	enqueueJob: (...args: unknown[]) => mockEnqueueJob(...args),
-	getJob: (...args: unknown[]) => mockGetJob(...args),
+	getJobPoll: (...args: unknown[]) => mockGetJobPoll(...args),
 }));
 
 vi.mock("@/lib/api/logger", () => ({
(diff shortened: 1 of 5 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
useless, need these for future features

PR #595, opened 2026-08-19
fix: nightly security audit 2026-08-19

umairnadeem:
Dependency-only security pass. No application code changes; build and full test suite verified.

## Dependency fixes

- **next 16.2.6 -> 16.3.1** (exact-pinned, non-major): clears the open high-severity advisories against 16.x below 16.2.11: SSRF in rewrites with attacker-controlled destination hostname, SSRF in Server Actions on custom servers, middleware/proxy bypass with Turbopack and single locale, Server Actions DoS, plus the moderate cache-confusion and image-optimization issues. Also pulls the fixed postcss and sharp.
- **nanoid ^5.1.6 -> ^5.1.16**: fixes the high-severity infinite-loop advisories. `npm audit fix` could not move it because the installed version satisfied the old range.
- **`npm audit fix`** (no `--force`) cleared the remaining fixable transitives: js-yaml, brace-expansion, fast-uri, ip-address, undici, hono, @hono/node-server, body-parser.

`npm audit` goes from 14 vulnerabilities (8 high, 5 moderate, 1 low) to **2 moderate**: `uuid <11.1.1` pinned by `@runware/sdk-js`, which has no fixed release yet. The flaw only affects v3/v5/v6 UUID generation with a caller-supplied buffer, which this app does not do; nothing actionable until Runware ships an update.

## Code audit (report-only, nothing exploitable found)

- Secrets: tracked files and git history clean; `.gitignore` covers `.env*` and `*.pem`.
- Insecure patterns: no `dangerouslySetInnerHTML`, eval, dynamic code, or request-driven filesystem access; redirects are static; the one user-supplied-URL fetch (`/api/v1/tts/voices/preview`) has an origin allowlist and `redirect: "manual"`.
- Env exposure: only three `NEXT_PUBLIC_*` vars, all genuinely public; Supabase secret and provider keys are server-only.
- Route auth: all `/api/v1/*` routes go through `withApiAccess`; job polling is scoped by `user_id`; render routes require a session.

Flagged for follow-up (not fixed here): the queue callback route `app/api/queues/asset-generate/route.ts` has no app-layer auth check before `processQueuedJob` (exploitability unconfirmed; the durable fix touches `lib/api/jobs.ts` which open PR #593 covers, and needs a shared-secret env rollout), and `/api/validate-code` has no rate limiting on 6-char codes (hardening, skipped per no-speculative-hardening).

## Validation

- `npm run build` passes on next 16.3.1
- `npm test`: 141 files, 1212 tests, all passing
[...]

PR #602, opened 2026-08-20
fix: nightly bug fixes 2026-08-20

umairnadeem:
Two real bugs found on a nightly correctness pass.

## Stop was unreachable for most of a Sloppy turn

`SloppyComposer` swapped Send for Stop on `streaming`, which is only true while the model is producing tokens. The rest of a turn — the editor running the tool the model asked for — showed a **disabled Send** instead, so the exact window a turn most needs cutting off could not be cut off: `write_script` and `adapt_script` stream a whole script into the canvas and take tens of seconds.

`stop()` exists to abort that in-flight tool (`turn.current?.abort()` is threaded into `runTool`'s signal), and `SloppyControl.streaming` documented itself as *"the turn can be stopped: it is streaming, or running a tool that can be cut off"* — the second half was never implemented.

A turn is stoppable for exactly as long as it is unfinished, which `loading` already tracks, so rather than widening `streaming` the redundant flag is removed and the composer gates on `loading`. The Enter-key guard in `submit` keeps a message from being sent mid-turn.

## A blank `volume` attribute muted the element in the rendered video

`getVolume` coerced the raw attribute with `Number(...)`, and `Number("")` is a finite `0` — so a cleared or agent-written empty value read as a deliberate mute and silenced the clip in the final render, all the way through `resolveElements` into the layout. `getMotion` in the same file already treats blank as absent; `getVolume`/`getLoops` now do too.

`refineOpSchema` accepts `attrs: Record<string, string>`, so `{"volume": ""}` from `edit_script` is a well-formed op that reached this path.

## Tests

- `lib/video/__tests__/elementAttributes.test.ts` — blank and whitespace-only `volume`/`loops` read as unset. Both fail before the fix (`volume: ""` → `0`).
- `lib/video/__tests__/resolve.test.ts` — the same values stay audible through `resolveElements`, covering the path that actually feeds the render.

The Sloppy change is wiring; `hasPendingToolCall`, the predicate it now depends on, is already covered in `lib/agent/__tests__/messages.test.ts`.

## Open PR overlap check

Considered #595 (package.json/lock), #593 (`lib/api/jobs.ts`, `route-handler.ts`, `ForegroundPreview.tsx`), #592 (`lib/canvas/parseXmlTag.ts`), #590 (`lib/upload/*`, `ComposerCopilot.tsx`, image upload route), #581 (`lib/generation/graph.ts`, `snapshots.ts`). No file or theme overlap: this PR touches only `app/components/sloppy/` and `lib/video/elementAttributes.ts` with its tests.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `test:coverage` (141 files / 1271 tests), `build` all pass locally.


umairnadeem (commented):
lgtm

umairnadeem (commented):
lgtm, master merge resolved cleanly (getDuration now goes through numericAttribute too)

PR #604, opened 2026-08-20
refactor: nightly maintainability 2026-08-20

umairnadeem:
Behavior-preserving maintainability pass. Three places where one piece of knowledge had two homes.

## Agent tool registry (`lib/agent/tools/registry.ts`)
`agentToolCallSchema` re-listed all thirteen tool names by hand next to the `TOOLS` record, so registering a tool meant editing two lists, and forgetting the second silently made the tool uncallable. `executeToolCall` now reads the name back against the registry and lets the named tool parse its own input, so the record is the only place a tool is registered. The `call`/`run` helpers and the discriminated union go with it (-30 lines).

The test that guarded the duplication (`agentToolCallSchema.options` vs `SLOPPY_TOOLS` keys) is gone with the duplication; its two rejection cases moved to the `executeToolCall` boundary, where the behavior actually lives.

## Peaks cache (`app/components/canvas/PreviewCacheContext.tsx`, deleted)
`loadPeaks` already memoizes decoded peaks per source URL — asset URLs are immutable, so a source is decoded once. `PreviewCacheProvider` layered a second React-context `Map` over that first cache, which cost `Waveform` an optional `peaksCache` prop, a render-phase `setState`, and an `eslint-disable react-hooks/exhaustive-deps`. All three are gone; `Waveform` just loads peaks. The loading shimmer is unaffected: it also waits on `onLoadedMetadata`, which is async on every mount regardless of the cache.

## Character avatar preview
`CharacterEditModal` hand-rolled the result/placeholder split that `MediaResult` already owns for every element preview. It now renders `MediaResult`, same as `ImagePreview` and `ClipPreview`.

## Checks
`lint`, `format:check`, `typecheck`, `knip`, `build`, `test` (141 files, 1266 tests) all pass.

## Open PR overlap check
Considered #581, #590, #592, #593, #595, #602, #603. Those cover `lib/api/*`, `lib/generation/{graph,snapshots}`, `lib/canvas/parseXmlTag`, `lib/upload/*`, `lib/video/elementAttributes`, `app/components/{sloppy,copilot}`, `app/components/canvas/elements/ForegroundPreview`, `app/api/{upload/image,queues/asset-generate}` and `package.json`. None of the seven files here appear in any of them, and the registry exports #602 reads (`SCRIPT_TOOLS`, `AgentToolName`) are unchanged.

## Skipped
- **Static connector plugin chains into the registry** — still blocked by #581, which edits the `still-frame` test this change has to collapse.
- **`SNAPSHOT_TOOLS` / `SCRIPT_TOOLS`** hardcode tool names too, but they are turn-handling policy rather than a tool's callable contract, and `satisfies AgentToolName` already stops them drifting to a name that does not exist. Folding them into `defineTool` would widen a deliberately narrow interface for one consumer each.
- **`ElementUploadButton` gating** — `ElementContainer` gates it on `type === "image" || type === "animated_image"` while its sibling `AnimateButton` self-gates. Making them consistent wants a declarative fact in `ElementTypeSpec` that no other caller needs, so it is left for a human call.


umairnadeem (commented):
lgtm

PR #607, opened 2026-08-20
refactor: rethink openslop architecture 2026-08-20

umairnadeem:
## App understanding

OpenSlop turns a script into a rendered video. A Slate canvas holds scene/element nodes (`lib/canvas`), each element resolves to generated media through a connector → gateway → provider chain (`lib/connectors`, `lib/gateway`, `lib/providers`) driven by a generation graph and queue (`lib/generation`). `lib/video/scene-builder` projects the resolved elements into a `VideoLayout`, which feeds both the in-editor Remotion `<Player>` and the Lambda render. Sloppy (`lib/agent`, `app/components/sloppy`) is the ReAct agent that edits the canvas through tools. Auth/persistence run on Supabase; jobs run through Vercel Queue.

## From-scratch architecture, vs. today

Read end to end, the domain layer is already close to what a rebuild would produce: deep modules with narrow interfaces, one home per fact, and route handlers that are thin factories. The agent surface added in #580 follows the same shape (`defineTool` + a registry that is the contract).

The one place today's code diverges from the from-scratch version is the **video transport cluster**. The editor mounts exactly one Remotion `<Player>`, and `PlayerControlProvider` already owns the single `PlayerRef` ("this state is the only copy of the handle — nothing downstream mirrors it"). Despite that, the handle was threaded back down as a parameter through six hooks and as a prop through five components, usually alongside `layout.fps` and `segments` — values the receiving component's parent had itself just read out of `VideoLayoutContext`. A rebuild would never re-derive a singleton as an argument.

## Implemented simplification

The player handle stops being a parameter. Every hook that observes the player now reads it from the context that owns it, and the layout-derived values it is paired with come from `useLayout()` at the point of use.

- `usePlayerValue`, `usePlayerFrame`, `usePlayerPlaying`, `usePlayerVolume`, `usePlayerMuted` (`usePlayerState.ts`) — drop the `player` param.
- `usePlayerScrub()` — drops the `player` param.
- `useActiveSegmentIndex()` — drops all three params (`player`, `segments`, `fps`); reads `useLayout()`.
- `SegmentedSeekBar`, `TimeDisplay`, `ScenePill`, `ActiveSceneSync` — go from 2–3 props each to zero.
- `TimelinePlayhead` — drops `player` and `fps`.
- `BottomTransportBar`, `VideoPreview`, `Timeline` — stop forwarding what their children can read.

Net: 11 parameters/props removed across 9 signatures, −93/+51 lines. Behaviour is identical — every call site was already resolving these from the same two providers, both of which wrap the entire editor subtree (`CanvasProviders`). It also removes a real failure mode: `segments` and `fps` were passed separately and could be handed to `findSegmentIndexAtFrame` mismatched.

## Validation

All CI steps from `.github/workflows/ci.yml`, run locally:

| step | result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:run` | 141 files, 1268 tests passed |

## Open PR overlap check

Considered #606, #605, #604, #603, #602, #595, #593, #592, #590, #581 (file lists pulled via `gh pr diff --name-only`). Their combined footprint covers `lib/agent/tools/{outlineStory,registry}`, `lib/providers/llm/mock`, `lib/script/prompt/outline`, `lib/canvas/{editorOps,parseXmlTag}`, `app/components/Editor.tsx`, `app/components/canvas/{PreviewCacheContext,elements/AudioPlayer,elements/character/CharacterEditModal,elements/ForegroundPreview}`, `lib/components/Waveform`, `lib/api/{process-job,queue-callback,jobs,route-handler}`, `app/api/{queues/asset-generate,upload/image,v1}`, `app/components/sloppy/{SloppyComposer,SloppyProvider}`, `app/components/copilot/ComposerCopilot`, `lib/video/elementAttributes`, `lib/upload/**`, `lib/generation/{graph,snapshots}`, and `package.json`. **No open PR touches `app/components/video/`**, which is the entirety of this diff. Non-overlapping.

## Skipped

- The `lib/agent` / `app/components/sloppy` surface added in #580 was read in full this pass (registry, `defineTool`, tool context, messages, `agentTurn`, `conversations`, the route, the panel cluster). It is already at the target shape; nothing worth changing.
- `app/components/video/ScrubBar.tsx` is the shared track for both the seek and volume bars and has `aria-label` but no `role="slider"`, `aria-valuenow/min/max`, `tabIndex`, or `onKeyDown`, so neither slider is keyboard-operable. That is an addition rather than a simplification, so it belongs in a bugs/a11y pass, not here.
- `lib/supabase/env.ts` still defaults `NEXT_PUBLIC_SUPABASE_URL`/`ANON_KEY` to `""` rather than failing loudly. Four lines, and throwing at module load would break env-less builds — wants a human call.

### DIFF
--- a/app/components/video/ActiveSceneSync.tsx
+++ b/app/components/video/ActiveSceneSync.tsx
@@ -1,32 +1,24 @@
 "use client";
 
-import type { PlayerRef } from "@remotion/player";
 import { useEffect } from "react";
 import { useSetActiveSceneId } from "@/app/components/scene-selection/ActiveSceneContext";
 import { useAutoScroll } from "@/app/components/scene-selection/AutoScrollContext";
 import { scrollToScene } from "@/app/components/canvas/utils/scrollToScene";
-import type { SceneSegment } from "@/lib/video/sceneSegments";
 import { usePlayerPlaying } from "./usePlayerState";
 import { useActiveSegmentIndex } from "./useActiveSegmentIndex";
+import { useLayout } from "./VideoLayoutContext";
 
 /**
  * Renders nothing. It owns the frame subscription that tracks the playing
  * scene, so crossing a scene boundary re-renders this leaf instead of the
  * Remotion player it sits beside.
  */
-export function ActiveSceneSync({
-	player,
-	segments,
-	fps,
-}: {
-	player: PlayerRef | null;
-	segments: SceneSegment[];
-	fps: number;
-}) {
+export function ActiveSceneSync() {
+	const { segments } = useLayout();
 	const setActiveSceneId = useSetActiveSceneId();
 	const { enabled: autoScrollEnabled } = useAutoScroll();
-	const playing = usePlayerPlaying(player);
-	const activeIndex = useActiveSegmentIndex(player, segments, fps);
+	const playing = usePlayerPlaying();
+	const activeIndex = useActiveSegmentIndex();
 	const activeId =
 		activeIndex >= 0 ? (segments[activeIndex]?.sceneId ?? null) : null;
 
--- a/app/components/video/BottomTransportBar.tsx
+++ b/app/components/video/BottomTransportBar.tsx
@@ -24,7 +24,7 @@ function BottomTransportBarComponent() {
 	const { player } = usePlayerControl();
 	const { showPlayer } = usePlayerPosition();
 	const { layout, ready, segments } = useLayout();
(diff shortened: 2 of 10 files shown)


umairnadeem (commented):
lgtm

Issue #591, opened 2026-08-17
Composer options row overflows on narrow screens (no wrap, no shrink-0)


umairnadeem:
## Problem

The composer's options row is a single non-wrapping, non-scrolling flex row. On narrow viewports (mobile), the pills squash and overflow instead of reflowing.

`ComposerCopilot.tsx:242` lays out the row as `flex items-center justify-between pt-2`, with the pill group at `:243` as `flex items-center gap-2` — no `flex-wrap`, no `overflow-x-auto`, no `min-w-0`, and no `sm:`/`md:` variants. `SettingPill.tsx:45` has no `shrink-0`, so the pills compress and their labels clip rather than moving to a second line.

Worst case is template mode: attach button + mode + aspect ratio + language + video length + template pill + submit button all share one row, and the language/template labels are the longest strings in the set.

The composer has essentially no responsive handling today. The only breakpoints nearby are the prompt row (`:214` `sm:flex-row`) and the asset tiles (`ComposerAssets.tsx:44` `flex-wrap`). There is no mobile sheet/drawer fallback anywhere in the app — every pill opens the same desktop `SelectMenu` popover.

## Reported by (verbatim feedback)

- "make the composer options (templates, duration, etc) responsive on smaller width screens like mobile. currently the options overflow"

## Relevant code

- `app/components/copilot/ComposerCopilot.tsx:242` — options row container, `flex items-center justify-between pt-2`
- `app/components/copilot/ComposerCopilot.tsx:243` — pill group, `flex items-center gap-2`, the element that overflows
- `app/components/copilot/ComposerCopilot.tsx:245-300` — AttachMenu, mode / aspect ratio / language / video length / template pills, submit button
- `app/components/copilot/SettingPill.tsx:45` — pill button, `inline-flex ... rounded-full px-2 py-0.5 text-label`, missing `shrink-0`
- `app/components/copilot/ComposerAssets.tsx:44` — the one part of the composer that already wraps (`flex flex-wrap gap-2`), a good precedent
- `app/components/ComposerHero.tsx:37` — page wrapper, `max-w-2xl px-4`

## Suggested direction

Solve it at the row, not per pill. Let the pill group wrap and keep the submit action anchored:

- Add `flex-wrap` + `min-w-0` to the pill group and `shrink-0` to `SettingPill`'s button, so pills move to a second line intact instead of compressing.
[...]

PR #608, opened 2026-08-20
fix(canvas): scope scroll-to-scene to the canvas scroll container

### LINKED ISSUE #601: Selecting a scene at narrow widths scrolls the whole app shell out of the fold (unscoped scrollIntoView), unrecoverable without a refresh
## Problem

At narrow viewport widths (~980px), with Sloppy open and either the timeline or storyboard dock showing, clicking a scene scrolls the canvas *and* shifts the whole app shell upward, pushing part of the UI above the fold. Nothing brings it back — there's no scrollbar to drag it down — so the only recovery is a page refresh.

`scrollToScene` is an unscoped `scrollIntoView`:

```ts
node?.scrollIntoView({ behavior: "smooth", block: "center" });
```

`scrollIntoView` doesn't stop at the nearest scroll container. It walks **every** scroll-container ancestor up to the viewport and aligns each one. And `overflow: hidden` boxes are scroll containers: they can't be scrolled by the user, but they scroll fine programmatically. The shell nests five of them between a scene node and the document — `app/components/PostPromptView.tsx:34,38,42,44,50` — so once any of them has scrollable overflow, `scrollIntoView` scrolls it and it stays scrolled forever. That is exactly the "shifts up and never recovers without a refresh" symptom: the user has no scrollbar on those boxes to undo it.

The reported conditions are what create the overflow in the first place. Sloppy is a `w-64 shrink-0` panel next to a `w-14`/`lg:w-[72px] shrink-0` rail (`app/components/canvas/panel/EditorSidebar.tsx:130-146`), and the side player is `shrink-0` at a resizable width defaulting to 560px (`app/components/video/PlayerPanel.tsx:12,72-80`). At ~980px those fixed-width columns exceed the row, and the timeline/storyboard dock adds a `shrink-0` block at the bottom (`app/components/video/BottomDock.tsx:30-56`). Wide screens have slack, so nothing overflows and the bug is invisible. This is why it only reproduces narrow, with Sloppy open, with a dock view showing.

The scroll container the call actually means is the canvas scroll area, `app/components/PostPromptView.tsx:44-47` (`flex-1 overflow-y-auto`). Everything above it should be left alone.

`SloppyPanel` has the same unscoped call on every message change (`app/components/sloppy/SloppyPanel.tsx:81`), so the transcript auto-scroll can shift the shell the same way. Worth fixing in the same pass.

- "in smaller width screens e.g. 980px, when youre in storyboard view and select a scene that gets scrolled to, the entire canvas shifts up and bleeds out of the fold. doesn't ever recover unless you refresh. this only happens with both sloppy open and either timeline or storyboard view and clicking to scroll on a scene somewhere. likely something to do with the scroll into view logic"

## Steps to reproduce

1. Open a project with enough scenes to make the canvas scrollable.
2. Narrow the window to ~980px.
3. Open the Sloppy panel and switch the bottom dock to storyboard (or timeline).
4. Click a scene far enough down that it has to scroll into view.
5. The whole shell shifts up; the toolbar/top of the app goes above the fold and stays there.

## Relevant code

- `app/components/canvas/utils/scrollToScene.ts:1-7` — the unscoped `scrollIntoView`; the single fix site
- `app/components/PostPromptView.tsx:34,38,42,44,50` — the nested `overflow-hidden` ancestors that get scrolled and can't be scrolled back, plus the `overflow-y-auto` canvas area (`:44-47`) that is the only container the call should touch
- `app/components/canvas/panel/EditorSidebar.tsx:130-146` — `w-14`/`lg:w-[72px]` rail + `w-64` panel, both `shrink-0`: the width Sloppy adds
- `app/components/video/PlayerPanel.tsx:12,72-80` — side player `shrink-0` at a 560px default width
- `app/components/video/BottomDock.tsx:30-56` — the dock block the timeline/storyboard view adds
- `app/components/canvas/elements/SceneContainer.tsx:102,158` — `data-scene-id`, the query target
- Call sites, all of which inherit the fix: `app/components/video/storyboard/Storyboard.tsx:57`, `app/components/video/timeline/Timeline.tsx:129`, `app/components/video/PlayerControls.tsx:57`, `app/components/video/ActiveSceneSync.tsx:38` (auto-scroll during playback — same bug, fires repeatedly while playing)
- `app/components/sloppy/SloppyPanel.tsx:81` — second instance of the same unscoped call

## Suggested direction

Fix it once in `scrollToScene` so all four call sites are covered; don't patch the layout to hide the overflow.

Scope the scroll to a single container instead of letting the browser walk ancestors. Two reasonable options:

1. Resolve the scene node's scroll container (the `overflow-y-auto` canvas area) and scroll *it* directly — `container.scrollTo({ top, behavior: "smooth" })` with `top` computed from the two `getBoundingClientRect`s plus the container's current `scrollTop`. Small, dependency-free, and testable as a pure `scrollTopFor(containerRect, nodeRect, scrollTop, block)` helper.
2. Use `scroll-into-view-if-needed`, which supports a `boundary` and stops walking there. Adds a dependency for something we'd otherwise write in a few lines.

Option 1 fits the repo better. Either way the container should be found from the node (nearest scrollable ancestor) rather than by threading a ref down from `PostPromptView` — the storyboard, timeline, transport bar, and playback sync all call this from outside the canvas tree.

While in there, `SloppyPanel:81` should scroll its own transcript container the same way rather than calling `scrollIntoView` on a sentinel div.

An `overflow-hidden` ancestor silently accumulating `scrollTop` is a whole class of bug, not one instance — worth a test asserting those ancestors stay at `scrollTop === 0` after a scroll-to-scene.

## Acceptance criteria

- [ ] At ~980px with Sloppy open and the storyboard dock showing, clicking a scene scrolls only the canvas area; the shell does not move.
- [ ] Same for the timeline dock, the transport bar's scroll-to-active-scene button, and auto-scroll during playback.
- [ ] After any scroll-to-scene, every `overflow-hidden` ancestor between the scene node and the document still has `scrollTop === 0` and `scrollLeft === 0`.
- [ ] Sloppy's transcript auto-scroll moves only the transcript.
- [ ] Scroll-position math is covered by a unit test on a pure helper (no jsdom layout dependence).
- [ ] Smooth-scroll behavior and the current centering are unchanged on wide screens.

## Related

- #591 (composer options row overflows on narrow screens) — same neighbourhood: fixed-width chrome that has no slack at ~980px

umairnadeem:
## Summary

`scrollToScene` called `scrollIntoView`, which aligns **every** scroll-container ancestor up to the viewport. `overflow: hidden` boxes are scroll containers too: they can't be scrolled by the user, but they scroll fine programmatically. `PostPromptView` nests five of them between a scene node and the document, so at ~980px with Sloppy open and a dock showing, clicking a scene shifted the whole app shell above the fold with no scrollbar to bring it back.

The fix scopes the scroll to a single container:

- New `lib/components/scrollIntoContainer.ts` resolves the nearest ancestor whose computed `overflow-y` is `auto` or `scroll` (so `overflow: hidden` ancestors are never candidates) and calls `scrollTo` on that one element. Falls back to the document scrolling element when nothing above the node is user-scrollable.
- `scrollTopFor(container, node, scrollTop, block)` is a pure function holding the scroll math, clamped at 0.
- `scrollToScene` delegates to it, so all four call sites inherit the fix without changing: `Storyboard`, `Timeline`, `PlayerControls`, and `ActiveSceneSync` (auto-scroll during playback).
- `SloppyPanel`'s transcript auto-scroll had the same unscoped call on every message change; it now scrolls its own transcript container.

No layout changes. Smooth behavior and centering are unchanged on wide screens.

#601 is a `size: S` bug with an obvious single fix site and a suggested direction the issue itself calls out as fitting the repo (scope the container, don't add `scroll-into-view-if-needed`). The change is 4 files, +86/-2, with no visual or product decisions involved.

## Validation

Every step from `.github/workflows/ci.yml`, run locally:

| step | result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:coverage` | pass (142 files) |
| `npm run test:e2e` | pass (3 tests) |

The scroll math is covered by 5 unit tests on `scrollTopFor` with no jsdom layout dependence (the repo has no DOM test environment configured, so container-selection behavior is not asserted in a test; the acceptance criteria only ask for the math).

## Open-PR overlap check

Checked all 11 open PRs with `gh pr diff <n> --name-only`. None touches `scrollToScene.ts`, `SloppyPanel.tsx`, or `lib/components/scrollIntoContainer.ts`. PR #607 imports `scrollToScene` in `Timeline.tsx` and `ActiveSceneSync.tsx` but does not modify it, so both branches merge cleanly and #607's call sites pick up this fix for free.

## Skipped candidates

No `size: xs` issues are open. Every other open issue is `size: M`+ or product-heavy (BYOK, local mode, community feed, i18n, publishing kit) or design-direction work (a11y audit #385, variant picker #451). #591 (composer options row overflow) is the closest sibling but is a visual/responsive-layout decision better made with `DESIGN.md` in hand and a human eye on the result. #587 already has open PR #590.

Closes #601

### DIFF
--- a/app/components/canvas/utils/scrollToScene.ts
+++ b/app/components/canvas/utils/scrollToScene.ts
@@ -1,6 +1,8 @@
+import { scrollIntoContainer } from "@/lib/components/scrollIntoContainer";
+
 export function scrollToScene(sceneId: string) {
 	const node = document.querySelector(
 		`[data-scene-id="${CSS.escape(sceneId)}"]`,
 	);
-	node?.scrollIntoView({ behavior: "smooth", block: "center" });
+	if (node) scrollIntoContainer(node, "center");
 }
--- a/app/components/sloppy/SloppyPanel.tsx
+++ b/app/components/sloppy/SloppyPanel.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import { memo, useEffect, useRef, useState } from "react";
+import { scrollIntoContainer } from "@/lib/components/scrollIntoContainer";
 import { cn } from "@/lib/utils";
 import { trailingAssistant } from "@/lib/agent/messages";
 import type { SloppyMessage } from "@/lib/agent/types";
@@ -78,7 +79,7 @@ export function SloppyPanel() {
 	const endRef = useRef<HTMLDivElement>(null);
 
 	useEffect(() => {
-		endRef.current?.scrollIntoView({ block: "end" });
+		if (endRef.current) scrollIntoContainer(endRef.current, "end", "auto");
 	}, [messages]);
 
 	return (
--- a/lib/components/__tests__/scrollIntoContainer.test.ts
+++ b/lib/components/__tests__/scrollIntoContainer.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from "vitest";
+import { scrollTopFor } from "../scrollIntoContainer";
+
+const extent = (top: number, height: number) => ({ top, height });
+
+describe("scrollTopFor", () => {
+	it("centers a node that sits below the container", () => {
+		expect(scrollTopFor(extent(100, 500), extent(400, 100), 0, "center")).toBe(
+			100,
+		);
+	});
+
+	it("keeps the container's existing scroll offset in the result", () => {
+		expect(
+			scrollTopFor(extent(100, 500), extent(400, 100), 250, "center"),
+		).toBe(350);
+	});
+
+	it("centers a node that sits above the current viewport", () => {
+		expect(
+			scrollTopFor(extent(100, 500), extent(-300, 100), 800, "center"),
+		).toBe(200);
+	});
+
+	it("aligns the node's bottom edge with the container's for block end", () => {
+		expect(scrollTopFor(extent(0, 400), extent(600, 100), 0, "end")).toBe(300);
+	});
+
+	it("never scrolls past the top of the container", () => {
+		expect(scrollTopFor(extent(100, 500), extent(120, 40), 0, "center")).toBe(
+			0,
+		);
+	});
+});
--- a/lib/components/scrollIntoContainer.ts
+++ b/lib/components/scrollIntoContainer.ts
@@ -1,0 +1,47 @@
+export type ScrollBlock = "center" | "end";
+
+type Extent = { top: number; height: number };
+
+export function scrollTopFor(
+	container: Extent,
+	node: Extent,
+	scrollTop: number,
+	block: ScrollBlock,
+): number {
+	const slack =
+		(container.height - node.height) * (block === "center" ? 0.5 : 1);
+	return Math.max(0, node.top - container.top + scrollTop - slack);
+}
+
+function scrollableAncestor(node: Element): Element | null {
+	for (let el = node.parentElement; el; el = el.parentElement) {
+		const { overflowY } = getComputedStyle(el);
+		if (overflowY === "auto" || overflowY === "scroll") return el;
+	}
+	return null;
+}
+
+/**
+ * `scrollIntoView` aligns every scroll-container ancestor, including
+ * `overflow: hidden` ones the user has no way to scroll back.
+ */
+export function scrollIntoContainer(
+	node: Element,
+	block: ScrollBlock,
+	behavior: ScrollBehavior = "smooth",
+) {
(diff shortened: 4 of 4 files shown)


umairnadeem (commented):
lgtm

PR #611, opened 2026-08-21
feat: let Sloppy measure how long each visual is on screen

umairnadeem:
## Summary
- Adds `measure_element_lengths`, an agent tool reporting each visual's seconds on screen and the dialogue ids that decide it, so Sloppy can answer "make each image last 4 seconds" by splitting or merging dialogue instead of guessing. Backed by a new pure module `lib/video/elementLengths.ts`.
- Renames `count_words` to `measure_total_length`, so the two tools read as one question at two anchors: the whole video, or each element.
- Fixes the Runware provider never stamping `durationSec` on a completed job (every generated video resolved to 0s), and makes the behavior that bug produced explicit as `buildVideoLayout`'s `trimVisualsToDialogue` option, default `true`. Ready to become a `VideoSettings` field; nothing turns it off yet outside tests.
- Teaches the system prompt the actual rule, and gives `toolPresentation` a fallback so transcripts written before the rename still render.

## Test plan
- [x] `npm test` — 1283 passing, including new coverage for `measureElementLengths` (dialogue spans across scenes, silent elements ignored, the 1s floor), both trim modes in `scene-builder`, the poll duration stamp, and `toolPresentation`'s renamed-tool fallback
- [ ] In the app: ask Sloppy "how long is each image shown?" and then "make each one about 4 seconds", confirm it calls the tool and edits dialogue rather than guessing
- [ ] Open a conversation saved before this branch and confirm the transcript renders instead of throwing in `<Tool>`

## Notes
- Rendered output is unchanged for Runware projects. Mock projects will differ: `MockVideo` reports real 15s-174s stock clips, which until now became scenes that long and are now trimmed to their dialogue.
- The existing `scene-builder` and `timelineRows` suites are built on fixtures where a visual's own length defines the span, which trimming makes moot; they now go through a local `untrimmed()` helper, with a new dedicated block covering the trimmed default.
- Reported seconds are per-sequence and do not subtract the 0.4s transition overlap that adjacent scenes share.

### DIFF
--- a/app/components/sloppy/__tests__/toolPresentation.test.ts
+++ b/app/components/sloppy/__tests__/toolPresentation.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, it } from "vitest";
+import type { ToolUIPart } from "ai";
+import type { SloppyTools } from "@/lib/agent/types";
+import { toolPresentation } from "../toolPresentation";
+
+const part = (type: string) =>
+	({
+		type,
+		toolCallId: "c1",
+		state: "input-available",
+		input: {},
+	}) as ToolUIPart<SloppyTools>;
+
+describe("toolPresentation", () => {
+	it("names the tool a live call runs", () => {
+		expect(toolPresentation(part("tool-read_script")).label).toBe(
+			"Reading the script",
+		);
+	});
+
+	it("still renders a tool a past build wrote into the transcript", () => {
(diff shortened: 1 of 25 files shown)


umairnadeem (commented):
lgtm, the rename + trim default look clean and the fallback covers old transcripts

PR #614, opened 2026-08-21
fix: nightly bug fixes 2026-08-21

umairnadeem:
## Bug

A saved script stores scenes as `--- Scene N ---` marker lines between serialized elements. `splitScenes` split the whole document on that marker with a `/m` regex, so a marker-looking line **inside an element's text** was treated as a scene boundary.

Element text can legitimately span lines (the LLM streams multi-line narration, and pasted scripts keep their line breaks). When such a line reads as a marker, reloading the project cut the element in half: its text was truncated at the marker and the rest of the scene came back as a separate scene with a dangling `</narration>`. Silent document corruption on the persistence path.

## Fix

Split a line at a time, tracking whether the previous line left an element tag open. Only markers that sit between elements separate scenes. `SCENE_MARKER_PATTERN` loses its `/m` flag since it is now matched against one line.

No format change: what gets written is byte-for-byte what it was, and existing scripts split exactly as before unless they were already being corrupted.

## Tests

- `splitScenes` keeps a marker line inside element text with its element
- `splitScenes` still splits on the marker that follows a multi-line element
- `deserializeWithScenes` round-trips text holding a line that reads as a scene marker

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test:coverage` all pass.

## Open PR overlap check

Considered #602, #603, #604, #605, #606, #607, #608. None touches `lib/project/serialize.ts` or `lib/canvas/osmlSerializer.ts`; #602 touches `lib/video/*` and `app/components/sloppy/*`, #605 `lib/canvas/editorOps.ts`, #608 `app/components/canvas/utils/*` and `lib/components/scrollIntoContainer.ts`. This PR is non-overlapping.

### DIFF
--- a/lib/canvas/osmlSerializer.ts
+++ b/lib/canvas/osmlSerializer.ts
@@ -4,7 +4,8 @@ import { getContentElements, isSceneElement } from "@/lib/canvas/scenes";
 import { withoutCaretMarker } from "./constants";
 import { escapeXml } from "./xmlEscape";
 
-export const SCENE_MARKER_PATTERN = /^---\s*Scene\s+\d+\s*---\s*$/m;
+/** Matched a line at a time, so a marker inside element text can be told apart. */
+export const SCENE_MARKER_PATTERN = /^---\s*Scene\s+\d+\s*---\s*$/;
 const sceneMarker = (n: number) => `\n--- Scene ${n} ---\n`;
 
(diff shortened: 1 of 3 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
overengineered

PR #615, opened 2026-08-21
refactor: nightly maintainability 2026-08-21

umairnadeem:
## What

Static connector plugin chains were declared imperatively in `ConfigProvider`, in four divergent styles, even though none of them has a runtime dependency. They now live in `DEFAULT_CONNECTOR_REGISTRY`, which already owns every other piece of a connector's static config.

- `image` → `buildImagePlugins()`, `animated_image` → `buildAnimatedImagePlugins()`, `video` → reference-images + dimensions, `tts` → metadata-voice + voice-search.
- `ConfigProvider` keeps only the one chain that needs a live project store (`createVoiceHydratePlugin(store)`).
- `DEFAULT_PROVIDER` moves from `registry.ts` to `types.ts` next to `ProviderKey`. It was a *value* import in `still-frame.ts`, which sits in the transitive closure of the chain builders, so leaving it in the registry would have made registry-imports-chains a real runtime cycle.
- `still-frame.test.ts` no longer hand-rolls the same `withRegistry(...).appendPlugins(...)` wiring in three places; it uses the registry directly.
- One line in ARCHITECTURE.md saying where chains are declared.

Append order per type is unchanged, so the resolved plugin list for every connector is byte-identical. No behavior change.

## Complexity delta

7 files, +36/−49. Three duplicated test fixtures collapsed; the plugin-composition surface in the React provider goes from 6 appends over 7 imports to 1 append over 1 import.

## Skipped

- `lib/supabase/env.ts`'s `""` env defaults (soft fail-loudly violation): making it throw is a behavior change, which this runbook forbids. Still open for a human call.
- `ScrubBar.tsx` has no `role="slider"`/keyboard support — belongs to the bugs/a11y nightly.
- Moving the `ConnectorRegistry` *type* into `types.ts` would clear the three new `import type`-only madge cycles (registry → chains → `characterAvatarNode` → registry, all runtime-erased, matching the four already on master), but it touches ~20 import sites including `lib/project/serialize.ts`, which PR #614 owns.

## Open PR overlap check

Considered #614, #608, #607, #606, #605, #604, #603, #602. Their footprint covers `lib/canvas/{osmlSerializer,editorOps}`, `lib/project/serialize`, `app/components/video/**`, `app/components/sloppy/**`, `lib/agent/tools/{registry,outlineStory}`, `lib/providers/llm/mock`, `lib/script/prompt/outline`, `lib/api/{process-job,queue-callback}`, `lib/video/elementAttributes`, `Editor.tsx`, `Waveform.tsx`, `CharacterEditModal`, `AudioPlayer`, `PreviewCacheContext`, `scrollToScene`/`scrollIntoContainer`. This PR touches none of them.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -38,7 +38,7 @@ Turns are stored as the SDK's UI message type, so the stream, the rows and the p
 
 The generation pipeline is split so providers and asset types can be swapped independently:
 
-- **Connectors** (`lib/connectors/`): what the editor calls. Model-agnostic, plugin-pipelined, return an `AssetResult`.
+- **Connectors** (`lib/connectors/`): what the editor calls. Model-agnostic, plugin-pipelined, return an `AssetResult`. `DEFAULT_CONNECTOR_REGISTRY` declares each type's static plugin chain; `ConfigProvider` appends only the chains that need a live project store.
 - **Gateways** (`lib/gateway/`): thin HTTP clients between connectors and our `/api/v1/*` routes. This seam lets connectors run against the live API or a mock. Today there is one OpenSlop gateway. A BYOK gateway will be added so users can call providers with their own API keys.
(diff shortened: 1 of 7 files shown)


umairnadeem (commented):
lgtm

PR #616, opened 2026-08-21
perf: nightly performance optimizations 2026-08-21

umairnadeem:
## What

The editor route was shipping 160KB of LLM few-shot prompt text to every user on first load.

`Template` carried `systemPrompt` and `exampleText` inline. The catalog is client data — `TemplateGallery`, `ComposerCopilot` and `applyTemplate` all read it — so those two fields landed in the `/projects/[id]` first-load bundle. `exampleText` alone is 160,258 characters across the 7 templates, 92% of `templates.ts`, and the only thing that ever reads it is `buildScriptPrompt`, on the generate path.

`lib/templates/templatePrompts.ts` now holds those two fields, keyed by template id, and `promptParts` loads it with a dynamic import. The catalog keeps everything the UI renders. The prompt payload arrives when the user generates a script from a template, behind a request that then waits seconds on an LLM stream, so the fetch is fully hidden.

`buildScriptPrompt` and `promptParts` became async; the one runtime caller (`ScriptProvider.runScript`) was already async.

## Measured

`npm run build`, first-load JS from `.next/diagnostics/route-bundle-stats.json`:

| route | raw before | raw after | gzip before | gzip after |
|---|---|---|---|---|
| `/projects/[id]` | 2,310,060 | 2,145,983 (-7.1%) | 671,155 | 620,189 (-7.6%) |

Every other route is byte-identical.

## No behavior change

The fields moved verbatim, `dedent` included. Checked with a throwaway test asserting each `getTemplatePrompt(id)` equals the old inline `systemPrompt`/`exampleText` string for all 7 templates — exact match, then deleted. `getTemplatePrompt` throws on an unknown id, matching `getTemplate`, and a new test keeps the catalog and the prompt table in lockstep now that the type no longer forces it.

## Checks

`npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run knip`, `npm run build`, `npm test` (145 files, 1288 tests) all green.

## Skipped

- Slate render path (`Canvas.tsx`, `SortableContent`, `withScenes` normalization) already follows the [performance walkthrough](https://docs.slatejs.org/walkthroughs/09-performance): file-level `renderElement`, `useSlateStatic` everywhere a component doesn't need the value, `useSlateSelector` on the two selectors that do, normalization rules bounded to `path.length === 1`. Nothing left worth the readability cost.
- `MotionLayer` subscribing to `useCurrentFrame()` when `effect === "none"` — its `children` element identity is stable, so React bails out of the subtree anyway. A guard here would be the kind of single-path rerender suppression that isn't worth a PR.
- `isNodeStale` re-walking each element's dependency graph on every queue notification. Real, but it's a shallow graph over a handful of elements once a second while generating, and the fix would be a memo shim of the sort that got closed on #577/#585.
- Moving script prompt assembly server-side. It would take the payload off the client entirely and shrink the request, but it changes the `/api/v1/llm` contract, which is out of scope for a performance-only pass.

## Open PR overlap check

Considered #602, #603, #604, #605, #606, #607, #608, #614, #615. This PR touches `lib/templates/**`, `lib/script/prompt/build.ts`, `lib/script/prompt/template.ts` and `lib/script/ScriptProvider.tsx`; none appear in any open PR's diff. The nearest neighbour is #606, which edits `lib/script/prompt/outline.ts` — a different file, and this change doesn't read it.


umairnadeem (commented):
lgtm

umairnadeem:
overengineered. this will be in a db eventually anyways

PR #618, opened 2026-08-21
feat: seed blank projects with a welcome narration

umairnadeem:
## What

Blank projects now open with a narration element that greets the user instead of an empty one:

> Welcome to OpenSlop! Add an element, or ask Sloppy to plan or write a story for you

`startBlank` also sets the project title to "Untitled" and feeds the blank script through `appendChunk`, so it reaches the canvas by the same OSML stream path a generated script does rather than a separate `setScript` seed.

## Testing

lint, format:check, typecheck, knip, build, and the full vitest suite (1286 tests) pass locally.

https://claude.ai/code/session_01ArqvtxiVArrPSyggQcQEaL

### DIFF
--- a/lib/project/serialize.ts
+++ b/lib/project/serialize.ts
@@ -6,7 +6,8 @@ import { makeNodeId } from "@/lib/canvas/nodeUtils";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 
 /** The document a project starts from when there is no generated script. */
-export const BLANK_SCRIPT = "<narration></narration>";
+export const BLANK_SCRIPT =
+	"<narration>Welcome to OpenSlop! Add an element, or ask Sloppy to plan or write a story for you</narration>";
 
 export function splitScenes(osml: string): string[] {
 	if (!osml.trim()) return [];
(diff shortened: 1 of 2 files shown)


lib/project/serialize.ts
```diff
@@ -6,7 +6,8 @@ import { makeNodeId } from "@/lib/canvas/nodeUtils";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 
 /** The document a project starts from when there is no generated script. */
-export const BLANK_SCRIPT = "<narration></narration>";
+export const BLANK_SCRIPT =
+	"<narration>Welcome to OpenSlop! Add an element, or ask Sloppy to plan or write a story for you</narration>";
```
umairnadeem, line 10: this welcome copy is a real narration body now, so it's indistinguishable from user content downstream: open a blank project, hit Generate all, and the TTS voices "Welcome to OpenSlop!..." into the video (empty narration used to get filtered out by the `inputs.prompt` check in useGenerateAll). is the plan that people overwrite it first? if so maybe carry it as placeholder/empty-state text on the element rather than as its text.

PR #617, opened 2026-08-21
refactor: drop write-only lastPrompt from project metadata

umairnadeem:
## Summary

`Metadata.lastPrompt` persisted the text of the most recent script prompt into every project's metadata, and nothing read it. Its own comment said so. The `messages` table already stores every user turn verbatim, keyed by conversation and ordered by `seq`, so prompt activity is durable there without duplicating user text into project metadata.

- Removed `lastPrompt` from `MetadataSchema` (`lib/project/types.ts`).
- Removed the `updateMetadata({ lastPrompt: ... })` write from `ScriptProvider.runScript`, along with the now-unused `useProject` subscription and callback dep.
- Removed `sourceText` from `lib/script/prompt/build.ts`; that write was its only caller, and knip flags dead exports.

Existing projects that already persisted a `lastPrompt` key still load: zod strips unknown keys rather than erroring, and the stale key drops out on the next metadata write.

## Why this issue was safe and small

Issue #609 is labeled `size: XS` with explicit acceptance criteria and named the exact three locations. The field is write-only, so removing it has no runtime consumer to break. Net diff is 14 deletions across 3 files, no behavior change.

## Validation

All CI-equivalent checks from `.github/workflows/ci.yml` were run locally and pass:

- `npm run lint` — clean
- `npm run format:check` — clean
- `npm run typecheck` — clean
- `npm run knip` — clean (confirms `sourceText` had no other consumer)
- `npm run build` — succeeds
- `npm run test:coverage` — 144 files, 1286 tests passed
- `npm run test:e2e` — 3 smoke tests passed

`grep -rn "lastPrompt\|sourceText"` over the repo returns nothing.

## Open PR overlap check

Reviewed all 10 open PRs (#602-#608, #614-#616) by changed files.

- #616 touches `lib/script/ScriptProvider.tsx` and `lib/script/prompt/build.ts`, but for an unrelated change (making `buildScriptPrompt` async and lazy-loading template prompts). Its edits are at different lines from these deletions, no textual conflict and no duplicated work. Flagging it here since it's the one file-level intersection.
- No other open PR touches `lib/project/types.ts` or the metadata schema.

## Skipped candidates

- #601 (unscoped `scrollIntoView`) — already implemented by open PR #608.
- #591 (composer options row overflow) — claimed by an outside contributor, and the maintainer said they'd hold off; not for an unattended run.
- Everything else open is `size: M` or larger, or turns on product/UX direction.

Closes #609

[...]

PR #619, opened 2026-08-22
fix: nightly security audit 2026-08-22

umairnadeem:
## Fix: API error responses leaked serialized exceptions

`runGuarded` in `lib/api/with-auth.ts` is the single error envelope for every API route. On a thrown error it returned:

```ts
serverError(`${label} failed: ${stringifyError(error)}`)
```

`stringifyError` is `JSON.stringify(serializeError(error))`, so the 500 body carried the full error object: message, **stack frames with absolute server file paths**, `cause` chains, and whatever upstream detail Supabase or a provider SDK attached. That body is returned to the caller on every route, including the unauthenticated public ones (`/api/validate-code`). Any user who can make a route throw (an upstream provider failure, a DB error) reads back internal server structure — CWE-209.

The fix returns `${label} failed` and nothing else. The error is still logged in full via `logger.error(error, ...)`, so nothing is swallowed and the failure is still loud.

### Audited, no change needed
- **Dependencies** — `npm audit`: one moderate, `uuid@9` via `@runware/sdk-js` (GHSA-w5hq-g745-h8pq). Not exploitable here: the bug is a missing bounds check on the caller-supplied `buf` argument to `v3/v5/v6`, which nothing in this tree reaches. No fix published upstream; forcing `uuid@11` on a dep declaring `^9.0.1` is a larger break than the risk. No open Dependabot alerts.
- **Secrets** — no hardcoded keys; only `.env.example` is tracked.
- **SSRF** — `/api/v1/tts/voices/preview` is already origin-allowlisted (`fetchAllowedVoicePreview`, HTTPS + host + port, `redirect: "manual"`). `referenceImages` URLs go to the model provider, not to a server-side fetch.
- **Supabase/RLS** — every table has RLS with `auth.uid() = user_id` policies; access codes are only reachable through the security-definer RPC. Service-role client is confined to queue processing and the dev-only impersonate route (404s in production).
- **Auth** — `withApiAccess` gates on `app_metadata.api_access` (not user-writable); job polling and conversations are scoped by `user_id`.
- **Injection/XSS** — no `dangerouslySetInnerHTML`, `eval`, or `new Function` in app code. Uploads sanitize filenames and land on a separate blob origin.
- **Headers** — nosniff / SAMEORIGIN / HSTS / Referrer-Policy already set in `next.config.ts`.

### Noted, not fixed here
`jobs.error` stores `stringifyError(...)` and `pollJob` returns it to the job owner — the same disclosure class, one tier lower (authenticated owner only). `lib/api/process-job.ts` is owned by open PR #603, so it is left alone per the overlap guard.

### Open PR overlap check
Considered #602, #603, #604, #605, #606, #607, #608, #614, #615, #616, #617. This PR touches `lib/api/with-auth.ts`, `lib/api/__tests__/route-handler.test.ts`, and `app/api/render/progress/__tests__/progress.test.ts` — none appear in any open PR's file list. #603 is the nearest (also `lib/api/`) but is confined to the queue callback path. Non-overlapping.

### Validation
`lint`, `format:check`, `typecheck`, `knip`, `build`, `test` all pass (1286 tests).

### DIFF
--- a/app/api/render/progress/__tests__/progress.test.ts
+++ b/app/api/render/progress/__tests__/progress.test.ts
@@ -106,7 +106,7 @@ describe("POST /api/render/progress", () => {
 		const res = await POST(makeRequest(validBody));
 
 		expect(res.status).toBe(500);
-		expect((await res.json()).error).toContain("output file");
+		expect((await res.json()).error).toBe("render/progress failed");
 	});
 
 	it("reports in-flight progress otherwise", async () => {
--- a/lib/api/__tests__/route-handler.test.ts
+++ b/lib/api/__tests__/route-handler.test.ts
@@ -8,6 +8,7 @@ import {
 	createSessionFormRouteHandler,
 	createSessionRouteHandler,
 } from "../route-handler";
+import { logger } from "../logger";
 
 vi.mock("../logger", () => ({
 	logger: { warn: vi.fn(), error: vi.fn() },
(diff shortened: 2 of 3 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
useless

PR #620, opened 2026-08-22
refactor: nightly maintainability 2026-08-22

umairnadeem:
## What

The project's asset dialogs (new character, character edit, narrator, art style) were a hook that returned four imperative openers plus a `dialogs` JSX blob. Both views that used it then threaded those openers down through layers that only forward them:

- `ComposerCopilot` → `AttachMenu` (3 props) and → `ComposerAssets` (3 props, none of which `ComposerAssets` uses) → the tiles that actually call them.
- `AssetsSection` → four tiles, one opener each.

`useAssetEditDialogs` is now `AssetEditProvider`: it mounts the dialogs once and publishes the openers on context, so each tile wires itself. `ArtStyleAssetTile`, `NarratorAssetTile` and `CharacterAssetTiles` lost their `onEdit` props entirely, and `AddCharacterTile` joins them beside the other self-wired tiles.

Three smaller fixes, all objective:

- `osmlStreamParser.appendNext` used `CANVAS_ELEMENT_TYPES.has(type as CanvasElementType)` and then cast again to build the node. `lib/canvas/guards.ts` already exports `isCanvasElementType`, which narrows; both casts are gone.
- `RunwareVideo._generate` did `jobId: job.metadata?.jobId ?? ""`. `toVideoJob` always sets the id, so the `""` was an unreachable silent-degradation path that the video job handler would later report as a generic failure. `toVideoJob` now returns a type that carries `metadata`, so the fallback is impossible rather than swallowed.
- The README's project-structure map was missing `lib/agent/` (Sloppy, the whole agent layer), `lib/templates/` and `lib/upload/`.

No behavior changes. Radix dialogs portal to `body`, so moving them from inside each view's wrapper to inside the provider renders identically.

## Complexity delta

15 props removed across 7 component signatures; two views stop forwarding state they don't read; two unchecked type casts and one silent `""` fallback removed. Net -10 lines.

## Open PR overlap check

Considered #602, #603, #604, #605, #606, #607, #608, #614, #615, #616, #617, #619. None touches any file here. #604 touches `character/CharacterEditModal.tsx`, which this PR only imports and leaves unchanged.

## Skipped, with reasons

- **Sloppy tool presentation isn't co-located** (icon + label live in a 93-line switch in `toolPresentation.tsx`, not on `defineTool`). Needs a `TOOLS` export from `lib/agent/tools/registry.ts`, which #604 is rewriting.
- **Two ways to reach the Slate editor in the canvas tree** (deep components use `useSlateStatic()`, the shell prop-drills `editor`). Must land as one sweep; two of its files (`Editor.tsx`, `SloppyProvider.tsx`) are held by #604 and #602.
- **`app/projects/[id]/loading.tsx` has drifted from the shell it mirrors** — it still draws the refine bar #580 removed, and its rail is short two entries. That's a visible change, so it belongs to the bugs/UI nightly.
- **`lib/supabase/env.ts` `""` defaults** — fixing the soft fail-loudly violation changes startup behavior. Wants a human call.
- **Transition labels live in `PropertiesPanel.tsx`** while sibling choice sets keep labels in `lib/`. Only the second occurrence; below the rule of three.

## Checks

`npm run lint`, `format:check`, `typecheck`, `knip`, `npm run build`, and `npm run test:run` (144 files, 1286 tests) all pass.


umairnadeem (commented):
lgtm

PR #622, opened 2026-08-24
feat: keep every take an element has produced

umairnadeem:
## Summary

- Per-element generation history: each card lists the takes it has produced with the prompt and settings behind each, and restoring one writes those inputs back into the document so the card, player, and timeline follow.
- Takes persist to a new `element_history` table on commit rather than through the debounced project autosave, so a take survives closing the tab mid-debounce and the saved document can never point at a take the database lacks.
- A take is keyed by the inputs that made it, so regenerating an unchanged element overwrites its take instead of adding an indistinguishable one. That also derives the active row from `resultInputs`, so no version id rides in the saved generation snapshot.
- Element attributes split into generation and layout buckets, applied once at the parse and write boundaries instead of re-derived by each consumer; restoring a take leaves `loops`, `volume`, and `motion` alone.
- Generation and history are separate modules: the queue announces finished takes and knows nothing about storage; history reads back only when a card opens it.
- Migration 005 creates the table with the update policy the overwrite needs; the redundant `with check` clauses in 002 and 004 are dropped since Postgres reuses `using` for the new-row check on update policies.

## Test plan

- [x] `npm run lint`, `tsc --noEmit`, `prettier --check`, `npm run build`
- [x] `npx vitest run` — 1331 passing
- [x] Manual: open a project, open an element's history, restore an older take — prompt, card preview, Remotion player, and timeline thumbnail all follow
- [x] Manual: regenerate with unchanged inputs — the existing take is overwritten in place (same position and date), no third entry appears
- [x] Manual: hard reload — history refetches the same takes, active row still derived correctly
- [ ] Apply migration 005 on a fresh database and confirm regenerate-then-reload persists the overwrite (the update policy is new and unexercised)

## Notes

- A failed history write toasts and is not retried, so the take stays on screen for the session but is absent after reload.
- Restoring a take older than its dependencies still reads as stale — that is the existing staleness rule (`nodeIdentity` folds in dependency results), not a gap in restore.


app/components/canvas/elements/character/CharacterEditModal.tsx
```diff
@@ -79,6 +84,11 @@ function CharacterEditDialogBody({
 		() => buildNode(forCharacterAvatar(name)),
 		[buildNode, name],
 	);
+	const restoreAppearance = useCallback(
+		(version: ElementVersion) =>
+			updateCharacter(name, characterFromAvatarInputs(version)),
```
umairnadeem, line 89: restore in this modal only rewrites the appearance text, the avatar image never comes back: the picture here is `avatarSnapshot.result` and only `useGenerate` ever calls `queue.restoreResult`, which doesn't run for the derived avatar node. so you pick an old take, see its thumbnail, and the card keeps showing the newest image while going stale (and `avatarUploaded: version.pinned` can now claim uploaded over a generated image). worth calling `queue.restoreResult(avatarElementId, version.inputs, version.result)` alongside the metadata update?

lib/generation/versions.ts
```diff
@@ -0,0 +1,69 @@
...
+	};
+
+	/** A replacement keeps the original id and date: the same take, remade. */
+	record(take: CommittedTake, createdAt: string): ElementVersion {
+		const key = keyOf(take);
+		const byInputs = keyBy(this.get(take.elementId), keyOf);
+		const previous: ElementVersion | undefined = byInputs[key];
+		const version: ElementVersion = {
+			...take,
+			id: previous?.id ?? crypto.randomUUID(),
```
umairnadeem: the overwrite only holds while the log is hydrated. reload, don't open history, hit Generate on unchanged inputs → `byInputs` is empty, so a fresh uuid is minted and the upsert inserts a second row with identical inputs (hydrate dedupes it on read, so the dupe is silent and the take's date/position jumps). a unique index on (project_id, element_id, inputs) with `onConflict` on that would make it idempotent regardless of what's in memory.

lib/generation/queue.ts
```diff
@@ -158,8 +175,15 @@ export class GenerationQueue {
...
-		this.snapshots.commit(elementId, result, inputs, connectorType, pinned);
+		const take = this.snapshots.commit(
+			elementId,
+			result,
+			inputs,
+			connectorType,
+			pinned,
+		);
 		this.snapshots.notify();
+		for (const listener of this.commitListeners) listener(take);
```
umairnadeem: an upload commits with `nodeInputs(node, this)`, ie the element's unchanged inputs, so `record` keys it the same as the take that generated it and replaces it in place (same id, so the upsert clobbers the row too). generate an image, then upload a replacement without touching the prompt → the generated take is gone from history for good, which is the one thing this PR promises not to do. `pinned` probably wants to be part of the take key, or the upload path should mint its own take.

app/components/canvas/elements/character/CharacterEditModal.tsx
umairnadeem: Fixed in 3e9dce1, though not in the modal.

The result is now restored in `VersionHistoryPopover` itself, which both callers share, so `onRestore` narrows to "write these inputs back where they were authored" — the document for an element card, `updateCharacter` for the avatar. Putting `restoreResult` in the modal would have left the same gap open for the next caller, and this bug was a caller forgetting exactly that step.

It also covers a case the modal-only fix would miss: a pinned element is never stale, so the `useGenerate` staleness path would never have restored it either.

`avatarUploaded: version.pinned` is consistent now that the image follows the version.

lib/generation/versions.ts
umairnadeem: Confirmed and fixed in 3e9dce1, though not with the unique index.

Two things blocked that route: a btree unique index over `inputs` blows the ~2704-byte entry limit once a prompt gets long, and PostgREST's `on_conflict` takes column names, so an expression index over a digest is not reachable from supabase-js.

Same root cause though — identity living in memory — so the row id is now derived from the version instead: a v5 uuid over `(projectId, elementId, versionKey)`. The same version always lands on the same row no matter what the client has read back, with plain `onConflict: "id"`. `projectId` is in there because avatar element ids (`avatar:<name>`) genuinely collide across projects.

`created_at` also came out of the write payload, so a remake keeps its original date instead of jumping the list — the other half of what you flagged.

lib/generation/queue.ts
umairnadeem: Right, and fixed in 3e9dce1 by the first option — `pinned` is part of the version key now (`versionKey` = `supplied:`/`generated:` + serialized inputs), so an upload files its own version and the generated one survives.

Two knock-ons that opens, closed with it:

- `restoreResult` takes the whole version and carries its `pinned`, so restoring a generated version unpins the element. Without that it stayed permanently unregeneratable, since `needsGeneration` returns false when pinned. The old test asserting `restoreResult` leaves `pinned` alone encoded the pre-fix model, so it is now the two directions of the new one.
- `useElementHistory` compares full version keys, so an upload and the generation sharing its inputs do not both highlight as active.

app/components/canvas/elements/DeleteButton.tsx
```diff
@@ -1,18 +1,22 @@
...
-			ariaLabel="Delete element"
-			onMouseDown={(e) => {
-				e.preventDefault();
-				removeElement(editor, element);
-			}}
-		/>
+		<SimpleTooltip label="Delete">
+			<HeaderIconButton
+				ariaLabel="Delete element"
+				className="text-destructive hover:bg-destructive/10 hover:text-destructive dark:hover:bg-destructive/20"
```
umairnadeem: this re-hand-rolls the destructive styling that `components/ui/delete-button.tsx` already owns (`text-destructive hover:bg-destructive/10 dark:hover:bg-destructive/20`), and `SceneDeleteButton` in this same PR still uses the shared one, so "what a delete button looks like" now has two homes. CONVENTIONS.md § Frontend: *"The design system is off the shelf and non-negotiable. Components ship it by default, with no one-off Tailwind in higher-level components."* if the header row needs a quieter h-6 variant, that wants to be a size/variant on `ui/delete-button` rather than a copy. same for `DuplicateButton`.

umairnadeem (commented):
lgtm, restore pulling the still along with the take is the right call. thanks for the revisions!

PR #624, opened 2026-08-24
fix(autosave): skip a save whose payload has not changed

### LINKED ISSUE #623: Autosave fires on project open with no user changes
## Problem

Opening a project triggers a full autosave round-trip roughly 2s later, even when the user never touches anything. Every open writes a new row state and pops the "Saved" toast, so the user is told their untouched project was saved.

The autosaver subscribes to every project-store emission and schedules a debounced save with no dirty check:

- `applyStoreSnapshot` hydrates the store on open by calling `updateMetadata`, `setReferenceImages`, and `markHydrated` (`lib/project/storeSnapshot.ts:36-38`) — three store emissions before the user does anything.
- `useMetadataSync` writes `collectWritableMetadata(nodes, metadata)` back into the store on mount (`app/components/canvas/hooks/useMetadataSync.ts:12-15`), another emission with identical content.
- `useAutosave` schedules a save on any store emission (`app/components/canvas/hooks/useAutosave.ts:43`), and `createAutosaver` never compares the pending payload against what was loaded (`lib/project/autosave.ts:84-87`).

- "prevent autosave immediately upon opening a project despite making no changes"

## Relevant code

- `lib/project/autosave.ts:55` — `createAutosaver`; `schedule` debounces then saves unconditionally
- `lib/project/storeSnapshot.ts:30` — `applyStoreSnapshot` emits three store updates during hydration
- `app/components/canvas/hooks/useAutosave.ts:43` — subscribes autosave to every store emission
- `app/components/canvas/hooks/useMetadataSync.ts:12` — writes metadata back on mount
- `lib/project/__tests__/autosave.test.ts` — where the no-op case should be covered

## Suggested direction

Make the autosaver skip a save whose payload matches the last known-saved payload, rather than special-casing hydration at each call site. In `createAutosaver`, keep the serialized `SaveProjectInput` of the last successful save (seeded from the hydrated state on first run) and return early in `save` when the newly built input is deeply equal. That kills the class of problem — hydration echoes, idempotent metadata sync, and any future emit-without-change path — instead of just the open case.

## Acceptance criteria

- [ ] Opening a project and making no changes issues no `saveProject` request and shows no "Saved" toast
- [ ] The first real edit after open still saves normally
- [ ] Generation-result saves still fire (result-version path in `useAutosave`)
- [ ] Test in `lib/project/__tests__/autosave.test.ts` covers "hydrate then schedule with unchanged state → no save"

---

dchaudhari7177:
Closes #623

Opening a project fired a full autosave ~2s later with no user input, writing a new row and popping the "Saved" toast on an untouched project.

Took the suggested direction — dedupe in `createAutosaver` rather than special-casing hydration at each call site, so hydration echoes, the idempotent metadata sync, and any future emit-without-change path are all covered at once. `createAutosaver` keeps the last saved `SaveProjectInput` and returns early when the newly built one is deeply equal (`lodash/isEqual`, alongside the `lodash/debounce` already imported).

### Seeding the baseline

The one design question was where the initial baseline comes from — it has to be the as-loaded state, or the first save after open isn't recognised as a no-op.

It can be captured eagerly at `createAutosaver` time, because `ProjectEditor` hydrates the store inside a `useState` initializer *before* children render (`app/projects/[id]/ProjectEditor.tsx:33-37`). By the time `useAutosave`'s `useMemo` runs, the store is already hydrated and the editor exists. So no subscription and no new lifecycle hook are needed.

When the store *isn't* hydrated at construction, `lastSaved` stays `null` and the first save runs unconditionally. That direction is deliberate: an unknown baseline should cost a redundant write, never a dropped edit.

### Three existing tests needed real edits

`coalesces a burst of changes into one save`, `flush runs a pending save immediately`, and `reports a failed save instead of throwing` all called `schedule()` without changing anything, so after this fix they were correctly skipped. They now make an actual store change first, via a small `edit()` helper that appends a reference image — which also matches what "a burst of changes" claims to be testing. `edit()` touches `referenceImages` rather than the title so the coalescing test's `name: "Moon Rabbit"` assertion still means something.

### New coverage

- **hydrate → schedule with unchanged state → no save** (the issue's requested case, using the real echo: metadata written back with identical content)
- **the first real edit after an unchanged open still saves**
- **a repeat of a payload just saved is skipped** (idempotent write after a genuine one)
- **a generation-result change still saves**, with the store untouched — this is the acceptance criterion most at risk from a store-only dirty check, and it passes because `generation` is part of the compared payload

Verified the new tests actually catch the bug — removing just the `isEqual` early-return line:

```
× does not save when the hydrated state is unchanged
× saves the first real edit after an unchanged open
× skips a repeat of a payload it just saved
3 failed | 8 passed
```

Note the generation test stays green there, as it should — it is guarding against over-correction, not the bug.

Full checks: `vitest run` **1313 passed (147 files)**, `tsc --noEmit` clean, `eslint` clean, `prettier` applied.

### DIFF
--- a/lib/project/autosave.ts
+++ b/lib/project/autosave.ts
@@ -32,8 +32,6 @@ export function buildProjectSave(
 export interface AutosaverOptions {
 	projectId: string;
 	store: ProjectStore;
-	/** The script as loaded, ie the one the server already holds. */
-	initialScript: string;
 	/**
 	 * Produces the script for the next save. Called only when a save runs, so
 	 * serializing stays off the per-keystroke path.
@@ -49,20 +47,21 @@ export interface Autosaver {
 	schedule: () => void;
 	/** Run any pending save immediately. */
 	flush: () => void;
+	/** Treat the current state as the one the server already holds. */
+	markSaved: () => void;
 }
 
 /**
  * Debounces edits into one save at a time. The queue keeps a slow save from
  * overlapping the next one, so the last scheduled state always lands last.
  *
- * A save whose payload the server already holds is dropped. Restoring the
- * loaded document into the empty editor is a Slate change like any other, so
- * without this an untouched project saves itself on open and reports "Saved".
+ * A save whose payload matches the last one is dropped. Restoring the loaded
+ * document into the empty editor is a Slate change like any other, so without
+ * this an untouched project saves itself on open and reports "Saved".
  */
 export function createAutosaver({
 	projectId,
 	store,
-	initialScript,
 	getScript,
 	getGeneration,
 	onSaved,
@@ -70,13 +69,11 @@ export function createAutosaver({
 }: AutosaverOptions): Autosaver {
 	const queue = new PQueue({ concurrency: 1 });
 
-	// The script is taken as loaded rather than from `getScript()`: Slate is
-	// rehydrated in an effect, so the editor is still empty while this runs.
-	let lastSaved = buildProjectSave(
-		extractStoreSnapshot(store),
-		initialScript,
-		getGeneration(),
-	);
+	const buildInput = (): SaveProjectInput =>
+		buildProjectSave(extractStoreSnapshot(store), getScript(), getGeneration());
+
+	/** Null until the loaded state is known: an unknown baseline has to save. */
+	let lastSaved: SaveProjectInput | null = null;
 
 	const save = async () => {
 		if (!store.getState().hydrated) {
@@ -84,11 +81,7 @@ export function createAutosaver({
 			return;
 		}
 		try {
-			const input = buildProjectSave(
-				extractStoreSnapshot(store),
-				getScript(),
-				getGeneration(),
-			);
+			const input = buildInput();
 			if (isEqual(input, lastSaved)) return;
 			await saveProject(projectId, input);
 			lastSaved = input;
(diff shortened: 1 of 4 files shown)


lib/project/autosave.ts
```diff
@@ -62,18 +68,32 @@ export function createAutosaver({
...
+
+	/**
+	 * Payload of the last save known to be on the server. Seeded from the loaded
+	 * state so the hydration echo has something to match; ProjectEditor hydrates
+	 * the store before the editor renders, so that state is available here.
+	 *
+	 * Left null while the store is unhydrated: an unknown baseline must save
+	 * rather than skip, so the worst case is a redundant write, never a lost edit.
+	 */
+	let lastSaved: SaveProjectInput | null = store.getState().hydrated
```
umairnadeem: afaict the seed calls `getScript()` while the slate editor is still empty: `useProjectRehydrate` fills `editor.children` in an effect, which runs after `useAutosave`'s `useMemo`. so on any non-empty project the baseline is `script: ""`, the open-time save doesn't match it, and #623 still fires. the tests don't catch it because `getScript` is a constant there. seeding the baseline script from the as-loaded `initialScript` (rather than the editor) would line it up.

umairnadeem (changes requested):

PR #626, opened 2026-08-24
fix: nightly security audit 2026-08-24

umairnadeem:
## What

A caller with only a browser session (no `api_access`) could reach a paid connector with parameters that never passed a route schema.

Two things lined up:

- `jobs_insert_own` let any `authenticated` role write a `jobs` row directly through the Supabase REST API, with an arbitrary `connector_type` and an arbitrary `request` jsonb. The policy only checked `auth.uid() = user_id`.
- `/api/queues/asset-generate` is unauthenticated by design: `@vercel/queue`'s `handleCallback` does not sign or verify a delivery, so anyone who knows a job id can ask the server to process it (already noted in `lib/api/queue-callback.ts`).

So the chain was: insert your own row → POST its id to the queue callback → `processQueuedJob` loads it with the service key and hands `job.request` to the connector as trusted, typed input. That bypasses both `withApiAccess` and every zod schema on `/api/v1/*`.

## Fix

Jobs are server-issued now. `createJob` writes with the service key (it already takes `userId` from the authenticated route), and `006_revoke_job_inserts.sql` drops the insert policy so nothing else can mint a row. `jobs_select_own` is untouched, so the poll path still reads through RLS.

Root cause rather than defense in depth: with no way to forge a row, `job.request` can only be something a route already parsed, and the unauthenticated callback can only ever replay work its owner paid for.

## Audited, no change needed

- **Dependencies** — one open Dependabot alert (#77, moderate): `uuid@9.0.1` under `@runware/sdk-js`. The advisory is a missing buffer bounds check in `v3`/`v5`/`v6` when `buf` is passed; the SDK only calls `v4`, so the path is unreachable. The fix is `uuid@11`, a two-major jump inside a transitive dep — forcing it via an override risks breaking the SDK at runtime for no security gain.
- **Secrets** — no keys committed; `.env*` ignored apart from `.env.example`. Every provider key is read server-side from `process.env`; no `NEXT_PUBLIC_` misuse (only the Supabase URL/anon key and the blob base URL, all public by design).
- **RLS** — `projects`, `jobs`, `conversations`, `messages` all enabled and scoped to `auth.uid()`; `access_codes` has no policy and is reached only through the `security definer` RPC.
- **Injection/XSS** — no `dangerouslySetInnerHTML`, `eval`, `new Function`, `innerHTML`, or `postMessage` handlers anywhere in `app/`, `lib/`, `components/`, `remotion/`.
- **Redirects** — every `redirect` target is a literal or `request.url`-derived origin; no user-controlled destination.
- **SSRF** — the voice-preview proxy already pins origin and uses `redirect: "manual"`; the only other server-side fetches are provider-issued URLs.
- **Headers** — nosniff, SAMEORIGIN, HSTS, and `strict-origin-when-cross-origin` are set globally in `next.config.ts`.
- **`/api/dev/impersonate`** — 404s when `NODE_ENV === "production"`, which covers Vercel preview builds too.

## Open PR overlap check

Considered #625 (`ProjectsList`, `ClipWaveform`), #624 (`lib/project/autosave`), and #622 (`lib/canvas`, `lib/generation`, `lib/connectors`, migrations 002/004/005). This PR touches `lib/api/jobs.ts` and adds migration `006`, neither of which appears in any of them.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, and `test` (1309 passed) all green.

### DIFF
--- a/lib/api/jobs.ts
+++ b/lib/api/jobs.ts
@@ -34,7 +34,7 @@ export async function createJob(input: {
 	connectorType: ConnectorType;
 	request: Record<string, unknown>;
 }): Promise<{ id: string }> {
-	const supabase = await createClient();
+	const supabase = createServiceClient();
 	const { data, error } = await supabase
 		.from("jobs")
 		.insert({
--- a/supabase/migrations/003_create_jobs.sql
+++ b/supabase/migrations/003_create_jobs.sql
@@ -27,6 +27,4 @@ alter table jobs enable row level security;
 
 create policy "jobs_select_own" on jobs
   for select using (auth.uid() = user_id);
-create policy "jobs_insert_own" on jobs
-  for insert with check (auth.uid() = user_id);
--- UPDATE/DELETE only via service role (bypasses RLS).
+-- INSERT/UPDATE/DELETE only via service role (bypasses RLS).


umairnadeem (commented):
lgtm

PR #627, opened 2026-08-24
refactor: nightly maintainability 2026-08-24

umairnadeem:
Two behaviour-preserving structural changes. No product behaviour changes.

## Sloppy's model choice gets its own boundary

`useSloppy()` bundled the agent's turn control (`send`, `stop`, `loading`, `writingScript`) with the LLM picker (`model`, `setModel`, `models`). Those are different concerns on different clocks: the control value changes on every streaming status transition, while the model is a setting read off the connector config that outlives any turn.

The practical cost was `ComposerCopilot` — the brief composer, which only ever wanted the model list — depending on the ReAct loop's provider and re-rendering with it.

New `SloppyModelProvider` owns the choice and sits just above `SloppyProvider`, which now consumes it like anyone else. `useSloppy()` is down to the four fields its name implies; `ComposerCopilot` reaches for `useSloppyModel()` and nothing else.

## Tool behaviour flags live with the tools

`registry.ts` kept two hand-written sets of tool names, `SNAPSHOT_TOOLS` and `SCRIPT_TOOLS`, alongside the comment that "registration is the contract: one entry is a tool's whole definition". They were the exception: adding a tool meant remembering to edit a set three files away from where the tool is defined.

`defineTool` now takes optional `snapshot` / `rewritesCanvas` flags, the six affected tools declare their own, and the registry derives both sets. Membership is pinned by a new test at that seam.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and the full suite (1311 tests) all pass.

## Open PR overlap check

Considered #626 (`lib/api/jobs.ts`, migration), #625 (`ProjectsList`, `ClipWaveform`), #624 (`lib/project/autosave`), #622 (canvas elements, `lib/generation`, `lib/canvas`, `lib/connectors`, `lib/video`, `components/ui`, docs). This PR touches `app/components/sloppy`, `app/components/copilot/ComposerCopilot.tsx`, `app/components/canvas/CanvasProviders.tsx` and `lib/agent/tools`, none of which appear in any open PR. Non-overlapping.

## Skipped

- `app/components/video/ScrubBar.tsx` has no `role="slider"` and no keyboard handling, so the seek and volume tracks are unreachable without a pointer. A real accessibility gap, but fixing it adds behaviour rather than preserving it — flagging for a human call instead of folding it into a refactor PR.

### DIFF
--- a/app/components/canvas/CanvasProviders.tsx
+++ b/app/components/canvas/CanvasProviders.tsx
@@ -11,6 +11,7 @@ import { ActiveSceneProvider } from "../scene-selection/ActiveSceneContext";
 import { AutoScrollProvider } from "../scene-selection/AutoScrollContext";
 import { ViewModeProvider } from "./ViewModeContext";
 import { ActiveCaptionFont } from "./CaptionFonts";
+import { SloppyModelProvider } from "../sloppy/SloppyModelProvider";
 import { SloppyProvider } from "../sloppy/SloppyProvider";
 import { EditorPanelProvider } from "./panel/EditorPanelContext";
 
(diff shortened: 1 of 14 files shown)


umairnadeem (commented):
lgtm

PR #629, opened 2026-08-24
refactor: rethink openslop architecture 2026-08-24

umairnadeem:
## App understanding

openslop turns a written brief into a rendered video. A project's script lives in a Slate document on the canvas (`app/components/canvas`), where each element (image, animated image, TTS, music, SFX) is generated through a declarative graph (`lib/generation`) that fans out to per-type connectors, gateways and vendor providers (`lib/connectors` / `lib/gateway` / `lib/providers`). `lib/video` resolves the document into a `VideoLayout` that both the Remotion player and the Lambda render consume. "Sloppy" (`lib/agent` + `app/components/sloppy`) is a ReAct agent that edits the same document through registered tools. Supabase carries auth, projects, and job rows.

## From-scratch architecture

The domain layer is already close to what a rebuild would produce: deep modules with narrow interfaces, one registry per variant family, zod parsing confined to boundaries, and a single source of truth for every derived value. The gap this pass found is in the **video shell's interaction layer**: dragging.

A from-scratch build has *one* drag protocol. This one had four hand-rolled ones, each re-deriving the same lifecycle:

- `ScrubBar` — pointer capture, plus a `draggingRef` that duplicates what capture already tells you.
- `timeline/TimelineRuler` — the same pointer-capture protocol again, verbatim, in a different coordinate space.
- `useResize` — the odd one out: legacy `mousemove`/`mouseup` listeners on `document`, a hand-rolled `cleaned` flag, a `cleanupRef`, an unmount effect to undo it, and a `sizeRef` mirrored from state by a second effect. 60 lines of listener bookkeeping to express "drag this edge".
- `components/ui/number-scrubber` — a fourth, left alone (design-system primitive).

## Implemented simplification

New `lib/components/usePointerDrag.ts` owns the protocol: the primary button captures the pointer on the surface it went down on, so a drag keeps reporting after it leaves that surface and ends exactly once. **Capture is the drag's state**, so no caller keeps a flag — that is what deletes `ScrubBar`'s `draggingRef` and all of `useResize`'s listener bookkeeping. It returns the four handler props to spread; it holds no state of its own, which keeps it directly testable.

Then:

- **`ScrubBar`** keeps only what is actually its own: the ratio math and the hover reporting. Its move handler composes hover with the shared drag rather than reimplementing the capture check.
- **`TimelineRuler`** drops its copy of the protocol and keeps its px-to-seconds mapping.
- **`useResize`** loses `attachResizeListeners`, the `ResizeListenerHost` structural type, the `cleaned` flag, `cleanupRef`, both effects, and the `sizeRef`/state mirror. Unmount mid-drag needs no cleanup now: capture is released by the browser and the handlers are React props. Dropping `useCallback` lets `onStart` read `size` from the current render, so the mirror ref disappears too. It now returns `handleProps` instead of `handleMouseDown`.
- **`ResizeHandle`** takes `handleProps` and gains `touch-none`, so the panel edges are draggable by touch like every other track in the app.

Behaviour is preserved at each call site: press still starts the drag and reports position immediately, moves are still only honoured while captured, and the resize handle still calls `preventDefault()` on press to keep the browser from starting a text selection — as caller policy, not baked into the shared mechanism.

Net -50 lines, and the app's only `document`-level mouse-listener lifecycle is gone.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and `test:coverage` (148 files, 1310 tests) all pass. Coverage 78.9% statements / 74.3% branches / 80.2% lines, above the configured thresholds. `attachResizeListeners`' tests are replaced by tests for the new seam; `clampResize`'s tests are untouched.

## Open PR overlap check

Considered #628 (canvas drag/sorting), #627 (sloppy providers + `lib/agent/tools`), #626 (`lib/api/jobs` + migration), #625 (`ProjectsList`, `ClipWaveform`), #624 (`lib/project/autosave`), #622 (canvas elements, `lib/generation`, `lib/canvas`, `lib/connectors`, `lib/video`, `components/ui`, docs). Verified file-by-file: zero intersection with the nine files here. Non-overlapping.

## Skipped

- **The two-ways-to-reach-the-Slate-editor sweep.** Deep components use `useSlateStatic()` while the shell prop-drills `editor` through layers that only forward it. Landing it needs `CanvasProviders.tsx` and `SloppyProvider.tsx` (#627) and `Canvas.tsx` (#628); the free-file half leaves the tree more inconsistent than it is now, so it has to land as one sweep.
- **Sloppy tool presentation is not co-located** — adding a tool touches the tool file, the `TOOLS` map, and a 93-line switch in `toolPresentation.tsx`. Needs a `TOOLS` export from `lib/agent/tools/registry.ts`, held by #627.
- **`components/ui/number-scrubber`** could adopt the same hook, but it is a design-system primitive with a click-to-type gesture layered on the drag; left for a deliberate pass rather than folded in here.
- **`app/components/video/ScrubBar.tsx` has no `role="slider"`, no `aria-valuenow`, no `tabIndex` and no key handling**, so neither the seek nor the volume track is keyboard-operable. Real gap, but fixing it adds behaviour — flagging for the a11y nightly, same as #627 did.
- **`app/projects/[id]/loading.tsx` has drifted from the shell it mirrors** (still draws the refine bar #580 removed; rail is two entries short of `EditorSidebar`), causing a layout shift on project load. A visible UI fix, not an architecture simplification.


umairnadeem (commented):
lgtm

PR #630, opened 2026-08-24
fix: wrap the composer options row on narrow screens

### LINKED ISSUE #591: Composer options row overflows on narrow screens (no wrap, no shrink-0)

VedantMadane:
## Summary

The composer's options row was a single non-wrapping flex row, so on narrow viewports the pills compressed and clipped their labels instead of reflowing. Fixed at the row, not per pill, matching how #551 fixed the element-card actions row.

## Changes

- `ComposerCopilot.tsx` — pill group gets `min-w-0 flex-wrap`, so pills move to a second line instead of squashing. The submit `ActionButton` stays outside the wrapping group and stays anchored at the end of the row.
- `SettingPill.tsx` — pill button gets `shrink-0 whitespace-nowrap`, so labels stay intact rather than compressing.

No `sm:`/`md:` variants needed: the row is unchanged above the wrap point, so desktop layout is untouched. `SelectMenu` popovers already stay on-screen via Radix collision detection, so no change there.

Fixes #591

### DIFF
--- a/app/components/copilot/ComposerCopilot.tsx
+++ b/app/components/copilot/ComposerCopilot.tsx
@@ -262,7 +262,7 @@ function Composer({ value, onValueChange, onSubmit }: ComposerCopilotProps) {
 					</div>
 				</div>
 				<div className="flex items-center justify-between pt-2">
-					<div className="flex items-center gap-2">
+					<div className="flex min-w-0 flex-wrap items-center gap-2">
 						{inputElement}
 						<AttachMenu openPicker={openPicker} uploading={uploading} />
 						<SettingPill
--- a/app/components/copilot/SettingPill.tsx
+++ b/app/components/copilot/SettingPill.tsx
@@ -50,7 +50,7 @@ export function SettingPill<T extends string>({
 				aria-label={`${name}: ${selected.label}`}
 				disabled={disabled}
 				className={cn(
-					"focus-ring inline-flex items-center gap-1 rounded-full bg-muted px-2 py-0.5 font-body text-label text-foreground transition-colors hover:bg-button-hover disabled:pointer-events-none disabled:opacity-50",
+					"focus-ring inline-flex shrink-0 items-center gap-1 rounded-full bg-muted px-2 py-0.5 font-body text-label whitespace-nowrap text-foreground transition-colors hover:bg-button-hover disabled:pointer-events-none disabled:opacity-50",
 					className,
 				)}
 				style={style}


app/components/copilot/ComposerCopilot.tsx
```diff
@@ -155,7 +155,7 @@ function TemplatePill({
 }) {
 	return (
 		<span
-			className=" relative inline-flex self-start shrink-0 items-center gap-1 overflow-hidden rounded-full py-0.5 pl-1 pr-2 font-body text-body text-foreground whitespace-nowrap sm:mt-px sm:self-auto"
+			className=" relative inline-flex self-start shrink-0 items-center gap-1 overflow-hidden rounded-full py-0.5 pl-1 pr-2 font-body text-body text-foreground whitespace-nowrap sm:mt-px sm:self-auto flex-wrap min-w-0"
```
umairnadeem: this one's a no-op afaict, and it contradicts itself: the pill already has `whitespace-nowrap` + `shrink-0`, so it never gets narrower than its content and `flex-wrap`/`min-w-0` can't ever kick in. mind dropping it? leaving `flex-wrap` here reads like the pill wraps when it doesn't.

app/components/copilot/ComposerCopilot.tsx
```diff
@@ -236,7 +236,7 @@ function Composer({ value, onValueChange, onSubmit }: ComposerCopilotProps) {
 					/>
 				</div>
 				<ComposerAssets uploadingCount={uploadingCount} />
-				<div className="flex flex-col gap-1 sm:flex-row sm:items-baseline">
+				<div className="flex flex-col gap-1 sm:flex-row sm:items-baseline flex-wrap min-w-0">
```
umairnadeem: this is the prompt row, not the options row #591 is about. it's also inert: the only flex child that can wrap is `flex-1` (basis 0), so it always fits the line, and `min-w-0` does nothing on a block-level child. perhaps just keep the :265 hunk, which is the actual fix.

umairnadeem (changes requested):

umairnadeem:
@VedantMadane thank you for your contribution! please contribute more as you please

PR #634, opened 2026-08-25
fix: animate the still an image already generated

### LINKED ISSUE #564: Animating an image discards its generated still instead of reusing it as the still dependency
## Problem

Clicking **Animate** on an image element that already has a generated result loses that still. The image was already generated and paid for, but after the element becomes an `animated_image` the still is regenerated from scratch instead of being reused as the animation's input frame.

The cause is the id mismatch between the two states of the element:

- A plain `image` element stores its result in `SnapshotStore` under its **own** element id.
- An `animated_image` takes its still from a **derived** node — `stillElementId(el.id) = derivedNodeId("still", el.id)` (`lib/connectors/animated_image/plugins/still-frame.ts:24`) — and `createStillFramePlugin` declares `forStillOf(element)` as its dependency.

So the moment the element is rewritten to `animated_image`, the snapshot the still node looks up (`still:<id>`) is empty, while the already-generated result sits under `<id>`. The still node has no result and regenerates.

`AnimateButton` (`app/components/canvas/elements/AnimateButton.tsx:26`) routes through the LLM refine pass (`animateImagePrompt`), which rewrites the tag — nothing in that path carries the existing image result onto the derived still node.

- "when animating an image element by clicking the button, we lose the still. the still should still be kept as the dependency of the node."

## Relevant code

- `app/components/canvas/elements/AnimateButton.tsx:26` — the button; fires `refineScript(animateImagePrompt(element.id))`
- `lib/script/refine/animatePrompt.ts` — the instruction the refine pass matches to rewrite `<image>` → `<animated_image>`
- `lib/connectors/animated_image/plugins/still-frame.ts:24` — `stillElementId` / `forStillOf`; the derived-node id the still result must live under
- `lib/generation/snapshots.ts` — `SnapshotStore`, keyed by element id (`update`, `commit`, `history`)
- `lib/canvas/osmlStreamParser.ts` — existing `TODO store and rehydrate element-scoped connector snapshot too`, same class of problem

## Suggested direction

When an image element with an existing result is animated, transplant that result onto the derived still node id (`still:<elementId>`) — copying `result`, `resultInputs`, `connectorType` and marking it so it isn't treated as stale — before/as the tag is rewritten. The still node then resolves from the snapshot it already has, and only the video generation runs.

Worth handling the transplant at the animate boundary rather than inside the plugin, so the still stays a real node with its own inputs (which is what makes replacing the still correctly stale the animation).

## Acceptance criteria

- [ ] Animating an already-generated image does not regenerate the still — the existing image is used as the animation's frame.
- [ ] The still remains a proper dependency node of the animated element: editing the still's prompt still regenerates it and stales the animation.
- [ ] The still/video toggle shows the carried-over still immediately, before the video finishes.
- [ ] Animating an image with **no** result yet behaves as today (still generates, then animates).

## Related

- #506 (animated-image still/video toggle doesn't show the still node's progress)
- #386 (animated image proceeds after still-image gen fails)

---

umairnadeem:
## Summary
- Animating an already-generated image reuses that image as the animation's still instead of regenerating it. The reuse is declared by the edit rather than performed ahead of it: a `set` op carries `deps`, a map of an input of the element it produces to the id of the element whose existing result should fill it. `applyScriptEdit` applies the ops, then seeds those dependencies against the element the whole edit left behind. The still stays a node with its own inputs, so it still regenerates (and stales the animation) when the element's prompt changes.
- The whole `AssetResult` moves rather than its url, so `pinned` and `durationSec` come with it. A frame the user uploaded stays pinned instead of coming back as an ordinary generated result that the next bit of drift overwrites. No url has to survive a trip through the model.
- Nothing is discarded. An edit that omits `deps` regenerates the still; it cannot leave a blank card with the blob reachable only from version history.
- Fixes the follow-on restore bug: restoring a version from before the animate threw "animated_image element is missing required videoPrompt attribute". Versions now record the element type they were generated as (`GenerationJob.elementType` → `CommittedVersion` → `element_history.element_type`), and restoring one retypes the element back to it. The column is nullable and the field optional, so a version stored without a type restores its inputs alone.
- `element_type` is folded into migration `005_create_element_history.sql` rather than added by a new one; if 005 already ran on a database, drop `element_history` and re-run it.
- The Animate button's message reads as a request rather than a tool call: "Animate the image in scene 9, reusing the frame it has already generated." Scene normalization keeps at most one visual per scene, so the scene number names the image without exposing an id.
- Unrelated one-liner bundled in at request: captions now default to `alignX: "center"`.

Closes #564

## Test plan
- [ ] Generate an image, click Animate: the still is not regenerated and the Still toggle shows the original frame immediately.
- [ ] Hit Generate on the resulting animated image; only the video generates.
- [ ] Edit the animated image's prompt; the still goes stale and regenerates.
- [ ] Upload a still, then animate: the carried-over frame stays pinned and is not overwritten by later drift.
- [ ] Stop the turn mid-animate, or have it fail: the element is still an image showing its picture, not a blank card.
- [ ] Open the animated image's history, restore the pre-animate version: the element becomes an image again with its original prompt, and Generate regenerates the image.
- [ ] Restore the animated version afterwards: the element retypes back to animated_image with its videoPrompt.
- [ ] Click Animate on an image with no result; behaves as before (still generates, then animates).
- [ ] `npm run lint && npm run format:check && npm run typecheck && npm run knip && npm run build && npm test` pass locally (1370 tests).

## Notes for review
- The reuse now depends on the agent emitting `deps`. If it does not, the still is generated a second time. That is a soft failure by design, and it replaces a hard one.
- `deps` resolves a name to a derived node id via `derivedNodeId(name, elementId)`, the same function that mints them. Today `still` is the only dependency keyed that way, so it is the only name a fill can take. A second one would want a real name on the plugin dependency contract rather than an id convention.
- `applyRefineOps` is still exported and ignores `deps`; `applyScriptEdit` is the entry point that does both, and `applyRefineOps` now says so.

https://claude.ai/code/session_01DprVAGVo4hfWaR5BdDoajJ

### DIFF
--- a/app/components/canvas/elements/AnimateButton.tsx
+++ b/app/components/canvas/elements/AnimateButton.tsx
@@ -3,6 +3,9 @@
 import { MagicVideo } from "@/components/ui/icon";
 import { Button } from "@/components/ui/button";
 import { animateImagePrompt } from "@/lib/script/refine/animatePrompt";
+import { carryOverStill } from "@/lib/connectors/animated_image/plugins/still-frame";
+import { useGenerationQueue } from "@/lib/generation/GenerationQueueProvider";
+import { useNodeBuilder } from "@/lib/generation/useNodeBuilder";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 import { useSloppy } from "@/app/components/sloppy/SloppyProvider";
 import { useEditorPanel } from "../panel/EditorPanelContext";
@@ -10,6 +13,8 @@ import { useEditorPanel } from "../panel/EditorPanelContext";
 export function AnimateButton({ element }: { element: CanvasContentElement }) {
 	const { send, loading } = useSloppy();
 	const { setActive } = useEditorPanel();
+	const queue = useGenerationQueue();
+	const buildNode = useNodeBuilder();
 
 	if (element.type !== "image") return null;
 
@@ -23,6 +28,7 @@ export function AnimateButton({ element }: { element: CanvasContentElement }) {
 			disabled={loading}
 			onMouseDown={(e) => e.preventDefault()}
 			onClick={() => {
+				carryOverStill(element, queue, buildNode);
 				setActive("sloppy");
 				void send(animateImagePrompt(element.id));
 			}}
--- a/app/components/canvas/elements/AnimateButton.tsx
+++ b/app/components/canvas/elements/AnimateButton.tsx
@@ -2,10 +2,9 @@
 
 import { MagicVideo } from "@/components/ui/icon";
 import { Button } from "@/components/ui/button";
+import { ReactEditor, useSlateStatic } from "slate-react";
 import { animateImagePrompt } from "@/lib/script/refine/animatePrompt";
-import { carryOverStill } from "@/lib/connectors/animated_image/plugins/still-frame";
-import { useGenerationQueue } from "@/lib/generation/GenerationQueueProvider";
-import { useNodeBuilder } from "@/lib/generation/useNodeBuilder";
+import { parentSceneId, sceneIndexOf } from "@/lib/canvas/scenes";
 import type { CanvasContentElement } from "@/lib/canvas/types";
 import { useSloppy } from "@/app/components/sloppy/SloppyProvider";
 import { useEditorPanel } from "../panel/EditorPanelContext";
@@ -13,8 +12,7 @@ import { useEditorPanel } from "../panel/EditorPanelContext";
 export function AnimateButton({ element }: { element: CanvasContentElement }) {
 	const { send, loading } = useSloppy();
 	const { setActive } = useEditorPanel();
-	const queue = useGenerationQueue();
-	const buildNode = useNodeBuilder();
+	const editor = useSlateStatic();
 
 	if (element.type !== "image") return null;
 
@@ -28,9 +26,13 @@ export function AnimateButton({ element }: { element: CanvasContentElement }) {
 			disabled={loading}
 			onMouseDown={(e) => e.preventDefault()}
 			onClick={() => {
-				carryOverStill(element, queue, buildNode);
+				const path = ReactEditor.findPath(editor, element);
 				setActive("sloppy");
-				void send(animateImagePrompt(element.id));
+				void send(
+					animateImagePrompt(
+						sceneIndexOf(editor.children, parentSceneId(editor, path)),
+					),
+				);
 			}}
 		>
 			<MagicVideo aria-hidden="true" />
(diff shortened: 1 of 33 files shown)


app/components/canvas/elements/AnimateButton.tsx
```diff
@@ -23,6 +28,7 @@ export function AnimateButton({ element }: { element: CanvasContentElement }) {
 			disabled={loading}
 			onMouseDown={(e) => e.preventDefault()}
 			onClick={() => {
+				carryOverStill(element, queue, buildNode);
```
umairnadeem: this discards the image's result up front but `send` is fire and forget: agent errors just toast, and `stop` aborts the turn. so if the tag never flips, the element is still an `image` with no snapshot, ie blank card and the blob only reachable from version history. can we carry over once the element actually becomes animated_image instead of optimistically here?

lib/connectors/animated_image/plugins/still-frame.ts
```diff
@@ -60,6 +61,22 @@ export const stillSnapshot = (
...
+ * still stays a node with its own inputs, so editing it still stales the animation.
+ */
+export function carryOverStill(
+	element: CanvasContentElement,
+	queue: GenerationQueue,
+	buildNode: NodeBuilder,
+): void {
+	const { result } = queue.getElementSnapshot(element.id);
+	if (!result) return;
+	queue.commitResult(buildNode(forStillOf(element)), result);
```
umairnadeem: if the image's result came from an upload it's `pinned`, and this commits it unpinned, so the carried-over still is a normal generated node again and the next bit of drift regenerates over the user's frame. perhaps pass `{ pinned }` through from the source snapshot.

supabase/migrations/005_create_element_history.sql
```diff
@@ -5,6 +5,7 @@ create table if not exists element_history (
   project_id uuid not null references projects(id) on delete cascade,
   element_id text not null,
   connector_type text not null,
+  element_type text,
```
umairnadeem, line 8: editing an applied migration in place means `supabase db push` skips 005 (already in the migration history), so prod never gets the column, and `COLUMNS` now selects `element_type` → every history read/write 400s. dropping the table also nukes everyone's takes. a 006 with `alter table element_history add column if not exists element_type text;` costs nothing and keeps the data?

lib/providers/llm/mock.ts
```diff
@@ -249,7 +250,9 @@ function mockCall(prompt: LanguageModelV3Prompt) {
 		};
 	}
 
-	const elementId = animateId ?? ELEMENT_ID.exec(last?.text ?? "")?.[1];
+	const elementId = (animating ? IMAGE_ID : ELEMENT_ID).exec(
```
umairnadeem: IMAGE_ID grabs the first `<image>` in the whole script, so the mock ignores the scene number: with no api key, Animate on scene 7's image animates scene 1's. the old prompt carried the id so it hit the right one. can the mock scan from the `--- Scene N ---` marker it was asked for?

supabase/migrations/005_create_element_history.sql
umairnadeem: assume it wasnt applied

PR #637, opened 2026-08-25
refactor: nightly maintainability 2026-08-25

umairnadeem:
## One way to reach the Slate editor

Twelve components already read the editor with `useSlateStatic()`. Seven others took it as a prop, and every one of them renders inside the same `<Slate>`, so both paths reached the identical object. The prop path also made `PostPromptView` and `CanvasProviders` accept an `editor` they only forwarded.

Now the editor is a prop in exactly one position: `Editor` hands it to `CanvasProviders`, which puts it in `<Slate>`. `VideoLayoutProvider`, `ViewModeProvider`, `SloppyProvider`, `PostPromptView`, `EditorToolbar`, `Canvas` and `Storyboard` each call `useSlateStatic()`. No component takes the editor just to pass it on.

`useSlateStatic` returns the same stable instance without subscribing to document changes, which is exactly what a prop gave them, so render behavior is unchanged.

## Supabase env fails loudly

`lib/supabase/env.ts` defaulted `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_ANON_KEY` to `""`. A deploy missing either one built a client against an empty URL and surfaced as an opaque fetch failure inside supabase-js. `service.ts` already threw a named error for its own key, so the two halves of the same config disagreed on what a missing variable means.

The module now exports `supabaseUrl()`, `supabaseAnonKey()` and `supabaseSecretKey()`, each throwing `"<NAME> is required"`. They read at call time rather than import time, so the failure lands where a client is constructed instead of at module load. `service.ts` drops its hand-rolled check and uses the same helper. Covered by `lib/supabase/__tests__/env.test.ts`.

## Docs

One line in `ARCHITECTURE.md` recording where the editor instance enters the tree, next to the existing `useEditorSession` entry.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, and 1357 tests across 153 files all pass.

## Open PR overlap check

Open PRs considered: #634 (animate-the-still: `lib/generation/**`, `lib/canvas/editorOps`, `lib/connectors/animated_image`, `lib/project/elementHistory`, canvas element buttons, `lib/video/captionStyle`) and #630 (`ComposerCopilot.tsx`). No file, module or theme in this PR is touched by either.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -72,6 +72,7 @@ Staleness falls out of the graph: a node needs generating when it has no result,
 - `lib/canvas/elementConnector.ts` answers "which connector, provider and model does this element use?" for both the UI and the queue. An element pins its provider when created, so this is also where a pin that no longer resolves falls back to the registry default.
 - `lib/script/` provides script context and refinement utilities.
 - `app/components/canvas/hooks/useEditorSession.ts` is the one place the Slate editor gets wired to a project: initial hydration, streaming script sync, metadata sync, and autosave. Views call it and render the editor; they never assemble it.
+- The editor instance is a prop in exactly one place: `Editor` hands it to `CanvasProviders`, which puts it in `<Slate>`. Everything below reaches it with `useSlateStatic()`, so no component takes it just to pass it on.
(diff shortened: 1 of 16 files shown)


umairnadeem (commented):
lgtm

PR #639, opened 2026-08-25
fix: align the left panel header with the canvas container

### LINKED ISSUE #633: Left panel header text is not flush with the canvas container
## Problem

The left panel's header text does not line up with the top of the canvas container. There is extra
space above it, so the two columns start at different heights and the editor reads as slightly out
of register.

The canvas container sits flush under the toolbar — `app/components/PostPromptView.tsx:39` has
`mr-2 mb-2` and no top margin. The side panel column next to it does have one:
`app/components/canvas/panel/EditorSidebar.tsx:150` is
`flex w-64 shrink-0 flex-col gap-3 pr-1 pt-4 pb-3 text-panel-fg`, and that `pt-4` is what drops the
header row (`:151-153`, the icon + `<h2>`) 16px below the canvas container's top edge.

The padding is not arbitrary, which is why this is worth fixing properly rather than deleting a
class. `UserProfile` is positioned out of flow — `fixed left-5 top-4 z-[100]`
(`app/components/UserProfile.tsx:54`) wrapping a `h-9 w-9` avatar (`:63`) — so it floats over the
top-left corner and every neighbor has to reserve space for it by hand. The icon rail does exactly
that at `EditorSidebar.tsx:133`: `pt-16 mr-2 pl-1 lg:w-[72px] lg:pt-4`. That `pt-16` (64px) is a
hardcoded clearance for a 52px-tall floating avatar, and it is the only such offset in the app —
`grep` for `pt-16` finds one hit. Shrinking the avatar shrinks the clearance the layout has to
absorb.

Separately, the profile dropdown is the least idiomatic overlay in the codebase. The shared
primitive already ships the house style — `components/ui/dropdown-menu.tsx:45` gives content
`rounded-md border border-border bg-popover p-1 shadow-elevation-3` with `zoom-in-95`, and `:77`
gives items `rounded-md px-2 py-1.5 text-label`. `CharactersPicker.tsx:78` uses it as intended:
`<DropdownMenuContent align="start" className="max-h-64 min-w-32">`, sizing only. `UserProfile`
overrides nearly every token instead:

- `:74` — `rounded-2xl p-0` plus `zoom-in-105` / `zoom-out-105`, against the primitive's `rounded-md p-1` and `zoom-95`
- `:72-73` — `sideOffset={-44} alignOffset={-8}`, negative offsets that pull the menu up *behind* the avatar
- `:76` — a hand-built `flex h-[52px] items-center gap-3 pr-4 pl-12` header row, where `pl-12` exists only to clear the avatar the menu is sliding out from under
- `:83,:87,:107,:117` — `rounded-xl` / `rounded-lg` and `my-1 py-2` on items and sub-content

DESIGN.md is explicit that overlays (modals, popovers, dropdowns, selects) are `rounded-md` and that
popovers and menus sit at `shadow-elevation-3`, and CONVENTIONS.md's frontend rules say the design
system ships by default with no one-off Tailwind in higher-level components. This menu is the
outlier on both counts, and its bespoke geometry is coupled to the avatar's current size — so
whoever shrinks the avatar has to touch it anyway.

- "remove the extra padding/margin on top of the sidepanel headers so the left panel header text is
  flush with the canvas container. this may involve making the profile icon smaller and perhaps even
  using a more idiomatic styling for the profile icon dropdown (consistent with other popovers in
  the app)"

## Relevant code

- `app/components/canvas/panel/EditorSidebar.tsx:150` — `pt-4` on the panel column: the padding to remove
- `app/components/canvas/panel/EditorSidebar.tsx:151-153` — the shared header row (icon + `<h2>`); one header serves every panel, so this is a single fix
- `app/components/canvas/panel/EditorSidebar.tsx:133` — `pt-16 ... lg:pt-4` on the icon rail: the hardcoded avatar clearance
- `app/components/PostPromptView.tsx:39` — the canvas container (`mr-2 mb-2`, no top margin): the edge to align to
- `app/components/PostPromptView.tsx:35-37` — `<UserProfile />` and `<EditorToolbar />` above the panel row
- `app/components/UserProfile.tsx:54` — `fixed left-5 top-4 z-[100]`: the out-of-flow positioning that forces the clearance
- `app/components/UserProfile.tsx:63` — `<Avatar className="h-9 w-9 ...">`: the size to reduce
- `app/components/UserProfile.tsx:69-80` — the non-idiomatic dropdown content, negative offsets, and hand-built header row
- `components/ui/dropdown-menu.tsx:45,77` — the primitive's defaults, which already match DESIGN.md
- `app/components/canvas/elements/CharactersPicker.tsx:78` — the idiomatic reference usage to copy
- `DESIGN.md` §"Spacing, radius, elevation" — overlays are `rounded-md`, menus sit at `shadow-elevation-3`
- `app/components/PrePromptView.tsx:13`, `app/components/projects/ProjectsList.tsx:54` — the other two mounts of `UserProfile`; check them after any size change

## Suggested direction

Work top-down, and stop as soon as it lines up:

1. Drop the `pt-4` on the panel column (`EditorSidebar.tsx:150`) so the header row starts at the same
   y as the canvas container's top edge. Keep `pb-3`; only the top is wrong.
2. Shrink the avatar (`UserProfile.tsx:63`, `h-9 w-9` → `h-8 w-8` or smaller) so the floating profile
   button fits the space the toolbar already occupies, and reduce the rail's `pt-16` clearance
   (`EditorSidebar.tsx:133`) to match. The two numbers are coupled: the clearance exists solely to
   clear the avatar, so they should move together.
3. Bring the dropdown onto the primitive's defaults: delete the `rounded-2xl p-0`, the `rounded-xl` /
   `rounded-lg` / `my-1 py-2` item overrides, and the `zoom-105` animations, keeping only genuine
   sizing (`w-56`). The negative `sideOffset` / `alignOffset` and the `pl-12` header row exist to
   make the menu emerge from behind the avatar; a normal `sideOffset` and a plain
   `DropdownMenuLabel` for the name is both simpler and consistent with every other menu.

Step 3 is a natural extension rather than a strict requirement — if it turns out larger than
expected, steps 1–2 alone fix the reported misalignment and step 3 can be split out.

## Acceptance criteria

- [ ] The left panel's header text is vertically flush with the top of the canvas container.
- [ ] The alignment holds at both sides of the `lg` breakpoint, where the rail's top padding changes.
- [ ] The profile avatar no longer overlaps the panel rail or panel header at any width.
- [ ] `UserProfile` is unchanged on the projects list and pre-prompt views (`ProjectsList.tsx:54`, `PrePromptView.tsx:13`).
- [ ] The avatar is sized via the `Avatar` `size` prop, not a hand-picked `h-*`/`w-*` (DESIGN.md components rule).
- [ ] The profile dropdown uses `components/ui/dropdown-menu` defaults, matching `CharactersPicker.tsx:78` and DESIGN.md (`rounded-md`, `shadow-elevation-3`), with no one-off radius, padding, or animation overrides.
- [ ] `app/projects/[id]/loading.tsx:118-119,123,134` is updated to match, so the layout does not jump when the skeleton is replaced by the editor.
- [ ] The avatar keeps its `focus-ring` and `aria-label="Account menu"`, and the menu is still keyboard-reachable.

## Notes for a first-time contributor

This is a CSS-only change across two files, with no state, data, or generation logic involved. The
whole fix is a handful of Tailwind classes. The one thing to watch: `UserProfile` is `fixed`, so it
does not push anything down — the padding in `EditorSidebar` is what keeps other elements out from
under it. Change the two together and check the layout at a narrow window and a wide one, since the
rail's padding differs across the `lg` breakpoint.

## Two things that make this simpler than it looks

**The avatar is already off-scale.** `components/ui/avatar.tsx:20` ships a `size` prop —
`size-8` by default, with `data-[size=sm]:size-6` and `data-[size=lg]:size-10`. `h-9 w-9`
(`UserProfile.tsx:63`) is a hand-picked 36px that matches none of them, and DESIGN.md's components
rule says dimensions come from the `size` prop, never `h-*` in `className`. So step 2 is not
inventing a number: it is deleting `className="h-9 w-9 cursor-pointer"` in favor of the primitive's
own size.

That also explains the geometry. `EditorToolbar.tsx:78` is `px-4 py-1.5` around `h-8` buttons and
`h-7` breadcrumb items, so the toolbar is roughly 44px tall. The avatar at `top-4` + `h-9` ends at
52px, about 8px past the toolbar and into the panel row below. Dropping to the default `size-8`
ends it at 48px, and `size="sm"` at 40px — inside the toolbar either way, which is what lets the
compensating padding go.

**The loading skeleton duplicates all of it and must move in tandem.** `app/projects/[id]/loading.tsx`
hand-copies the same magic numbers:

- `:118-119` — `fixed top-4 left-4 z-[100]` and `Skeleton "h-9 w-9 rounded-full"` (note `left-4` here vs `left-5` in the real component — already out of sync)
- `:123` — the fake toolbar's `pl-16`
- `:134` — the fake rail's `pt-16 ... lg:pt-4`

If these are not updated together, the layout will visibly jump as the skeleton swaps for the real
editor. There is no shared constant for any of this today — grep finds no header-height or
panel-padding token, just these duplicated literals in two files. Extracting one is optional, but
keeping the two files consistent is not.

## Scope note

There is no right-panel header to fix. The right pane is `SidePlayerPanel`, which lives *inside* the
canvas card (`PostPromptView.tsx:55`) and is just the vertically-centered player
(`app/components/video/PlayerPanel.tsx:80`). The header at `EditorSidebar.tsx:151-153` is the only
side panel header, shared by every panel — which is what keeps this XS.

## Related

- #559 (closed) — prior panel-header control cleanup in the same area

umairnadeem:
## Summary

The left panel's header row sat 16px below the top of the canvas card, so the two columns started at different heights. Fixed top-down, per the issue's suggested direction:

- `EditorSidebar.tsx` — dropped the panel column's `pt-4` so the shared header (icon + `<h2>`) is flush with the canvas card's top edge. `pb-3` is untouched; only the top was wrong.
- `UserProfile.tsx` — the avatar now uses the `Avatar` primitive's own size (`size-8`) instead of a hand-picked `h-9 w-9`, per DESIGN.md's rule that dimensions come from the `size` prop. At `top-4` it now ends at 48px, inside the rail's normal `pt-4`.
- `EditorSidebar.tsx` — with the smaller avatar, the rail's hardcoded 64px clearance collapses: `pt-16 ... lg:pt-4` becomes a single `pt-4`, so the rail no longer shifts across the `lg` breakpoint. This was the only `pt-16` in the app.
- `UserProfile.tsx` — the dropdown is back on the `components/ui/dropdown-menu` defaults (`rounded-md`, `shadow-elevation-3`, `zoom-95`, standard `sideOffset`), matching `CharactersPicker.tsx:78` and DESIGN.md. The `rounded-2xl p-0` content, `rounded-xl`/`rounded-lg`/`my-1 py-2` item overrides, `zoom-105` animations, negative `sideOffset`/`alignOffset`, `z-[90]`/`z-[91]` hacks, and the hand-built `pl-12` header row are gone; the name is a plain `DropdownMenuLabel` + `DropdownMenuSeparator`. Only genuine sizing (`w-56`) remains. Item icons drop their `mr-2` and rely on the primitive's `gap-2`, matching `action-menu.tsx`.
- `app/projects/[id]/loading.tsx` — same geometry in the skeleton so the layout does not jump on swap: `left-4` → `left-5` (it was already out of sync), `h-9 w-9` → `size-8`, the fake toolbar's `pl-16` removed, the fake rail's `pt-16 ... lg:pt-4` → `pt-4`, and the fake bar `h-10` → `h-8` so the skeleton toolbar is 44px like the real one.

The toolbar is 44px tall (`py-1.5` + `h-8` row; `Breadcrumbs` is `absolute` and out of flow). The avatar at `top-4` + `size-8` ends at 48px, and the panel row's `pt-4` puts the first rail item at 60px, so nothing overlaps at any width.

`size: XS`, `good first issue`, `bug`, with explicit acceptance criteria and named line numbers. CSS-only across three files — no state, data, or generation logic. `UserProfile` is unchanged in behavior, so its other two mounts (`ProjectsList.tsx:54`, `PrePromptView.tsx:13`) only pick up the smaller avatar and the idiomatic menu. The `focus-ring`, `aria-label="Account menu"`, and keyboard reachability are all preserved.

## Validation

Full CI-equivalent suite, all green:

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` — 152 files, 1351 tests passed

No unit test added: the change is Tailwind classes with no logic to assert.

## Open PR overlap check

Open PRs at selection time: #638 (`app/components/video/*`), #637 (`Editor/EditorToolbar/PostPromptView/canvas/sloppy/lib/supabase`), #634 (generation + connectors), #630 (`ComposerCopilot.tsx`, which already covers #591). None touch `UserProfile.tsx`, `EditorSidebar.tsx`, or `app/projects/[id]/loading.tsx`. #637's `PostPromptView.tsx` edit is prop-threading only and changes no classNames; this PR does not touch that file at all. Re-checked against the final diff.

## Skipped candidates

- #591 — already covered by open PR #630.
- Everything else open is `size: M` or larger, or product/design-direction work (mobile editor, BYOK, variant history, publishing kit), which this run skips by policy.

Closes #633

### DIFF
--- a/app/components/UserProfile.tsx
+++ b/app/components/UserProfile.tsx
@@ -13,6 +13,7 @@ import {
 	DropdownMenuContent,
 	DropdownMenuItem,
 	DropdownMenuLabel,
+	DropdownMenuSeparator,
 	DropdownMenuSub,
 	DropdownMenuSubContent,
 	DropdownMenuSubTrigger,
@@ -58,70 +59,58 @@ function UserProfile() {
 						type="button"
 						suppressHydrationWarning
 						aria-label="Account menu"
-						className="relative z-[91] rounded-full focus-ring"
+						className="cursor-pointer rounded-full focus-ring"
 					>
-						<Avatar className="h-9 w-9 cursor-pointer">
+						<Avatar>
 							{avatarUrl && <AvatarImage src={avatarUrl} alt={email} />}
 							<AvatarFallback className="text-label">{initials}</AvatarFallback>
 						</Avatar>
 					</button>
 				</DropdownMenuTrigger>
-				<DropdownMenuContent
-					align="start"
-					side="bottom"
-					sideOffset={-44}
-					alignOffset={-8}
-					className="z-[90] w-56 origin-center rounded-2xl p-0 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-105 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-105"
-				>
-					<div className="flex h-[52px] items-center gap-3 pr-4 pl-12">
-						<span className="truncate pl-1 text-body font-medium text-foreground">
-							{name || email.split("@")[0]}
-						</span>
-					</div>
-					<div className="border-t border-border p-1">
-						<DropdownMenuSub>
-							<DropdownMenuSubTrigger className="my-1 cursor-pointer rounded-xl py-2">
-								<Contrast className="mr-2 h-4 w-4" />
-								Appearance
-							</DropdownMenuSubTrigger>
-							<DropdownMenuSubContent className="rounded-xl">
-								<DropdownMenuLabel className="text-label text-muted-foreground">
-									Mode
-								</DropdownMenuLabel>
-								{THEME_MODES.map((m) => (
-									<DropdownMenuItem
-										key={m.value}
-										onClick={() => setTheme(m.value)}
-										className="cursor-pointer rounded-lg"
-									>
-										<span className="flex w-4 items-center justify-center">
-											{theme === m.value && <Check className="h-3.5 w-3.5" />}
-										</span>
-										{m.label}
-									</DropdownMenuItem>
-								))}
-							</DropdownMenuSubContent>
-						</DropdownMenuSub>
-						{IS_DEV && (
-							<DropdownMenuItem
-								onClick={() => setImpersonateOpen(true)}
-								className="my-1 cursor-pointer rounded-xl py-2"
-							>
-								<User className="mr-2 h-4 w-4" />
-								Impersonate user
-								<Badge variant="caution" className="ml-auto">
-									Dev
-								</Badge>
-							</DropdownMenuItem>
-						)}
+				<DropdownMenuContent align="start" className="w-56">
+					<DropdownMenuLabel className="truncate">
+						{name || email.split("@")[0]}
+					</DropdownMenuLabel>
+					<DropdownMenuSeparator />
+					<DropdownMenuSub>
+						<DropdownMenuSubTrigger className="cursor-pointer">
+							<Contrast />
+							Appearance
+						</DropdownMenuSubTrigger>
+						<DropdownMenuSubContent>
+							<DropdownMenuLabel className="text-label text-muted-foreground">
+								Mode
+							</DropdownMenuLabel>
+							{THEME_MODES.map((m) => (
+								<DropdownMenuItem
+									key={m.value}
+									onClick={() => setTheme(m.value)}
+									className="cursor-pointer"
+								>
+									<span className="flex w-4 items-center justify-center">
+										{theme === m.value && <Check className="size-3.5" />}
+									</span>
+									{m.label}
+								</DropdownMenuItem>
+							))}
+						</DropdownMenuSubContent>
+					</DropdownMenuSub>
+					{IS_DEV && (
 						<DropdownMenuItem
-							onClick={handleLogout}
-							className="my-1 cursor-pointer rounded-xl py-2"
+							onClick={() => setImpersonateOpen(true)}
+							className="cursor-pointer"
 						>
-							<LogOut className="mr-2 h-4 w-4" />
-							Log out
+							<User />
+							Impersonate user
+							<Badge variant="caution" className="ml-auto">
+								Dev
+							</Badge>
 						</DropdownMenuItem>
-					</div>
+					)}
+					<DropdownMenuItem onClick={handleLogout} className="cursor-pointer">
+						<LogOut />
+						Log out
+					</DropdownMenuItem>
 				</DropdownMenuContent>
 			</DropdownMenu>
 			{IS_DEV && (
--- a/app/components/canvas/panel/EditorSidebar.tsx
+++ b/app/components/canvas/panel/EditorSidebar.tsx
@@ -130,7 +130,7 @@ function EditorSidebarComponent() {
 		<div className="flex shrink-0">
 			<nav
 				aria-label="Panels"
-				className="flex w-14 shrink-0 flex-col items-center gap-1 pt-16 mr-2 pl-1 lg:w-[72px] lg:pt-4"
+				className="flex w-14 shrink-0 flex-col items-center gap-1 pt-4 mr-2 pl-1 lg:w-[72px]"
 			>
 				<RailItem icon={Home} label="Home" href="/" />
 				<div className="my-1 h-px w-full bg-border" />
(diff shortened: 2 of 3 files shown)


umairnadeem (commented):
lgtm

PR #644, opened 2026-08-26
fix: nightly bug fixes 2026-08-26

umairnadeem:
Two real correctness bugs. No behaviour change beyond the fixes.

## A chosen model is silently reset to the connector default on reload

`createCanvasNode` stamped `model` and `provider` from the registry default *after* merging the caller's attributes, so any incoming `model` was overwritten:

```ts
const attributes = { ...schema.defaultAttributes, ...opts.attrs };
if (connectorConfig?.defaultModel) {
  attributes.model = connectorConfig.defaultModel;   // clobbers opts.attrs.model
  attributes.provider = provider;
}
```

`ModelBadge` lets a user pick any model in `config.models` per element, and it is written to `generationAttributes.model`, serialized into OSML, and persisted. Reopening the project runs the OSML back through `parseOSML` → `createCanvasNode`, which resets the model to the connector default. So the picked model does not survive a reload, and because `generationAttributes` feed `nodeInputs`, the element also comes back *stale* — the next Generate quietly runs on the wrong model.

Only the registry can say what is generatable, so that is what the resolution now asks: keep the authored provider/model when the registry has it, otherwise fall back to the type's default. A model retired from the registry resolves to the default instead of being carried into a generation that would fail — the edge case is defined out rather than guarded downstream. The resolved model is also what now picks the attribute schema, matching `resolveElementSchema`.

## A failed element shows the old error through its whole retry

`enqueueGraph` moves a node to `queued` without clearing `error`, and `commit` is the only thing that ever nulls it. `PlaceholderOverlay` renders `error` whenever it is non-null, independent of status — so after a generation fails, hitting Generate again leaves the red error banner sitting over the spinner for the entire run, and it only disappears if the retry succeeds. If the retry fails too, the message never visibly changes.

Queuing now clears the error: the previous failure is what this run is answering.

## Tests

- `queue.test.ts` — re-enqueuing an element that failed clears its error and shows it generating.
- `createCanvasNode.integration.test.ts` — a registry-known model/provider survives; an unknown model and an unknown provider each fall back to the default; and an OSML serialize → parse round trip keeps the chosen model (the reload path from the first bug).

All four fail on master and pass here. Full suite: 155 files, 1381 tests passing.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test:run` all pass.

## Open PR overlap check

`gh pr list --state open` returned an empty list both before starting and again before committing — there are no open PRs in the repo, so this work overlaps nothing.

### DIFF
--- a/lib/canvas/__tests__/createCanvasNode.integration.test.ts
+++ b/lib/canvas/__tests__/createCanvasNode.integration.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from "vitest";
 import { createCanvasNode } from "../createCanvasNode";
+import { parseOSML } from "../osmlStreamParser";
+import { serializeOSML } from "../osmlSerializer";
+import { SCENE_TYPE, type SceneElement } from "@/lib/canvas/types";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 import { flatAttributes } from "@/lib/video/elementAttributes";
 
@@ -23,7 +26,7 @@ const connectors: ConnectorRegistry = {
 	image: {
 		openslop: {
 			defaultModel: "image-model",
-			models: ["image-model"],
+			models: ["image-model", "image-model-pro"],
 			isDefault: true,
 			apiKey: "",
 		},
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
overengineered, will fix this with BYOK

PR #645, opened 2026-08-26
refactor: nightly maintainability 2026-08-26

umairnadeem:
Three cleanups under one theme: let each component own the condition for its own existence, and drop the checks the types already make impossible. No behaviour change.

## The upload button decides for itself whether it applies

`ElementContainer` listed the two element types that can take an uploaded picture:

```tsx
{(element.type === "image" || element.type === "animated_image") && <ElementUploadButton />}
```

while `ElementUploadButton` already resolved its own target generically (`stillDependency(node) ?? node`). So the rule was split: the container knew *which* elements, the button knew *what* to write, and a new picture-bearing type would have to be added in both places.

`pictureNode` in `still-frame.ts` now answers the whole question — the still behind an animated image, an image element itself, `null` for an element that makes no picture — by reading `outputKind` off the element-type registry rather than naming types. It generalises the `stillDependency` helper that already lives there. The container renders `<ElementUploadButton />` unconditionally and the button returns `null` when it doesn't apply.

The set of types that render the button is unchanged: image and animated_image render it, clip and the audio types don't. Three tests cover that.

## The seek bar's settle timer becomes a debounced value

`SegmentedSeekBar` held the hover position, the hovered segment index, the settled segment index, a timer ref and an unmount effect, and rebuilt all of it inside an `onHoverChange` callback. Only two things are ever read: the segment under the pointer (derivable) and the segment the pointer has held for 80ms (a debounce of the hover itself).

It now keeps the hover and a debounced copy of it, derives both segments, and passes `setHover` straight to `ScrubBar`. The timer restarts on every pointer move exactly as before, so the thumbnail still needs 80ms of stillness, and the tooltip still unmounts the instant the pointer leaves.

## Two branches the types make unreachable

`VideoComposition` guarded a `Sequence` without an `element` — `Sequence.element` is required, and `timelineRows.ts` reads it unguarded — and had a `default: return null` on a switch over `LayerType`, which is `"audio" | "visual"` and already exhaustive.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test:coverage` all pass. 155 files, 1382 tests.

## Open PR overlap check

One open PR: #644 (`lib/canvas/createCanvasNode.ts`, `lib/generation/queue.ts` and their tests). Re-checked before committing — no file or theme overlap with this branch.

### DIFF
--- a/app/components/canvas/elements/ElementContainer.tsx
+++ b/app/components/canvas/elements/ElementContainer.tsx
@@ -128,8 +128,7 @@ export function ElementContainer({
 							contentEditable={false}
 						>
 							<ElementStaleIndicator />
-							{(element.type === "image" ||
-								element.type === "animated_image") && <ElementUploadButton />}
+							<ElementUploadButton />
 							<AnimateButton element={element} />
 							<ElementGenerateButton />
 						</div>
--- a/app/components/canvas/elements/ElementUploadButton.tsx
+++ b/app/components/canvas/elements/ElementUploadButton.tsx
@@ -2,15 +2,16 @@
 
 import { useGenerationQueue } from "@/lib/generation/GenerationQueueProvider";
 import { isGenerationActive } from "@/lib/generation/snapshots";
-import { stillDependency } from "@/lib/connectors/animated_image/plugins/still-frame";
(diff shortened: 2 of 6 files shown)


umairnadeem (commented):
lgtm

Issue #632, opened 2026-08-25
Let Sloppy change an already-selected voice, not just its description


umairnadeem:
## Problem

Once a voice has actually been used, Sloppy can no longer change it. Every trait edit it makes is
silently inert.

The voice pipeline has two distinct notions of a voice, and the agent can only write one of them:

- **Traits** (`gender`, `age`, `pitch`, `accent`, `description`, `language`) — a *description* of a
  voice (`lib/project/types.ts:20-27`, `VOICE_TRAITS` at `:34-41`).
- **Identity** (`voiceId`, `resolvedVoiceId`) — the *actual* voice
  (`lib/project/types.ts:29-32`).

Sloppy's two voice tools, `set_narrator` (`lib/agent/tools/setNarrator.ts:15-18`) and
`set_character` (`lib/agent/tools/setCharacter.ts:24`), both write traits only — their input is
`VOICE_TRAITS` from `lib/agent/tools/inputs.ts:10-20`, which has no `voiceId`. `set_narrator`'s own
description states the model of the world: *"A voice is described, never picked: the traits resolve
to a real voice when speech is generated."*

That is true exactly once. The resolution path makes identity sticky:

1. `voice-search` turns traits into a real voice, but only when none is set yet —
   `lib/connectors/tts/plugins/voice-search.ts:21` is `if (params.voiceId) return params;`, and on a
   miss it *drops the traits entirely* (`:37` omits `VOICE_DESCRIPTOR_KEYS`).
2. `voice-hydrate` then writes that result back as `resolvedVoiceId` so later generations skip the
   search (`lib/connectors/tts/plugins/voice-hydrate.ts:7-10,18-30`).
3. From then on `metadata-voice` supplies it on every call —
   `lib/connectors/tts/plugins/metadata-voice.ts:26`: `voiceId: fields.voiceId ?? resolvedVoiceId`.

So after the first line of speech is generated, step 1's guard is permanently satisfied. Sloppy can
set `description: "gravelly old sailor"` on a narrator that resolved to a bright young voice, get
back `"Set the narrator's voice."`, and nothing changes — not on the next generation, not ever.
Nothing in the codebase clears `resolvedVoiceId`; grep finds no writer other than `voice-hydrate`
itself.

A human is not stuck this way. `MetadataVoicePicker` writes an explicit identity —
`app/components/canvas/elements/character/VoiceMetadataFields.tsx:102`,
[...]

Issue #610, opened 2026-08-21
Let Sloppy read element generation state (generated vs stale vs ungenerated, durations, errors)


umairnadeem:
## Problem

Sloppy is blind to generation state. Everything it knows about the canvas comes from two
places, and neither carries any:

- The per-request context block (`lib/agent/context.ts:15-31,66-89`) is settings only: title,
  style, language, length, aspect ratio, narrator, character list, reference image count.
- `read_script` (`lib/agent/tools/readScript.ts:31-52`) returns characters plus OSML from
  `serializeOSMLWithScenes`, which emits `id` + custom attributes + body text and nothing else
  (`lib/canvas/osmlSerializer.ts:19-26`).

Meanwhile the app already tracks rich per-element state that never reaches the model.
`ElementSnapshot` (`lib/generation/snapshots.ts:6-15`) holds `status`, `result`, `error`,
`resultInputs`, `connectorType`, `pinned`, and `seconds`; `isNodeStale`
(`lib/generation/graph.ts:140-145`) already computes exactly the "generated but out of date"
bit; and `AssetResult.durationSec` (`lib/connectors/types.ts:82-88`) carries how long a
generated audio or video asset actually runs, which is what drives the timeline
(`lib/video/sceneSegments.ts:13,86`). `AgentToolContext` (`lib/agent/tools/context.ts:8-27`)
has no accessor for any of it: `avatarUrl(name)` is the only capability that reads a result at
all, and it is narrowed to avatars.

The consequences show up constantly:

- Sloppy cannot answer "is this generated yet?", "which elements are stale?", or "how long is
  this scene?" It either guesses or has to tell the user to look at the canvas.
- The LIMITS block (`lib/agent/prompt.ts:47-58`) makes it instruct the user to hit Generate All
  for "any stale elements" without ever knowing whether there are any, or which.
- Timing reasoning is done from a hardcoded rule of thumb in the prompt rather than from data.
  The working tree currently carries a hand-written "OpenSlop quirks" note in `prompt.ts`
  explaining that a visual element's on-screen time is set by the dialogue elements after it.
  That rule is only necessary because the agent can't see the durations themselves, and it
  still can't say how long anything is, only how it is derived.
- After `edit_script`, Sloppy has no way to check what its own edit invalidated.

This is the textual half of giving the agent eyes. #598 is the pixel half.

## Reported by (verbatim feedback)

[...]

Issue #598, opened 2026-08-20
Let Sloppy view generated image element results and give feedback on them


umairnadeem:
## Problem

Sloppy has eyes, but only for two hardcoded sources. `view_reference_images` shows the project's uploaded refs (`lib/agent/tools/viewReferenceImages.ts`) and `view_avatar` shows one character's avatar (`lib/agent/tools/viewAvatar.ts`). Both hand the model real image parts via `imagePart(url)` (`lib/agent/tools/defineTool.ts:30-34`), and the LLM provider already accepts image input (`lib/providers/llm/anthropic.ts:27-46,99-107`).

What Sloppy cannot see is **anything the project actually generated**. There is no tool that looks at an image element's rendered output, so the agent writes a prompt, the image comes back wrong, and the agent has no idea. It can talk about the script but never about the result. `AgentToolContext` exposes `avatarUrl(name)` and nothing more general (`lib/agent/tools/context.ts:6-27`), even though `view_avatar` already reaches a generated result through `characterAvatarUrl` (`lib/project/characterAvatar.ts:14-18`) — the plumbing exists, it is just narrowed to avatars.

This is the missing half of the loop: an agent that authors visual prompts should be able to look at what those prompts produced and say whether it landed.

**Scope: image elements only.** Video and audio results are explicitly out of scope for this issue.

## Reported by (verbatim feedback)

- "also another feature request to let sloppy view specific image element results and give feedback on them (video/sound are out of scope for now)"

## Relevant code

- `lib/agent/tools/viewAvatar.ts:5-32` — the exact pattern to generalize: id in, `imagePart` out, `toModelOutput` returning text + image content
- `lib/agent/tools/context.ts:13` — `avatarUrl(name)` is the only result-reading capability; needs a general `elementImageUrl(elementId)` sibling
- `lib/agent/tools/useAgentTools.ts:32` — where `avatarUrl` is implemented against the generation queue; the same `queue` handle can serve any element id
- `lib/generation/snapshots.ts:6-22` — `ElementSnapshot { status, result, resultInputs, pinned }`, read via `queue.getElementSnapshot(id)` (`lib/generation/queue.ts:53`)
- `lib/connectors/assetUrl.ts:16-21` — `getPrimaryUrl(result, "image")` is the one supported way to pull an image URL off a result
- `lib/canvas/types.ts:20-63` — `ELEMENT_TYPES[type].outputKind === "image"` is the guard that keeps this tool to images
[...]

PR #648, opened 2026-08-26
feat: copy and paste OSML through the canvas clipboard

### LINKED ISSUE #491: Copy/paste OSML from the canvas (clipboard round-trip)

umairnadeem:
## Summary

Wires OSML into the canvas clipboard, using the serializer and parser that already exist.

- **Copy/cut** — `setFragmentData` writes the selection to `text/plain` as OSML, so element types, custom attributes and scene markers survive outside the app. Slate's `application/x-slate-fragment` blob is untouched, so in-app copy/paste stays lossless and keeps flowing through `withFlatPaste` / `withNodeId`.
- **Paste** — overrides `insertTextData` rather than `insertData`, so a Slate fragment (in-app paste) still wins automatically and only unrecognised plain text reaches the new path. Text that looks like OSML is parsed with `parseOSML`; anything else falls through to the default plain-text behaviour.
- Pasted elements get fresh ids via the existing `stripIds` helper, so a copy never shares generation state with its original.
- `serializeOSMLWithScenes` now also emits content elements that sit outside a scene instead of silently dropping them, making it a single entry point for a whole document or an arbitrary clipboard fragment. Autosave output is unchanged, since normalization guarantees top-level nodes are scenes.

Scene markers are deliberately not reinstated on paste: `withScenes` derives scene boundaries from foreground elements, so inserting the elements alone reproduces the source grouping.

#491 (`size: M`, `priority: medium`). Both halves of the conversion were already written and tested; the change is one new plugin plus one branch in the serializer. No new representation, no second parser, no product or design decisions, and no auth-gated flows to validate.

## Validation

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` — 1389 tests, 156 files, thresholds met
- `npm run test:e2e` — 3 Playwright smoke tests

New coverage: `app/components/canvas/__tests__/withOSMLClipboard.test.ts` (copy writes OSML, attributes survive, collapsed selection defers to the default, paste rebuilds elements with attributes, fresh ids, scene splitting, prose and non-canvas markup defer). Plus `looksLikeOSML` cases in `osmlStreamParser.test.ts` and bare/mixed-fragment cases in `osmlSerializer.test.ts`.

## Open PR overlap check

Re-listed at selection time and again before committing. Open PRs #644, #645, #646, #647 touch `lib/generation/*`, `lib/canvas/createCanvasNode.ts`, `app/components/canvas/elements/*`, `app/components/canvas/dnd/*`, `remotion/*`, `lib/connectors/animated_image/*` and `lib/agent/*`. No file in this diff appears in any of them. #647 already claimed #597, so this pass took a different issue.

## Skipped candidates

- #597 — already claimed by open PR #647
- #632, #610, #598 — each needs new agent tools plus `lib/agent/tools/registry.ts`, which #647 is editing
- #492, #345 — overlap `lib/connectors/animated_image/*` and `app/components/canvas/elements/ElementContainer.tsx` (#645)
- #594, #296, #398 — product decisions (what a template clear should wipe, a composer opt-out toggle, a blocked prerequisite in #370)
- #461, #316 — word-level timing data and audio DSP; neither validatable without a render
- #254 — `good first issue`, left for outside contributors

Closes #491

### DIFF
--- a/app/components/canvas/plugins/withOSMLClipboard.ts
+++ b/app/components/canvas/plugins/withOSMLClipboard.ts
@@ -1,10 +1,31 @@
-import { Range, Transforms } from "slate";
+import { Editor, Point, Range, Transforms } from "slate";
 import type { CanvasEditor } from "@/lib/canvas/types";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 import { serializeOSMLWithScenes } from "@/lib/canvas/osmlSerializer";
 import { looksLikeOSML, parseOSML } from "@/lib/canvas/osmlStreamParser";
-import { isParsedContentElement } from "@/lib/canvas/guards";
+import { isContentElement, isParsedContentElement } from "@/lib/canvas/guards";
 import { stripIds } from "@/lib/canvas/nodeUtils";
+
+/**
+ * A fragment keeps the ancestor chain from the root, so serializing a few words
+ * inside one element would still put its tags on the clipboard. Copying tags is
+ * only what the user meant once the selection swallows a whole element.
+ */
+function coversWholeElement(editor: CanvasEditor, at: Range): boolean {
+	const [start, end] = Range.edges(at);
+	for (const [, path] of Editor.nodes(editor, {
+		at,
+		match: isContentElement,
+	})) {
+		const element = Editor.range(editor, path);
+		if (
+			Point.compare(start, element.anchor) <= 0 &&
+			Point.compare(end, element.focus) >= 0
+		)
+			return true;
+	}
+	return false;
+}
 
 /**
  * Makes OSML the canvas clipboard format: copy writes it to `text/plain` so a
@@ -25,6 +46,7 @@ export const withOSMLClipboard =
 			setFragmentData(data, originEvent);
 			const { selection } = editor;
 			if (!selection || Range.isCollapsed(selection)) return;
+			if (!coversWholeElement(editor, selection)) return;
 			const osml = serializeOSMLWithScenes(editor.getFragment());
 			if (osml) data.setData("text/plain", osml);
 		};
--- a/app/components/canvas/plugins/withOSMLClipboard.ts
+++ b/app/components/canvas/plugins/withOSMLClipboard.ts
@@ -1,41 +1,22 @@
-import { Editor, Point, Range, Transforms } from "slate";
+import { Editor, Range, Transforms } from "slate";
 import type { CanvasEditor } from "@/lib/canvas/types";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 import { serializeOSMLWithScenes } from "@/lib/canvas/osmlSerializer";
-import { looksLikeOSML, parseOSML } from "@/lib/canvas/osmlStreamParser";
+import { parseOSML } from "@/lib/canvas/osmlStreamParser";
+import { splitScenes } from "@/lib/project/serialize";
 import { isContentElement, isParsedContentElement } from "@/lib/canvas/guards";
 import { stripIds } from "@/lib/canvas/nodeUtils";
 
-/**
- * A fragment keeps the ancestor chain from the root, so serializing a few words
- * inside one element would still put its tags on the clipboard. Copying tags is
- * only what the user meant once the selection swallows a whole element.
- */
-function coversWholeElement(editor: CanvasEditor, at: Range): boolean {
-	const [start, end] = Range.edges(at);
-	for (const [, path] of Editor.nodes(editor, {
-		at,
-		match: isContentElement,
-	})) {
-		const element = Editor.range(editor, path);
-		if (
-			Point.compare(start, element.anchor) <= 0 &&
-			Point.compare(end, element.focus) >= 0
-		)
-			return true;
-	}
-	return false;
-}
+/** A fragment keeps the ancestor chain, so a few words would still carry their tags. */
+const coversWholeElement = (editor: CanvasEditor, at: Range) =>
+	Array.from(Editor.nodes(editor, { at, match: isContentElement })).some(
+		([, path]) => Range.surrounds(at, Editor.range(editor, path)),
+	);
 
 /**
- * Makes OSML the canvas clipboard format: copy writes it to `text/plain` so a
- * selection survives outside the app, and paste parses it back into elements.
- * The Slate fragment is left untouched, so in-app copy/paste stays lossless and
- * keeps flowing through `withFlatPaste` and `withNodeId`.
- *
- * Scene markers are not reinstated on paste: `withScenes` derives scene
- * boundaries from foreground elements, so inserting the elements alone
- * reproduces the source grouping.
+ * Puts OSML on `text/plain`. The Slate fragment is left untouched, so in-app
+ * copy/paste stays lossless and keeps flowing through `withFlatPaste` and
+ * `withNodeId`.
  */
 export const withOSMLClipboard =
 	(connectors: ConnectorRegistry) =>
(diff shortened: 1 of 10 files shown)


app/components/canvas/plugins/withOSMLClipboard.ts
```diff
@@ -0,0 +1,44 @@
...
+	(connectors: ConnectorRegistry) =>
+	(editor: CanvasEditor): CanvasEditor => {
+		const { setFragmentData, insertTextData } = editor;
+
+		editor.setFragmentData = (data, originEvent) => {
+			setFragmentData(data, originEvent);
+			const { selection } = editor;
+			if (!selection || Range.isCollapsed(selection)) return;
+			const osml = serializeOSMLWithScenes(editor.getFragment());
+			if (osml) data.setData("text/plain", osml);
```
umairnadeem: this clobbers `text/plain` for *any* expanded selection, including a few words inside one narration. `Editor.fragment` keeps the ancestor chain from the root, so selecting "three words" mid-narration and hitting cmd+C puts `--- Scene 1 ---\n<narration id="n1">three words</narration>` on the clipboard instead of the three words. paste that into slack or a search bar and it's tags. maybe gate on the selection covering at least one whole content element before overwriting?

lib/canvas/osmlSerializer.ts
```diff
@@ -40,6 +46,8 @@ export function serializeOSMLWithScenes(descendants: Descendant[]): string {
 			for (const child of node.children) {
 				osml += serializeElement(child);
 			}
+		} else if (isContentElement(node)) {
```
umairnadeem: afaict this branch is unreachable from either caller: autosave passes `editor.children` (normalization guarantees scenes at the top) and the clipboard passes `editor.getFragment()`, which preserves the ancestor chain from the root, so its top-level nodes are always scenes too. so "content elements outside a scene (a partial clipboard selection)" in the doc comment isn't a state that happens, and the tests only reach it by hand-building the array. drop it?

app/components/canvas/plugins/withOSMLClipboard.ts
```diff
@@ -0,0 +1,45 @@
...
+			if (!coversWholeElement(editor, selection)) return;
+			data.setData("text/plain", serializeOSMLWithScenes(editor.getFragment()));
+		};
+
+		editor.insertTextData = (data) => {
+			const elements = splitScenes(data.getData("text/plain"))
+				.flatMap((sceneOsml) => parseOSML(sceneOsml, connectors))
+				.filter(isParsedContentElement)
+				.map(stripIds);
+			if (elements.length === 0) return insertTextData(data);
```
umairnadeem, line 39: prose sitting next to a tag gets silently eaten on this path. paste `notes: <image id="i1">a sunset</image> fix this` and you end up with just the image: `parseOSML` drops text before the first tag (`updateCurrent` no-ops with no current node) and leaves the trailing text sitting in the buffer, and text between two tags folds into the previous element's body. only handling it when the text is *all* elements (i.e. bail to the default if anything outside the tags survives) would avoid quietly losing half a paste.

PR #654, opened 2026-08-27
feat: canvas version history with preview and restore

### LINKED ISSUE #258: Canvas history (versioning & diffing)
## Problem

There's no history of canvas changes. Users want versioning and the ability to diff versions (and presumably restore), plus a lightweight **activity log** they can open to see the recent changes they've made.

- "Canvas history with versioning and diffing"
- "also maybe a window log you can open to see recent changes you made"

## Relevant code

- `app/components/canvas/Canvas.tsx`
- `lib/project/store.ts`, `lib/project/api.ts` — project store/save

## Suggested direction

- Persist canvas snapshots/versions with a diff and restore UI.
- Surface a recent-changes activity log (openable window) for at-a-glance change tracking.

Related: #267 (in-app source script window) references this for the "recent changes log" part of the same feedback.

---

umairnadeem:
Closes #258 (versioning and restore; diffing is out of scope).

## What this adds

A **History** tab in the editor sidebar listing every autosaved version of the canvas script, with read-only preview and restore.

- **New `canvas_versions` table** (migration 006) holding one OSML script per version, with RLS mirroring `element_history`.
- **Folding.** Consecutive autosaves overwrite the open version for 10 minutes. An idle gap, a reload, or a restore starts a new one, so the list reads as checkpoints instead of one row per keystroke.
- **Preview.** Clicking a version flushes any pending save, suspends autosave, and shows the old script on a read-only canvas. A banner names what is on screen and offers the two ways out.
- **Restore.** The previewed script becomes current and saves forward as its own version. Nothing is deleted, so a restore can itself be undone by restoring the version you were on.
- **Back to latest** puts back exactly what the canvas held before the preview, from an in-memory stash — so edits can never be lost by looking at history.

## Scope

Versions store the **script only**. Generated media and project metadata are not snapshotted per version, so history stays cheap and restoring never rolls back generations. Element ids round-trip through OSML, so a restored script reattaches whatever media those elements already have.

## Structure

- `lib/project/canvasVersions.ts` — `CanvasHistory`: the list, the fold window, and the preview/restore lifecycle. Talks to storage and to the document through two small interfaces, so it is unit-tested with no Supabase and no Slate.
- `lib/project/canvasHistory.ts` — the Supabase implementation of that storage port, zod-parsed at the boundary.
- `lib/project/applyScript.ts` — the editor-replacement logic lifted out of `useProjectRehydrate`, now shared by project load and version restore.
- `lib/project/autosave.ts` — gains `suspend`/`resume` and an `onScriptSaved` subscription (the same shape as `queue.onCommitted`).
- UI: `HistoryPanel` registers through the existing `PANELS` record; `VersionBanner` and `VersionPreviewLock` are leaf client components.

## Known trade-off

Each autosave now writes the project row and then the version row. Within a fold window that is one extra full-script write per 2s of active editing. Throttling the version write separately would need a flush-on-exit path to keep the newest checkpoint from lagging the live project, so I left it simple. There is also no retention rule yet (reads are capped at 100 rows), matching `element_history`.

## Testing

`lint`, `format:check`, `typecheck`, `knip`, `build`, 1388 unit tests, and the Playwright smoke tests all pass. New coverage: the fold window, preview/back-to-latest/restore transitions, load failure and retry, and the autosaver's suspend/resume/subscription seams.

https://claude.ai/code/session_01G65tVUHz7Vj2uE5qsg9QqH

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -92,6 +92,7 @@
 	--amber-400: #c38a40;
 	--amber-500: #a77231;
 	--amber-700: #6f4509;
+	--amber-800: #4a2d00;
 	--amber-500-a: #f6bd6026;
 
 	--radius: 0.625rem;
@@ -152,6 +153,9 @@
 	--success-foreground: #fff;
 	--caution: var(--amber-500);
 	--caution-foreground: #fff;
+	--version-preview: var(--amber-100);
+	--version-preview-foreground: var(--amber-700);
+	--version-preview-link: var(--amber-800);
 	--border: var(--grey-100-a);
 	--input: var(--grey-30-a);
 	--ring: var(--blurple-500);
@@ -234,8 +238,11 @@
 	--green-600: #53a57b;
 	--green-500-a: #00945426;
 
+	--amber-100: #623b00;
 	--amber-400: #8b5b20;
 	--amber-500: #b57e38;
+	--amber-700: #e4c69a;
+	--amber-800: #f8e0bd;
 	--amber-500-a: #a7723159;
 
 	/* Surface overrides (warm near-blacks) */
@@ -349,6 +356,9 @@
 	--color-success-foreground: var(--success-foreground);
 	--color-caution: var(--caution);
 	--color-caution-foreground: var(--caution-foreground);
+	--color-version-preview: var(--version-preview);
+	--color-version-preview-foreground: var(--version-preview-foreground);
+	--color-version-preview-link: var(--version-preview-link);
 	--color-border: var(--border);
 	--color-input: var(--input);
 	--color-ring: var(--ring);
--- a/app/globals.css
+++ b/app/globals.css
@@ -153,9 +153,8 @@
 	--success-foreground: #fff;
 	--caution: var(--amber-500);
 	--caution-foreground: #fff;
-	--version-preview: var(--amber-100);
-	--version-preview-foreground: var(--amber-700);
-	--version-preview-link: var(--amber-800);
+	--caution-soft: var(--amber-500-a);
+	--caution-soft-foreground: var(--amber-700);
 	--border: var(--grey-100-a);
 	--input: var(--grey-30-a);
 	--ring: var(--blurple-500);
@@ -356,9 +355,8 @@
 	--color-success-foreground: var(--success-foreground);
 	--color-caution: var(--caution);
 	--color-caution-foreground: var(--caution-foreground);
-	--color-version-preview: var(--version-preview);
-	--color-version-preview-foreground: var(--version-preview-foreground);
-	--color-version-preview-link: var(--version-preview-link);
+	--color-caution-soft: var(--caution-soft);
+	--color-caution-soft-foreground: var(--caution-soft-foreground);
 	--color-border: var(--border);
 	--color-input: var(--input);
 	--color-ring: var(--ring);
--- a/lib/project/applyScript.ts
+++ b/lib/project/applyScript.ts
@@ -3,7 +3,10 @@ import { HistoryEditor } from "slate-history";
 import type { ConnectorRegistry } from "@/lib/connectors/registry";
 import { deserializeWithScenes } from "./serialize";
 
-/** Outside undo history: loading or restoring is not something to undo into. */
+/**
+ * Replaces the document outside undo history and clears the stack, whose
+ * entries point at paths this swap removes.
+ */
 export function applyScriptToEditor(
 	editor: Editor,
 	script: string,
@@ -24,6 +27,8 @@ export function applyScriptToEditor(
 
 	if (HistoryEditor.isHistoryEditor(editor)) {
 		HistoryEditor.withoutSaving(editor, replaceChildren);
+		editor.history.undos = [];
+		editor.history.redos = [];
 		return;
 	}
 
--- a/lib/project/applyScript.ts
+++ b/lib/project/applyScript.ts
@@ -1,0 +1,31 @@
+import { Editor, Transforms } from "slate";
+import { HistoryEditor } from "slate-history";
+import type { ConnectorRegistry } from "@/lib/connectors/registry";
+import { deserializeWithScenes } from "./serialize";
+
+/** Outside undo history: loading or restoring is not something to undo into. */
+export function applyScriptToEditor(
+	editor: Editor,
+	script: string,
+	connectors: ConnectorRegistry,
+): void {
+	const scenes = deserializeWithScenes(script, connectors);
+
+	const replaceChildren = () => {
+		Editor.withoutNormalizing(editor, () => {
+			Transforms.removeNodes(editor, {
+				at: [],
+				match: (_node, path) => path.length === 1,
+			});
+			Transforms.insertNodes(editor, scenes, { at: [0] });
+		});
+		Editor.normalize(editor, { force: true });
+	};
+
+	if (HistoryEditor.isHistoryEditor(editor)) {
+		HistoryEditor.withoutSaving(editor, replaceChildren);
+		return;
+	}
+
+	replaceChildren();
+}
--- a/lib/project/applyScript.ts
+++ b/lib/project/applyScript.ts
@@ -27,8 +27,7 @@ export function applyScriptToEditor(
 
 	if (HistoryEditor.isHistoryEditor(editor)) {
 		HistoryEditor.withoutSaving(editor, replaceChildren);
-		editor.history.undos = [];
-		editor.history.redos = [];
+		editor.history = { undos: [], redos: [] };
 		return;
 	}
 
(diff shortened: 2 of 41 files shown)


lib/project/autosave.ts
```diff
@@ -105,5 +114,18 @@ export function createAutosaver({
 		markSaved: () => {
 			lastSaved = buildInput();
 		},
+		suspend: () => {
+			schedule.flush();
```
umairnadeem, line 107: `schedule.flush()` isn't a snapshot, it just enqueues `save`, and `save` calls `getScript()` (serializes `editor.children`) whenever it actually runs. if a save is already in flight when the user clicks a version, the flushed one queues behind it and by then `preview` has already written the old script into the editor, so the project row gets clobbered with the previewed version. reload during the preview and the live edits are gone. maybe pass the script into the flush so suspend captures it synchronously?

lib/project/autosave.ts
```diff
@@ -105,5 +114,18 @@ export function createAutosaver({
...
+			suspended = true;
+		},
+		resume: () => {
+			suspended = false;
```
umairnadeem, line 111: anything scheduled while suspended is dropped and `resume` doesn't replay it. a generation landing mid-preview schedules a save, the debounce fires, `suspended` swallows it, and nothing re-schedules, so close the tab while previewing and that result never hit the project row. want a `pendingWhileSuspended` flag that `resume` re-schedules on?

lib/project/canvasHistory.ts
```diff
@@ -0,0 +1,117 @@
...
+			],
+		});
+	};
+
+	preview = async (id: string): Promise<void> => {
+		if (this.state.previewId === id) return;
+		const content = await this.storage.read(id);
+		if (!this.stash) this.stash = this.document.read();
+		this.autosaver.suspend();
+		this.document.write(content);
```
umairnadeem, line 96: `document.write` calls `queue.replaceSnapshots`, which starts with `cancelAll()`, so just peeking at a version aborts every in-flight generation and clears `pending`. back-to-latest doesn't bring them back either, `settled()` forces the stashed statuses to idle, so kick off a video gen, click View, and it's gone with no trace. cancelling on `restore` makes sense, but on `preview` it contradicts the "edits can never be lost by looking at history" promise.

app/globals.css
```diff
@@ -152,6 +153,9 @@
 	--success-foreground: #fff;
 	--caution: var(--amber-500);
 	--caution-foreground: #fff;
+	--version-preview: var(--amber-100);
```
umairnadeem: `DESIGN.md` § Color says *"**State:** `--destructive` (red), `--success` (green), `--caution` (amber)"* and this quietly adds a fourth amber state family that isn't documented anywhere. either reuse `--caution` for the banner or add `--version-preview*` to DESIGN.md, otherwise there are two ambers and no rule for picking one.

lib/project/applyScript.ts
```diff
@@ -0,0 +1,31 @@
...
+				at: [],
+				match: (_node, path) => path.length === 1,
+			});
+			Transforms.insertNodes(editor, scenes, { at: [0] });
+		});
+		Editor.normalize(editor, { force: true });
+	};
+
+	if (HistoryEditor.isHistoryEditor(editor)) {
+		HistoryEditor.withoutSaving(editor, replaceChildren);
```
umairnadeem, line 29: `withoutSaving` keeps the swap itself out of the stack, but the undos already on `editor.history` survive and now point at a document that's gone. this only ran at load before (empty stack), now preview/restore run it mid-session: view an older version, hit cmd+Z, and slate replays inverse ops at paths that no longer exist. clear `editor.history.undos/redos` here? load and restore both start a fresh doc anyway.

lib/project/autosave.ts
umairnadeem: Fixed in 2fa489f9. Rather than passing the script into `flush`, the payload is now read when the debounce fires instead of when the save runs: `schedule` calls `buildInput()` itself and the queued task just persists that frozen input.

Suspend-time capture falls out of that for free, and it fixes the general case too, not just preview: any save queued behind a slow one now writes the state it was actually scheduled for.

Covered by "saves the document as it stood when suspend was called", which I checked fails against the old implementation.

lib/project/autosave.ts
umairnadeem: Fixed in 2fa489f9, but as one line rather than a flag: `resume` now just calls `schedule()`. The `isEqual(input, lastSaved)` check already makes it a no-op when nothing changed, so tracking a `pendingWhileSuspended` bit was more bookkeeping than the problem needed.

One caveat worth recording: the specific scenario you describe can't happen any more. Preview still cancels in-flight generations (see the other thread), so nothing can land mid-preview. The line still catches the store and queue subscribers that fire during `document.write` while suspended, but it's insurance now rather than a fix for a live bug. Easy to drop if you'd rather close this as won't-fix.

lib/project/canvasHistory.ts
umairnadeem: Not fixing this one. Recording it as a known trade-off instead.

I tried the obvious split (preview swaps the script only, restore adopts the version whole) and it costs more than it saves:

- Versions stop rolling back metadata, so restoring an old script wouldn't bring back characters, style, narration or videoSettings. An old script can then reference a character that's since been renamed or deleted.
- It also stops rolling back which element version was current.
- It creates a failure case the current design doesn't have: with generations surviving a preview, a result landing mid-preview has nowhere to go, because the stash is frozen at preview start. Back-to-latest would clobber it.

That last point is the interesting one. Cancelling on preview is exactly what makes the frozen stash a *complete* description of what to return to. The single-write model is coherent because of the cancellation, not in spite of it.

So `preview` keeps doing `document.write(content)`. The cost stands as you wrote it: kick off a video gen, click View, and it's gone. If it bites, the cheap fix is gating the View button on `queue.isBusy()` rather than restructuring the write path.

Leaving this thread open so the trade-off stays visible rather than quietly closed.

app/globals.css
umairnadeem: Fixed in 2fa489f9, taking the "reuse `--caution`" option.

The three `--version-preview*` tokens are gone. The banner now uses `--caution-soft` / `--caution-soft-foreground`, which mirrors the existing `--accent-soft` and hangs off the same amber ramp, so there's one amber again. Dropped the separate link token entirely: the underline plus `hover:decoration-2` already carries the affordance, so it didn't need its own colour.

DESIGN.md § Color documents the soft pairing and the rule for picking: one ramp per state, never add a second red, green, or amber.

lib/project/applyScript.ts
umairnadeem: Good catch, this was a real crash. Fixed in 15398699 by clearing the stack, as you suggested.

Confirmed the reasoning first: I built a two-scene doc, edited inside the second scene so the undo entry pointed at path `[1, 0, 0]`, then applied a one-scene script. The entry survives and `[1]` no longer exists. `applyScriptToEditor` now clears `editor.history.undos` and `redos` after the swap, which is the right place for it since any caller doing a full replacement invalidates the stack regardless of why.

Covered by "drops undo entries that point into the replaced document", which fails against the previous code.

lib/project/canvasHistory.ts
```diff
@@ -0,0 +1,117 @@
...
+			this.setState({ versions: await this.storage.list(), status: "ready" });
+		} catch (err) {
+			console.error("Failed to read canvas history", err);
+			this.setState({ status: "failed" });
+		}
+	};
+
+	record = async (content: ProjectContent): Promise<void> => {
+		const open =
+			Date.now() < this.foldUntil ? this.state.versions[0] : undefined;
```
umairnadeem: the open version is tracked positionally, so an in-flight `load()` can make this fold into the wrong row: `list()` runs before a `create` commits, resolves and replaces `versions` with the DB list (no new row), and the next save still has `foldUntil` set → `update(versions[0].id, ...)` clobbers an old checkpoint with current content. holding the open version's id on the class instead of reading index 0 defines it out of existence.

umairnadeem: Confirmed and fixed in d5239920. This one loses data, so thanks for spotting it.

I reproduced the exact sequence before changing anything: a stubbed `list()` that resolves after `create` commits but returns rows without the new one. The stale list replaces `versions`, the fold window is still open, and the next `record` updates `versions[0]`, which is now an older checkpoint. The test asserts the pre-existing row still holds its original content and fails against the previous code.

Took your suggestion and went one step further: rather than keeping an `openVersionId` alongside `foldUntil`, the two are now one field, since they're the same fact.

```ts
private open: { id: string; until: number } | null = null;
```

`record` reads `this.open` instead of index 0, and `restore` sets it to `null` instead of zeroing `foldUntil`. Folding no longer depends on what the list happens to hold, and there's no way for the id and the window to disagree.

Two things this deliberately doesn't cover, both benign:

- The stale `list()` still drops the new row from the displayed list until the next save prepends it back. Cosmetic and self-healing, so I left it rather than adding load/record sequencing.
- Two `record` calls can overlap, since the `onProjectSaved` listener doesn't await, and both would then create rather than fold. Worst case is an extra checkpoint row, not a lost one.

PR #657, opened 2026-08-27
update styling of timecode text

umairnadeem:
Make it consistent with the rest of the app


umairnadeem (commented):
lgtm, clean migration. verified no `font-mono`/`tabular-nums`/plex refs left anywhere, and the new Sloptastic.woff2 still carries `tnum` (only the name table changed vs the old one).

PR #659, opened 2026-08-27
fix: nightly bug fixes 2026-08-27

umairnadeem:
Four real correctness bugs plus tests for the two that sit on pure paths.

## Bugs fixed

**`lib/canvas/parseXmlTag.ts` — attribute values lost their whitespace.**
The parser split the tag on `/\s+/` and rejoined with a single space before running the attribute regex, so `prompt="line one\nline two"` round-tripped as `line one line two` and any run of spaces collapsed to one. The regex now scans the trimmed tag directly and the tag name is taken from the first whitespace boundary.

**`lib/providers/cache.ts` — a cache row with no duration rehydrated as `NaN`.**
`audioBundleCache.fromMetadata` did `Number(m.duration)` unguarded. Legacy `audioUrl` rows written before the rename carry no `duration` key, so a hit produced `durationSec: NaN`, which travels into the timeline as a NaN-length sequence. It now forces a miss, which is exactly what the `fromMetadata` contract documents for an unrehydratable row.

**`app/components/AuthForm.tsx` — a thrown magic-link call left the form stuck.**
`setLoading(false)` sat after the await with no `try`. If `sendMagicLink` threw (client construction, network), loading stayed `true` forever, the submit button stayed disabled, and no error was shown. Now wrapped in `try/catch/finally` matching the existing `handleGoogle` shape.

**`lib/components/Waveform.tsx` — unhandled `play()` rejection toasted at the user.**
`audio.play()` was called without handling its promise. `GlobalErrorToaster` listens on `unhandledrejection`, so every pause-during-play raised a spurious "The play() request was interrupted…" toast. The rejection is now handled: `AbortError` is normal use and ignored, anything else goes through `toastError`.

## Tests

- `lib/canvas/__tests__/parseXmlTag.test.ts` — whitespace runs and newlines survive inside an attribute value; attributes following a multi-line value still parse.
- `lib/providers/__tests__/cache.test.ts` — `fromMetadata` misses on a row with a missing or non-numeric `duration`.

Skipped tests for `AuthForm` and `Waveform`: vitest here has no jsdom environment and no rendering harness (all existing tests are pure-function), so covering them would mean standing up a component-test setup, which is out of scope for a bug-fix pass.

## Open PR overlap check

Considered #654 (canvas version history — `lib/project/**`, `lib/generation/**`, canvas hooks/panel/elements, `lib/format.ts`), #657 (timecode styling — `app/components/video/**`, `components/ui/*`, `app/globals.css`, `DESIGN.md`), #658 (`app/components/video/PlayerPositionContext.tsx`). This PR touches `app/components/AuthForm.tsx`, `lib/canvas/parseXmlTag.ts`, `lib/components/Waveform.tsx`, `lib/providers/cache.ts` and their tests — no file or theme overlap.

## Validation

`npm run lint`, `npx tsc --noEmit`, `npm test` (156 files, 1394 tests), and `npm run build` all pass.

### DIFF
--- a/app/components/AuthForm.tsx
+++ b/app/components/AuthForm.tsx
@@ -41,18 +41,23 @@ export default function AuthForm({
 		setLoading(true);
 		setError("");
 
-		const { error } = await sendMagicLink({
-			email,
-			shouldCreateUser,
-			data: otpData,
-		});
+		try {
+			const { error } = await sendMagicLink({
+				email,
+				shouldCreateUser,
+				data: otpData,
+			});
 
-		if (error) {
-			setError(error);
-		} else {
-			setSent(true);
+			if (error) {
+				setError(error);
+			} else {
+				setSent(true);
+			}
+		} catch (cause) {
+			setError(errorMessage(cause));
+		} finally {
+			setLoading(false);
 		}
-		setLoading(false);
 	};
 
 	const handleGoogle = async () => {
--- a/lib/canvas/__tests__/parseXmlTag.test.ts
+++ b/lib/canvas/__tests__/parseXmlTag.test.ts
@@ -26,6 +26,22 @@ describe("parseXmlTag", () => {
 		});
 	});
 
+	it("preserves runs of whitespace and newlines inside attribute values", () => {
+		expect(parseXmlTag('image prompt="a  b\nc"')).toEqual({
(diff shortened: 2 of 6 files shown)


umairnadeem (commented):
lgtm

Issue #651, opened 2026-08-26
Add an on-theme 404 page: the page you asked for was generated, and it came out wrong


umairnadeem:
## Problem

We have no `not-found.tsx` anywhere in `app/`. `app/projects/[id]/page.tsx:29` already calls `notFound()` for a missing project, and any bad URL hits the same path, so today both render Next's stock default: white background, Helvetica, `404 | This page could not be found.` It ignores the theme, ignores our fonts, and is the one screen in the product that looks like it was shipped by a different company.

For an app called OpenSlop this is a wasted joke. The 404 is the cheapest surface in the product to be memorable on, and right now it is the only one with no personality at all.

## Reported by (verbatim feedback)

- "add an openslop github issue to create a funny, on-theme 404 page ideally sloppy themed and AI-forward. be creative"

## Relevant code

- `app/` — no `not-found.tsx` at any level; Next falls back to its built-in default
- `app/projects/[id]/page.tsx:29` — `if (!project) notFound()`, the one in-app caller today
- `app/loading.tsx` — the closest sibling for a full-page route-level surface, and the layout pattern to copy
- `app/layout.tsx:44` — `metadata`; the 404 route should export its own title
- `app/globals.css` — `.dot-grid-bg`, `.grain`, media-type tint tokens
- `DESIGN.md` — Sloptastic for the headline, `text-display`, disciplined-violet rule

## Suggested direction

Add `app/not-found.tsx` as a server component. Structural bones stay boring and correct — `min-h-screen bg-background text-foreground`, the `dot-grid-bg` backdrop as an `aria-hidden pointer-events-none` layer, content centered, one `accent` CTA back to the gallery and one `ghost` back to home. Tokens only, both themes, no one-off hex.

All the fun rides on top of that. The concept: **the page you asked for was generated, and it came out wrong.** Not "not found" — *found, but slop*. Lean into the failed-generation bit rather than a generic sad-robot page.

Ideas worth stealing from, pick the ones that land:

- **A 404 that hallucinated itself.** The big Sloptastic display number renders as `404`, but a couple of glyphs are subtly off — a mirrored `4`, six fingers on a hand-drawn zero. The kind of mistake a diffusion model makes and nobody catches.
- **Confidence score.** A small mono line under the headline: `confidence: 0.03` or `this page is 12% real`. IBM Plex Mono, `text-muted-foreground`. Sells the bit in six characters.
[...]

Issue #398, opened 2026-07-10
Model selection dropdown in character & narration edit modals


umairnadeem:
## Problem

The **character edit modal** and **narration edit modal** expose voice settings only — there's no way to pick the generation **model** from inside either modal. The element container **header** already has an editable model dropdown (`ModelBadge` → `AttributeBadge` → `SelectMenu`), but it isn't surfaced in the modals, so a user editing a character/narration can't change its model without leaving the modal.

This is the narrow, shippable slice of #370 (which additionally covers per-element *provider* selection and generalizing the control to every element type).

## Reported by (verbatim feedback)

- "add a model selection dropdown to character and narration edit modals (re-use relevant parts of the model dropdown components that exist on each element container header)"

## Reuse note (important nuance)

The header's `ModelBadge` / `AttributeBadge` write to the **Slate node** via `updateElementAttrs` and require `useSlateStatic()`, so they only work inside the canvas editor tree — they can't drop into a modal as-is. The modals persist to the **project store** (`MetadataVoice` / `MetadataCharacter`) via `setCharacter` / `setNarration`, and those types have **no `model` field today**. So the pieces to reuse are the lower-level `SelectMenu` (or the `EnumField` wrapper) fed by the same `connectorConfig[connector][provider].models` list the header badge uses, plus a new persisted field on the voice/character metadata.

## Relevant code

- `app/components/canvas/elements/ModelBadge.tsx:12` — header's editable model dropdown; the pattern to mirror (enum options from `connectorConfig[connector][provider].models`)
- `components/ui/select-menu.tsx:19` — `SelectMenu<T>` primitive, decoupled from Slate — directly reusable in a modal
- `app/components/canvas/elements/character/VoiceMetadataFields.tsx:15` — shared `VoiceSection` used by both modals; natural host for the model dropdown row (already uses `EnumField`)
- `app/components/canvas/elements/character/fields.tsx:77` — `EnumField<T>` wrapper for in-modal fields, styled consistently with the other fields
- `app/components/canvas/elements/character/CharacterEditModal.tsx:38` — character modal; already has `connectorConfig` in scope
- `app/components/canvas/elements/character/NarratorEditModal.tsx:15` — narration modal (`VoiceSection` only)
- `lib/project/types.ts:29-46` — `MetadataVoice` / `MetadataCharacter`; add the persisted `model` field here
[...]

Issue #366, opened 2026-06-25
Generate one scene at a time (per-scene generate button)


umairnadeem:
## Problem

Generation is all-or-nothing at the project level: the only top-level trigger is **Generate All** (`EditorToolbar`), which queues every stale element across every scene. Users want to generate **one scene at a time** — produce/iterate a single scene, review it, then move on — instead of regenerating the whole project.

## Reported by (verbatim feedback)

- "Want to generate one scene at a time"

## What already exists

Per-**element** generation is already built: `useGenerate` + `ElementGenerateButton` schedule a single element's job. The missing layer is per-**scene** generation (a scene = a `SceneElement` with a `.children` array of content elements).

## Relevant code

- `app/components/canvas/hooks/useGenerateAll.ts:1-31` — current "Generate All"; filters `getContentElements()` and batches all jobs
- `app/components/canvas/hooks/useGenerate.ts:14-54` — existing **per-element** generate hook to mirror at scene scope
- `app/components/canvas/elements/SceneContainer.tsx:55-88` — scene header (`rightSlot`, line ~80) where a "Generate Scene" button would sit, next to Play-from-here / Delete
- `lib/canvas/scenes.ts:1-24` — scene model; `getContentElements()` filters scene children
- `app/components/EditorToolbar.tsx:68-83` — global "Generate All" button for reference

## Suggested direction

- Add a `useGenerateScene` hook: mirror `useGenerateAll` but filter to a single `scene.children` instead of all content elements.
- Wire a "Generate scene" button into the `SceneHeader` `rightSlot`, following the existing `PlayFromHereButton` / `DeleteButton` pattern.
- Reuse the existing queue + scheduler (`lib/generation/scheduleGeneration.ts`, `lib/generation/queue.ts`); per-scene scheduling is idempotent via the queue state machine, so it composes with the global path.

## Acceptance criteria

- [ ] Each scene has its own "Generate" control in its header.
- [ ] It queues only that scene's stale elements, not the whole project.
- [ ] The existing global "Generate All" continues to work unchanged.

## Related

- #306 (start per-node generation as it streams from the LLM) — per-node granularity, complementary
- #345 (collapse individual element containers) — same scene-as-a-unit interaction surface

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #661, opened 2026-08-27
feat: fit clip durations to the dialogue under them

### LINKED ISSUE #599: Fit foreground element durations to the dialogue that runs under them
## Problem

A foreground element's duration is authored blind. `duration` is a plain enum attribute on animated images and clips (`lib/connectors/attributes/common.ts:35-41`, defaulting to `"10"`), picked from a dropdown with no relationship to how long the dialogue under it actually runs.

The timeline then papers over the mismatch instead of fixing it. In `buildVideoLayout`, a foreground element opens a scene at its own `durationSec`, and every following overlay (narration / character TTS) stretches that scene with `Math.max` (`lib/video/scene-builder.ts:118-141`). So the audio is never cut — but a 5s clip sitting under 14s of narration simply runs out, and the scene holds on a dead frame for nine seconds. Generate a longer clip than the speech needs and you've paid for video nobody sees.

Everything needed to compute the right number already exists and is already trusted:

- `countSpokenWords` filters to `connector === "tts"` elements and counts their body text (`lib/canvas/spokenWords.ts:9-17`)
- `NARRATION_WORDS_PER_MINUTE = 180` is the established words→seconds conversion (`lib/video/videoLength.ts:18`)
- `count_words` already applies both to judge total runtime against the project's length target (`lib/agent/tools/countWords.ts:24-40`)

The only missing piece is doing it **per foreground element** rather than per project, and writing the answer back.

- "using the existing word count functionality, we should count the word counts for all dialogue for a scene (i.e. all narration/character tts in between two visual/foreground elements) and assign the leading visual/foreground element a duration just long enough to cover the dialogue (with some leeway). this could be facilitated with a tool by sloppy."

## Relevant code

- `lib/canvas/spokenWords.ts:9-17` — `countSpokenWords`: counts across the whole canvas; the grouping this needs is the same filter applied per run of elements
- `lib/video/videoLength.ts:18` — `NARRATION_WORDS_PER_MINUTE`, the single conversion constant both the estimate and this feature should share
- `lib/video/scene-builder.ts:118-141` — the run semantics are already encoded here: foreground opens a scene, following overlays extend it. "Dialogue between two foreground elements" is exactly this grouping
- `lib/canvas/types.ts:20-63,72-75` — `ELEMENT_TYPES` roles + `FOREGROUND_TYPES` (`image`, `animated_image`, `clip`); overlay TTS types are `narration` and `character`
- `lib/canvas/types.ts:77-79` — `DURATION_OPTIONS` is 4–15 seconds; a hard ceiling any computed duration must clamp to, and a real limit worth surfacing
- `lib/connectors/animated_image/attributes.ts:17` / `lib/connectors/video/attributes.ts:5` — where `duration` lives; note `image` elements have no duration attribute at all, so a still already stretches to fit
- `lib/video/resolve.ts:42` — the rendered `durationSec` comes off the *generated result*, so `duration` is a generation input: changing it stales the node and requires a regenerate
- `lib/script/refine/types.ts:26-31` — `set` ops already write arbitrary attrs, so applying the result needs no new mutation path
- `lib/agent/tools/countWords.ts:24-40` — the tool shape to follow: read canvas, compute, return a sentence the model can act on
- `lib/agent/tools/registry.ts:20-33` — registration point for a new tool

## Suggested direction

**1. A pure helper first, in `lib/`.** Something like `dialogueRuns(descendants)` that walks the content elements and returns, per foreground element, the spoken-word count of the TTS elements that follow it up to the next foreground element. This is the reusable knowledge; it belongs beside `countSpokenWords`, not inside a tool. Reuse `NARRATION_WORDS_PER_MINUTE` to turn words into seconds, add a leeway (a small fixed pad plus a percentage, one named constant), then clamp to `DURATION_OPTIONS`.

**2. A `fit_durations` tool for Sloppy.** It reports the suggested duration per foreground element against the current one, and applies the changes through the existing attribute-set path. Keep it a normal tool: input can scope to a scene or an element id, default to the whole canvas.

**3. Report, don't silently regenerate.** `duration` is a generation input (`lib/video/resolve.ts:42`), so changing it stales the animated image or clip. The tool should say which elements changed and that they need regenerating, consistent with how the agent already flags stale elements after edits — not kick off generation itself.

**4. Say what it couldn't do.** Dialogue longer than 15s cannot be covered by one element under `DURATION_OPTIONS`. The tool should name those elements rather than silently clamping, since the real fix is splitting the dialogue across visuals — a suggestion the agent can then act on. Stills should be reported as already-fitting rather than treated as failures.

Because the helper is pure and separate, the same computation can later back a manual "fit to dialogue" button on the element card without duplicating anything.

## Acceptance criteria

- [ ] A pure, tested helper maps each foreground element to the spoken-word count of the dialogue that runs under it, using the same TTS filter and words-per-minute constant as `countSpokenWords` / `count_words`.
- [ ] Suggested durations include a leeway defined by one named constant and clamp to `DURATION_OPTIONS`.
- [ ] Sloppy can fit durations across the canvas or a single scene in one tool call.
- [ ] The tool reports what changed and that those elements need regenerating; it does not trigger generation.
- [ ] Elements whose dialogue exceeds the maximum duration are named explicitly rather than silently clamped.
- [ ] Image (still) foregrounds are handled correctly — they carry no duration attribute and already stretch.

## Related

- #251 (animated-image & clip controls, incl. durations over 5s) — the manual side of the same knob; raising the ceiling there directly widens what this can fit
- #600 (explicit timing behavior for clips and animated images) — the timeline-side counterpart: this issue picks a `duration` to generate, #600 decides what the timeline does when the generated clip and the dialogue still disagree. #600 also fixes Runware returning `durationSec: 0`, which is why a mismatch is currently invisible
- #597 (`set_caption_style` tool) — same pattern: an existing typed setting given an agent tool

---

umairnadeem:
## Summary

`duration` on an `animated_image` or `clip` was authored blind: a dropdown with no relationship to how long the dialogue under the element actually runs. A 5s clip under 14s of narration runs out and holds a dead frame; a 15s clip under 3s of dialogue is generated video nobody sees.

This adds `fit_durations`, a Sloppy tool that sets each clip's `duration` to just cover the speech that follows it, up to the next visual.

- `lib/video/durationFit.ts` — pure helper. Turns the spoken-word count already measured per visual into the duration option that covers it, using the existing `secondsForWords` / `NARRATION_WORDS_PER_MINUTE` conversion, one named leeway constant (`DURATION_FIT_LEEWAY_SEC`), and a clamp to `DURATION_OPTIONS`. Stills carry no duration and already stretch, so they are excluded rather than treated as failures.
- `lib/video/elementLengths.ts` — `ElementLength` now carries the element's own `durationSec` where its type has one, so the fit can be compared against what is set today. `ownSeconds` became `ownDuration`; the layout math is unchanged.
- `lib/agent/tools/fitDurations.ts` — the tool. Scopes to a `scene` or `element_ids`, defaults to the whole canvas, applies through the existing `editScript` `set` path, and reports what changed. It does not trigger generation: it says the elements are stale and need regenerating, matching how the agent already flags edits.
- Elements whose dialogue outruns the longest option are named explicitly with the suggestion to split the dialogue across more visuals, rather than silently clamped.

The helper is pure and separate from the tool, so a manual "fit to dialogue" button on the element card can back onto the same computation later.

Note that acceptance criterion 1 (a pure helper mapping each foreground element to the dialogue under it) already landed as `measureElementLengths` / `measure_element_lengths`; this builds on it rather than duplicating it.

`size: M`, `priority: medium`, unambiguous acceptance criteria, and an established pattern to follow (`set_caption_style`, #647). Everything it needs already exists: the per-element word counts, the words-to-seconds constant, and the attribute-set mutation path. No new context capability, no UI decisions, no product judgment.

## Validation

Full CI-equivalent suite, all green:

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` — 157 files, 1400 tests passed

New coverage: 5 tests over `durationFits` (leeway, rounding up to a whole option, the no-dialogue floor, the clamp plus the falls-short flag, stills excluded) and 4 over the tool through `executeToolCall` (the ops it emits and its report, the named too-long element, the no-change path, the stills-only path).

## Open PR overlap check

Checked #660, #659, #658, #657, #654 by `gh pr diff --name-only`. No file, module, or theme in common: #660 is `lib/video/render-api.ts`, #659 is `lib/canvas/parseXmlTag.ts` / `lib/providers`, #658 is `PlayerPositionContext`, #657 is timecode styling, #654 is `lib/generation` / `lib/project` canvas history.

## Skipped

- #492 (separate still/video model selection) — spans connectors, the OSML attribute contract shown to the LLM, and the model badge UI, with two open design questions the issue itself flags as unsettled.
- #632 (agent voice reselection) — needs a new search capability on the tool context plus a decision on when `resolvedVoiceId` is cleared.
- #461 (karaoke word highlight) — presentational, and picks a highlight style the issue leaves open.
- #316, #345, #365, #366, #398 and the rest of `size: M` — broader UX or product direction.

Left for outside contributors: #651 and #254 (`good first issue`), and every `size: XS`.

Closes #599

### DIFF
--- a/lib/agent/tools/fitDurations.ts
+++ b/lib/agent/tools/fitDurations.ts
@@ -1,0 +1,86 @@
+import dedent from "dedent";
+import { z } from "zod";
+import {
+	DURATION_FIT_LEEWAY_SEC,
+	LONGEST_DURATION_SEC,
+	durationFits,
+	type DurationFit,
+} from "@/lib/video/durationFit";
+import type { ElementLength } from "@/lib/video/elementLengths";
+import { defineTool } from "./defineTool";
+
+const sec = (seconds: number) => `${seconds.toFixed(1)}s`;
+
+const where = ({ sceneNumber, type, id }: ElementLength) =>
+	`Scene ${sceneNumber} ${type} ${id}`;
+
+const changeLine = ({ length, duration, dialogue }: DurationFit) =>
+	`${where(length)}: ${length.durationSec}s to ${duration}s, for ${sec(dialogue)} of dialogue.`;
+
+const shortLine = ({ length, needed }: DurationFit) =>
+	`${where(length)} needs ${sec(needed)} but ${LONGEST_DURATION_SEC}s is the longest a clip can be generated at; split the dialogue after it across more visuals.`;
+
+export const fitDurations = defineTool({
+	description: dedent`
+	  Fit every animated_image and clip to the dialogue that runs under it: set each one's
+	  \`duration\` to just cover the speech that follows it, up to the next visual, plus
+	  ${DURATION_FIT_LEEWAY_SEC}s of leeway. Shorter clips stop dead under long dialogue;
+	  longer ones are generated video nobody sees.
+
+	  Scope it with \`scene\` or \`element_ids\`, or send neither for the whole canvas.
+
+	  Images carry no duration and already stretch, so they are left alone. Changing a
+	  duration stales the element: tell the user to regenerate it afterwards.
+	`,
+	input: z.object({
+		scene: z.number().int().positive().optional(),
+		element_ids: z.array(z.string()).min(1).optional(),
+	}),
+	output: z.string(),
+	execute: async ({ scene, element_ids }, ctx) => {
+		const ids = element_ids && new Set(element_ids);
+		const scoped = ctx
+			.measureElementLengths()
+			.filter(
+				(length) =>
+					(scene === undefined || length.sceneNumber === scene) &&
+					(!ids || ids.has(length.id)),
+			);
+
+		const fits = durationFits(scoped);
+		if (fits.length === 0) {
+			return "No animated_image or clip in scope. Images carry no duration and already stretch to the dialogue under them.";
+		}
+
+		const changes = fits.filter(
+			({ length, duration }) => length.durationSec !== duration,
+		);
+		const short = fits.filter(({ short }) => short);
+		const stills = scoped.length - fits.length;
+
+		const { applied, failures } = changes.length
+			? ctx.editScript(
+					changes.map(({ length, duration }) => ({
+						op: "set" as const,
+						id: length.id,
+						attrs: { duration: String(duration) },
+					})),
+				)
+			: { applied: 0, failures: [] };
+
+		return [
+			changes.length === 0
+				? `All ${fits.length} already cover their dialogue; nothing to change.`
+				: `Fitted ${applied} of ${changes.length}. Those elements are stale now and need regenerating.`,
+			...changes.map(changeLine),
+			...short.map(shortLine),
+			stills > 0 &&
+				`${stills} image still${stills === 1 ? "" : "s"} left alone.`,
+			failures.length > 0 &&
+				`Failed: ${failures.join("; ")}. Read the script again; the ids may be stale.`,
+		]
+			.filter(Boolean)
+			.join("\n");
+	},
+	rewritesCanvas: true,
+});
--- a/lib/agent/tools/fitDurations.ts
+++ b/lib/agent/tools/fitDurations.ts
@@ -21,6 +21,39 @@ const changeLine = ({ length, duration, needed }: DurationFit) =>
 const shortLine = ({ length, needed }: DurationFit) =>
 	`${where(length)} needs ${sec(needed)} but ${DURATION_MAX}s is the longest a clip can be generated at; split the dialogue after it across more visuals.`;
 
+const report = (...lines: (string | false)[]) =>
+	lines.filter(Boolean).join("\n");
+
+const isUnfitted = ({ length, duration }: DurationFit) =>
+	length.durationSec !== duration;
+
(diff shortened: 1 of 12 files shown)


lib/agent/tools/fitDurations.ts
```diff
@@ -0,0 +1,86 @@
...
+						op: "set" as const,
+						id: length.id,
+						attrs: { duration: String(duration) },
+					})),
+				)
+			: { applied: 0, failures: [] };
+
+		return [
+			changes.length === 0
+				? `All ${fits.length} already cover their dialogue; nothing to change.`
```
umairnadeem: a clip that's already clamped at 15s but needs 61s lands in `short` and not in `changes`, so this reports "All 1 already cover their dialogue" and then immediately contradicts itself with the shortLine. your own test fixture (clip1, scene 2) hits it. maybe key the no-change line off `changes.length === 0 && short.length === 0`?

PR #662, opened 2026-08-27
feat: select the still-image model separately on animated_image

### LINKED ISSUE #492: Select the still-image model and the video model separately on animated_image

umairnadeem:
## Summary

An `animated_image` runs two generations (a still, then an image-to-video animation of it) but only exposed one model choice: `model`, the video model. The still frame silently fell through to the image registry default, so the generation that decides composition, likeness and style was unpickable.

This adds a `stillModel` attribute alongside `model`:

- `stillElement` strips `stillModel` from the still's attributes and writes it back as the still's `model`, so each generation only ever sees a model name its own connector understands. Unset keeps today's behaviour exactly (the image connector default), so saved projects are unaffected.
- The still-frame plugin drops `stillModel` from the video params, so a still model name never reaches the video provider.
- `ModelBadge` now renders through a shared `ModelAttributeBadge`, with a second thin wrapper (`StillModelBadge`) for the still. One badge component serves both selectors; an animated image shows "Image model" and "Video model", every other element is unchanged.
- `AttributeBadge` gained an optional `fallback` so the still badge can display the connector default while the attribute is unset, rather than an em dash.

Because the still's `model` is part of the still node's inputs, changing it invalidates the still and restages the animation through the existing dependency-graph rules; no extra invalidation wiring.

Issue #492 is `size: M`, `priority: high`, with unambiguous acceptance criteria and a single change point named in the body. The code area is small and fully covered by unit tests.

## Deliberately out of scope

Two open questions the issue raised, and one criterion left undone:

- **`stillProvider`** is not introduced. Only `openslop` is registered per connector type, so a still model alone is sufficient today; the issue explicitly allows deferring this.
- **The OSML spec is unchanged**, so the LLM does not set `stillModel`. It reads as a user preference, and hand-authored OSML still round-trips it (the parser passes unknown attributes through).
- **Carrying the model pick across an `image` ↔ `animated_image` type switch** (the last acceptance-criteria box) is not implemented. `createCanvasNode` stamps the connector default over any incoming `model`, so the `animated_image` → `image` direction cannot be expressed through `preservedAttributes` without changing element-creation semantics, and doing only the other direction would be asymmetric. Worth a follow-up on its own, since the same clobber also drops a `model` attribute from pasted OSML.

## Validation

- `npm run build` — succeeded
- `npm run test:coverage` — 156 files, 1396 tests passing, including 5 new still-model tests (translation, default fall-through, no cross-connector model leak, restaging)
- Playwright smoke tests not run locally (need Supabase env); no routes or app copy changed

## Open PR overlap check

Re-listed open PRs (#654, #657, #658, #659, #660, #661) and diffed their file lists. None touches `lib/connectors/animated_image/**`, `lib/connectors/types.ts`, `ModelBadge.tsx`, `AttributeBadge.tsx` or `ElementContainer.tsx`. #661 (issue #599) claims the agent-tool surface, which is why the agent-facing issues were passed over.

Closes #492


app/components/canvas/elements/ModelBadge.tsx
```diff
@@ -35,12 +59,38 @@ export function ModelBadge({ element }: { element: CanvasContentElement }) {
...
 
 	return (
-		<AttributeBadge
+		<ModelAttributeBadge
 			element={element}
 			attrKey="model"
-			spec={spec}
-			className="text-label-xs"
+			// An animated image picks two models, so its own badge says which.
+			label={type === "animated_image" ? "Video model" : "Model"}
```
umairnadeem: afaict these labels never render: `AttributeBadge` only shows `spec.label` when there's no icon, and both badges pass `Codesandbox`. so an animated image ends up with two identical-looking pills and "Image model" vs "Video model" only lives in the tooltip/aria-label. different icon for the still one?

lib/canvas/createCanvasNode.ts
```diff
@@ -14,31 +17,24 @@ type Opts = {
...
-		connectors,
-		connector,
-	);
-	const schema = resolveAttributeSchema(
-		connector,
-		provider,
-		connectorConfig?.defaultModel,
-	);
+	const model =
+		opts.attrs?.model ?? defaultModelFor(connector, opts.projectModels);
```
umairnadeem: incoming `model` is taken on faith here while the pinned one goes through `catalog.has`. paste OSML with `model="Slop Video v1"` onto an image element and that name now sticks on the node forever, so every generate 400s at `modelField`. before this diff the default clobbered it. same guard as `defaultModelFor` would do it: `catalog.has(opts.attrs?.model)` else default.

lib/project/types.ts
```diff
@@ -77,6 +77,8 @@ export const MetadataSchema = z.object({
 	narration: MetadataVoiceSchema.default({}),
 	characters: z.record(z.string(), MetadataCharacterSchema).default({}),
 	videoSettings: VideoSettingsSchema,
+	/** The model each connector type generates with, when the project pins one. */
+	connectorModels: z.record(z.string(), z.string()).default({}),
```
umairnadeem, line 81: nothing writes `connectorModels`, afaict, no UI, no tool, no migration. so it's a persisted metadata field plus a `ConnectorModels` type plus a `projectModels` param on 6 signatures that can only ever be `{}`. wfm to drop it and have `defaultModelFor(type)` just read the catalog, then add it back when something can actually pin a model?

umairnadeem: future proofing

lib/connectors/animated_image/plugins/still-frame.ts
```diff
@@ -37,14 +34,14 @@ export const stillElementId = (elementId: string) =>
...
 	return {
 		...element,
 		id: stillElementId(element.id),
 		type: "image",
 		generationAttributes: {
-			...omit(element.generationAttributes ?? {}, VIDEO_ONLY_KEYS),
-			// The element's own provider generates the video, not the still.
-			provider: DEFAULT_PROVIDER,
+			...omit(attributes, VIDEO_ONLY_KEYS),
+			...(stillModel && { model: MODEL_CATALOGS.image.resolve(stillModel) }),
```
umairnadeem, line 44: `model` gets normalized on the way in by `createCanvasNode.resolve`, but `stillModel` doesn't, so this is the only place the bad name gets caught. paste `<animated_image stillModel="Slop Video v1">` and the node keeps that forever: the "Image model" badge shows a video model that's not even in its own options list, while generation quietly runs the image default. same guard at creation time would keep the stored value honest.

umairnadeem (commented):
lgtm. resolving at creation catches the stillModel case and the rest of the enums come along for free, and since the filter only runs on create, fit_durations and the popover paths are untouched.

PR #665, opened 2026-08-28
chore: rebalance per-media generation concurrency limits

umairnadeem:
## Summary
- Raise in-flight limits for `video`, `image`, and `animated_image` from 3 to 4
- Drop `tts`, `music`, and `sfx` from 2/5/5 to 1 so audio jobs stop competing with visual generation for shared provider rate limit headroom

## Test plan
- [ ] Generate a script with mixed media and confirm visual jobs saturate at 4 while audio runs one at a time
- [ ] No provider 429s during a full generation run

### DIFF
--- a/lib/generation/concurrency.ts
+++ b/lib/generation/concurrency.ts
@@ -5,16 +5,15 @@ export type ConcurrencyLimits = Record<AssetConnectorType, number>;
 
 /**
  * Hardcoded until BYOK, where these become per-user settings: the limits exist
- * to keep our shared provider keys under their rate limits, so heavier media
- * gets a smaller share.
+ * to keep our shared provider keys under their rate limits
  */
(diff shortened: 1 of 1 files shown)


lib/generation/concurrency.ts
```diff
@@ -5,16 +5,15 @@ export type ConcurrencyLimits = Record<AssetConnectorType, number>;
 
 /**
  * Hardcoded until BYOK, where these become per-user settings: the limits exist
- * to keep our shared provider keys under their rate limits, so heavier media
- * gets a smaller share.
+ * to keep our shared provider keys under their rate limits
  */
 const DEFAULT_CONCURRENCY_LIMITS: ConcurrencyLimits = {
 	video: 3,
```
umairnadeem, line 11: description says video/image/animated_image go 3 -> 4, but the diff leaves all three at 3 and only touches the audio types. did the visual bump get dropped, or is the description stale?

PR #663, opened 2026-08-28
feat: add on-theme 404 for failed generations

VedantMadane:
﻿## Summary

- Add `app/not-found.tsx` so unknown URLs and `notFound()` from the project page use the OpenSlop design system instead of Next stock white Helvetica 404.
- Concept: the page was generated and came out wrong (hallucinated display 404, confidence line, deadpan excuse from request headers, failed element card).
- Accent CTA and ghost exit back to `/`; semantic tokens only; server component; metadata title `404 · OpenSlop`.

Closes #651

## Test plan

- [x] eslint + prettier on `app/not-found.tsx` (lint-staged / husky)
- [ ] Visit a nonsense path (e.g. `/this-is-not-a-route`) in light and dark
- [ ] Open a missing project id so `notFound()` hits the same page
- [ ] Tab to both exit buttons; confirm keyboard reachability
- [ ] Confirm no raw hex / no `text-[Npx]` arbitraries

## Screenshots

Light/dark screenshots pending preview deploy (Vercel) or maintainer check. Structure matches `loading.tsx` + DESIGN.md tokens (`dot-grid-bg`, `text-display`, `font-numeric`, `bg-element-card`, media-image tint).

PR #668, opened 2026-08-28
fix: nightly bug fixes 2026-08-28

umairnadeem:
Three correctness bugs, each with a test that fails on master and passes here.

## Scene seek skipped a scene whose first foreground was ungenerated

`findSceneSequence` took `children.find(isForeground)` and looked that one element up in the sequence index. An element with no generated result never reaches the layout, so a scene whose *first* foreground is ungenerated resolved to `undefined` — while `buildSceneSegments` still emitted a segment for that scene off its second foreground. The seek bar showed the scene, "play from here" and the scene container's seek did nothing.

Now it returns the first foreground that actually made it into the layout.

## A retry displayed the previous attempt's error

`enqueueGraph` set `status`, `seconds` and `connectorType` but left `error` in place. After a failure, hitting Generate again put the element in `generating` with the old error string still on its snapshot, which `useGenerate` hands straight to the card. Cancelling that retry calls `resetToIdle`, which preserves `error`, so the element sat idle showing a failure from a superseded attempt.

Queuing a node for a new attempt now clears `error`. `blockedByError` / `releaseBlocked` are unaffected: the errors they read are set within the same run, after the enqueue.

## The OSML stream parser did not report appended nodes

`parseBuffer` set `updated = true` only in the text branch, never when a tag opened a new node, and `useOSMLStreamParser` gates `setNodes` on that boolean. A chunk that only opened tags left the parser's node list ahead of React state. Body-less tags — the `metadata_*` elements, which carry their whole payload in attributes — never signalled at all, so they only reached the UI if a later chunk happened to contain text.

## Tests

5 cases added alongside the existing suites, all verified failing against the pre-fix sources:

- `lib/video/__tests__/sceneSegments.test.ts` — foreground present in the index, an ungenerated one skipped, none in the layout at all
- `lib/generation/__tests__/queue.test.ts` — error cleared while a retry is in flight, and no error left behind when that retry is cancelled
- `lib/canvas/__tests__/osmlStreamParser.test.ts` — an open tag and a body-less metadata tag each signal an update, a partial tag does not

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, and the full vitest suite (159 files / 1438 tests) all pass.

## Open PR overlap check

Open PRs at branch time: #663 (`app/not-found.tsx`) and #662 (canvas element badges, `lib/connectors/animated_image/**`, `lib/connectors/types.ts`). This PR touches only `lib/video/sceneSegments.ts`, `lib/generation/queue.ts`, `lib/canvas/osmlStreamParser.ts` and their tests — no file or theme overlap with either.

### DIFF
--- a/lib/canvas/__tests__/osmlStreamParser.test.ts
+++ b/lib/canvas/__tests__/osmlStreamParser.test.ts
@@ -221,3 +221,27 @@ describe("parseOSML", () => {
 		expect(getElementText(nodes[0])).toContain("Once upon a time");
 	});
 });
+
+describe("OSMLStreamParser update signalling", () => {
+	it("reports an update for a chunk that only opens a tag", () => {
+		const s = new OSMLStreamParser();
+		expect(s.appendChunk("<narration>", connectors)).toBe(true);
+		expect(s.getNodes()).toHaveLength(1);
+	});
+
+	it("reports an update for a body-less metadata tag", () => {
+		const s = new OSMLStreamParser();
+		expect(
+			s.appendChunk(
+				'<metadata_narration voiceId="v1"></metadata_narration>',
+				connectors,
+			),
+		).toBe(true);
+		expect(s.getNodes()).toHaveLength(1);
+	});
+
+	it("reports no update for a chunk that adds nothing", () => {
+		const s = new OSMLStreamParser();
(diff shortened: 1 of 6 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
overengineered

PR #669, opened 2026-08-28
refactor: nightly maintainability 2026-08-28

umairnadeem:
No behavior changes. Three related cleanups around the two history features, plus the docs they left stale.

## `CanvasProviders` owns the editor session

`CanvasHistoryProvider` landed with #654 at the top level, so `Editor` had gone back to destructuring the session and assembling a provider itself — the exact shape `CanvasProviders` exists to prevent, per its own doc comment.

`CanvasProviders` now calls `useEditorSession()` and composes `CanvasHistoryProvider` with the rest of the stack. `Editor` is down to the view switch and takes no props from the session at all, so the editor instance is no longer a prop anywhere.

## Element history: one status, not two booleans

`ElementHistory` exposed `isLoaded` and `isFailed` as independent booleans tracked in two structures, which made `loaded && failed` representable and pushed a four-arm nested ternary into the popover. The sibling feature added a week later (`CanvasHistory`) and `ClipWaveform` both already model this as a `status` union.

`ElementHistory.status(elementId)` now returns `"loading" | "ready" | "failed"`, `useElementHistory` returns it instead of the pair, and the popover reads it with early returns. The forwarding `ElementHistoryBody` layer is gone: the popover holds its own chrome and mounts `ElementVersionList` (which owns the state machine and the subscription) only while open, as before.

## Docs

`ARCHITECTURE.md` claimed the editor is a prop handed to `CanvasProviders`, didn't mention version history in the editor-state section, and was missing `canvas_versions` from the data table. Fixed, plus one line on `projectDocument` / `canvasHistory`.

## Checks

`lint`, `format:check`, `typecheck`, `knip`, `build`, and `vitest run` (159 files / 1430 tests) all green.

## Open PR overlap check

Considered #668 (`osmlStreamParser`, `generation/queue`, `video/sceneSegments`), #663 (`app/not-found.tsx`), #662 (`animated_image` still-frame, `AttributeBadge` / `ElementContainer` / `ModelBadge`, `connectors/types`). No file or theme here overlaps any of them.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -71,21 +71,23 @@ Staleness falls out of the graph: a node needs generating when it has no result,
 - `MetadataSchema` (`lib/project/types.ts`) is where a metadata default lives, not the reader. `videoSettings` in particular is total once parsed — `VideoSettingsSchema` fills every knob — so the aspect ratio, transition, length, and caption settings are read straight off the store (`useVideoSetting`) with no fallback in sight, and written back through `useUpdateVideoSettings` rather than a hand-assembled metadata patch.
 - `lib/canvas/elementConnector.ts` answers "which connector, provider and model does this element use?" for both the UI and the queue. An element pins its provider when created, so this is also where a pin that no longer resolves falls back to the registry default.
 - `lib/script/` provides script context and refinement utilities.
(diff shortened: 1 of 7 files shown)


umairnadeem (commented):
lgtm

PR #671, opened 2026-08-28
refactor: nightly conventions audit 2026-08-28

umairnadeem:
Nightly CONVENTIONS.md compliance audit.

Delta-audited every non-test source file merged to master since the 2026-08-27 sweep (#654, #657, #658, #659, #660, #661, #665), minus the files covered by open PRs #662, #663, #668 and #669.

## Fixed

- `components/ui/number-scrubber.tsx` defined a private `clamp` identical to the one `lib/utils` already exports and that `SeekTooltip`, `Waveform` and `elementAttributes` all import. One home per piece of knowledge; no behavior change.

## Flagged, not fixed

- `lib/project/relativeTime.ts` is a pure display formatter with no project knowledge, living apart from its siblings in `lib/format.ts` (`formatDateTime`, `formatBytes`, `truncateMiddle`) — `CanvasHistoryPanel` imports from both. Moving it would touch `ElementHistoryPopover.tsx`, which open PR #669 owns, so it is left for a follow-up.

## Verified clean

The new canvas-version-history surface is exemplary against all seven rules: `CanvasHistory` never has to choose between continuing and beginning a version (errors defined out of existence), `canvasVersionStorage` throws on every Supabase error and parses each row with zod, `ProjectDocument` gives history one read/write unit, and the panel is registered in `EditorSidebar`'s `PANELS` record rather than special-cased. Also clean: `lib/video/durationFit.ts`, `elementLengths.ts`, `lib/agent/tools/fitDurations.ts` (registered in the `TOOLS` record), `lib/generation/concurrency.ts`, `lib/format.ts`, `storeSnapshot.ts`, `snapshots.ts`, `PlayerPositionContext`, the timeline components and `components/ui/{color-field,disclosure}`.

Repo-wide invariant greps are unchanged: bare `catch {` is still exactly 3 intentional sites, zero catch-return-null, zero `as any`, zero non-null assertions, zero provider-string sniffing, and all 4 `switch` statements are exhaustive discriminated-union dispatch. Two standing flags were resolved by merged work: `Waveform`'s unhandled `play()` rejection (#659) and the duplicated inputs/result zod schemas in `lib/project/elementHistory.ts`, now imported from `lib/generation/snapshots.ts`.

lint, typecheck, format:check, 1430 tests and `npm run build` all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #672, opened 2026-08-28
refactor: rethink openslop architecture 2026-08-28

umairnadeem:
## App understanding

OpenSlop turns a written brief into a rendered video. A project's script lives in a Slate document on the canvas (`app/components/canvas`). Each element resolves into a dependency graph (`lib/generation`) that fans out through connectors, gateways and vendor providers (`lib/connectors` / `lib/gateway` / `lib/providers`). `lib/video` resolves the document into a `VideoLayout` that both the Remotion player and the Lambda render read. Sloppy (`lib/agent` + `app/components/sloppy`) is a ReAct agent that edits the same document through registered tools. Supabase carries auth, projects, conversations and job rows.

## From-scratch architecture

A rebuild would land close to what is already here. The domain layer is deep modules with narrow interfaces, one registry per variant family, zod parsing confined to the boundary, and one home for every derived value. The repo's canonical way to declare a variant family is a single registry entry that carries behaviour *and* how the thing shows up: `lib/canvas/elementConfigs.tsx` does it for element types, `lib/connectors/attributes/schema.ts` does it for connector attributes (`icon`, `label`, `edit`).

The agent tool registry was the one family that did not follow it. A tool declared its schema, its executor and its flags in one file, then its icon and label lived in a 93-line exhaustive switch under `app/components/sloppy/toolPresentation.tsx`. Adding a tool meant three edits in two layers, and a rename in one place silently changed the other.

## Implemented simplification

`defineTool` now takes the whole tool, including how a call reads while it runs:

- `icon` and `label` join `description`, `input`, `output`, `execute` and the flags. `label` is a string, or a function of the call's input for the three tools that read one (`edit_script`, `view_avatar`, `set_character`). `defineTool` normalises both into one presenter, so callers never branch on the shape.
- The registry gets one seam, `presentToolCall(toolName, input)`, next to the existing `executeToolCall`. It returns null for a name this build no longer offers, which is what a stored transcript can still hold. It reuses the same erased view of `TOOLS` that `DISPATCH` already uses, for the same reason: the SDK widens a streaming call's name and input.
- `app/components/sloppy/toolPresentation.tsx` drops from 98 lines to 22. All it owns now is the fallback for a renamed-away tool, which is a UI decision.

Behaviour is identical. Every icon and every label string is carried over unchanged, including the `1 change` / `N changes` pluralisation and the two "no name yet" fallbacks. Both existing presentation tests pass untouched, which is the point: the seam moved, the behaviour did not.

Adding a tool is now one file plus its line in `TOOLS`. The exhaustiveness that the switch bought is now bought by the required `icon`/`label` fields on `defineTool`, so a new tool without a face is still a type error.

## Validation

| Command | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:coverage` | pass, 159 files / 1430 tests, 79.1% statements, 74.4% branches, 80.4% lines |
| `npm run test:e2e` | pass, 3 smoke tests |

## Open PR overlap check

Considered #671 (`components/ui/number-scrubber.tsx`), #669 (`ARCHITECTURE.md`, `Editor.tsx`, `CanvasProviders.tsx`, `ElementHistoryPopover.tsx`, `useElementHistory.ts`, `lib/generation/history.ts`), #668 (`osmlStreamParser.ts`, `lib/generation/queue.ts`, `lib/video/sceneSegments.ts`), #663 (`app/not-found.tsx`), #662 (`AttributeBadge.tsx`, `ElementContainer.tsx`, `ModelBadge.tsx`, `lib/connectors/animated_image/*`, `lib/connectors/types.ts`). This branch touches only `lib/agent/tools/**` and `app/components/sloppy/toolPresentation.tsx`. Zero intersection.

## Skipped

- **The two-ways-to-reach-the-Slate-editor sweep** carried over from #629. Re-checked and it is already resolved: `Editor` hands the editor to `CanvasProviders` and nothing below takes it as a prop. Nothing left to do.
- **`ARCHITECTURE.md`** would read better with a line saying a tool's face is part of its definition. `ARCHITECTURE.md` belongs to #669, so it is left alone.
- **Model choice has two faces**: `SettingPill` in the hero composer and `ModelPicker` in `SloppyComposer` render the same choice two ways. That is a deliberate difference between a hero and a side panel, not duplication worth collapsing.
- **`app/projects/[id]/loading.tsx` has drifted from the shell it mirrors**, so the project load shifts layout. Still a real gap, still a visible UI fix rather than an architecture change.

## Risks

Low. The change is a move, not a rewrite: no schema, prompt, executor or flag changed, and no tool name changed. The one thing worth a reviewer's eye is that `lib/agent/tools/*` now imports icon components from `@/components/ui/icon`, which the agent route pulls in server-side. They are plain components that are never rendered there, and the same import already exists in `lib/canvas/elementConfigs.tsx` and `lib/connectors/attributes/schema.ts`.

### DIFF
--- a/app/components/sloppy/toolPresentation.tsx
+++ b/app/components/sloppy/toolPresentation.tsx
@@ -1,98 +1,22 @@
 "use client";
 
 import type { ToolUIPart } from "ai";
+import { Robot } from "@/components/ui/icon";
 import {
-	Eye,
-	Film,
-	Hourglass,
-	Mic,
-	Pencil,
-	Robot,
-	SlidersHorizontal,
-	TextBox,
-	Translate,
-	User,
-	type IconComponent,
-} from "@/components/ui/icon";
-import { SLOPPY_TOOLS } from "@/lib/agent/tools/registry";
+	presentToolCall,
+	type ToolPresentation,
+} from "@/lib/agent/tools/registry";
 import type { SloppyTools } from "@/lib/agent/types";
 
-type ToolPresentation = { icon: IconComponent; label: string };
-
-type ToolPart = ToolUIPart<SloppyTools>;
-
-const OFFERED: ReadonlySet<string> = new Set(
-	Object.keys(SLOPPY_TOOLS).map((name) => `tool-${name}`),
-);
-
 /** A stored transcript can hold a tool this build has renamed away, and it still has to render. */
-const pastTool = (type: string): ToolPresentation => ({
-	icon: Robot,
-	label: type.replace(/^tool-/, "").replaceAll("_", " "),
-});
-
-export function toolPresentation(part: ToolPart): ToolPresentation {
-	return OFFERED.has(part.type) ? offeredTool(part) : pastTool(part.type);
-}
-
-/** Exhaustive over the tools this build offers: a new one without a case is a type error. */
-function offeredTool(part: ToolPart): ToolPresentation {
-	switch (part.type) {
-		case "tool-read_script":
-			return { icon: Eye, label: "Reading the script" };
-		case "tool-edit_script": {
-			const count = part.input?.ops?.length ?? 0;
-			return {
-				icon: Pencil,
-				label:
-					count === 1
-						? "Editing the script (1 change)"
-						: `Editing the script (${count} changes)`,
-			};
+export function toolPresentation(
+	part: ToolUIPart<SloppyTools>,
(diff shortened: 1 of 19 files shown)


umairnadeem (commented):
lgtm

PR #673, opened 2026-08-28
feat: suggested prompts behind a lightbulb in the Sloppy composer

### LINKED ISSUE #664: Suggested prompts for Sloppy behind a lightbulb button in the composer
## Problem

Sloppy can do a lot that a user never finds out about. `fit_durations` sizes every clip to the dialogue under it, `set_caption_style` restyles captions by preset or by field, `edit_script` rewrites any element, `measure_element_lengths` reads real on-screen times. None of that is discoverable from the composer, which is one textarea with the placeholder "Write or change the script…" (`app/components/sloppy/SloppyComposer.tsx:80`). A user who does not already know the vocabulary types nothing, or types something Sloppy cannot act on.

There is no suggested-prompt surface anywhere in the Sloppy panel today. The composer row holds exactly two controls: the model picker and send/stop (`app/components/sloppy/SloppyComposer.tsx:84-100`).

- "create a openslop gh issue to add suggestions to sloppy using an idea (lightbulb) icon button. some common suggestions can be \"make sure all videos are long enough to cover their dialogue\" or \"each scene should last around 5 seconds\" or \"suggest some improvements to the video\" or \"how can i make the video more engaging\" or \"add some cliffhangers to end of each 3rd scene\" or \"change the caption style to fit the video's mood\" etc. these should be worded in extremely dead simple language, be concise, short, easy to understand at a 4th grade reading level and be within Sloppy's capabilities"

## Relevant code

- `app/components/sloppy/SloppyComposer.tsx:52-101` — the composer; `submit()` already funnels text into `send(message)`, so a suggestion is just `send(text)` with no new plumbing
- `app/components/sloppy/SloppyComposer.tsx:84-85` — the control row a lightbulb button sits in, next to `ModelPicker`
- `app/components/copilot/ActionButton.tsx:3-24` — the 24px icon-button shape send/stop already use
- `app/components/sloppy/SloppyProvider.tsx:119-134` — `send` pins the model and opens a fresh AbortController; a suggestion goes through it untouched
- `app/components/ComposerHero.tsx:15-21` — existing precedent for a non-typed programmatic `send(...)`
- `app/components/TemplateGallery.tsx:26-60` — the closest existing thing, but it is hero-only and prefills the box instead of sending
- `app/components/sloppy/SloppyPanel.tsx:14,49-53` — the empty state is one plain sentence, the other place these suggestions could surface
- `components/ui/select-menu.tsx` / `components/ui/action-menu.tsx` — the existing menu primitives; the trigger is a child, so the lightbulb button drops straight in
- `components/ui/icons.css:456` — `--lightbulb-icon` already exists in the token set; only the `icon("lightbulb")` export in `components/ui/icon.tsx` is missing
- `lib/agent/tools/registry.ts:22-38` — the full capability list every suggestion must map onto
- `lib/agent/tools/fitDurations.ts` — backs "make every video long enough for its words"
- `lib/agent/tools/setCaptionStyle.ts` — backs "match the caption style to the mood"
- `lib/agent/tools/measureElementLengths.ts` — how Sloppy grounds a scene-length request instead of guessing

## Suggested direction

Add a lightbulb icon button to the composer control row that opens a menu of preset prompts. Picking one sends it as a normal user message, exactly as if it were typed. No new message kind, no special-case path through `SloppyProvider`.

- Keep the list in one exported const (a `SUGGESTIONS` array of plain strings) next to the composer, so the copy has one home and stays reviewable as a list.
- Every suggestion must be answerable by a tool that exists today. Do not ship copy for a capability Sloppy lacks.
- Wording rule: 4th-grade reading level, under ~9 words, no jargon ("clip", "element", "OSML", "duration" are all out).

Starting set:

- Make every video long enough for its words
- Make each scene about 5 seconds long
- Tell me how to make this video better
- Make this video more fun to watch
- End every third scene on a cliffhanger
- Match the caption style to the mood

Menu items show that plain wording; the message Sloppy receives can be the same string, since each one already reads as a clear instruction.

Two capability notes that constrain the copy:

- `duration` is an enum of whole seconds 4 through 15 (`lib/canvas/types.ts:77-79`), so "about 5 seconds" lands on a real value, but suggestions must never name a length outside that range.
- There is no add-scene op. `edit_script` inserts elements only (`lib/script/refine/types.ts:12-46`); scenes are structural markers (`lib/canvas/osmlSerializer.ts:35-39`). The cliffhanger suggestion works because it rewrites the text of existing scenes, not because Sloppy can add one.

## Acceptance criteria

- [ ] A lightbulb button sits in the composer control row, labelled for screen readers, keyboard reachable, and opens a menu of suggestions.
- [ ] Picking a suggestion sends it as a normal user turn and closes the menu; the transcript shows it like any typed message.
- [ ] The button is disabled (or the menu inert) while Sloppy is loading, matching the send button.
- [ ] Suggestion copy lives in one array, every entry maps to a registered tool, and every entry is under ~9 words at a 4th-grade reading level.
- [ ] `--lightbulb-icon` is exported from `components/ui/icon.tsx` rather than inlined as an SVG.

## Related

- #597 (set_caption_style — makes the caption-mood suggestion actionable)
- #610 (Sloppy reading generation state — would unlock further suggestions later)

---

umairnadeem:
## Summary

A lightbulb button in the Sloppy composer control row drops a random starter prompt into the textarea. Clicking again rerolls. The prompt is left there to edit, so nothing sends until the user hits send.

- `app/components/sloppy/suggestions.ts` — six starters, each answerable by a tool that exists today (`fit_durations`, `edit_script`, `read_script`, `adapt_script`, `set_caption_style`), plus `nextSuggestion(current)`, which never hands back the one already in the box.
- `app/components/sloppy/SloppyComposer.tsx` — `SuggestionButton` next to `ModelPicker`. Disabled while Sloppy is loading, and disabled once the user has typed anything of their own, so a click can never destroy real input. The trigger styling the model picker already used is now a shared `controlClassName`.
- `components/ui/icon.tsx` — exports `Lightbulb`; `--lightbulb-icon` already existed in the token set.

## Why fill instead of a menu

A menu asks the user to read six prompts and pick one before they know what any of them do. Fill-and-edit gets a concrete sentence in front of them in one click, and treats the suggestion as a starting point rather than a command. Not sending on pick also means a suggestion is never an accidental agent turn.

## Validation

- `npm run lint`, `format:check`, `typecheck`, `knip`, `build` — pass
- `npm run test:run` — pass, including a new `suggestions.test.ts` covering the no-repeat guarantee

Closes #664

### DIFF
--- a/app/components/sloppy/SloppyComposer.tsx
+++ b/app/components/sloppy/SloppyComposer.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState } from "react";
+import { useRef, useState } from "react";
 import {
 	ChevronDown,
 	Codesandbox,
@@ -8,14 +8,13 @@ import {
 	Lightbulb,
 	SquareFilled,
 } from "@/components/ui/icon";
-import { ActionMenu } from "@/components/ui/action-menu";
 import { SelectMenu } from "@/components/ui/select-menu";
 import { cn } from "@/lib/utils";
 import { PanelCard } from "../canvas/panel/PanelCard";
 import { ActionButton } from "../copilot/ActionButton";
 import { useSloppyModel } from "./SloppyModelProvider";
 import { useSloppy } from "./SloppyProvider";
-import { SUGGESTIONS } from "./suggestions";
+import { nextSuggestion, SUGGESTIONS } from "./suggestions";
 
 const controlClassName =
 	"focus-ring inline-flex cursor-pointer items-center gap-1.5 rounded-md px-2 py-1 text-label text-muted-foreground transition-colors hover:bg-button-hover hover:text-foreground disabled:pointer-events-none disabled:opacity-40";
@@ -56,33 +55,24 @@ function ModelPicker({
 	);
 }
 
-function SuggestionPicker({
+function SuggestionButton({
 	onPick,
 	disabled,
 }: {
-	onPick: (suggestion: string) => void;
+	onPick: () => void;
 	disabled: boolean;
 }) {
 	return (
-		<ActionMenu
-			side="top"
-			items={SUGGESTIONS.map((suggestion) => ({
-				key: suggestion,
-				label: suggestion,
-				onSelect: () => onPick(suggestion),
-			}))}
-			contentClassName="max-w-72"
+		<button
+			type="button"
+			aria-label="Suggest a prompt"
+			disabled={disabled}
+			onMouseDown={(event) => event.preventDefault()}
+			onClick={onPick}
+			className={controlClassName}
 		>
-			<button
-				type="button"
-				aria-label="Suggested prompts"
-				disabled={disabled}
-				onMouseDown={(event) => event.preventDefault()}
-				className={controlClassName}
-			>
-				<Lightbulb className="h-3.5 w-3.5 shrink-0" aria-hidden="true" />
-			</button>
-		</ActionMenu>
+			<Lightbulb className="h-3.5 w-3.5 shrink-0" aria-hidden="true" />
+		</button>
 	);
 }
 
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
lgtm

umairnadeem (commented):
lgtm, the fill-and-edit version reads better than the menu

PR #679, opened 2026-08-29
fix(loading): add 4th RailItemSkeleton in left rail for History panel

vjymisal0:
Fixes #678.

### Summary of Changes

- Added a fourth `RailItemSkeleton` to the top rail group in `app/projects/[id]/loading.tsx`.
- Aligns the loading placeholder count with `RAIL_KEYS` (which now has 4 unpinned items: `layout`, `captions`, `properties`, `history`), preventing visual layout shifting when the real `EditorSidebar` mounts.

PR #674, opened 2026-08-28
feat: generate one scene at a time

umairnadeem:
## Summary

Each scene header now has its own **Generate** button that queues only that scene's stale elements, so you can produce and iterate one scene at a time instead of regenerating the whole project.

The element-filtering knowledge (has a prompt + `needsGeneration`) now lives in one place: a `useGenerateElements` hook that takes the scope at call time. `useGenerateAll` passes the whole document's content elements; `SceneGenerateButton` passes `scene.children`. Both go through the same queue path, and the queue's state machine keeps overlapping scopes idempotent.

- `app/components/canvas/hooks/useGenerateElements.ts` (new) — the shared filter + `enqueueGraph`
- `app/components/canvas/hooks/useGenerateAll.ts` — now a thin scope over it, behavior unchanged
- `app/components/canvas/elements/SceneGenerateButton.tsx` (new) — header icon button, same `TooltipIconButton` + caret-preserving `onMouseDown` pattern as `PlayFromHereButton` / `SceneDeleteButton`
- `app/components/canvas/elements/SceneContainer.tsx` — wires it into the scene header's `rightSlot`

No new state or subscriptions on the scene, so nothing is added to the canvas render path.

## Selected issue and why it was safe

#366 (`size: M`, `priority: medium`). The issue spells out the target behavior, the acceptance criteria are unambiguous, and every piece it needs already exists (`useGenerate`, `useGenerateAll`, the queue, the scene header slot). No product or design decisions: the button follows the established scene-header pattern.

## Validation

- `npm run lint` — pass
- `npm run format:check` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:coverage` — 159 files, 1432 tests, all pass (includes 2 new cases covering the scene-scoped path)
- `npm run test:e2e` — 3 passed

## Open PR overlap check

Re-listed open PRs (#662, #663, #668, #669, #671, #672, #673) and diffed their files. None touches the generate hooks, `SceneContainer`, or the scene header buttons. #662 touches `ElementContainer.tsx`, not `SceneContainer.tsx`; #668 touches `lib/generation/queue.ts`, which this PR does not modify.

## Skipped candidates

- #667, #651, #254 — `good first issue`, reserved for outside contributors
- #664 — already claimed by #673 earlier tonight; #492 by #662
- #461 — already implemented on master (per-word spans and an `activeWord` caption style); the issue is stale, not open work
[...]

PR #680, opened 2026-08-29
fix: nightly bug fixes 2026-08-29

umairnadeem:
## The bug

`updateSession` in `lib/supabase/middleware.ts` builds `supabaseResponse` and lets the Supabase SSR client's `setAll` write rotated auth cookies onto it. When an already-authenticated user hits `/login` or `/signup`, the function returned a brand-new `NextResponse.redirect(...)` and threw `supabaseResponse` away.

Any token `getUser()` rotated during that request never reached the browser, so the client held on to a stale, already-consumed refresh token. That is the classic Supabase SSR "random logout" failure: the next request presents a token the server has already invalidated. The pass-through path was always correct, which is why nothing surfaced it.

## The fix

Copy `supabaseResponse`'s cookies onto the redirect before returning it.

## Tests

`lib/supabase/__tests__/middleware.test.ts` now captures the cookie adapter handed to `createServerClient`, so `getUser` can rotate a cookie the way a real refresh does:

- `keeps refreshed cookies on the auth-route redirect` — fails on master (`undefined` instead of `"refreshed"`), passes here.
- `keeps refreshed cookies on a pass-through response` — passed before and after; it pins the path that already worked so the two cannot drift apart.

The shared `beforeEach` was hoisted out of the existing describe rather than duplicated.

## Verification

`npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run knip`, `npm run build` all clean. 160 test files / 1437 tests passing.

## Open PR overlap check

Considered open PRs #679, #674, #668, #662 and their touched files. This PR only touches `lib/supabase/middleware.ts` and its test, neither of which appears in any of them. Non-overlapping.

## Scope note

Two scoped audit passes were run over the non-overlapping surface. The first (video layout, captions, timeline, waveform/peaks, prefetch, agent tools, providers, clients) turned up no reproducible bugs; every suspicion was probed with a scratch test and discarded. The second (render/upload/auth routes, auth libs, proxy, render client) produced the fix above. Suspicions deliberately left alone as speculative rather than defects: `runRender` polling past unmount (bounded, and the fix is an abort-signal design change), and `useElementWidth` measuring border-box on attach vs content-box on resize (both consumers are padding- and border-free).

### DIFF
--- a/lib/supabase/__tests__/middleware.test.ts
+++ b/lib/supabase/__tests__/middleware.test.ts
@@ -3,10 +3,21 @@ import { NextRequest } from "next/server";
 
 const mockGetUser = vi.fn();
 
+type CookieAdapter = {
+	setAll: (
+		cookies: { name: string; value: string; options?: object }[],
+	) => void;
+};
+
+let cookieAdapter: CookieAdapter | null = null;
+
 vi.mock("@supabase/ssr", () => ({
-	createServerClient: vi.fn(() => ({
-		auth: { getUser: mockGetUser },
-	})),
+	createServerClient: vi.fn(
+		(_url: string, _key: string, options: { cookies: CookieAdapter }) => {
+			cookieAdapter = options.cookies;
+			return { auth: { getUser: mockGetUser } };
+		},
+	),
 }));
 
 import { updateSession } from "../middleware";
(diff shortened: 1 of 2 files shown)


umairnadeem (commented):
lgtm

umairnadeem:
doesn't happen in the wild, closing

PR #681, opened 2026-08-29
refactor: nightly maintainability 2026-08-29

umairnadeem:
## What changed

Two related cleanups in the editor's view-state layer.

### The player's placement is one value

`PlayerPositionContext` modelled a three-way choice as `position: "top" | "right"` plus `visible: boolean`, and exposed six members to do it. Every consumer then re-derived the tri-state:

- `LayoutPanel` mapped `visible ? position : "hidden"` back into a toggle, branched in its change handler to split one selection across `setVisible` and `setPosition`, and built its own options array inline.
- `PostPromptView` read `visible && isTop` / `visible && !isTop`.

The sibling `BottomViewContext` already models the same shape of choice — `"timeline" | "storyboard" | "hidden"` — as one union, with `BottomViewToggle` as a one-liner over it. This makes the player match:

- `PlayerPlacementContext` holds `placement: "top" | "right" | "hidden"`. "Hidden but on the right" can no longer be represented, so the mapping and the branching handler go away.
- Hiding keeps the preferred side internally, so `showPlayer()` restores it, and a narrow viewport still forces `top`. The interface stays one union plus `showPlayer()`.
- `PlayerPlacementToggle` mirrors `BottomViewToggle`, so `LayoutPanel` is now purely declarative: every field in it is a component or a two-line `MediaToggle`. It drops from 84 to 62 lines and loses its local type, its derived value, its handler and its options array.
- The context value is memoized. It was the one provider in the tree building a fresh object literal on every render.

### One boundary for canvas-scoped providers

`CanvasProviders` documents itself as composing every canvas-scoped provider "so the top-level view stays a flat orchestrator rather than a provider pyramid", but `PlayerPositionProvider` and `BottomViewProvider` sat outside it in `Editor.tsx`, and the pyramid inside was 11 deep.

Both are editor view-state and neither is read by `PrePromptView`, so they move in. `composeProviders` (`lib/components/composeProviders.tsx`) then nests the prop-free providers outermost-first, turning the pyramid into a list:

```tsx
const CanvasScopedProviders = composeProviders(
	RenderProvider,
	VideoLayoutProvider,
	PlayerPlacementProvider,
	...
);
```

`Editor.tsx` is now a flat orchestrator. `CanvasHistoryProvider` takes a prop, so it stays wrapped at the use site.

Nesting order is preserved exactly. `ActiveCaptionFont` moves from beside `VideoLayoutProvider` to inside the composed stack; it renders `null` and reads only the project store, so its position in the tree is immaterial.

## Behaviour

Unchanged. The placement refactor is a re-encoding of the same state machine: hidden keeps the preferred side, narrow viewport forces top, `showPlayer()` unhides, and the "Right" option is disabled on a narrow viewport. Every aria label and every icon is carried over.

## Validation

| Command | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:run` | pass, 160 files / 1435 tests |
| `npm run test:e2e` | pass, 3 smoke tests |

## Open PR overlap check

Considered #680 (`lib/supabase/middleware.ts`), #679 (`app/projects/[id]/loading.tsx`), #674 (`EditorToolbar.tsx`, `SceneContainer.tsx`, `SceneGenerateButton.tsx`, `useGenerateAll.ts`, `useGenerateScope.ts`), #668 (`osmlStreamParser.ts`, `lib/generation/queue.ts`, `lib/video/sceneSegments.ts`) and #662 (99 files across `lib/connectors/**`, `lib/canvas/**`, `lib/api/**`, `lib/project/**`, `app/api/v1/*`, `SloppyModelProvider.tsx`, `Storyboard.tsx`, canvas element cards). This branch touches `Editor.tsx`, `PostPromptView.tsx`, `CanvasProviders.tsx`, `panel/LayoutPanel.tsx`, `video/PlayerPlacement*`, `video/PlayerControlContext.tsx`, `video/BottomTransportBar.tsx`, `lib/components/composeProviders.tsx` and `ARCHITECTURE.md`. Zero intersection.

## Skipped

- **`RenderProvider`'s toast effect** is a four-branch chain over `RenderState`. A table would not read better than the early returns already there.
- **`Timeline` walks `rows` twice**, once for lanes and once for the gutter, with `ROW_HEIGHT[row.kind]` and the lane border duplicated across both. The two must stay in step or the timeline misaligns, so it is worth one home — but it is a five-line change and would put this branch inside a file next to #674's scene work.
- **`ComposerHero` owns the composer's text** only so `TemplateGallery` can set it. Lifting is the right call for two siblings; nothing to fix.
- **`LayoutPanel` and `CaptionsPanel` differ** on hoisting their `MediaToggle` options to module scope. Cosmetic, and the inline ones are two options long.

## Risks

Low. No new state, no new effect, and no change to what any provider provides. The one thing worth a reviewer's eye is that `PlayerPlacementProvider` now mounts under `<Slate>` rather than above `CanvasProviders`; it holds no editor-derived state, and `PlayerControlProvider`, its only non-UI consumer, was already inside.


umairnadeem (commented):
lgtm

PR #683, opened 2026-08-29
feat: enqueue messages sent to Sloppy while a turn is in flight

### LINKED ISSUE #666: Enqueue messages sent to Sloppy while a turn is in flight
## Problem

While Sloppy is mid-turn you cannot say anything to it. Pressing Enter in the composer does nothing at all: `submit()` early-returns on `loading`, with no toast, no shake, no hint. The send button has also been swapped for a Stop button, so there is not even an affordance to press. The typed text just sits there while the turn streams.

This is the wrong shape for how people actually steer an agent. You watch it start down the wrong path and want to say "no, make it 30 seconds" or "actually drop the narrator" immediately, while you are thinking it. Today the options are to interrupt with Stop (losing the turn's work) or to sit and wait for it to finish, then remember what you wanted to say.

- "also create another feature suggestion in openslop to enqueue messages sent to sloppy"

## Relevant code

- `app/components/sloppy/SloppyComposer.tsx:59` — `if (!hasText || loading) return;` silently drops the submit mid-turn
- `app/components/sloppy/SloppyComposer.tsx:86` — send button is replaced by Stop while `loading`, so there is no send affordance at all
- `app/components/sloppy/SloppyProvider.tsx:114` — `working` is derived from `status` plus pending tool calls; it is the single gate the composer reads
- `app/components/sloppy/SloppyProvider.tsx:121` — `send` sets up the abort controller and calls `sendMessage` immediately, assuming no turn is in flight
- `app/components/sloppy/SloppyPanel.tsx` — where a pending-queue affordance would render, below the transcript and above the composer

## Suggested direction

Give the queue one home in `SloppyProvider`, not in the composer. `send` becomes an enqueue: if no turn is in flight it dispatches immediately as it does today, otherwise it pushes onto a pending list. When `working` falls to false, the provider shifts the next message off the queue and sends it. The composer then stops caring about `loading` for submit entirely, and its guard collapses to `if (!hasText) return`.

Expose the pending list on the existing `SloppyControl` context so the panel can render it, keeping the composer dumb.

Details worth getting right:

- **Order is preserved.** Queued messages send one at a time, in the order typed, each as its own turn.
- **Editable before it fires.** A queued message is still just text, so let the user remove one or click to pull it back into the textarea. Once it has been dispatched it is a normal transcript message.
- **Stop clears the queue.** If the user hits Stop they are abandoning the direction, so pending follow-ups should not then fire against a turn that no longer exists. Worth confirming this is the behavior people want rather than assuming.
- **Errors halt the queue.** If a turn fails, do not keep firing queued messages into a broken conversation. Surface the failure and leave the rest pending.
- **The button stays Stop while working.** Send is the Enter key; the queued items render as their own affordance rather than fighting for the button slot.

## Acceptance criteria

- [ ] Pressing Enter mid-turn queues the message instead of silently doing nothing, and clears the textarea
- [ ] Queued messages send in order, one per turn, as each preceding turn completes
- [ ] Pending queued messages are visible in the panel, distinct from sent transcript messages
- [ ] A pending message can be removed, and can be pulled back into the composer for editing
- [ ] Stop clears the pending queue
- [ ] A failed turn halts the queue rather than dispatching the next message
- [ ] Queue state lives in `SloppyProvider`; `SloppyComposer.submit()` no longer gates on `loading`

## Related

- #246 (conversational plan mode — a back-and-forth interview makes mid-turn steering much more common)
- #598, #610 (Sloppy reading generation state — longer tool-heavy turns mean longer windows where the composer is locked)

---

umairnadeem:
## Summary

Pressing Enter mid-turn used to do nothing at all: `submit()` early-returned on `loading`, with no toast and no hint, and the send button had already been swapped for Stop. Typed text just sat there.

The queue now has one home in `SloppyProvider`. `send` dispatches immediately when nothing is in flight and otherwise pushes onto a pending list; when the turn settles the provider shifts the next message off and sends it as its own turn. `SloppyComposer.submit()` no longer looks at `loading` at all.

- **Order is preserved.** One queued message per turn, in the order typed.
- **Stop clears the queue.** Abandoning the direction abandons the follow-ups.
- **Errors halt the queue.** A failed turn leaves `status` on `"error"`, which holds the queue there until the user sends something themselves.
- **The ReAct loop is not a gap.** Between a tool result and the follow-up step the SDK sends for it, the chat briefly reads as idle. The drain reuses `lastAssistantMessageIsCompleteWithToolCalls` — the same predicate `sendAutomaticallyWhen` uses to continue the loop — so a queued message cannot land mid-loop. A `dispatching` ref covers the microtask between `sendMessage` and `status` reaching `"submitted"`.
- **Pending messages are still just text.** Each renders above the textarea with a remove button, and clicking one pulls it back out of the queue and into the composer for editing. Both routes are the same `dropQueued(id)` mechanism; the composer owns the policy of what to do with the text.

The Stop button keeps the button slot while working, as the issue asked. The suggestion lightbulb no longer disables on `loading` either, since the composer is no longer locked mid-turn.

`size: M`, `priority: medium`, unambiguous acceptance criteria, and the whole change is two files in one self-contained component tree. No API, schema, or generation-pipeline surface.

## Validation

- `npm run lint` — pass
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:run` — 1435 passed / 160 files

Not added: unit tests. The repo has no React Testing Library and no component or hook tests, so the queue's behavior has no existing seam to test against; the change is provider state plus presentation, with no pure helper worth extracting for its own sake. `npm run test:e2e` was skipped locally (needs Supabase env vars); the change touches no public route the smoke tests cover.

One targeted `react-hooks/set-state-in-effect` disable: the chat is the external system this effect synchronizes against, and shifting the queue when a turn settles is the synchronization itself, not an incidental cascading render.

## Open PR overlap check

Diffed against every open PR (#682, #681, #680, #679, #674, #668, #662). Zero file overlap — #662 touches `SloppyModelProvider.tsx` but not the composer, panel, or provider.

Closes #666

### DIFF
--- a/app/components/sloppy/SloppyComposer.tsx
+++ b/app/components/sloppy/SloppyComposer.tsx
@@ -7,6 +7,7 @@ import {
 	CornerDownLeft,
 	Lightbulb,
 	SquareFilled,
+	X,
 } from "@/components/ui/icon";
 import { SelectMenu } from "@/components/ui/select-menu";
 import { cn } from "@/lib/utils";
@@ -76,6 +77,47 @@ function SuggestionButton({
 	);
 }
 
+/**
+ * Messages typed while a turn was running. Editing one pulls it back out of the
+ * queue and into the textarea, so a pending message is only ever text.
+ */
+function QueuedMessages({ onEdit }: { onEdit: (text: string) => void }) {
+	const { queued, dropQueued } = useSloppy();
+
+	if (queued.length === 0) return null;
+
+	return (
+		<ul aria-label="Queued messages" className="flex flex-col gap-1">
+			{queued.map(({ id, text }) => (
+				<li
+					key={id}
+					className="flex items-center gap-1 rounded-md bg-secondary px-2 py-1"
+				>
+					<button
+						type="button"
+						aria-label={`Edit queued message: ${text}`}
+						onClick={() => {
+							dropQueued(id);
+							onEdit(text);
+						}}
+						className="focus-ring min-w-0 flex-1 cursor-pointer truncate rounded-sm text-left text-label text-muted-foreground transition-colors hover:text-panel-fg"
+					>
+						{text}
+					</button>
+					<button
+						type="button"
+						aria-label={`Remove queued message: ${text}`}
+						onClick={() => dropQueued(id)}
+						className="focus-ring shrink-0 cursor-pointer rounded-sm p-0.5 text-muted-foreground transition-colors hover:text-panel-fg"
+					>
+						<X className="h-3 w-3" aria-hidden="true" />
+					</button>
+				</li>
+			))}
+		</ul>
+	);
+}
+
 export function SloppyComposer() {
 	const { send, stop, loading } = useSloppy();
 	const { model, setModel, models } = useSloppyModel();
@@ -90,14 +132,20 @@ export function SloppyComposer() {
 	};
 
 	const submit = () => {
-		if (!hasText || loading) return;
+		if (!hasText) return;
 		const message = value;
 		setValue("");
 		send(message);
 	};
 
+	const edit = (text: string) => {
+		setValue(text);
+		textareaRef.current?.focus();
+	};
+
 	return (
 		<PanelCard>
+			<QueuedMessages onEdit={edit} />
 			<textarea
 				ref={textareaRef}
 				rows={2}
@@ -121,7 +169,7 @@ export function SloppyComposer() {
 					<ModelPicker model={model} onChange={setModel} options={models} />
 					<SuggestionButton
 						onPick={suggest}
-						disabled={loading || (hasText && !showsSuggestion)}
+						disabled={hasText && !showsSuggestion}
 					/>
 				</div>
 				{loading ? (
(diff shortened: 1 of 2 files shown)


app/components/sloppy/SloppyComposer.tsx
```diff
@@ -90,14 +132,20 @@ export function SloppyComposer() {
...
 	const submit = () => {
-		if (!hasText || loading) return;
+		if (!hasText) return;
 		const message = value;
 		setValue("");
 		send(message);
 	};
 
+	const edit = (text: string) => {
+		setValue(text);
```
umairnadeem, line 142: clicking a queued message clobbers whatever is already in the textarea: type a follow-up, click a pending one to tweak it, and the draft you were typing is gone with no undo. maybe only pull it back when `value` is empty, or append instead of replace?

umairnadeem:
Closing: the queueing approach here is more machinery than this is worth right now (provider-owned queue, drain predicate tied to the ReAct loop, a dispatching ref, pending-message UI). Not worth carrying that complexity for the payoff. Downgrading #666 to low priority / larger scope instead of tackling it now.

Issue #667, opened 2026-08-28
History panel offers no way back to the latest version while previewing an older one


umairnadeem:
## Problem

In the canvas history side panel, the newest version row renders no actions at all. `RowState` is `"current" | "previewing" | "older"`, and only `previewing` gets a Restore button and only `older` gets a View button, so the `current` row is inert.

That means while you're previewing an older version, the only way back to latest is the "return to latest" link inside the yellow banner. The history panel — the surface you're already looking at, with the latest version listed right at the top — offers nothing.

## Reported by (verbatim feedback)

- "we should be able to 'view' and 'restore' the latest version even if we're on it. usecase: if we're previewing an older version, there should be a way to get back to the latest version using the history side panel (and not just the banner)"

## Relevant code

- `app/components/canvas/panel/CanvasHistoryPanel.tsx:22` — `RowState` union; comment says the newest version "is neither viewed nor restored"
- `app/components/canvas/panel/CanvasHistoryPanel.tsx:56-78` — buttons gated so `current` renders nothing
- `app/components/canvas/panel/CanvasHistoryPanel.tsx:130-133` — `rowState()` maps index 0 to `current`
- `app/components/canvas/CanvasVersionBanner.tsx:49` — `history.backToLatest` is today's only escape hatch
- `lib/project/canvasHistory.ts:100` — `backToLatest()` already exists, so the panel just needs to call it

## Suggested direction

Let the top row carry a View action whenever it isn't the one being previewed, wired to `backToLatest()` rather than `preview()`. Restore stays scoped to the row you're actually previewing, since restoring the version you're already on is a no-op.

Simplest shape: drop the special-casing of index 0 for the View affordance and let the row's action derive from whether it's the previewed row, with the newest row's view handler being `backToLatest`.

## Acceptance criteria

- [ ] While previewing an older version, the newest row in the history panel exposes a View action that returns to latest
- [ ] With no preview active, the newest row still reads as current and shows no redundant action
- [ ] Restore remains on the previewed row only
- [ ] Keyboard and hover reveal behavior matches the existing `older` rows

## Related

- #654 (canvas version history with preview and restore)

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #684, opened 2026-08-29
feat: let Sloppy look at a generated image element

### LINKED ISSUE #598: Let Sloppy view generated image element results and give feedback on them

umairnadeem:
## Summary

Sloppy could see uploaded reference images and a character's avatar, but nothing the project actually generated. It wrote an image prompt, the image came back wrong, and it had no way to know.

`view_image` closes that loop. Given an element id from `read_script`, it hands the model the generated picture alongside the prompt that produced it, so intent and result are comparable in one turn.

- New `elementImage(id)` capability on `AgentToolContext`, implemented in `useAgentTools` off the same `queue` handle `avatarUrl` already uses. It projects the element down to `{ type, prompt, status, url }`: no urls beyond the picture itself, no `resultInputs`, no internal identity strings.
- The tool guards the three distinct cases separately, so each is an observation the model can act on: no such id (read the script again), an element with no picture at all (clip and audio results stay out of scope), and a result that is not there yet (the error names the status, so "still generating" reads differently from "never generated").
- Registered with `snapshot: true`, like `view_avatar` and `view_reference_images`: a viewed image is only true for its own turn, since a node can regenerate underneath the agent.
- One ROLE bullet in `prompt.ts` tells Sloppy to look before commenting on how an image turned out.

## Animated images

The first cut gated on `ELEMENT_TYPES[type].outputKind === "image"`, which excluded `animated_image` (its output kind is `video`). That read as a missing feature from the outside: asked to look at the first scene, the model was refused, said nothing about it, and quietly went and looked at the first scene that held a plain image instead.

An animated image does have a picture, the still frame it animates, so it is now viewable. There is no ready-made set of "visual" types to gate on: `layer: "visual"` and `FOREGROUND_TYPES` both include `clip`, which has no still and no poster frame, so either one would promise a picture that does not exist.

`pictureElementId(element)` in `still-frame.ts` answers it instead: the still node for an animated image, the element itself for an image, nothing for anything else. It sits beside the `pictureNode(node)` the upload button already uses, which gives the same answer off a built node, and the two now share the one `outputKind` rule.

Reading the still node matters beyond the type check. The animation's own result carries the frame it rendered from, which goes stale the moment the still is replaced or an uploaded one takes over, so the tool reads the same source the card preview does, for the same reason.

`ElementImage.picture` is absent rather than empty when a type makes no picture, so "generates no picture" and "not generated yet" are two shapes rather than two readings of a missing url, and each keeps its own error.

## Cleanups this pulled in

- `imageOutput` in `defineTool.ts` replaces the `{ type: "content", value: [...] }` literal that `view_avatar`, `view_reference_images` and now `view_image` had each written out. Third occurrence, so it earned a helper.
- The prompt shown beside the picture comes from `getPromptText`, the same text the generator feeds the connector, rather than the serializer's `getElementBodyText`, which differs by a trim.
- The ROLE bullet in `prompt.ts` is one line again; the type rules live in the tool description, which the model sees in the same request.

#598 is `size: M`, `priority: high`, with unambiguous acceptance criteria and an existing tool (`view_avatar`) named as the exact pattern to generalize. The change is additive: one new tool file plus one context capability, no existing behavior altered.

## Validation

- `npm run build` — succeeded
- `npm test` — 160 test files, 1439 tests passing, including 4 `view_image` cases (image handed over with its prompt, unknown id, an element with no picture, not-generated-yet reporting the status); the animated-image mapping is asserted where it lives, in `still-frame.test.ts`
- Playwright smoke tests not run locally; this change adds no route or page copy, so CI covers it

## Open PR overlap check

Diffed against every open PR (#662, #668, #674, #679, #680, #681, #682, #683). None touches `lib/agent/`; the closest, #683, is confined to `app/components/sloppy/`. No overlap.

## Skipped candidates

- #667, #254 — `good first issue`, reserved for outside contributors
- #461 (karaoke word highlighting) — the only other `size: S`; needs word-level timing design across the player and canvas, not a single focused PR
- #666, #492, #366 — already claimed by open PRs #683, #662, #674
- #632 (agent-changeable voice identity) — `priority: high` but genuinely two features (clearing `resolvedVoiceId` on a trait edit plus a new voice-search tool) with cache-invalidation consequences the issue itself flags; too broad for an unattended run
- #610 (agent reads generation state) — overlaps in-flight unlanded work on `measure_element_lengths` and asks for a projection whose shape is still open

Closes #598

### DIFF
--- a/lib/agent/__tests__/registry.test.ts
+++ b/lib/agent/__tests__/registry.test.ts
@@ -558,6 +558,26 @@ describe("executeToolCall", () => {
 		});
 	});
 
+	it("hands over the still frame an animated image animates", async () => {
+		const outcome = await executeToolCall(
+			{ toolName: "view_image", input: { id: "anim-1" } },
+			context({
+				elementImage: () => ({
+					type: "animated_image",
+					prompt: "a wolf at the door",
+					status: "idle",
+					url: "https://example.com/still.png",
+				}),
+			}),
+		);
+
+		expect(outcome.ok && outcome.output).toEqual({
+			id: "anim-1",
+			prompt: "a wolf at the door",
+			url: "https://example.com/still.png",
+		});
+	});
+
 	it("says an element id is not on the canvas rather than inventing a result", async () => {
 		const outcome = await executeToolCall(
 			{ toolName: "view_image", input: { id: "nope" } },
--- a/lib/agent/__tests__/registry.test.ts
+++ b/lib/agent/__tests__/registry.test.ts
@@ -545,8 +545,7 @@ describe("executeToolCall", () => {
 				elementImage: () => ({
 					type: "image",
 					prompt: "a wolf at the door",
-					status: "idle",
-					url: "https://example.com/wolf.png",
+					picture: { status: "idle", url: "https://example.com/wolf.png" },
 				}),
 			}),
 		);
@@ -565,8 +564,7 @@ describe("executeToolCall", () => {
 				elementImage: () => ({
 					type: "animated_image",
 					prompt: "a wolf at the door",
-					status: "idle",
-					url: "https://example.com/still.png",
+					picture: { status: "idle", url: "https://example.com/still.png" },
 				}),
 			}),
 		);
@@ -588,15 +586,14 @@ describe("executeToolCall", () => {
 		expect(!outcome.ok && outcome.errorText).toContain("no element nope");
 	});
 
-	it("refuses an element whose result is not an image", async () => {
+	it("refuses an element that generates no picture", async () => {
 		const outcome = await executeToolCall(
 			{ toolName: "view_image", input: { id: "clip-1" } },
 			context({
 				elementImage: () => ({
 					type: "clip",
 					prompt: "a wolf running",
-					status: "idle",
-					url: undefined,
+					picture: null,
 				}),
 			}),
 		);
@@ -612,8 +609,7 @@ describe("executeToolCall", () => {
 				elementImage: () => ({
 					type: "image",
 					prompt: "a wolf at the door",
-					status: "generating",
-					url: undefined,
+					picture: { status: "generating", url: undefined },
 				}),
 			}),
 		);
--- a/lib/agent/__tests__/registry.test.ts
+++ b/lib/agent/__tests__/registry.test.ts
@@ -28,6 +28,7 @@ const context = (over: Partial<AgentToolContext> = {}): AgentToolContext => ({
 	generateText: async () => "an outline",
 	referenceImages: () => [],
 	avatarUrl: () => undefined,
+	elementImage: () => undefined,
 	readMetadata: () => metadata,
 	editScript: () => ({ applied: 0, failures: [] }),
 	writeScript: async () => {},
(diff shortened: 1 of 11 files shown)


umairnadeem (commented):
lgtm, clean generalization of view_avatar

umairnadeem (commented):
lgtm. checked that the still really is prompted by the element body (videoPrompt drives the animation), so handing back getPromptText next to the still frame lines up, and swapping to the same helper resolveGraph uses kills the drift risk.

PR #687, opened 2026-08-30
fix: nightly bug fixes 2026-08-30

umairnadeem:
Nightly correctness pass. Six real bugs, each with a test that fails without the fix.

## Bugs fixed

**Runware video generation blocked the queue worker and uploaded every video twice** (`lib/providers/video/runware.ts`)
`videoInference` was called without `skipResponse`, so the SDK polled the task to completion (up to 10 min) before returning instead of handing back the ack. `submit()` therefore blocked one queue-worker invocation for the whole generation, and because the returned object already carried `videoURL`, `store()` streamed the finished video into Vercel Blob — a bundle `videoHandler` then discarded, since it only reads `metadata.jobId` and returns `pending`. The next poll fetched and re-uploaded the same video. Past the SDK's 10-minute poll timeout it resolved with `[]` and threw, failing a job the app's own 15-minute deadline would still have been waiting on.

**Runware images were JPEG bytes served as `image/png`** (`lib/providers/image/runware.ts`)
The request named no `outputFormat` (the SDK omits the field entirely when unset, so Runware applied its JPG default) while the result was hardcoded `format: "png"`, writing JPEG bytes to `output.png` with `contentType: image/png`. Browsers sniff past it; anything that trusts the declared type does not — notably feeding a generated image back as a reference image. The request now names the format the bundle labels the bytes with.

**Audio drifted off its picture as scene count grew** (`lib/video/scene-builder.ts`)
The renderer positions scene *i* at `Σ round(dⱼ·fps)` (one `TransitionSeries.Sequence` per scene, each rounded on its own) while layered narration/music/sfx are positioned at `round(Σ dⱼ · fps)`. Those agree only if every duration is already a whole number of frames. The file already fixed one instance of this for the transition overlap; scene durations got no such treatment. Five scenes of 60.5 frames each put the voice-over 2 frames (~83 ms) ahead of its scene by the fourth, and the error accumulates. Now every duration is snapped to the frame grid once, on the way in, so `round(Σ)` and `Σ round` are equal by construction. This also removes the path where `trimSequencesAt` could emit a 0-frame `Sequence`, which Remotion throws on.

**"Use references" stayed disabled after uploading a reference** (`app/components/canvas/elements/style/ArtStyleModal.tsx`)
`artStyleReferences` spans two stores — `referenceImages` from the project store and uploaded avatars from the generation queue — but was read entirely inside `useQueueSelector`, which only subscribes to the queue. Following the tooltip's own instruction to upload first left the button disabled until an unrelated queue tick. Split into `uploadedAvatarUrls` (queue) so each half gets its own subscription.

**Cancel in the "Still" tab discarded the animation instead** (`app/components/canvas/elements/preview/AnimatedImagePreview.tsx`)
`AnimatedImageMedia` swapped status/error/url per mode but passed one `onDiscard` through unswapped, and `queue.discard` is strictly per-id. Switching to "Still" while it generated and clicking Cancel aborted the animated element and left the still spinning. `onDiscard` now lives in `ModeState` alongside the state it belongs to.

**Cross-scene drop gap outlived the pointer** (`app/components/canvas/dnd/useDragAndDrop.ts`)
dnd-kit fires `onDragOver` with `over: null` when the pointer leaves every droppable, and the handler returned before reaching its only `set(null)`. The 3rem gap stayed open in the receiving scene for the rest of the drag. Clearing moved to the top of the handler; `DragTransferStore.set` now skips the notify when the value is unchanged, so the extra call costs nothing on the drag path.

## Tests

- `scene-builder`: frame-grid drift across 5 off-grid scenes (fails on master with a 1-frame divergence), and no sequence rounding down to 0 frames
- `runware-video`: `skipResponse: true` is sent; a task-less ack fails loudly
- `runware-image`: requested format matches the declared filename and content type
- `deriveArtStyle`: `uploadedAvatarUrls` excludes the project-store half
- `dragTransfer`: a no-change write does not notify
- `cancelOffset`: updated for the per-mode `onDiscard`

## Skipped

- **Inert "Get Started" CTA** (`app/page.tsx:68`) — a `<Button>` with no handler, not in a form; submission actually happens inside `AccessCodeInput` when the 6th character lands. Real, but removing or wiring a CTA is a product call, not a nightly one.
- **Stale timeline clip labels** — `getLayoutKey` omits prompt text by design, to keep the Player's `inputProps` off the per-keystroke path. Fixing it properly means changing where the label is read from; too large for this pass.
- **`app/api/upload/image/route.ts`** — an authenticated caller can name an image part `manifest.json` and clobber the bundle manifest. Needs a hand-crafted multipart body; filed as low severity rather than fixed here.

## Open PR overlap check

Considered #684, #683, #682, #681, #680, #674, #668, #662. None of the 18 files here appears in any of their diffs. Closest adjacency: #668 touches `lib/video/sceneSegments.ts` and #662 touches `lib/connectors/animated_image/plugins/still-frame.ts` and `lib/project/characterAvatar.ts` — this PR changes `lib/video/scene-builder.ts` and `lib/project/deriveArtStyle.ts`, and only imports from the other two. `lib/generation/queue.ts` (#668, #662) is called but not modified.

Lint, format, typecheck, knip, build, and the full suite (160 files, 1442 tests) pass.

### DIFF
--- a/app/components/AccessCodeInput.tsx
+++ b/app/components/AccessCodeInput.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useEffect, useRef, useState, useCallback } from "react";
+import { useRef, useState, useCallback } from "react";
 import { useRouter } from "next/navigation";
 import { apiJson } from "@/lib/clients/http";
 import { errorMessage } from "@/lib/errors";
@@ -23,13 +23,8 @@ export default function AccessCodeInput() {
 	const resetWithError = useCallback((message: string) => {
 		setError(message);
 		setValues(emptyAccessCode());
+		inputRefs.current[0]?.focus();
 	}, []);
-
-	// The boxes are disabled while loading, so focus has to wait for the render
-	// that re-enables them.
-	useEffect(() => {
-		if (error && !loading) inputRefs.current[0]?.focus();
-	}, [error, loading]);
 
 	const submitCode = useCallback(
 		async (code: string) => {
--- a/app/components/AccessCodeInput.tsx
+++ b/app/components/AccessCodeInput.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useRef, useState, useCallback } from "react";
+import { useEffect, useRef, useState, useCallback } from "react";
 import { useRouter } from "next/navigation";
 import { apiJson } from "@/lib/clients/http";
 import { errorMessage } from "@/lib/errors";
@@ -23,8 +23,13 @@ export default function AccessCodeInput() {
 	const resetWithError = useCallback((message: string) => {
 		setError(message);
 		setValues(emptyAccessCode());
-		inputRefs.current[0]?.focus();
 	}, []);
+
+	// The boxes are disabled while loading, so focus has to wait for the render
+	// that re-enables them.
+	useEffect(() => {
+		if (error && !loading) inputRefs.current[0]?.focus();
+	}, [error, loading]);
 
 	const submitCode = useCallback(
 		async (code: string) => {
--- a/app/components/AuthForm.tsx
+++ b/app/components/AuthForm.tsx
@@ -15,7 +15,6 @@ interface AuthFormProps {
 	subtitle: string;
 	submitLabel: string;
 	sentKind?: string;
-	initialError?: string;
 	shouldCreateUser?: boolean;
 	otpData?: Record<string, unknown>;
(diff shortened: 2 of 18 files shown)


umairnadeem (commented):
lgtm. verified the runware SDK returns the raw ack (with taskUUID) when skipResponse is set, so the video path does what you describe.

PR #688, opened 2026-08-30
refactor: nightly maintainability 2026-08-30

umairnadeem:
Four stores hand-roll the same subscription list, and three providers hand-roll the same `useSyncExternalStore` binding over it. This collapses both onto one seam.

## What was duplicated

`SnapshotStore`, `ElementHistory` and `CanvasHistory` each carried their own `listeners` Set with an identical `subscribe`/`notify` pair. On the React side, `GenerationQueueProvider`, `ElementHistoryProvider` and `CanvasHistoryProvider` each restated:

```ts
useSyncExternalStore(store.subscribe, () => selector(store), () => selector(store))
```

Four copies of one mechanism, and a fifth arrives with every new store.

## The change

New `lib/store/`:

- `emitter.ts` — `createEmitter()`, the subscription list. Stores hold one and expose `subscribe = this.emitter.subscribe`.
- `createStoreContext.tsx` — `createRequiredContext` plus the selector hook, returning `[Context, useStore, useStoreSelector]`. Adding a store is now a provider and nothing else.

Composition, not a base class: the stores own an emitter rather than inherit one, so nothing about their own shape changes. `GenerationQueue` already delegated `subscribe` to its `SnapshotStore` and is untouched.

Behaviour is identical — same Set semantics, same `useSyncExternalStore` arguments, same public store surface. 1438 tests pass, including the existing suites over the queue, element history and canvas history.

`DragTransferContext` has a fourth copy of the emitter, and it is left alone: that file is reserved by #687.

## Complexity delta

6 files touched, 87 lines removed against 32 added. Three classes lose a private field and two methods each; three providers lose their `useSyncExternalStore` wiring. New seam is 40 lines with a 3-case test.

## Open PR overlap check

Considered #687 (auth/projects UI, `DragTransferContext`, runware providers, `scene-builder`), #684 (`lib/agent/*`), #683 (Sloppy composer/provider), #682 (waveform/viewport), #681 (`PostPromptView`, video player contexts, `composeProviders`, `ARCHITECTURE.md`), #680 (`lib/supabase/middleware`), #674 (generate-scope hooks), #668 (`osmlStreamParser`, `lib/generation/queue.ts`, `sceneSegments`), #662 (connectors, `lib/canvas`, api routes, `lib/project/{applyScript,serialize,types,characterAvatar,projectDocument}`).

[...]

PR #689, opened 2026-08-30
perf: nightly performance optimizations 2026-08-30

umairnadeem:
Every Lambda renderer paid for the whole caption library before it could capture a frame. Now it loads the one face the composition's own props name.

## What was wrong

`remotion/index.ts` is the entry point deployed to the Remotion site (`scripts/deploy-remotion-site.ts`). It imported `remotion/loadCaptionFonts.ts`, which at module scope did:

```ts
const handle = delayRender("Loading caption fonts");
loadCaptionFonts((file) => staticFile(`fonts/${file}`)).then(
  () => continueRender(handle),
  (error) => cancelRender(error),
);
```

`loadCaptionFonts` with no font list means **the entire library**: 12 families, 14 woff2 files, **651 KiB**. Each one is fetched from the site bundle, parsed by `FontFace.load()` and added to `document.fonts` — and the `delayRender` handle holds the render until the last of them resolves.

A render is split into chunks across concurrent renderer functions, and every one of them loads the bundle in its own browser page. So the full library is fetched, parsed and registered once per invocation, and every invocation waits on it before its first frame.

## Only one face is ever drawn

`CaptionOverlay` sets `font-family` from `captionFontStack(style.font)`, and `style` is `captionStyle` off `VideoLayout` — an ordinary input prop. It is the composition's only text, so exactly one family renders:

- default style (`inter`): **352 KiB, 1 file** instead of 651 KiB, 14 files
- a display face like `anton`: **18 KiB, 1 file**

## What changed

- New `remotion/components/CaptionFont.tsx`: a renderless component that holds the same `delayRender`/`continueRender`/`cancelRender` contract as before, for one font.
- `remotion/Root.tsx` renders it beside `VideoComposition`, so the load happens where the composition's props say which face it needs, instead of blind at the bundle entry.
- `remotion/loadCaptionFonts.ts` and its side-effect import are gone.

The delay contract is unchanged, so a frame still cannot be captured mid-load and bake in a fallback face. `lib/video/captionFonts.ts` already took a `fonts` argument — this only starts passing it.

This also makes the two registration sites consistent. The editor was already precise: `ActiveCaptionFont` registers just the active face, and `CaptionFontLibrary` pulls the rest only when the captions panel mounts. The renderer was the odd one out.

`@remotion/player` in the app imports `VideoComposition` directly, not `Root`, so the preview path is untouched — and the editor has already registered the active face there anyway.

## Validation

`npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and `npm test` (160 files, 1435 tests) all pass.

## Open PR overlap check

Considered #688, #687, #684, #683, #682, #681, #680, #674, #668, #662. None of them touch `remotion/**` or `lib/video/captionFonts.ts`. #682 is the other open perf PR and works on waveforms (`ClipWaveform`, `Waveform`, `useNearViewport`) — no overlap.


umairnadeem (commented):
lgtm, nice find. delay contract is preserved and the Player path is untouched since it imports VideoComposition directly.

PR #690, opened 2026-08-30
refactor: nightly conventions audit 2026-08-30

umairnadeem:
Nightly CONVENTIONS.md compliance sweep. Files covered by an open PR were left alone.

## Fixed

**The panel rail's shape had three copies.** `PanelKey`'s union in `EditorPanelContext`, the `partition` over `PANELS` in `EditorSidebar`, and one `<RailItemSkeleton />` per panel typed out by hand in `app/projects/[id]/loading.tsx`. The third copy already drifted once — 3ab3a4e1 added the missing History slot after the fact. `panelKeys.ts` now holds the rail order and the pinned keys, and all three read it, so the skeleton counts itself.

**Every `<Separator />` in the app carried the same one-off class.** All three usages hand-wrote `-mx-3 my-2 data-[orientation=horizontal]:w-[calc(100%+1.5rem)]` to bleed past their popover's padding — one of them via a local `RULE` const. That's a `bleed` prop on the design-system component now, per "no one-off Tailwind in higher-level components".

No behavior change. lint, typecheck, format:check, 1435 tests and `npm run build` all pass.

## Flagged, not fixed (blocked by open PRs)

- `lib/project/elementHistory.ts:112` — `elementHistoryStorage.write` catches and toasts, so a failed write resolves; its sibling `canvasVersionStorage` throws and lets the caller decide. Policy in a mechanism module. Fixing it means changing the caller in `lib/generation/history.ts`, which #688 owns.
- `dot-grid-bg pointer-events-none fixed inset-0 -z-10` appears verbatim in `not-found.tsx`, `PrePromptView`, `PostPromptView` and `loading.tsx`. Wants one `<PageBackdrop />`. `PostPromptView` is in #681.
- `-mx-1 flex min-h-0 flex-1 flex-col gap-4 overflow-y-auto px-1` is the scroll body of `CharacterEditModal`, `NarratorEditModal` and `ArtStyleModal`. `ArtStyleModal` is in #687.
- `lib/components/Waveform.tsx:92` — a decode failure is only `console.error`'d, so the waveform silently stays empty. In #682.

## Flagged, design-level

- `lib/project/autosave.ts:89` — an unhydrated store aborts the save with a console line and no `onError`, so the user keeps typing into an autosaver that isn't saving.

## Clean

Rules 1, 2, 3, 4, 6 and 7 turned up nothing else across the 402 non-test source files outside open PRs. No empty catches, no catch-return-null, no `as any`, no provider-sniffing chains (the only `switch` is discriminated-union narrowing), API routes are thin, and the agent-tool and connector layers are registry-driven throughout.

### DIFF
--- a/app/components/canvas/elements/ElementHistoryPopover.tsx
+++ b/app/components/canvas/elements/ElementHistoryPopover.tsx
@@ -32,8 +32,6 @@ const PROMPT_LENGTH = 72;
 
 const SKELETON_ROWS = [0, 1, 2];
 
-const RULE = "-mx-3 my-2 data-[orientation=horizontal]:w-[calc(100%+1.5rem)]";
-
 const Dot = () => (
 	<span aria-hidden="true" className="text-muted-foreground">
 		·
@@ -170,7 +168,7 @@ export function ElementHistoryPopover({
 					<span className="text-label font-semibold">Generation history</span>
 					<CloseButton onClick={close} />
 				</div>
-				<Separator className={RULE} />
+				<Separator bleed />
 				<div
 					aria-live="polite"
 					className="-mx-1 flex max-h-80 flex-col overflow-y-auto overscroll-contain"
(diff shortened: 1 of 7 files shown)


umairnadeem (commented):
lgtm. panelKeys as the source of truth type-checks both directions (extra key in PANELS errors, missing key errors too), so the skeleton cant drift again.

PR #691, opened 2026-08-30
feat: upload a script as a PDF from the composer

umairnadeem:
## Summary

Users who already have a finished script as a PDF had to copy-paste it out by hand — a script could only enter the app as text typed into the hero composer. This adds an **Upload script (PDF)** entry to the composer's attach menu, alongside the existing reference-image upload.

The file posts to a new `POST /api/upload/script`, which validates it and extracts the text with [`unpdf`](https://github.com/unjs/unpdf). The composer then drops into "Paste a script" mode with the extracted text in the textarea, so the user reviews and edits before submitting rather than generating blind.

- `lib/upload/scriptPdf.ts` — `extractScriptText(file)`, mirroring `uploadImage`
- `lib/upload/useScriptUpload.tsx` — picker hook mirroring `useImageUpload`; hands back text, never the file
- `app/api/upload/script/route.ts` — `createSessionFormRouteHandler`, same shape as `upload/image`
- `lib/api/request-schema-fields.ts` — `pdfFile(maxBytes)` next to the existing `imageFile(maxBytes)`

Validation lives only on the route, so the accept/size rule has one home. Failures surface through the existing `apiJson` error envelope → `toastError`, and a PDF with no readable text (a scan) is a loud 400 rather than an empty script. The attach button reuses its existing spinner while extraction is in flight.

`unpdf` is a new dependency (1 package, no transitive deps). PDF text extraction is not something to hand-roll, and `unpdf` is the serverless-friendly pdfjs wrapper.

## Why this issue was safe and small

`size: M`, `priority: medium`, unambiguous acceptance criteria, an existing upload pattern to mirror, and no product or design decisions left open. Not labeled `good first issue` or `size: XS`.

## Validation

All CI-equivalent checks from `.github/workflows/ci.yml`, all passing:

- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run knip`
- `npm run build`
- `npm run test:coverage` — 162 files, 1444 tests passed (9 new)
- `npm run test:e2e` — 3 passed

Extraction was also verified end to end against a real one-page PDF outside the test suite. `npm audit`'s 2 moderate findings are pre-existing (`@runware/sdk-js` → `uuid`), not from `unpdf`.

## Open PR overlap check

Diffed against the file lists of all 12 open PRs (#690, #689, #688, #687, #684, #683, #682, #681, #680, #674, #668, #662). No file in this change appears in any of them.

## Issues skipped

[...]

Issue #685, opened 2026-08-29
Context-aware ghost-text suggestion in the Sloppy composer (Tab to accept)


umairnadeem:
## Problem

The composer's only prompt help is the lightbulb (#664, PR #673): a random pick from six fixed starter prompts. It is blind to the conversation. After Sloppy finishes a turn and ends with "want me to tighten the pacing?", the composer sits empty with the same generic placeholder, and the lightbulb still offers "Make each scene about 5 seconds long" no matter what just happened.

The next thing a user wants to type is almost always implied by the turn that just ended: the assistant's closing line, the tools it ran, or the prompt it just answered. None of that reaches the composer today.

## Reported by (verbatim feedback)

- "create a suggestion for openslop where we pre-populate (as a placeholder) in the sloppy composer a context-aware suggestion of what to type next. this suggestion can be from the previous prompt result, or sloppy's ending message. this will be claude-style where the user can press tab on the composer to populate the suggestion. this would replace the idea button on the composer or supersede it."

## Relevant code

- `app/components/sloppy/SloppyComposer.tsx:101-119` — the textarea; `placeholder="Write or change the script…"` at :115, `onKeyDown` at :109 handles Enter only
- `app/components/sloppy/SloppyComposer.tsx:58-73,122-125` — `SuggestionButton` and its `disabled` rule, the surface this replaces or supersedes
- `app/components/sloppy/SloppyComposer.tsx:85-90` — `showsSuggestion` / `suggest()`, the current random-pick wiring
- `app/components/sloppy/suggestions.ts:6-18` — `SUGGESTIONS` + `nextSuggestion`, the static pool that becomes the cold-start fallback
- `app/components/sloppy/SloppyProvider.tsx:29-35,119-134` — `useSloppyMessages` already exposes the full transcript to any client component; `send` takes plain text, so accepting a suggestion needs no new plumbing
- `lib/agent/types.ts:6-10` — `sloppyMetadataSchema`, the per-turn metadata channel a server-generated suggestion would ride on
- `lib/api/agentTurn.ts:90-96` — `messageMetadata` on the `finish` part, where `workSeconds` is stamped and where a `followUp` would be stamped
- `app/components/sloppy/SloppyPanel.tsx:14,51` — `EMPTY_HINT`, the fresh-project copy the cold-start suggestion sits next to

## Suggested direction

[...]

Issue #653, opened 2026-08-27
Move metadata out of OSML into typed JSON


umairnadeem:
## Problem

Four of the OSML tags are not script at all. `metadata_title`, `metadata_style`, `metadata_narration` and `metadata_character` are structured project data wearing prose clothing, and carrying them in the same XML stream as the canvas elements costs us a parser branch, a projection module, and a second node kind that every consumer has to filter back out.

They never reach the document. `useScriptSync` drops them with `isParsedContentElement`, and `serializeOSMLWithScenes` only walks scenes and content elements, so `projects.script` never contains a `metadata_*` tag. They exist purely as a transport detail of the initial generation stream, and the round trip is lossy in shape: attributes go out as strings and come back through `MetadataVoiceSchema.parse(attrs)` to become typed again. `metadata_narration` is an empty tag whose entire payload is attributes.

Every other path that writes this data is already typed JSON: the `set_metadata`, `set_narrator` and `set_character` tools take zod-validated input. Only the from-scratch generation path routes it through XML.

## Reported by (verbatim feedback)

- "file an issue to move metadata out of OSML into typed JSON"

Context: this came out of evaluating whether to retire OSML entirely in favour of a JSONL canvas. The conclusion was no for the prose script (streaming, token cost and prose escaping all favour tags), but yes for exactly this seam.

## Relevant code

- `lib/canvas/osmlStreamParser.ts:83` - the non-canvas pass-through branch that exists only for `metadata_*`
- `lib/canvas/osmlMetadata.ts:16` - `METADATA_TAGS` readers projecting tag attributes back into typed metadata
- `lib/canvas/osmlMetadata.ts:55` - `collectWritableMetadata`, including the "user already set this" carve-outs
- `lib/canvas/guards.ts:22` - `isParsedContentElement`, the filter every consumer needs because the node list is two kinds in a trench coat
- `app/components/canvas/hooks/useMetadataSync.ts:14` - projects streamed nodes into the project store
- `app/components/canvas/hooks/useScriptSync.ts:16` - drops the metadata nodes on the way to Slate
- `lib/script/prompt/osml.ts:112-146` - the four metadata sections of the spec
- `lib/script/ScriptProvider.tsx:69` - `runScript`, the raw `connector.stream` call both `write_script` and `adapt_script` go through

## Suggested direction

Have the script generation emit the metadata as typed JSON before the prose, and leave OSML as pure canvas markup.

[...]

Issue #631, opened 2026-08-24
Placeholder clips on the timeline for ungenerated assets


umairnadeem:
## Problem

The timeline only ever shows what has already been generated. `resolveElements` drops any
element without a result before layout is ever built (`lib/video/resolve.ts:23` —
`if (!snapshot.result) continue;`), so `VideoLayout.series` / `layout.sequences` contain
generated assets only. `buildTimelineRows` reads straight off that layout
(`app/components/video/timeline/timelineRows.ts:45-52`), which means an ungenerated element has
no clip at all: it is not a greyed-out block, it is a gap.

The practical effect is that the timeline is useless as a structural view until a project is
fully generated. Halfway through generating, scene 4's narration is a clip and scene 5's is
nothing, so the strip reads as if the script ends early. On a brand-new script every row is
empty and the panel falls back to `<TimelineEmpty />`
(`app/components/video/timeline/Timeline.tsx:146-148`) even though the script already describes
a complete structure.

The storyboard panel already made the opposite call and keeps ungenerated scenes visible
(`app/components/video/storyboard/storyboardScenes.ts:13,21-25` — `start: null` for a scene with
nothing on the timeline yet). The timeline should be consistent with it.

## Reported by (verbatim feedback)

- "add placeholder clips on the timeline for ungenerated assets with a hardcoded placeholder
  duration. these clips won't actually play on the player and the timestrip would also
  skip-over them (as well as playback) so they're only for visuals"

## Relevant code

- `lib/video/resolve.ts:23` — the `!snapshot.result` skip that makes ungenerated elements invisible to the whole layout pipeline
- `app/components/video/timeline/timelineRows.ts:45-52` — `sequencesForRole`, which derives every lane's clips from `VideoLayout` alone
- `app/components/video/timeline/TimelineClip.tsx` — the clip block that would need an ungenerated variant
- `app/components/video/timeline/Timeline.tsx:102-122,146-148` — layout consumption, seek mapping, and the empty state
- `lib/video/sceneSegments.ts:66-80` — `buildSceneSegments`, the seek bar's spans; must keep excluding placeholders
- `lib/video/elementLengths.ts:58` — `measureElementLengths`, which already estimates a script-only on-screen duration per visual without any generated asset
- `app/components/video/storyboard/storyboardScenes.ts:13` — existing precedent for keeping ungenerated items visible
- `lib/generation/snapshots.ts` — `ElementSnapshot.status`, the generated/ungenerated signal

## Suggested direction

[...]

PR #692, opened 2026-08-30
fix: say why an element is stale instead of always blaming the prompt

### LINKED ISSUE #537: Show why an element is stale (which prompt, attribute, or upstream dependency changed)
## Problem

`StaleIndicator` tells the user *that* an element is stale and nothing else. Its tooltip is a
hardcoded string — "Prompt changed — regenerate to update"
(`app/components/canvas/elements/GenerateButton.tsx:15`) — which is not only uninformative but
frequently wrong: staleness has three independent causes and only one of them is the prompt.

`needsGeneration` compares the whole of `nodeInputs(node, results)` against the snapshot's
`resultInputs` (`lib/generation/graph.ts:138-141`), and `GenerationInputs` has three parts
(`lib/generation/inputs.ts:6-13`):

- `prompt` — the element's authored text
- `attributes` — model, provider, duration, motion, overlays, characters, and the rest
- `dependencies` — the resolved identity of every upstream node

So an element can go stale because someone changed a character's appearance three cards away,
or switched the model in Settings, or because the still frame behind an animated image was
re-rendered — and the badge says "Prompt changed." The user is told to regenerate without being
told what regenerating would actually change, which is the information they need to decide
whether it's worth the time and cost.

The dependency case is the one that hurts most, because it's the only cause with no visible
locus. A prompt edit at least happened under the user's cursor; an upstream avatar change is
invisible, arrives asynchronously, and can light up many cards at once with no explanation.

### The data is already there

Nothing new has to be recorded. `resultInputs` is the exact input set the current result was
produced from, and it's already read directly for user-facing purposes — `CharacterEditModal`
pulls `resultInputs.attributes.appearance` to power its Revert button
(`CharacterEditModal.tsx:98`, `:148-159`). That's the precedent: this issue generalizes it from
one hand-picked attribute on one surface to the full diff on every surface.

- "we want to see why an element is stale. we should show the attributes that changed in a
  UX-friendly and intuitive way."

## Relevant code

- `app/components/canvas/elements/GenerateButton.tsx:13-28` — `StaleIndicator` / `ElementStaleIndicator`; the hardcoded tooltip to replace.
- `lib/generation/graph.ts:126-148` — `needsGeneration` / `isNodeStale`; the boolean that would become a reason.
- `lib/generation/inputs.ts:6-27` — `NodeInputs` / `GenerationInputs` / `serializeInputs`; the three buckets to diff.
- `lib/generation/snapshots.ts:6-14` — `ElementSnapshot.resultInputs`, the recorded before-state.
- `lib/generation/graph.ts:100-104` — `resultIdentity`: a dependency's identity is joined asset URLs, so a dependency diff can say *which* upstream changed but not *how*.
- `lib/generation/graph.ts:84-87` — `derivedNodeId`; dependency keys are ids like `~avatar:Red` and `~still:<elementId>`, not labels. Mapping id → human name is the real work here.
- `lib/project/characterAvatar.ts:8`, `lib/connectors/animated_image/plugins/still-frame.ts:26` — the two derived-node kinds that need labels.
- `app/components/canvas/elements/ElementGenerationContext.tsx` — one staleness check per card; where a reason would be computed and shared.
- `app/components/canvas/elements/character/CharacterEditModal.tsx:98`, `:148-159` — the existing revert-from-`resultInputs` precedent.
- `app/components/canvas/elements/style/ArtStyleModal.tsx` — third surface rendering `StaleIndicator`; must get this for free.

## Suggested direction

Turn the staleness boolean into a described reason, computed once in the domain layer and
rendered by the existing badge.

1. **A pure diff helper in `lib/generation/`** over `(currentInputs, resultInputs)` returning a
   structured description: which of prompt / attributes / dependencies changed, and for
   attributes the changed keys with before and after values. Pure and directly testable, in the
   spirit of keeping the imperative bits small and covered.

2. **Label the dependency keys.** A raw `~avatar:Red` is not a UX. The node kinds are a closed
   set (`avatar`, `still`), so a small registry mapping kind → human label is enough, and it
   belongs next to `derivedNodeId` rather than being re-derived per surface. Since
   `resultIdentity` is only an asset URL join, a dependency change can honestly say "Red's
   avatar changed" without claiming to know how — that's the right level of detail anyway.

3. **Render it from the existing badge, not a new surface.** `StaleIndicator` already owns this
   spot; the reason should replace its hardcoded tooltip so all three surfaces (element cards,
   character modal, art-style modal) inherit it with no per-surface wiring. Keep the badge terse
   — a short one-line reason for the common single-cause case, with the attribute-level
   before/after available on hover or expansion rather than crammed into the pill. Follow
   DESIGN.md; no one-off Tailwind in the higher-level components.

4. **Attribute values need display names too.** `attributes` keys are raw (`videoPrompt`,
   `durationSec`); they should render with the same labels Settings already shows, sourced from
   one place rather than a second hand-written map.

Out of scope: a full before/after diff view of generated results (#258, #329). This is about
explaining the input drift that made the element stale.

## Acceptance criteria

- [ ] The stale affordance states the actual cause instead of always claiming the prompt changed.
- [ ] Prompt, attribute, and dependency causes are each distinguishable.
- [ ] An attribute cause names the changed attribute(s) with their before and after values, using the same display names Settings uses.
- [ ] A dependency cause names the upstream in human terms ("Red's avatar", "still frame"), not a node id.
- [ ] Multiple simultaneous causes are summarized without overflowing the badge.
- [ ] Element cards, the character edit modal, and the art-style modal all show it from the shared component, with no per-surface wiring.
- [ ] The reason is accessible: available to screen readers, not hover-only, and not conveyed by color alone.
- [ ] Unit tests for the diff helper covering each bucket, multiple simultaneous changes, and the no-change case.

## Related

- #233 (revert appearance from `resultInputs` while stale — the existing single-attribute precedent this generalizes)
- #536 (uploaded results are never stale; that fix and this one share the staleness predicate and the badge)
- #329, #258 (diff view for refine changes / canvas history — result diffing, distinct from input drift)
- #295 (a staleness bug that was hard to diagnose precisely because the cause was never surfaced)

---

umairnadeem:
## Summary

`StaleIndicator`'s tooltip was a hardcoded `"Prompt changed — regenerate to update"`, but staleness has three independent causes and only one of them is the prompt. `needsGeneration` diffs the whole of `nodeInputs` — prompt, attributes, and the resolved identity of every dependency — so an element could go stale because a character's appearance changed three cards away, or the model changed in Settings, and the badge still said the prompt changed.

The dependency case is the one with no visible locus, so it gets a name rather than a shrug.

- New pure `staleReason(node, results)` in `lib/generation/staleReason.ts`. It reuses `isNodeStale` for the boolean and diffs `nodeInputs` against the snapshot's `resultInputs` to describe what actually moved. Returns `null` when the element is not stale, so it replaces the boolean at the two call sites rather than sitting beside it.
- A node carries the `label` a dependent uses to name it, set where the node is minted: `forCharacterAvatar` → "Red's avatar", `forStillOf` → "the still frame", `forArtStyle` → "the art style", `forVoice` → "<name>'s voice". `staleReason` reads `dep.label` and never parses an id.
- Attribute keys are already user-facing names, so they are humanised (`videoPrompt` → "video prompt") rather than mapped.
- Past three causes the rest are counted off ("… and 2 more") so the badge stays a badge.

No visual change: same badge, same tooltip component, only the text is now true.

#537, `size: M`, `priority: medium`. It was picked because the fix is almost entirely a pure domain function over data that is already recorded (`resultInputs` is the exact input set the current result came from, and `CharacterEditModal` already reads it for its Revert button). No new state, no new surface, no schema change, no design decision — the suggested direction in the issue spells out the shape, and the rendering stays in the badge that already exists.

## Validation

| Check | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:run` | 1446 passed (161 files), incl. 11 new `staleReason` tests |

New tests cover: not stale → `null`; prompt-only change; attribute change not blamed on the prompt; an attribute the element no longer carries; an upstream avatar named by its character; a project source node; a dependency that is itself stale though its output is unchanged; several causes joined; the count-off past three.

## Open-PR overlap check

Diffed this branch's file list against `gh pr diff --name-only` for all 13 open PRs (#662, #668, #674, #680–#684, #687–#691). Zero overlap. Notably `ElementContainer.tsx` (touched by #662) and `lib/video/sceneSegments.ts` (touched by #668) are untouched here.

## Candidates skipped

- #461 (`size: S`) — already implemented on master (`CaptionWord.active`, `style.activeWord`, `captionPresets`, the Captions panel control). The issue is stale, not open work.
- #667, #254 — `good first issue`, reserved for outside contributors.
- #365, #366, #492, #598, #666 — already claimed by open PRs #691, #674, #662, #684, #683.
- #610, #632 — both would edit `lib/agent/tools/context.ts`, `registry.ts`, `useAgentTools.ts` and `prompt.ts`, all in flight on #684.
- #398 — blocked on its stated prerequisite #370, which establishes the persisted metadata field it writes to.
- #594 — the issue itself says gap 2 "still needs designing", and an unmerged branch already covers gap 1.
- #345, #316 — would edit `ElementContainer.tsx` / the SFX connector files that #662 has open.
- #267, #631, #653, #685 — new product surface or larger architecture than one unattended run should take.

Closes #537

### DIFF
--- a/app/components/canvas/elements/GenerateButton.tsx
+++ b/app/components/canvas/elements/GenerateButton.tsx
@@ -10,9 +10,9 @@ import {
 } from "@/lib/generation/snapshots";
 import { useElementGeneration } from "./ElementGenerationContext";
 
-export function StaleIndicator() {
+export function StaleIndicator({ reason }: { reason: string }) {
 	return (
-		<SimpleTooltip label="Prompt changed — regenerate to update">
+		<SimpleTooltip label={reason}>
 			<span className="inline-flex items-center gap-1 rounded-full border border-tertiary/50 bg-tertiary/10 px-2 py-0.5 text-label-xs font-medium text-tertiary">
 				<AlertCircle className="h-3 w-3" />
 				Stale
@@ -22,9 +22,9 @@ export function StaleIndicator() {
 }
 
 export function ElementStaleIndicator() {
-	const { stale } = useElementGeneration();
-	if (!stale) return null;
-	return <StaleIndicator />;
+	const { staleReason } = useElementGeneration();
+	if (!staleReason) return null;
+	return <StaleIndicator reason={staleReason} />;
 }
 
 function generateLabel(status: GenerationStatus, hasResult: boolean) {
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -17,7 +17,7 @@ import {
 	useGenerationQueue,
 	useQueueSelector,
 } from "@/lib/generation/GenerationQueueProvider";
-import { isNodeStale } from "@/lib/generation/graph";
+import { staleReason } from "@/lib/generation/staleReason";
 import { isGenerationActive } from "@/lib/generation/snapshots";
 import { forCharacterAvatar } from "@/lib/connectors/image/plugins/characterAvatarNode";
 import { useNodeBuilder } from "@/lib/generation/useNodeBuilder";
@@ -100,7 +100,8 @@ function CharacterEditDialogBody({
 		queue.enqueueGraph([avatarNode]);
 	};
 
-	const isStale = isNodeStale(avatarNode, queue);
+	const staleWhy = staleReason(avatarNode, queue);
+	const isStale = staleWhy !== null;
 
 	const generating = isGenerationActive(avatarSnapshot.status);
 	const hasAppearance = Boolean(character.appearance?.trim());
@@ -158,7 +159,7 @@ function CharacterEditDialogBody({
 								elementId={avatarElementId}
 								onRestore={restoreAppearance}
 							/>
-							{isStale && <StaleIndicator />}
+							{staleWhy && <StaleIndicator reason={staleWhy} />}
 							<GenerateButton
 								status={avatarSnapshot.status}
 								hasResult={Boolean(avatarUrl)}
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -100,8 +100,8 @@ function CharacterEditDialogBody({
 		queue.enqueueGraph([avatarNode]);
 	};
 
-	const staleWhy = staleReason(avatarNode, queue);
-	const isStale = staleWhy !== null;
+	const reason = staleReason(avatarNode, queue);
+	const isStale = reason !== null;
 
 	const generating = isGenerationActive(avatarSnapshot.status);
 	const hasAppearance = Boolean(character.appearance?.trim());
@@ -159,7 +159,7 @@ function CharacterEditDialogBody({
 								elementId={avatarElementId}
 								onRestore={restoreAppearance}
 							/>
-							{staleWhy && <StaleIndicator reason={staleWhy} />}
+							{reason && <StaleIndicator reason={reason} />}
 							<GenerateButton
 								status={avatarSnapshot.status}
 								hasResult={Boolean(avatarUrl)}
--- a/app/components/canvas/hooks/useGenerate.ts
+++ b/app/components/canvas/hooks/useGenerate.ts
@@ -3,7 +3,8 @@ import {
 	useGenerationQueue,
 	useQueueSelector,
 } from "@/lib/generation/GenerationQueueProvider";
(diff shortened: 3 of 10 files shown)


umairnadeem (commented):
lgtm, the reason is derived from the same resultInputs the boolean was, so no new state to drift. nice.

PR #697, opened 2026-08-31
refactor: nightly maintainability 2026-08-31

umairnadeem:
Two structural cleanups that were found on 2026-08-29 but blocked behind open PRs (#662, #681). Both have since merged, so this picks them up. No behavior changes.

## The bottom dock's views get one home

`BottomView` was a variant family spread over four files: the union in `BottomViewContext.tsx`, `FILLS_DOCK` in `BottomDock.tsx`, the view→panel record in `PostPromptView.tsx`, and the label/icon options in `BottomViewToggle.tsx`. Adding a view meant four edits, and nothing forced them to agree.

`app/components/video/bottomViews.tsx` is now the registry, in the same shape as `elementConfigs.tsx` and `SLOPPY_TOOLS`: one entry per view carrying label, icon, panel and `fillsDock`, with `BottomView` derived as its keys. `BottomDock` renders its own panel instead of taking `children`, so `PostPromptView` drops to `<BottomDock />` and the toggle derives its options.

The registry holds the panels as stable module-scope elements, so switching to a resizable view still lets React bail out of re-rendering the panel on every resize frame, exactly as the old `children` prop did.

## Selecting a scene gets one rule

`Timeline.select` and `Storyboard`'s `onSelect` each spelled out "set active, scroll the canvas, seek the player", and had already drifted: the storyboard seeked raw through `player.seekTo` with no clamp and outside the `usePlayerScrub` seam that silences media mid-seek.

`useSelectScene()` is now the single path, and both callers go through it. Behavior is identical today — storyboard starts always come from `segments` so the clamp is a no-op, and `usePlayerScrub.seekTo` only silences when a scrub is in flight, which a click never is — but the two can no longer drift apart.

## Docs

Two bullets in `ARCHITECTURE.md` under Editor state, alongside the existing `PlayerPlacementContext` note.

## Complexity delta

62 lines removed against 26 added across the 6 touched files, plus the two new modules. Four sources of truth for `BottomView` collapse to one; two copies of scene selection collapse to one. `PostPromptView` loses a hook, a record and three imports; `Storyboard` loses four imports and a five-line inline handler.

## Open PR overlap check

No open PRs in the repo at the time of this run (`gh pr list --state open` returned empty), so nothing to overlap with.

## Skipped

Nothing else in the surveyed surface met the bar for a substantial, non-behavioral change.

## Validation

`lint`, `format:check`, `typecheck`, `knip`, `build`, `test` (1509 passed) and `test:e2e` (3 passed) all green.


umairnadeem (commented):
lgtm

PR #699, opened 2026-08-31
refactor: give the dialog scroll body one home

umairnadeem:
Nightly CONVENTIONS.md compliance sweep (fifteenth).

## Violation fixed

**Rule 5 (one canonical way / DRY is about knowledge).** The three modals with a scrolling middle each hand-wrote the same class string:

- `app/components/canvas/elements/character/NarratorEditModal.tsx:44`
- `app/components/canvas/elements/style/ArtStyleModal.tsx:86`
- `app/components/canvas/elements/character/CharacterEditModal.tsx:147`

```tsx
<div className="-mx-1 flex min-h-0 flex-1 flex-col gap-4 overflow-y-auto px-1">
```

`components/ui/dialog.tsx` already owns `DialogHeader` and `DialogFooter`, so the body was the one slot of the header/body/footer shape with no home — the layout survived as a convention three files had to remember, and the design system's "no one-off Tailwind in higher-level components" rule was being broken at each site. Adds `DialogBody` and uses it.

Rule of three is met exactly. This was flagged in the 2026-08-30 sweep but blocked on PR #687; that has since merged.

## No behaviour change

`DialogBody` emits the identical class string through the same `cn()` path as its sibling slots, so the rendered output at all three sites is byte-identical. `shrink-0` is left on the headers and footers untouched.

## Rest of the sweep

The delta since the last sweep (15 files from #689 and #692) was clean for all seven rule categories. Repo-wide re-greps are unchanged from the last sweep: 3 bare `catch {}` (the same intentional ones in `lib/supabase/server.ts` and `lib/api/parse.ts`), 3 `switch` statements (all discriminated-union dispatch), zero catch-return-null, zero `as any`, zero non-null assertions, zero provider sniffing.

`npm run lint`, `typecheck`, `format:check`, `test:run` (1509 passed) and `build` all green.

### DIFF
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -5,6 +5,7 @@ import { Trash2 } from "@/components/ui/icon";
 import { Button } from "@/components/ui/button";
 import { CloseButton } from "@/components/ui/close-button";
 import {
+	DialogBody,
 	DialogContent,
 	DialogDescription,
 	DialogFooter,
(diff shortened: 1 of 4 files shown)


umairnadeem (commented):
lgtm

PR #700, opened 2026-08-31
fix: don't leave an interrupted Sloppy turn spinning forever

### LINKED ISSUE #693: Sloppy panel stuck on "Thinking…" forever after an interrupted turn
## Problem

A Sloppy turn that dies mid-stream (abort, navigation away, or a provider error before the first `finish` part) leaves the panel showing a **"Thinking…" disclosure forever**, across reloads, with no error and no way to clear it.

Nothing is actually running when this happens. On an affected project the panel reads `aria-busy="false"`, there is no stop button, and `loading` is `false`. The spinner is a stale render of a **persisted** message, not live state.

The stored transcript's last row is a stub:

```json
{ "id": "hlDu1S-fubRpZgHISRXSC", "role": "assistant",
  "parts": [ { "type": "step-start" },
             { "type": "reasoning", "state": "streaming", "text": "" } ] }
```

Empty reasoning, still in `state: "streaming"`, and no `metadata` — so the `finish` part never fired and the turn was cut off right after `reasoning-start`.

Two halves cause it:

1. **Server persists a half-open stream.** `onEnd` writes `responseMessage` unconditionally. The AI SDK fires `onEnd` on abort too and hands over `isAborted` and `finishReason`, both ignored, so a part frozen at `state: "streaming"` is written to `messages` verbatim.
2. **Client renders the pending label from `part.state` alone.** `Reasoning`'s label is `streaming ? "Thinking…" : "Done thinking"`, where `streaming` is just `part.state === "streaming"`. A restored transcript is by definition never streaming, so a saved-mid-stream part spins forever. The panel's real `working` flag has no say in it.

Because the stuck state lives in Postgres and not in React, a reload does not clear it.

Scope check, two things this does **not** break: `convertToModelMessages` drops the contentless assistant message, so the next turn still goes out fine, and `hasPendingToolCall` only reads live chat messages, so nothing is genuinely blocked. The defect is cosmetic but permanent, and it makes a dead turn look like a hung one.

- "bug: why is http://localhost:3000/projects/8cf75bd6-7889-408c-ad78-e6f14c01c06a sloppy panel stuck on thinking? use browser if need be"
- "this probably still happens on master too"

Confirmed on `master`: neither `Reasoning.tsx` nor the `onEnd` handler differ from `master`.

## Relevant code

- `lib/api/agentTurn.ts:99` — `onEnd: ({ responseMessage }) => saveConversationMessage(...)` persists unconditionally, ignoring the `isAborted` and `finishReason` the callback provides.
- `app/components/sloppy/Reasoning.tsx:24` — `label={streaming ? "Thinking…" : "Done thinking"}`, with `pending={streaming}` on the next line.
- `app/components/sloppy/SloppyMessage.tsx:35` — `streaming={part.state === "streaming"}` reads the persisted part state directly; the turn-level `streaming` prop `AgentTurn` already receives is not consulted.
- `app/components/sloppy/SloppyPanel.tsx:61` — the panel does track live-ness (`working && index === messages.length - 1`) and threads it into `AgentTurn`, so the signal needed for the client-side guard is already in hand.
- `lib/api/conversations.ts:64` — `saveConversationMessage` upserts on `message_id`, so the stub row is a normal upsert and will be overwritten by a later turn with the same id, never cleaned up on its own.

## Suggested direction

Fix both halves; either alone leaves a hole.

**Server — don't persist a half-open turn.** In `onEnd`, skip the write when the turn produced nothing, and when `isAborted` close out non-terminal parts (`state: "streaming"` → `"done"`) before saving, so an interrupted turn is stored as interrupted rather than as eternally live. Keeping the normalization in one place at the persistence boundary means every future part type inherits it, rather than each renderer defending itself.

**Client — a restored part can never be pending.** Gate the pending label on the turn actually being live rather than on `part.state` alone, so any stub already in the database renders inert. `AgentTurn` already receives `streaming`; thread it into the `Reasoning` call.

A one-off cleanup for stub rows already written would clear existing affected projects.

## Acceptance criteria

- [ ] An aborted or errored Sloppy turn never persists a message part left in a non-terminal state.
- [ ] A turn that streamed no content at all is not written to `messages`.
- [ ] Reloading a project whose last turn was interrupted shows no "Thinking…" spinner.
- [ ] A reasoning part shows the pending label only while its turn is genuinely in flight, regardless of the stored `state`.
- [ ] Existing stub rows render inert without a migration (client guard covers them).
- [ ] A regression test covers a restored transcript ending in a `state: "streaming"` reasoning part.

## Related

- #666 (touches the same turn-lifecycle surface in `SloppyProvider`; independent of this bug, which predates it)

---

umairnadeem:
## Summary

A Sloppy turn that died mid-stream (abort, navigation away, provider error before `finish`) left the panel showing a **"Thinking…" disclosure forever**, across reloads, with no error and no way to clear it. Nothing was actually running: the spinner was a stale render of a *persisted* stub row.

**A restored part can never be pending.** `SloppyMessage`'s `Step` read `part.state === "streaming"` directly and ignored the turn-level `streaming` flag `AgentTurn` already receives. It now takes a `live` prop and only shows the pending label when the turn is genuinely in flight, so stub rows already in the database render inert with no migration needed.

**A restored thought can never auto-open, either.** `reasoningOpen` opened the newest thought in a turn, and an interrupted turn's newest part *is* the half-finished thought. Killing only the spinner would have left the abandoned half-thought force-expanded as the largest thing in the panel — the same "this didn't finish cleanly" impression, minus the shimmer. Auto-opening is a live-streaming affordance, so `reasoningOpen` now takes `live` as well. A normally-completed turn ends in a text part, so its reasoning is superseded and collapses either way; nothing about live streaming changes.

**`Reasoning` is handed its open state instead of deriving it.** It called `reasoningOpen(superseded, text)` itself, which is now the one place that needs `live`. Passing `open` down keeps the policy next to the flag it depends on and leaves the component purely presentational — same prop count, one less thing it knows.

**Mock LLM thinks for longer.** The mock's reasoning phase was three deltas wide, so there was no practical way to interrupt a turn mid-thought by hand. `MOCK_THOUGHT` now streams ~2s per step at the existing 20ms cadence. Deliberately *not* by raising `chunkDelayInMs`: that is paid on every chunk of every step, so a bump to 60ms made a whole mock turn ~7x slower (and the `MAX_TOOL_CALLS` worst case ~72s) to buy a window the text alone buys for free. The string also stays under `LONG_REASONING`, so the thought still streams visibly.

## Scope

An earlier revision also normalized turns at the persistence boundary (`settleTurn` in `lib/agent/messages.ts`, called from `onEnd`) so a half-open turn never got written in the first place. That half is dropped: the client fix already makes both existing and future stub rows render correctly, and the stored data stays a faithful record of what the provider actually sent.

Known and deliberately left alone: an interrupt can also strand a tool part in `input-streaming`/`input-available`. That renders as a bare collapsed disclosure with no outcome row — silent rather than misleading-in-motion, and no spinner — so it is a separate design call, not this bug.

## Validation

Full CI-equivalent suite from `.github/workflows/ci.yml`, all green:

- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run knip` — clean
- `npm run build` — succeeded
- `npm run test` — 167 files, 1513 tests passed

New and updated tests:

- `app/components/sloppy/__tests__/SloppyMessage.test.tsx` — a restored transcript ending in a `state: "streaming"` reasoning part renders "Done thinking" with no shimmer and leaves the half-thought collapsed, while the same parts render as "Thinking…" and expanded when the turn really is live.
- `app/components/sloppy/__tests__/turnDisplay.test.ts` — `reasoningOpen` shuts a restored thought the turn was interrupted in the middle of.

Closes #693

### DIFF
--- a/app/components/sloppy/Reasoning.tsx
+++ b/app/components/sloppy/Reasoning.tsx
@@ -2,18 +2,16 @@
 
 import { Sparkles } from "@/components/ui/icon";
 import { Disclosure, DisclosureText } from "@/components/ui/disclosure";
-import { reasoningOpen } from "./turnDisplay";
 
 export function Reasoning({
 	text,
 	streaming,
-	superseded,
+	open,
 }: {
 	text: string;
 	streaming: boolean;
-	superseded: boolean;
+	open: boolean;
 }) {
-	const open = reasoningOpen(superseded, text);
 	return (
 		<Disclosure
 			// Keyed on the auto state so reasoning that streams in open shuts itself
--- a/app/components/sloppy/SloppyMessage.tsx
+++ b/app/components/sloppy/SloppyMessage.tsx
@@ -6,7 +6,7 @@ import { PanelCard } from "../canvas/panel/PanelCard";
 import { Reasoning } from "./Reasoning";
 import { Tool } from "./Tool";
 import { WorkedStatus } from "./TurnStatus";
-import { userText, type TurnPart } from "./turnDisplay";
+import { reasoningOpen, userText, type TurnPart } from "./turnDisplay";
 
 export function UserMessage({ message }: { message: SloppyMessage }) {
 	return (
@@ -26,14 +26,23 @@ function ReplyText({ text }: { text: string }) {
 	);
 }
 
-function Step({ part, superseded }: { part: TurnPart; superseded: boolean }) {
+function Step({
+	part,
+	live,
+	superseded,
+}: {
+	part: TurnPart;
+	/** A restored transcript is never live, however its parts were stored. */
+	live: boolean;
+	superseded: boolean;
+}) {
 	if (isTextUIPart(part)) return <ReplyText text={part.text} />;
 	if (part.type === "reasoning") {
 		return (
 			<Reasoning
 				text={part.text}
-				streaming={part.state === "streaming"}
-				superseded={superseded}
+				streaming={live && part.state === "streaming"}
+				open={reasoningOpen(live, superseded, part.text)}
 			/>
 		);
 	}
@@ -57,6 +66,7 @@ export function AgentTurn({
 				<Step
 					key={`${message.id}-${index}`}
 					part={part}
+					live={streaming}
 					superseded={part !== newest}
 				/>
 			))}
--- a/app/components/sloppy/__tests__/SloppyMessage.test.tsx
+++ b/app/components/sloppy/__tests__/SloppyMessage.test.tsx
@@ -1,0 +1,35 @@
+import { describe, expect, it } from "vitest";
+import { renderToStaticMarkup } from "react-dom/server";
+import type { SloppyMessage } from "@/lib/agent/types";
+import { AgentTurn } from "../SloppyMessage";
+
+const interrupted: SloppyMessage = {
+	id: "m2",
+	role: "assistant",
+	parts: [
+		{ type: "step-start" },
+		{ type: "reasoning", text: "half a thought", state: "streaming" },
+	],
+};
+
+const render = (message: SloppyMessage, streaming: boolean) =>
+	renderToStaticMarkup(<AgentTurn message={message} streaming={streaming} />);
+
+describe("AgentTurn", () => {
+	it("shows a restored half-open turn as finished, not as still thinking", () => {
+		const markup = render(interrupted, false);
+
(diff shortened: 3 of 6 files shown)


umairnadeem (commented):
lgtm

PR #701, opened 2026-09-01
BYOK connectors and a third-party gateway

### LINKED ISSUE #331: BYOK: bring-your-own-key + third-party gateway (ElevenLabs/Speechify/Gemini TTS, Fal.ai, Anthropic/OpenAI LLMs)

umairnadeem:
Closes #331.

Lets people generate on their own provider keys, routed through a third-party gateway beside the hosted one. Scoped to the vendors we already run on — Anthropic, Runware, Cartesia, ElevenLabs — so no new provider integrations ride along.

## A model name is the whole routing decision

`ModelCatalog` maps each name to the provider serving it and the id that provider's API takes, so `providerForModel` picks the connector, the gateway family and whose key runs the generation. Nothing stores a provider separately.

A request keeps the **name** until the code that actually reaches a vendor: the sync routes translate at their call site, and an asset job stores the name for the queue worker to resolve. Translating at the HTTP boundary would throw the routing away — two models can share a vendor id (`Slop Image v1` *is* `Seedream 5 Lite`, on our key rather than yours), so the id alone can never say whose key to read.

## Keys

Stored per user in `provider_keys`, one row per provider. The key itself lives in Supabase Vault; the row keeps only the last four characters and the last verification result. Reads and writes go through `security definer` functions that only the service role may execute, so a key is decrypted on the server for the one request about to use it and never returns to a client. `anon` and `authenticated` have no table-level grants; `authenticated` gets column-level `SELECT` on everything except `secret_id`, under a select-own policy.

## Routes

`/api/third-party/*` mirrors `/api/v1/*` but is session-authenticated rather than API-access gated, and derives the provider from the model instead of accepting one. Asset generation creates the same `jobs` row, so polling, timeouts, cancel and history are unchanged. The same provider classes serve both families — only where the key comes from differs.

## Resolution and UI

A model resolves **element override → project default → account default → what OpenSlop recommends**, and the badge says which scope supplied it. Settings open from the URL, so a menu item, a "connect" affordance beside an unavailable model, and a pasted link (`?settings=models&provider=elevenlabs`) are all the same way in.

- Settings modal: nav plus a raised pane, with account defaults grouped as Images / Videos / Voice / Sound / Music / Text, and a searchable provider browser that matches on provider *or* model name.
- Project-level defaults get their own rail panel.
- Model pickers show unavailable models greyed with an inline connect affordance, and cost/speed chips.

## Notes

- `007_create_provider_keys.sql` is the only new migration. `003` gains nothing — an earlier draft added a `jobs.provider` column and it was removed once the name-based routing landed.
- Verified end to end against production: vault store/read/rotate/delete, key validation against the real vendor APIs, and the full resolution chain in the browser in both themes.
- `Disconnect` has no confirmation dialog yet, though `ConfirmDeleteDialog` exists if we want one.

https://claude.ai/code/session_01UyKaXtMQjZmyum9MRsCfzF


app/components/connectors/useConnectors.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Whether a generation can run on a provider at all. The hosted one always can;
+ * the rest need a key, whether or not it has been verified since.
+ */
+export function useCanGenerateWith(): (provider: ProviderKey) => boolean {
+	const connectors = useAccount((state) => state.connectors);
+	return useMemo(() => {
+		const stored = new Set(connectors.map((row) => row.provider));
+		return (provider: ProviderKey) =>
+			provider === MANAGED_PROVIDER || stored.has(provider);
```
umairnadeem: this says the hosted provider is always reachable, but `/api/v1/*` is still `withApiAccess` and 403s without the grant, which `ConnectorsTab` itself acknowledges (`user.app_metadata?.api_access` gates the OpenSlop row). for a BYOK-only account the catalog default for every type is a Slop model, so the picker shows them available and every generate 403s with no connect hint. worth folding api_access into this so hosted models grey out too?

app/components/connectors/useConnectors.ts
```diff
@@ -0,0 +1,34 @@
...
+ * not it has been verified since.
+ */
+export function useMissingKey(): (
+	provider: ProviderKey,
+) => BYOKProvider | null {
+	const connectors = useAccount((state) => state.connectors);
+	return useMemo(() => {
+		const stored = new Set<ProviderKey>(connectors.map((row) => row.provider));
+		return (provider: ProviderKey) =>
+			provider === MANAGED_PROVIDER || stored.has(provider) ? null : provider;
```
umairnadeem: `connectors` starts empty until `loadConnectors` lands, so every byok provider reads as missing in that window: open a model picker early and a model you *do* have a key for shows the Connect pill, and clicking it navigates to settings instead of picking the model. `loaded` is on the store for exactly this but nothing reads it, maybe gate here on it (treat unknown as not-missing until loaded)?

lib/api/providers.ts
```diff
@@ -56,3 +69,101 @@ export const getTTSProvider = defineProvider(
...
+}
+
+/** Checks a key against its provider and records what it found on the account. */
+export async function verifyConnector(
+	userId: string,
+	provider: BYOKProvider,
+	key: string,
+): Promise<ValidationResult> {
+	const result = await validateKey(provider, key);
+	await setConnectorStatus(userId, provider, result.ok ? "valid" : "invalid");
```
umairnadeem: a probe that times out or DNS-fails falls into the catch above and lands here as `invalid`, so the row flips to "Not working" for a key that's perfectly fine. the comment right above says "we could not tell" is its own answer, and `unverified` is already in the status enum, so imo the couldn't-reach case should leave/set `unverified` rather than accuse the key.

umairnadeem (commented):
lgtm, the provider/model pair refactor reads well. the route-families split in particular makes the byok/hosted difference easy to follow.

app/components/models/KeyStatusBadge.tsx
```diff
@@ -0,0 +1,20 @@
+import { Badge } from "@/components/ui/badge";
+import { AlertCircle, CheckCircle, Hourglass } from "@/components/ui/icon";
+import type { ConnectorStatus } from "@/lib/connectors/connectorRecord";
+
+const STATUS = {
+	valid: { variant: "default", Icon: CheckCircle, label: "Connected" },
+	// The same pill a stale element wears: something to look at, not an error.
```
umairnadeem, line 7: remove

app/components/connectors/useConnectors.ts
umairnadeem: overengineering, nvm

umairnadeem:
one finding from the latest round (inline is blocked, you've got a pending draft review sitting on the PR):

`app/globals.css:160/299` `--tertiary-foreground: #d29cb8` on the new solid `--tertiary: #772a4d` is ~4.1:1, and the tertiary badge paints at `text-badge` (10px), so it's under WCAG 1.4.3's 4.5:1 for normal text. lightening the foreground a couple steps (~#e0b4c8 gets ~5.2:1) clears it without touching the fill.

lib/project/types.ts
```diff
@@ -78,7 +90,7 @@ export const MetadataSchema = z.object({
 	characters: z.record(z.string(), MetadataCharacterSchema).default({}),
 	videoSettings: VideoSettingsSchema,
 	/** The model each connector type generates with, when the project pins one. */
-	connectorModels: z.record(z.string(), z.string()).default({}),
+	models: connectorModelsSchema.default({}),
```
umairnadeem, line 127: renaming the persisted key resets every saved project's model pins: the `store` column parses through `MetadataSchema`, so old rows carrying `connectorModels` hit `.default({})` and silently fall back to catalog defaults. either a `z.preprocess` that reads the old key, or a data migration?

lib/user/UserProvider.tsx
```diff
@@ -3,16 +3,30 @@
...
 	children,
 }: {
 	user: User;
 	children: ReactNode;
 }) {
-	return <UserContext value={user}>{children}</UserContext>;
+	return (
+		<UserContext value={user}>
+			<AccountStoreProvider
+				models={accountModelsSchema.parse(user.user_metadata?.models)}
```
umairnadeem: same shape for account defaults: `persistModels` now writes `user_metadata.models`, so anyone with `connectorModels` already stored loses their defaults on next login (and `accountModelsSchema` is `.catch({})` so it fails quietly). `user_metadata?.models ?? user_metadata?.connectorModels` would cover the existing rows.

lib/connectors/animated_image/attributes.ts
```diff
@@ -2,13 +2,14 @@ import { Image } from "@/components/ui/icon";
...
+import { modelDefs } from "../attributes/model";
 import { referenceImagesDef } from "../attributes/referenceImages";
 
 export const ANIMATED_IMAGE_ATTRIBUTES = AttributeSchema.from([
-	modelDef("animated_image", { label: "Video model" }),
-	modelDef("image", {
-		key: "stillModel",
+	...modelDefs("animated_image", { label: "Video model" }),
+	...modelDefs("image", {
+		key: "imageModel",
```
umairnadeem: `stillModel`/`stillProvider` are persisted in the OSML of saved scripts, so existing animated_image elements drop their pinned still model and quietly revert to the image default (old attr just rides along unrecognized). alias the old keys on parse, or is a one-time rewrite planned?

umairnadeem (commented):
lgtm on the latest pass, the modality-typed `Providers` map reads well and drops the hosted/vendor split nicely. still looking for the persisted-key renames from the earlier round though.

umairnadeem (commented):
lgtm on the audit pass. one `vendorParams`, the hosted row built server-side beside the user, and the model picks coming off the schema all read cleaner than the hand-rolled versions. still nothing for the persisted-key renames from the earlier round.

umairnadeem (commented):
lgtm, the `connector` -> `type` rename on the model pick is clean, no stragglers left.

lib/connectors/tts/plugins/voice-hydrate.ts
```diff
@@ -14,20 +16,16 @@ export function createVoiceHydratePlugin(
...
-				updateMetadata({ narration: { resolvedVoiceId: params.voiceId } });
-			}
+		beforeGenerate(params, ctx) {
+			const { voiceId, name } = params;
+			if (!voiceId) return params;
+			const model = requireModel(ctx, "voice-hydrate");
+			const { metadata, updateCharacter, setNarration } = store.getState();
+			const voice = metadataVoiceFor(metadata, name);
+			if (!voice || voiceIdOn(voice, model) === voiceId) return params;
+			const next = { ...voiceOnModel(voice, model), resolvedVoiceId: voiceId };
```
umairnadeem: generating clobbers a hand-picked voice now that the element owns its pair: `voiceOnModel` nulls `voiceId` whenever the voice sits on a different pair, and the element's header model select makes that trivial to hit. change one narration element's model, generate, and the narrator's picked voice in metadata is gone, plus every other narration element (still on the old pair) sees `voiceIdOn` -> undefined and re-searches, so the project's narration drifts to different voices. imo only remember `resolvedVoiceId` on the pair that ran and leave a picked `voiceId` alone rather than re-pinning the whole voice.

lib/project/types.ts
```diff
@@ -59,6 +68,33 @@ export const voiceSearchParamsSchema = voiceTraitsSchema.extend({
 
 export type MetadataVoice = z.infer<typeof MetadataVoiceSchema>;
 
+const voiceIsOn = (voice: MetadataVoice, model: ModelRef) =>
+	hasModel("tts", voice) && sameModel(voice, model);
+
+/** The id a voice has on a model: picked, or else found by an earlier search. */
+export const voiceIdOn = (
```
umairnadeem, line 75: same shape as the other persisted-key cases: every project saved before this branch has `narration.voiceId` (and character ones) with no provider/model beside it, so `voiceIsOn` is false and the pick reads as nothing. the editor shows no voice selected, and the first generate runs a fresh search, then hydrate stamps the pair next to the old `voiceId` so only *that* element ends up on the found voice while every later one flips back to the pick. imo a preprocess on `MetadataVoiceSchema` that stamps `DEFAULT_TTS_MODEL` onto a voice carrying an id but no pair would cover the old rows.

umairnadeem: overengineering, we don't want backwards compatibility

lib/project/types.ts
umairnadeem: irrelevant, backwards compatibility isn't required

lib/user/UserProvider.tsx
umairnadeem: backwards compatibility isn't required

lib/connectors/animated_image/attributes.ts
umairnadeem: backwards compatibility isn't required

lib/agent/client.ts
```diff
@@ -1,9 +1,15 @@
-import { OpenSlopClient } from "@/lib/clients/openslop";
+import { ApiClient } from "@/lib/clients/apiClient";
+import type { ModelRef } from "@/lib/connectors/types";
+import { apiPrefixFor, OPENSLOP_API_PREFIX } from "@/lib/gateway/prefix";
 import type { SloppyMessage } from "./types";
 
-export const AGENT_PATH = "/api/v1/agent";
+/** The transcript is ours to serve whoever the turns ran on. */
+export const AGENT_PATH = `${OPENSLOP_API_PREFIX}/agent`;
```
umairnadeem, line 7: this GET is still `withApiAccess`, so an account without the grant (the byok-only case the hosted row's `invalid` status exists for) runs its turns fine on `/api/third-party/agent` but 403s here on every editor load: "Could not load Sloppy" toast and an empty transcript for a conversation that's sitting in the db. the rows are the user's own, imo the transcript GET should be session-gated (or the third-party family gets its own GET and this picks by model).

PR #703, opened 2026-09-01
refactor: nightly maintainability 2026-09-01

umairnadeem:
No behavior changes. Four pieces of shared knowledge that had more than one home, or no home at all.

## Architecture

**`DragTransferContext` moves onto the `lib/store` seam.** It was the last hand-rolled observable store in the tree: its own `Set<() => void>`, its own notify loop, and its own `useSyncExternalStore` + `useCallback` wiring in `useDropIndex`. It now composes `createEmitter` and `createStoreContext`, so `useDropIndex` is one selector call and how a store reaches React stays decided in one place. Public API (`createDragTransferStore`, `dropIndexIn`, `useDropIndex`, `DragTransferContext`) is unchanged, and `dragTransfer.test.ts` covers it as-is.

**`MediaWithSkeleton` gets a shared home.** It lived in `app/components/canvas/elements/`, but `app/components/video/timeline/TimelineClip.tsx` imported it across features via `@/app/components/canvas/elements/MediaWithSkeleton`. It is the media sibling of `ImageWithShimmer` and is now beside it in `lib/components/`, marked `"use client"` to match. Four call sites updated; the canvas→video reach-in is gone.

## Simplification

**`soundwaveMaskStyle(heights)` owns "the style a soundwave needs".** Two call sites spread `SOUNDWAVE_MASK_STYLE` and then restated `maskImage`/`WebkitMaskImage` themselves, and the audio placeholder had drifted off it entirely — hand-writing `maskSize`/`WebkitMaskSize` and silently dropping `maskRepeat`. One helper in `lib/components/soundwave.ts` now returns the whole style, and the placeholder picks up the repeat rule it had lost (a no-op at `maskSize: 100% 100%`, which is why the drift went unnoticed).

**`SceneContainer.useSceneState` splits into `useDropPadding`.** The hook returned two unrelated things, so the collapsed scene built and memoized a `childIds` array on every render that only the expanded scene ever read. Each path now computes what it uses.

## Complexity delta

- 4 files simplified, 1 moved, 1 helper added.
- One hand-rolled subscription store removed (~12 lines of listener bookkeeping).
- `useDropIndex`: 7 lines and two React primitives down to one selector.
- One cross-feature `app/` → `app/` import removed.
- Net −24 lines.

## Skipped

- **A payload-carrying emitter for `GenerationQueue.commitListeners` and `Autosaver.savedListeners`.** Two hand-rolled `Set<(payload) => void>` remain, but two uses is under the rule of three and the shapes differ; a generic `createPayloadEmitter<T>` would be speculative today.
- **Timeline geometry context.** `pxPerSec` threads to five children in `Timeline.tsx`. It is one prop each and reads clearly; a context here would be churn, not a boundary.
- **A shared async-list shell for `CanvasHistoryPanel` and `ElementHistoryPopover`.** Same loading/failed/empty/list shape, but only two uses and different row semantics.
- **`AssetTile`'s `removeAffordance`/`fallback` string props.** Boolean-prop-ish, but both are genuinely presentational variants of one tile; splitting it would duplicate more than it saves.

## Open PR overlap check

One open PR: **#701** (BYOK connectors and a third-party gateway), which touches the connector/gateway/API/settings surface plus `app/components/canvas/dnd/SortableContent.tsx` and `app/components/video/storyboard/Storyboard.tsx`. Every file in this PR was checked against `gh pr diff 701 --name-only`: zero intersection. `SortableContent.tsx` consumes `useDropIndex` but is not modified here — the hook's signature is unchanged.

## Validation

`lint`, `format:check`, `typecheck`, `knip`, `build`, and `npm test` (1513 tests, 167 files) all pass.

### DIFF
--- a/app/components/canvas/dnd/DragTransferContext.tsx
+++ b/app/components/canvas/dnd/DragTransferContext.tsx
@@ -1,5 +1,5 @@
-import { useCallback, useSyncExternalStore } from "react";
-import { createRequiredContext } from "@/lib/components/createRequiredContext";
+import { createStoreContext } from "@/lib/store/createStoreContext";
+import { createEmitter, type Emitter } from "@/lib/store/emitter";
 
 export type DragTransfer = {
 	itemId: string;
@@ -11,7 +11,7 @@ export type DragTransfer = {
 export type DragTransferStore = {
 	get: () => DragTransfer;
 	set: (next: DragTransfer) => void;
-	subscribe: (onChange: () => void) => () => void;
+	subscribe: Emitter["subscribe"];
 };
 
 /**
@@ -21,21 +21,16 @@ export type DragTransferStore = {
  * re-render only when its own answer flips.
  */
 export function createDragTransferStore(): DragTransferStore {
+	const { subscribe, notify } = createEmitter();
 	let transfer: DragTransfer = null;
-	const listeners = new Set<() => void>();
 	return {
 		get: () => transfer,
 		set: (next) => {
 			if (transfer === next) return;
 			transfer = next;
-			for (const listener of listeners) listener();
+			notify();
 		},
-		subscribe: (onChange) => {
-			listeners.add(onChange);
-			return () => {
-				listeners.delete(onChange);
-			};
-		},
+		subscribe,
 	};
 }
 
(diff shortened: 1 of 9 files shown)


umairnadeem (commented):
lgtm

PR #705, opened 2026-09-01
refactor: give the dot grid one home

umairnadeem:
## What

The `.dot-grid-bg` texture was hand-written at five call sites as a bare `div` carrying `aria-hidden`, `pointer-events-none` and one of two placement class sets. DESIGN.md stated the rule in prose (`Render it aria-hidden and pointer-events-none, behind content`) precisely because no component held it. That is CONVENTIONS' "each piece of knowledge gets one home", and the rule of three is met twice over.

`components/ui/dot-grid.tsx` now owns it, with the two placements named rather than spelled out per site:

- `viewport` (`fixed inset-0 -z-10`) — pages that do not scroll: `not-found`, `PrePromptView`, `PostPromptView`, `projects/[id]/loading`.
- `container` (`absolute inset-0 z-0`) — `ProjectsList`, where the grid paints over the full height of a scrolling ancestor whose content carries its own `z-10`.

Every call site keeps the exact classes it had, so this is a no-behavior-change refactor.

## Flagged, not fixed

`app/not-found.tsx` renders the grid at `-z-10` inside a `relative … bg-background` parent. Per the CSS paint order, a negative-z-index descendant paints before its ancestor's own background, so the grid is very likely invisible on the 404 page. The other three `viewport` sites have no background on the positioned parent, so they are fine. Changing 404 to the `container` placement would make the grid appear, which is a visual change, not a refactor — leaving it for a human call.

## Audit context

This is the sixteenth nightly CONVENTIONS.md sweep. The delta since the last one was a single commit (#700, five source files across `app/components/sloppy/*` and `lib/providers/llm/mock.ts`) and it audits clean — `Reasoning` taking `open` as a prop instead of deriving it is the policy/mechanism split done right. Repo-wide invariant greps are unchanged for the tenth run: three intentional bare `catch {}` (`lib/supabase/server.ts:18`, `lib/api/parse.ts:31,46`), three exhaustive discriminated-union `switch`es, zero catch-return-null, zero `as any`, zero non-null assertions, zero provider-string sniffing.

Checks: lint, typecheck, `format:check`, `build`, and 1513 tests across 167 files all pass.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -47,7 +47,7 @@ Eases `--ease-casual` / `--ease-productive` / `--ease-expressive`, durations `--
 Two opt-in treatments, both quiet:
 
 - `.grain`: film grain (`--grain-opacity`) on raised surfaces (dialogs, panel cards, the canvas card, onboarding, the inline copilot). Needs a positioned element and paints through `::after`. Not on dense element cards.
-- `.dot-grid-bg`: the "+" grid behind the editor and gallery. Render it `aria-hidden` and `pointer-events-none`, behind content.
+- `.dot-grid-bg`: the "+" grid behind the editor and gallery. Reach for `<DotGrid />` (`components/ui/dot-grid.tsx`), which renders it `aria-hidden` and `pointer-events-none`, behind content.
 
 ## Components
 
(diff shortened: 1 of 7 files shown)


umairnadeem (commented):
lgtm

PR #706, opened 2026-09-01
refactor: rethink openslop architecture 2026-09-01

umairnadeem:
## App understanding

openslop is a Next.js app for generating short videos from a script. A Slate canvas (`app/components/canvas`) holds the OSML script; `lib/generation` schedules per-element generation through a connector/gateway/provider stack (`lib/connectors` → `lib/gateway` → `lib/providers`); `lib/video` builds a layout that Remotion renders and that the player/timeline/storyboard stack under `app/components/video` draws. Persistence is Supabase, with autosave and canvas/element version history in `lib/project`.

## From-scratch architecture

The current shape is already close to what I'd build: a registry-driven connector contract, one `routeHandler` builder behind every API route, a `defineTool` registry for the agent, and small single-purpose canvas helpers. `npm run knip` is clean and no module in the free surface exceeded ~250 lines. So this pass is narrow rather than structural: it takes the one duplicated-knowledge finding that was recorded but blocked in a previous pass and closes it.

**A timeline lane's chrome gets one home.** `Timeline.tsx` renders each row twice — the clip track in grid column 2, the icon gutter cell in column 1 — and both copies carried the `ROW_HEIGHT` lookup and the `border-b border-border/50` rule. They must stay in step or the timeline visibly misaligns, which is exactly the kind of fact that should not live in two places. A shared `Lane` component now owns the height and the rule; both call sites pass their own positioning classes.

**Scene numbering moves into the row table.** Whether a lane shows scene numbers was a `row.id === "foreground"` check inside `Track`, away from `ROWS` in `timelineRows.ts` where the lane's label, icon and kind already live. It becomes a `numbered` field on the row, so `Track` reads the row instead of re-deriving the role's behaviour.

No behaviour change: same markup, same classes, same scene numbers on the same lane.

## Validation

All CI steps from `.github/workflows/ci.yml` run locally and green:

- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run knip` — clean
- `npm run build` — succeeded
- `npm run test:run` — 167 files, 1513 tests passed

## Open PR overlap check

Considered #701 (BYOK connectors, 208 files), #703 (nightly maintainability), #704 (generation graph perf), #705 (dot grid) — 224 files locked in total, spanning essentially the whole `lib/` domain layer plus the canvas, settings, connectors, copilot and sloppy UI. Neither file touched here (`app/components/video/timeline/Timeline.tsx`, `timelineRows.ts`) appears in any of them. #703 touches `TimelineClip.tsx` and `ClipWaveform.tsx` in the same directory but not these two files, and this change does not alter the `TimelineClip` props contract.

## Skipped

- Everything under `lib/` and the canvas/settings/connectors UI: locked by #701/#703.
- `lib/project/autosave.ts` hand-rolls a payload-carrying listener set that `lib/store/emitter.ts` cannot absorb without making `Emitter` generic. ~10 lines for a speculative generic; left alone.
- The `lib/agent` tool registry, `lib/project` history/store modules, `lib/canvas` helpers and `app/components/canvas/hooks` were all read end to end this pass and found small, deep and idiomatic with nothing worth changing.

### DIFF
--- a/app/components/video/timeline/Timeline.tsx
+++ b/app/components/video/timeline/Timeline.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useMemo, useRef, useState } from "react";
+import { useMemo, useRef, useState, type ReactNode } from "react";
 import {
 	Timeline as TimelineIcon,
 	ZoomIn,
@@ -27,11 +27,36 @@ import {
 } from "./timelineRows";
 import { useTimelineZoom } from "./useTimelineZoom";
 
-const ROW_HEIGHT: Record<LayerType, string> = { visual: "h-20", audio: "h-16" };
+const LANE_HEIGHT: Record<LayerType, string> = {
+	visual: "h-20",
+	audio: "h-16",
+};
 const GUTTER_PX = 40;
 /** Runway past the last tick, so its label isn't cut off by the end. */
 const RULER_TAIL_PX = 32;
 const MIN_CLIP_WIDTH_PX = 2;
+
+/**
+ * A lane's height and rule, shared by the track and the gutter cell beside it.
+ * They are drawn in separate grid columns and have to line up exactly.
+ */
+function Lane({
+	kind,
+	className,
+	children,
+}: {
+	kind: LayerType;
+	className?: string;
+	children: ReactNode;
+}) {
+	return (
+		<div
+			className={cn("border-b border-border/50", LANE_HEIGHT[kind], className)}
+		>
+			{children}
+		</div>
+	);
+}
 
 function Track({
 	row,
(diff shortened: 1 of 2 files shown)


app/components/video/timeline/Timeline.tsx
```diff
@@ -233,14 +254,12 @@ export function Timeline() {
 					>
 						{rows.map((row) => (
 							<SimpleTooltip key={row.key} label={row.label}>
-								<div
-									className={cn(
-										"flex items-center justify-center border-b border-border/50 text-muted-foreground",
-										ROW_HEIGHT[row.kind],
-									)}
+								<Lane
```
umairnadeem: `SimpleTooltip` renders `TooltipTrigger asChild`, so radix clones this child with its pointer/focus handlers, `aria-describedby` and a ref. `Lane` only destructures `kind`/`className`/`children` and drops the rest, so the gutter tooltips silently stop opening (it was a plain `div` before). spread them through: `function Lane({ kind, className, ...props })` onto the div.

Issue #702, opened 2026-09-01
Vox-style explainer documentary template


umairnadeem:
## Problem

There is no template for the Vox-style explainer documentary, and it is not a format the user can reliably reach by prompt wording alone. `TEMPLATES` ships seven presets (`pov-life`, `sleep-story`, `finance-tips`, `true-crime`, `pov-financial-lifestyle`, `celebrity-death`, `stick-explainer`), none of which produce the editorial-explainer look: an authoritative narrator walking through one question, archival and photographic material rather than illustrated characters, on-screen text and data callouts, maps and charts, and a clean typographic identity.

The nearest preset is `stick-explainer`, which is explainer-shaped but locked to a crude hand-drawn doodle art style and chibi stick figures. A user who wants Vox has to fight that style rather than pick a different one.

Without a template, all of the direction lives in the free-text brief: art style, narration voice, image cadence, and script structure are each a separate thing the user has to guess at and restate. That is exactly the "I'm not sure how to direct it" failure below.

## Reported by (verbatim feedback)

- "so I've been trying to see if the machine can produce VOX style content"
- "I'm not sure how to direct it"

Reference videos supplied by the reporter:
- https://youtu.be/k5TB9Ey-AYc
- https://youtu.be/Hh0zvOU1eeE

## Relevant code

- `lib/templates/templates.ts:34` — `TEMPLATES`, the static preset list a new entry is added to
- `lib/templates/templates.ts:17-32` — the `Template` contract: `promptPrefix`, `systemPrompt`, `style`, `referenceImages`, `narration`, `length`, `showcase`
- `lib/templates/templates.ts:1490-1519` — `stick-explainer`, the closest existing preset and the best structural model for this one (narrator-only, no `characters`, style + reference images + a short `systemPrompt`)
- `lib/templates/applyTemplate.ts:7` — applies the preset onto the project (reference images, style, narration, length)
- `app/components/TemplateGallery.tsx:31` — the gallery renders any template carrying a `showcase`, so the new preset needs one to be discoverable
- `lib/script/prompt/template.ts:4` — `templatePrompt` tells the model to "pastiche this story format" and hands it the template's `exampleText` verbatim, so `exampleText` is the highest-leverage field: a hand-written Vox-shaped sample script does more than any amount of `systemPrompt` wording
[...]

Issue #695, opened 2026-08-31
Suggest 3 next steps as buttons under Sloppy's ending response


umairnadeem:
## Problem

When Sloppy finishes a turn, the conversation dead-ends. The transcript renders the final assistant message and stops; the composer sits empty with the same generic placeholder. The only prompt help anywhere is the lightbulb (#664, PR #673), which picks at random from six fixed starter prompts and is blind to what just happened.

The model almost always ends a turn knowing what could come next — it just offered to tighten the pacing, or it rewrote three scenes and the obvious follow-up is to fit the durations. None of that reaches the user. They have to work out the vocabulary themselves and type it from scratch.

## Reported by (verbatim feedback)

- "create a new openslop gh issue to suggest some next possible steps in Sloppy with each ending response. these 3 should show up as buttons below Sloppy's ending responses that the user can click. These should be compact, have a clear icon that indicates what they are, and pressing them should simply populate the sloppy composer allowing the user to edit"

## Relevant code

- `app/components/sloppy/SloppyPanel.tsx:56-63` — `Transcript` maps messages to `Row`; `index === messages.length - 1` already marks the trailing message, so "is this the ending response" needs no new state
- `app/components/sloppy/SloppyMessage.tsx:44-67` — `AgentTurn` already gates trailing UI on a settled turn with `{!streaming && worked && <WorkedStatus …/>}`. That is the pattern and the insertion point
- `app/components/sloppy/SloppyProvider.tsx:110-113` — the settled/in-flight truth source: `working = status === "submitted" || status === "streaming" || hasPendingToolCall(messages)`. Note the ReAct loop: streaming having stopped does **not** mean the turn is done
- `app/components/sloppy/SloppyComposer.tsx:58-77` — `SuggestionButton`, the compact icon button to extract and reuse; currently private to the composer
- `app/components/sloppy/SloppyComposer.tsx:19` — `controlClassName`, the existing compact-chip styling these buttons should use
- `app/components/sloppy/SloppyComposer.tsx:81` — **the blocker**: `const [value, setValue] = useState("")` is local to the composer. Nothing outside it can populate the box today
- `app/components/sloppy/SloppyComposer.tsx:95-100` — `submit()` clears `value` and calls `send(message)`; a populated draft flows through this untouched
- `lib/agent/types.ts:6-9` — `sloppyMetadataSchema`, currently just `workSeconds`; the per-turn channel a server-emitted follow-up rides on
[...]

Issue #694, opened 2026-08-31
Cancelling a canvas generation doesn't stop its poll loop


umairnadeem:
## Problem

Cancelling a generation job in the canvas does not stop the poll. The `AbortController` created per job is only used as an "ignore the result" flag — its signal is never threaded into the connector, so the polling `while` loop keeps hitting `/api/v1/<type>/<jobId>` once a second until the job hits a terminal status or the 15-minute timeout.

`cancel()` / `discard()` both abort the controller, stop the elapsed ticker, and drop the node from `pending`. The signal is then read only as a guard in `handleJobSuccess` / `handleJobError` / `finalizeJob` (`if (controller.signal.aborted) return;`), so the in-flight promise runs to completion and its result is silently dropped. Unmounting the canvas has the same problem: the provider's cleanup calls `cancelAll()`, which leaves every poll loop running.

Net effect: cancel looks instant in the UI but the client keeps polling (and the provider keeps billing) for up to 15 minutes per cancelled job.

## Reported by (verbatim feedback)

- "when you cancel a generation job in the canvas, the poll still keeps going"

## Relevant code

- `lib/providers/poll.ts:3` — the poll loop; takes no `AbortSignal` and has no cancellation input at all
- `lib/generation/queue.ts:181` — `abortJob()` aborts the controller but nothing downstream ever reads it
- `lib/generation/queue.ts:251` — job start; the controller is created here but not passed to `generateForElement`
- `lib/generation/generateForElement.ts:7` — the seam where the signal is lost: no `signal` param, and `connector.generate` is called without one
- `lib/connectors/asset-base.ts:31` — `_generate` calls `awaitCompletion`; stop condition is only `isTerminal(status)`
- `lib/connectors/types.ts:126` — the connector options type already accepts `signal?: AbortSignal` (the LLM connector honors it)
- `lib/generation/GenerationQueueProvider.tsx:36` — unmount cleanup calls `cancelAll()`, same non-teardown

## Suggested direction

Thread the existing per-job `controller.signal` through the layers that already have a slot for it: `queue.runJob` → `generateForElement` → `connector.generate` → `BaseAssetConnector._generate` → `awaitCompletion`, and have the poll loop bail on `signal.aborted` and forward the signal to the underlying `fetch`. No new state or cancellation machinery — the controller and the `signal?: AbortSignal` option both already exist; the plumbing just stops halfway.

## Acceptance criteria

[...]

PR #707, opened 2026-09-01
fix: crossfade looping SFX so repeats don't click

### LINKED ISSUE #316: Crossfade looping SFX to remove the audible gap between loops

umairnadeem:
## Summary

Looping sound effects restarted with a hard cut. `pushSequence` laid down `element.loops` copies butted end to end (`start + i * durationSec`), so each repeat began at a waveform discontinuity — the audible click/gap in the issue. Nothing faded: `AudioSequence` only built a fade envelope for `role === "background"`, so effects played at flat gain right through the seam.

Now each copy starts one crossfade early and carries a matching fade envelope, so the tail of one iteration sums with the head of the next across the overlap.

- **`lib/video/audioFade.ts` (new)** owns the whole policy: `LOOP_CROSSFADE_SEC` (0.15s default), `loopCrossfadeSec` (gated on `role === "effect" && loops > 1`, capped at half a copy), `loopStrideSec`, and `audioFadeSec`. One home means the builder's stride and the composition's envelope cannot drift apart — if they disagree, the crossfade doesn't sum to unity.
- **`lib/video/scene-builder.ts`** — `pushSequence` takes a stride instead of assuming the clip length. The stride is snapped to the frame grid the same way durations are.
- **`remotion/compositions/VideoComposition.tsx`** — `AudioSequence` asks `audioFadeSec(element)` for its fade instead of branching on the role inline. `audioVolume` already returns a flat scalar for a zero fade, so one-shots are byte-for-byte unchanged.

Music beds keep their existing `AUDIO_FADE_SEC` edge fade and their existing back-to-back loop stride; only effects flagged as looping change.

## Selected issue

#316 — `size: M`, `bug`, `priority: low`. Picked because the fix is a localized behaviour change in `lib/video` + `remotion/` with an unambiguous outcome, the fade machinery it needs (`fadeRamp`, `audioVolume`) already exists, and it is fully verifiable from the existing test suite without auth, secrets or a render.

## Acceptance criteria

- [x] No audible gap or click when a looping SFX repeats — copies overlap by the crossfade and fade through it rather than cutting.
- [x] Crossfade is applied only to looping SFX, not one-shots — `loopCrossfadeSec` returns 0 unless `role === "effect" && loops > 1`; covered by tests.
- [x] Crossfade duration is configurable with a reasonable default — `LOOP_CROSSFADE_SEC`, read by both the builder and the composition, clamped to half a copy so a short clip can't fade through itself end to end.

## Validation

All CI-equivalent checks from `.github/workflows/ci.yml`, all green:

| Check | Result |
| --- | --- |
| `npm run lint` | pass |
| `npm run knip` | pass |
| `npm run build` | pass |
| `npm run test:coverage` | 168 files, 1522 tests passed |

New coverage in `lib/video/__tests__/audioFade.test.ts` for the gate, the half-copy cap and the background/one-shot cases. Existing loop-position expectations in `scene-builder.test.ts` and `timelineRows.test.ts` updated to the new stride.

## Open PR overlap check

Checked #701, #703, #704, #705, #706 at selection time and again before pushing. No file in this diff appears in any of them. #706 touches `app/components/video/timeline/timelineRows.ts`; this PR touches only that module's test file, and only the two loop-position assertions.

## Skipped candidates

- #667, #254 — `good first issue`, reserved for outside contributors.
- #461 — already shipped on master.
- #694 (cancel doesn't stop the poll loop) — the best bug on the board, but threading the signal runs straight through `lib/gateway/openslop/base.ts` and `lib/connectors/types.ts`, both of which #701 rewrites (it deletes the former outright).
- #631, #398, #345 — overlap #706, #701 and #703 respectively.
- #653 — architecture migration, too large.
- #594, #702, #695, #685, #610, #632 — product/design direction, or acceptance criteria that can only be judged on generated output.

Closes #316

### DIFF
--- a/app/components/video/timeline/__tests__/timelineRows.test.ts
+++ b/app/components/video/timeline/__tests__/timelineRows.test.ts
@@ -165,10 +165,14 @@ describe("buildTimelineRows", () => {
 		const effects = buildTimelineRows(layout).filter(
 			(row) => row.id === "effect",
 		);
-		// Copies at 0 and 4 fall inside a 5s video; the one at 8 does not.
-		expect(
-			effects.flatMap((row) => row.clips).map((clip) => clip.start),
-		).toEqual([0, 4]);
+		// Two copies fall inside a 5s video; the third starts past the end. The
+		// second starts a crossfade early, so it lands short of the 4s clip length.
+		const starts = effects
+			.flatMap((row) => row.clips)
+			.map((clip) => clip.start);
+		expect(starts).toHaveLength(2);
+		expect(starts[0]).toBe(0);
+		expect(starts[1]).toBeCloseTo(4 - 4 / 24, 5);
 	});
 
 	it("draws a clip that overruns the end only as far as the video goes", () => {
--- a/lib/video/__tests__/audioFade.test.ts
+++ b/lib/video/__tests__/audioFade.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it } from "vitest";
+import {
+	audioFadeSec,
+	LOOP_CROSSFADE_SEC,
+	loopCrossfadeSec,
+	loopStrideSec,
+} from "../audioFade";
+import { AUDIO_FADE_SEC } from "../transitions";
+import type { ResolvedElement } from "../types";
+
+function el(overrides: Partial<ResolvedElement>): ResolvedElement {
+	return {
+		id: "e1",
+		type: "sound",
+		role: "effect",
+		layer: "audio",
+		sceneId: "s1",
+		sceneNumber: 1,
+		prompt: "",
+		url: "https://example.com/e1",
+		durationSec: 4,
+		loops: 1,
+		volume: 10,
+		motion: "none",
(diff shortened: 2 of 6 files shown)


umairnadeem (commented):
lgtm

PR #710, opened 2026-09-02
refactor: nightly maintainability 2026-09-02

umairnadeem:
## Summary

Two "one home" fixes, no behavior change.

**One generation lifecycle.** `useGenerateNode(spec)` is now the hook behind `useGenerate`. It takes any `NodeSpec` (the same shape `forElement` and `forCharacterAvatar` already produce) and owns node building, the snapshot subscription, staleness and enqueue/discard. `CharacterEditModal` re-implemented every one of those steps inline for the avatar node; it now calls `useGenerateNode(forCharacterAvatar(name))` and keeps only its own policy (appearance required, uploaded flag). The two existing `useGenerate` callers are untouched.

**Persisted shapes derive from their schemas.** `NodeInputs`, `GenerationInputs`, `GenerationStatus` and `ElementSnapshot` were hand-typed twins of the Zod schemas that parse them, held together by `satisfies z.ZodType<T>`. That check is one-directional: adding an optional field to the type compiles fine and is then silently stripped on every project reload. Each is now `z.infer` of its schema, and `GenerationInputsSchema` lives next to its type in `lib/generation/inputs.ts`. `CanvasElementTypeSchema` replaces the same 4-line enum cast copied into `refine/types.ts` and `elementHistory.ts`.

Net: 7 files, +66 / -100.

## Notes

- The modal used to call `staleReason(node, queue)` unsubscribed during render, so its stale badge only repainted when the avatar's own snapshot changed. Through the hook it subscribes via `useQueueSelector` like every element card does. The avatar node's prompt is the character name, so the hook's empty-prompt guard never fires for it.
- `AssetResult` is the one remaining `satisfies` twin. Closing it needs `lib/connectors/types.ts`, which PR #701 changes, so it is deferred.

## Skipped

- Voice-trait keys (`VOICE_TRAITS` x2, `VOICE_DESCRIPTOR_KEYS`, `SEMANTIC_FIELDS`): overlapping key sets but different policies (persisted, agent input, search filter, semantic ranking), not one fact.
- OSML attribute vocabulary in `osml.ts` vs `editScript.ts`: merging changes rendered LLM prompt text, not behavior-preserving.
- `TopPlayerPanel`/`SidePlayerPanel` registry: the two layouts differ in wrapper structure and handle order; a record would trade two plain components for one branchy one.
- Player nullability facade across 7 video files: leaves two ways to reach the player unless every raw consumer migrates; too wide for a nightly.

## Merge-conflict guard

```
PR #709: clean
PR #707: clean
PR #706: clean
PR #705: clean
PR #704: clean
PR #703: clean
PR #701: clean
```

## Checks

lint, format:check, typecheck, knip, build, vitest (167 files / 1513 tests) all pass.


umairnadeem (commented):
lgtm

PR #713, opened 2026-09-02
refactor: rethink openslop architecture 2026-09-02

umairnadeem:
Nightly architecture rethink. Ten open PRs (#701 BYOK connectors alone reserves ~200 files) were treated as off limits; everything here is outside their footprint.

## App understanding

OpenSlop turns a prompt into a short video. An LLM streams OSML into a Slate canvas; each element (narration, character, image, animated image, clip, sound, music) resolves into a generation graph run client-side against `/api/v1/*`, results land in Vercel Blob, project state autosaves to Supabase with version history, Sloppy edits the script through a tool loop, and Remotion renders the composition on Lambda.

## From-scratch architecture

The same three layers (connectors, gateways, providers) and the same generation graph, but with every "variant family" declared exactly once and everything else derived: an element type's spec (`ELEMENT_TYPES`) drives its config, its preview, and its connector; every subscribable store uses one emitter; every label accessor lives beside its type. The current codebase is already very close to that. This PR closes the remaining places where the same fact was declared twice.

## Implemented

- **Preview registry derives from `ELEMENT_TYPES`.** `ImagePreview` and `ClipPreview` were the same component with one differing literal, and `ELEMENT_PREVIEWS` restated each type's `outputKind`. Now `PREVIEWS_BY_KIND` is keyed by `ResultKind`, `animated_image` is the one explicit override, and the per-type record is generated the way `ELEMENT_CONFIGS` already is. Flipping an `outputKind` in `types.ts` can no longer leave the preview rendering the wrong media.
- **`createEmitter<T>` carries a payload.** The generation queue's commit listeners and the autosaver's saved listeners hand-rolled the same listener set the emitter already owns. Third use, so it earned the generic; the no-payload callers are unchanged.
- **Transition labels live with the type.** `transitionLabel()` in `lib/video/transitions.ts` mirrors `videoLengthLabel`, `captionFontLabel` and `captionPresetLabel`; `PropertiesPanel` no longer owns a domain table.
- **Upload URL goes through `AssetBundle.resolve`.** The route re-derived the blob path layout by hand from the type and provider it had just passed to `upload()`. Its test now spies on the real `upload` so the pure URL path runs for real.
- **`markHydrated` removed.** A second writer of `hydrated` that only one test called; `applyStoreSnapshot` is the writer.

Small fixes found on the way:

- `restoreResult` dropped `connectorType`, so a restored version onto an unhydrated snapshot could never become the project thumbnail.
- `getAspectRatioValue` sized the player from the clip-generation dimensions rather than the output frame. Numerically identical today, wrong the day a ratio's generation size crops differently.
- `GenerationQueue` now declares `implements NodeResults`, which it already satisfied structurally.

## Validation

All run locally on the branch, all passing:

```
npm run lint
npm run format:check
npm run typecheck
npm run knip
npx vitest run        # 167 files, 1513 tests
npm run build
```

`/simplify` ran as four review angles over the diff (reuse, simplification, efficiency, altitude). Three reported clean. The simplification angle preferred a flat literal registry; kept the derived form because the flat literal is the second declaration this change removes and it matches the `elementConfigs.tsx` precedent. The composition, web-design and React best-practice reviews were not run separately: the only component change is a registry file that renders no new UI.

## Merge-conflict guard

```
PR #712: clean
PR #711: clean
PR #710: clean
PR #709: clean
PR #707: clean
PR #706: clean
PR #705: clean
PR #704: clean
PR #703: clean
PR #701: clean
```

## Skipped, for a human call

- **Seek silencing is decided three ways** (`usePlayerScrub.seekTo` only when a drag started, `startPlaybackAt` always, `BottomTransportBar` prev/next never). Prev/next scene while playing leaks audio from the old position. Not fixed here: silencing during a one-shot seek without the pause/rAF-play sequence risks leaving media paused against a running picture (the #425 class of bug) and needs browser verification.
- **`getLoops` clamps to 1000 while the picker offers 1 to 8**, and `VOLUME_MIN/MAX` are hand-typed twins of `VOLUME_OPTIONS`. Deriving bounds from the option lists is the right shape but changes behaviour for hand-authored `loops` above 8.
- **`clampedAttribute` falls back to a module constant** (volume 10) while each connector declares its own default (sfx and music 2, video 5). Un-materialized attributes play five times too loud. Behaviour change and a per-element schema lookup on the layout path, so left for review.
- **Music and sfx ElevenLabs providers restate the audio cache policy** (eight identical lines each). Two uses, below the rule of three.
- **`projectModels` is threaded as an optional parameter through ten modules** with one consumer; every intermediate file is reserved by #701.
- **Generate-button state machine** is implemented in `EditorToolbar` and `SceneGenerateButton` and has drifted (the toolbar never shows the regenerate icon its label describes). Unifying is a visible change; flagged rather than shipped.
- Blocked by reserved files and recorded so they aren't re-derived: `connectorType` is dead payload on the asset queue message, `createSSEResponse` has no production caller, `ElementSnapshot` is a hand-typed twin of its schema, `staleReason.changedInputs` re-implements the graph staleness rule, `ElementHistory` spreads one lifecycle across four collections, `templateId` has two store writers.


umairnadeem (commented):
lgtm. the `restoreResult` connectorType fix and the `resolve("image")` swap on the upload route both look right to me, and the preview derivation kills a real duplicate declaration without losing exhaustiveness.

PR #194, opened 2026-05-24
Render videos via Remotion Lambda instead of Vercel sandbox

umairnadeem:
## Summary
- Replace the Vercel-sandbox render route (capped at the 800s function timeout) with two short-lived routes — `POST /api/render` invokes Remotion Lambda, `POST /api/render/progress` polls render state — and a client-side 5s polling loop in `useRendering`. Long renders no longer time out.
- Extract the Remotion webpack tweaks (Tailwind v4 + `@` alias) into `remotion/webpack-override.ts` so the same config is used by the CLI (`remotion.config.ts`) and by programmatic `deploySite` calls. This matches the `template-next-app-tailwind` pattern.
- Add `npm run deploy:remotion-site` (wired into the Vercel build via `vercel.json` → `turbo run build deploy:remotion-site`) and `npm run deploy:remotion-function` for one-off Lambda function (re)deploys. Lambda sizing lives in `lib/video/lambda-config.ts` (`RAM=3008, DISK=10240, TIMEOUT=240`).
- Delete dead sandbox/snapshot code: `create-snapshot.ts`, `app/api/render/{helpers,restore-snapshot}.ts`, the old SSE render-route test, the `SSEMessage` type. Drop `@remotion/vercel` + `@vercel/sandbox` from dependencies.

## Test plan
- [ ] `npm run typecheck && npm run lint && npm run format:check && npm run test:run` (already green locally — 766 tests pass)
- [ ] `npm run deploy:remotion-function` (one-time; only needed when Remotion is upgraded or sizing changes)
- [ ] `npm run deploy:remotion-site` once locally, then `npm run dev` and click Export Video on an existing project. Watch network tab: one `POST /api/render` returns `{renderId, bucketName}`, then `POST /api/render/progress` polls every ~5s, download URL surfaces in the UI. MP4 plays.
- [ ] Render a previously-timing-out long composition (>13 min) and confirm it completes.
- [ ] Verify the Vercel preview build runs `deploy:remotion-site` and prints a `serveUrl` in build logs.
- [ ] Verify the deployed Lambda function concurrency limit is sufficient (request raise from 10 → 1000 via AWS Service Quotas if not).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #714, opened 2026-09-03
refactor: nightly maintainability 2026-09-03

umairnadeem:
## Summary
- `.env.example` is now the one annotated env contract (it was missing the required `SUPABASE_SECRET_KEY` plus the ElevenLabs and Pinecone keys; the README was missing the OpenAI embedding and Remotion AWS keys). The README points at it instead of restating a second, drifting list.
- `DEFAULT_MOTION` and `isMotionEffect` move from `motionEffects.ts` (imports `remotion`) to `motionEffectNames.ts`, so reading element attributes no longer pulls Remotion into the canvas and server bundles. That is the split the names module exists to enforce.
- `LAMBDA_FUNCTION_SPEC` is the one home for the Lambda function's identity; the deploy script and `getFunctionName` both read it, so they cannot drift into a name that does not exist.
- Dead or duplicated app-internal logic removed: `createSSEResponse` (orphaned since #194), the hand-rolled `streamToBuffer` (now `Response.arrayBuffer`), the second `TOOLS` widening in the tool registry, `GenerationIndicator`'s never-passed `onClick`/`idleLabel`, `ElementHistory`'s unreachable `SESSION_ONLY` default, and `useRendering`'s re-typed `RenderUpdate` members.
- `pickThumbnailUrl` reads through `getPrimaryUrl` like every other result reader; `createProject` parses its Supabase row with zod and infers `ProjectRow` from the schema; `ResizeHandle` reads its axis from a two-row table instead of six ternaries.

No behavior change. The only observable difference is that a malformed `projects` row now throws at the boundary instead of flowing through untyped.

## Test plan
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run knip`
- [x] `npm run build`
- [x] `npx vitest run --coverage`: 167 files, 1508 tests pass

## Merge-conflict guard
```
PR #713: clean
PR #712: clean
PR #711: clean
PR #710: clean
PR #709: clean
PR #707: clean
PR #706: clean
PR #705: clean
PR #704: clean
PR #703: clean
PR #701: clean
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,24 +1,31 @@
 # Copy to `.env.local` and fill in. Do not commit secrets.
 
-# --- Supabase (public; safe to expose to the browser) ---
+# --- Supabase: required for auth and the database ---
 NEXT_PUBLIC_SUPABASE_URL=
 NEXT_PUBLIC_SUPABASE_ANON_KEY=
+SUPABASE_SECRET_KEY=            # generation jobs read and write with it
 
-# --- Vercel Blob storage ---
-BLOB_READ_WRITE_TOKEN=
-NEXT_PUBLIC_BLOB_URL=
+# --- Vercel Blob: where real providers store what they generate ---
+# BLOB_READ_WRITE_TOKEN=
+# NEXT_PUBLIC_BLOB_URL=
 
-# --- Embeddings (used by voice similarity ranking) ---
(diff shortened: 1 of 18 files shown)


umairnadeem (commented):
lgtm, the motion split and lambda spec both read well

PR #722, opened 2026-09-04
Revert "perf: fetch card images at the size they render" (#698)

umairnadeem:
Reverts #698.

Routing generated stills through `next/image`'s optimizer broke image previews on real projects. Runware stills are stored as ~6-7 MB PNGs (2848x1600), and the optimizer's upstream fetch runs under a hard-coded 7 s abort signal that also covers the body read. A blob CDN serving those PNGs at 0.5-2 MB/s trips it constantly, and the abort falls through to the generic catch, so `/_next/image` answers 500 for most stills in a project with a few dozen scenes. There is no config knob for that timeout.

Card, storyboard, history and timeline previews go back to `unoptimized` and load the raw blob URL directly. Confirmed by rendering `MediaWithSkeleton` and checking the markup carries the blob URL with no `/_next/image` or `srcset`.

The size win #698 was after is real but has to come from the source: emitting WebP instead of PNG from the Runware provider, or writing a preview alongside each still at upload time. Either lands separately.

### DIFF
--- a/app/components/canvas/elements/AssetTile.tsx
+++ b/app/components/canvas/elements/AssetTile.tsx
@@ -63,7 +63,7 @@ export function AssetTile({
 						src={previewUrl}
 						alt={name ?? ""}
 						fill
-						sizes="80px"
+						unoptimized
 						className="object-cover"
 					/>
 				) : (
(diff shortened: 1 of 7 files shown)


umairnadeem (commented):
lgtm, clean 1:1 revert of #698 and the remaining MediaWithSkeleton callers all line up with the dropped prop.

PR #712, opened 2026-09-02
refactor: nightly conventions pass (sleep helper, dead metadata guards, comment trims)

umairnadeem:
Nightly CONVENTIONS.md audit over the files no open PR touches. Safe no-behavior-change fixes only.

## Fixed
- **One canonical way**: three hand-rolled `new Promise(setTimeout)` sleeps (poll, render-client, mock LLM, mock-utils jitter) now share `sleep()` in `lib/utils.ts`.
- **Define errors out of existence**: `deriveProjectName` and `spokenLanguage` took `Metadata | undefined` though every caller passes a `Metadata`; params are now required and the `?.`/`??` guards are gone. `projectPreamble` stopped guarding `narration`/`characters`, which `MetadataSchema` already defaults.
- **Dumb units**: mock LLM dropped `last?.` after its own null guard.
- **Comments**: removed history narration and an issue number in `sceneSegments`, `timestamps`, `useScriptSync`, `supabase/service`.

## Flagged (design-level, not applied)
- `queue.getElementSnapshot(id?: string)` widens the queue contract for one optional UI prop (AssetTile, still-frame plugin).
- Element-type capability sniffing in `ElementCharacters`, `AnimateButton`, `thumbnail.ts` instead of the `ELEMENT_TYPES` spec.
- Scope-state decision tree re-derived in `EditorToolbar`, `useGenerateScope`, `SceneGenerateButton` (a single `kind` discriminant would make it config).
- `asset-bundle.baseUrl = env ?? ""` masks a missing required env var; `lib/supabase/env.ts` has the canonical `required()` pattern.
- `console.error`-only catches on user-facing paths in `Waveform`, `useAssetPrefetch`, `CaptionFonts` while `toastError` is the documented surface.
- `autosave` logs and swallows the not-hydrated invariant.
- `lib/providers/cache.ts` reads `PINECONE_API_KEY` inside the mechanism (policy in mechanism).
- `SortableItem` has 8 optional props, two for a single caller.
- `lib/auth/session.ts` mixes return-`{error}` and throw contracts; `AuthForm` handles both.
- `validate-code` route maps an unparsed RPC status through `?? "Invalid access code"`.

## Checks
lint, typecheck, format:check, test:run (1512 passed), build: all green.

## Merge-conflict guard
```
PR #711: clean
PR #710: clean
PR #709: clean
PR #707: clean
PR #706: clean
PR #705: clean
PR #704: clean
PR #703: clean
PR #701: clean
[...]

PR #723, opened 2026-09-04
refactor: nightly maintainability 2026-09-04

umairnadeem:
## Summary

Nightly maintainability pass for 2026-09-04, the first over the BYOK connector surface after #701 merged. Rebased onto master at 54208ead.

- `AssetResultSchema` moves into `lib/connectors/types.ts` beside the types it describes; `AssetResult` and `TextTimestamp` are now `z.infer` of it. `snapshots.ts` and `elementHistory.ts` import the schema from its owner instead of `snapshots.ts` carrying a second copy held in line with a `satisfies` check.
- Both waveforms decode their peaks through one `usePeaks` hook in `lib/components/usePeaks.ts`. `ClipWaveform`'s inline copy is gone, and `Waveform` builds its mask declaratively via `soundwaveMaskStyle`, dropping the ref and the imperative mask-image writes. The hook keys its state by source, so a changed source reads as loading in the same render.
- `listProviderKeys` zod-parses its rows instead of casting them, reusing the `byokProviderField` the request schemas already use; `KeyStatus` is derived from a `KEY_STATUSES` list. `providerKeyRpc` returns `unknown` and `readProviderKey` parses what it gets back. `generation-schema.ts` reuses `byokProviderField` too.
- `KeyStatusBadge` drops the comment asked to be removed in the #701 review.

No behavior change. Lint, format, typecheck, knip, build and the full test suite pass on the rebased branch.

## Not taken (needs a human call)

- `BaseAssetConnector.resolveBundle` casts the manifest into `AssetResult`; parsing instead would reject `NaN` durations, which is a behavior change.
- `ProviderKeyRecord` crosses HTTP as an `apiJson<T>` cast like every other own-API response.
- `app/projects/[id]/page.tsx` awaits `listProviderKeys` after the user fetch; it needs `user`, so it is not a simple `Promise.all` fold.
- `HostedProviderCard`'s `?? "invalid"` fallback hides that the hosted row is always present.

## Merge-conflict guard

This branch shares no files with any open PR. The guard script still reports #712 because #712 conflicts with master itself on `lib/providers/llm/mock.ts` (rewritten by #701); pushed on request.

### DIFF
--- a/lib/api/providerKeys.ts
+++ b/lib/api/providerKeys.ts
@@ -1,8 +1,10 @@
 import type { User } from "@supabase/supabase-js";
-import type {
-	ProviderKeyRecord,
-	KeyStatus,
-	ValidationResult,
+import { z } from "zod";
+import {
+	KEY_STATUSES,
+	type KeyStatus,
+	type ProviderKeyRecord,
+	type ValidationResult,
 } from "@/lib/connectors/providerKey";
 import {
 	MANAGED_PROVIDER,
@@ -12,6 +14,7 @@ import {
 import type { Provider } from "@/lib/connectors/types";
 import { createClient } from "@/lib/supabase/server";
 import { createServiceClient } from "@/lib/supabase/service";
+import { byokProviderField } from "./request-schema-fields";
 
 /**
  * The account has no key for a provider it was asked to generate with. An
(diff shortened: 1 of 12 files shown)


lib/api/providerKeys.ts
```diff
@@ -25,16 +28,20 @@ export class MissingProviderKeyError extends Error {
...
 }
 
-type ProviderKeyRow = {
-	provider: string;
-	last4: string;
-	status: KeyStatus;
-	verified_at: string | null;
-	created_at: string;
-};
+const ProviderKeyRowsSchema = z.array(
```
umairnadeem, line 31: fyi this conflicts with #725 even though both descriptions say no overlap: #725 types `ProviderKeyRow.provider` as `BYOKProvider` and adds `hasApiAccess` in this file, and drops the `// TTS types` header in `lib/connectors/types.ts` right where `TextTimestamp` moves into the schema. `git merge-tree` flags both files, so whichever merges second needs a rebase (this file's zod parse should win).

umairnadeem: Rebased #723 onto #725's branch and set it as the base, so this PR is stacked on #725 and the two files merge cleanly: the zod parse wins in providerKeys.ts (the BYOKProvider row typing is subsumed by the parse, hasApiAccess kept), and types.ts drops both the header and the old TextTimestamp alias. Once #725 lands, GitHub retargets this to master; if #725 is squash-merged, one `git rebase --onto origin/master origin/umair/nightly-conventions-20260904-0541` cleans up the diff.

PR #730, opened 2026-09-05
refactor: nightly maintainability 2026-09-05

umairnadeem:
## Summary

Nightly maintainability pass for 2026-09-05. This takes the items the 2026-09-03 pass listed as blocked behind #701, now that it and the rest of the BYOK stack have merged.

- **Job rows are parsed, not cast.** `JobRowSchema` in `lib/api/jobs.ts` is now the one description of the `jobs` table. It is backed by a `BundleResponseSchema` beside the type it already described (`BundleResponse` is now `z.infer` of it) and a `JOB_STATUSES` list that `JobStatus` derives from. `getJob` and `loadJobForProcessing` return parsed rows, matching how `createProject` and `listProviderKeys` already read theirs.
- **`AssetBundle` takes only what it reads.** The constructor accepts the two manifest fields `resolve` and `resolveBundle` use, so `fromResponse` hands the response straight through instead of fabricating `version: 1` and an empty `createdAt`. Eight test fixtures drop the same filler, and the `fromResponse` test stops asserting on it.
- **`GenerationIndicator` only knows active statuses.** `isGenerationActive` is a type predicate over a new `ActiveGenerationStatus`. Every caller already renders the indicator behind that check, so the prop narrows and the idle icon, animation and "Generate" label rows, which no caller could reach, are gone. The three per-status tables and the label function fold into one `STATUS` table.
- **`VideoJob.metadata` is required.** Every producer (`toVideoJob`, both mocks) already sets it, so the optional chains in `BaseVideoProvider.poll`, `MockVideo.store` and the video job handler go, along with the leading `jobId` in `poll` that the spread always overwrote. `VideoPoll`'s pending branch carries the metadata unconditionally.
- **Project columns come from the schema.** `PROJECT_ROW_COLUMNS` is derived from `ProjectRowSchema.shape` and used by both reads, and the home page parses its rows through the same schema instead of handing an untyped array to `ProjectsList`.

No behavior change. Lint, format, typecheck, knip, build and the full test suite (176 files, 1589 tests) pass.

## Not taken

- `render-client.ts` still casts the progress response via `apiJson<RenderProgress>`. #723 settled that own-API responses cross HTTP as casts everywhere; parsing this one alone would be the odd one out.
- The hosted and BYOK vendor tables still each name the six real provider classes. Folding them trades a compile-time guarantee in the hosted table for a runtime lookup, for about ten net lines. Sub-nightly on its own.
- `HANDLERS` stays a `Partial` with the "No job handler registered" throw. Keying it by a narrower connector type would need `JobRow.connector_type` to exclude `llm` and `animated_image`, which is an architecture call rather than a cleanup.

## Merge-conflict guard

```
PR #729: clean
```

### DIFF
--- a/app/components/canvas/elements/GenerationIndicator.tsx
+++ b/app/components/canvas/elements/GenerationIndicator.tsx
@@ -1,27 +1,24 @@
-import {
-	Hourglass,
-	Loader2,
-	Wand2,
-	type IconComponent,
-} from "@/components/ui/icon";
+import { Hourglass, Loader2, type IconComponent } from "@/components/ui/icon";
 import { SimpleTooltip } from "@/components/ui/tooltip";
 import { cn } from "@/lib/utils";
-import type { ElementSnapshot } from "@/lib/generation/snapshots";
-
-type Status = ElementSnapshot["status"];
+import type { ActiveGenerationStatus } from "@/lib/generation/snapshots";
 
 type Size = "sm" | "md";
 
-const ICONS: Record<Status, IconComponent> = {
-	idle: Wand2,
-	queued: Hourglass,
-	generating: Loader2,
-};
-
-const ICON_ANIMATION: Record<Status, string> = {
-	idle: "",
-	queued: "animate-pulse",
-	generating: "animate-spin",
+const STATUS: Record<
+	ActiveGenerationStatus,
+	{ icon: IconComponent; animation: string; label: (seconds: number) => string }
+> = {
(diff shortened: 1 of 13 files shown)


umairnadeem (commented):
lgtm. JobRowSchema lines up with 003_create_jobs.sql (metadata not null default {}, result nullable) and every GenerationIndicator caller is already behind isGenerationActive, so the narrowing is safe.

PR #55, opened 2026-04-06
Add Remotion video rendering pipeline with player UI

umairnadeem:
## Summary
- Integrate [Remotion](https://www.remotion.dev/) for server-side video rendering from generated assets (images, TTS, music, SFX)
- Add scene builder (`lib/video/`) that resolves asset bundles into timed video scenes with proper audio duration metadata
- Add `/api/render` route with SSE progress streaming for real-time render feedback
- Add resizable video player UI (`app/components/video/`) with preview, playback controls, and render trigger
- Extract `PrePromptView` and `PostPromptView` from `Editor` for cleaner component separation
- Audio providers (ElevenLabs, Cartesia, mocks) now expose duration metadata for accurate scene timing

## Test plan
- [ ] Verify scene builder tests pass (`lib/video/__tests__/scene-builder.test.ts`)
- [ ] Verify existing connector and provider tests still pass
- [ ] Test video preview renders in the player panel after content generation
- [ ] Test render API streams progress events via SSE
- [ ] Test resizable panel drag behavior in the video layout
- [ ] Verify pre/post prompt views render correctly in the editor

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #732, opened 2026-09-05
refactor: rethink openslop architecture 2026-09-05

umairnadeem:
## App understanding

OpenSlop is a Next.js 16 + Supabase video-storytelling editor. A Slate canvas of scenes and content elements (tts, image, animated_image, video, music, sfx) is generated through a dependency graph (`lib/generation`), each element type running through a plugin-pipelined connector (`lib/connectors`) that posts to our own routes via a gateway (`lib/gateway`). On the server, a route family (hosted `/api/v1` on our keys, BYOK `/api/third-party` on a stored user key) resolves the picked `{ provider, model }` pair to a vendor model id with `vendorParams` and hands it to a vendor adapter in `lib/providers` (Anthropic, Runware, Cartesia, ElevenLabs). Remotion renders the result; Sloppy is the tool-calling agent over the same editor ops.

## From-scratch architecture

Rebuilt today, the seam between routes and vendor adapters would be one total contract: a provider takes exactly the request its vendor's API takes, with the model id already resolved, and nothing below the route ever chooses a model. Plugin hooks would always receive a context, and a key save or test would always answer with its validation. The current code already had every piece of this in place (`vendorParams`, `VendorParams`, `contextFor`, `verifyProviderKey`) but typed the receiving side as optional, so each vendor re-declared a fallback model, every hook guarded a context that is never missing, and the client defaulted a missing validation to ok.

## Implemented

- **Providers take the vendor request.** `VendorParams<TReq>` loses its `extends ModelRef` constraint, so `ImageProvider`, `SFXProvider`, `MusicProvider`, `TTSProvider` and `LLMProvider` are typed on `VendorParams<XGenerateParams>` with a required `model: string`. That is the type the job handlers and LLM routes were already passing. The Runware image, Cartesia, ElevenLabs music and Anthropic adapters drop their hand-copied model fallbacks (`bytedance:seedream@5.0-lite`, `sonic-3.5`, `music_v1`, `claude-opus-5`), which duplicated knowledge homed in `lib/connectors/*/models.ts` and could drift from the catalog silently. Mocks are untouched: they accept the wider connector params and still satisfy the interfaces.
- **`AnthropicLLM implements LLMProvider` directly** (flag #55). It only extended `BaseProvider` to stub `toFiles()`, override `store()` to a no-op and carry a dead `blobConfig`. `MockLLM` already implemented the interface directly.
- **Plugin hooks always get a context** (flag #57). `ConnectorPlugin` hooks, the `run*` runners and the `require*` helpers take a required `PluginContext`; `BaseConnector` always built one. The `ctx?.` guards in `plugins.ts`, `still-frame.ts`, `character-references.ts` and `voice-search.ts` go away. Context fields stay optional since they vary per connector.
- **A key check answers with a required validation** (flag #61). `ProviderKeysView` is the list alone; `ProviderKeyCheck` adds `validation`. The two POST routes return `providerKeyCheck`, DELETE returns `providerKeysView`, and `accountStore` reads `.validation` off the check instead of defaulting a missing one to `{ ok: true }`, which would have told the user a key works when the server said nothing.
- ARCHITECTURE.md states the provider contract in one sentence.

Net: 27 files, +190/-132 (most of the additions are test fixtures now passing a vendor model id).

## Validation

- `npm run typecheck` clean
- `npm run lint` clean
- `npm run format:check` clean
- `npm run knip` clean
- `npm run build` succeeds
- `npx vitest run`: 176 files, 1589 tests passed
- `madge --circular lib app`: 19 cycles, same as master, all type-only

## Skipped

- `lib/providers/video/{base,runware,mock}.ts` keep their `params.model || "bytedance:seedance@2.0-fast"` fallback: all three files are changed by open PR #730. Retyping `VideoProvider` on `VendorParams<VideoGenerateParams>` is a 4-line follow-up once #730 lands.
- `HANDLERS` Partial + "No job handler registered" throw (flag #49): `lib/api/jobs.ts` is in #730.
- `static attributesFor(_model)` (flag #56): every connector ignores the model today, but `editorOps.mergeAttrs` documents it as a deliberately kept per-model seam; left for a human call.
- `lib/providers/image/runware.ts` and `video/runware.ts` still default `width`/`height`: those are dimension defaults, not model knowledge, and the video one is reserved anyway.

## Merge-conflict guard

```
PR #731: clean
PR #730: clean
PR #729: clean
```

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -30,7 +30,7 @@ Users see models and providers; developers also see connectors. Each word means 
 
 - **Connectors** (`lib/connectors/`): what the editor calls. One class per media type, plugin-pipelined, serving every provider of that type.
 - **Gateways** (`lib/gateway/`): thin HTTP clients from connectors to our own routes. The provider a model names picks the route family: `/api/v1` for models OpenSlop hosts, `/api/third-party` for models a user brings a key for.
-- **Providers** (`lib/providers/`): server-side vendor adapters (Runware, ElevenLabs, Cartesia, Anthropic). The same classes serve both families; only where the key comes from differs.
+- **Providers** (`lib/providers/`): server-side vendor adapters (Runware, ElevenLabs, Cartesia, Anthropic). The same classes serve both families; only where the key comes from differs. A provider takes the request as its vendor's API does, the model id already resolved by the route, so no vendor falls back to a model of its own.
 
 ## Models and provider keys
 
--- a/app/api/providers/[provider]/route.ts
+++ b/app/api/providers/[provider]/route.ts
@@ -1,9 +1,10 @@
 import { NextResponse } from "next/server";
 import { z } from "zod";
 import {
-	providerKeysView,
 	deleteProviderKey,
 	MissingProviderKeyError,
+	providerKeyCheck,
+	providerKeysView,
 	readProviderKey,
 } from "@/lib/api/providerKeys";
 import { verifyProviderKey } from "@/lib/api/providers/byok";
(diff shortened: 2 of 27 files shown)


umairnadeem (commented):
lgtm. checked the VendorParams narrowing end to end (llm-routes + jobVendorParams both resolve the model before the provider sees it, contextFor always hands plugins a ctx), all reads clean.

PR #733, opened 2026-09-05
feat: let Sloppy read element generation state

### LINKED ISSUE #610: Let Sloppy read element generation state (generated vs stale vs ungenerated, durations, errors)

### LINKED ISSUE #720: Let Sloppy read element generation state and per-element generation history (and restore a past version)
## Problem

Every generation an element has ever had is now recorded and persisted, but Sloppy cannot see any of it, and it still cannot tell whether an element is generated at all.

What the app knows per element today:

- **Current state.** `ElementSnapshot` (`lib/generation/snapshots.ts:21-32`) carries `status`, `result`, `error`, `resultInputs`, `connectorType`, `pinned`. `needsGeneration` / `isNodeStale` (`lib/generation/graph.ts:153-173`) decide "ungenerated" vs "stale" vs "up to date", and `staleReason` (`lib/generation/staleReason.ts:46-60`) already renders *why* in the user's words.
- **Full history.** `ElementHistory` (`lib/generation/history.ts`) keeps an append-only `VersionLog` of `ElementVersion`s per element — `createdAt`, `connectorType`, `elementType`, `inputs` (prompt + attributes + dependency identities), `result`, `pinned` (`lib/generation/versions.ts:7-17`). It is persisted to the `element_history` table (`lib/project/elementHistory.ts`) and restorable with `restoreElementVersion` (`lib/generation/restore.ts:19`).
- **A UI for it.** `ElementHistoryPopover` (`app/components/canvas/elements/ElementHistoryPopover.tsx`) lists versions with a thumbnail, relative time, pinned marker, truncated prompt, and a restore button.

What reaches Sloppy: none of it. `AgentToolContext` (`lib/agent/tools/context.ts:20-41`) has `avatarUrl` and `elementImage` (which peeks at `status` + image url for exactly one picture), and `read_script` (`lib/agent/tools/readScript.ts:44-54`) returns characters + OSML, which by design carries no state (`lib/canvas/osmlSerializer.ts`).

So a user who asks Sloppy any of these gets a guess or a redirect to the canvas:

- "Is scene 3 generated yet?" / "What's still missing?" / "Which ones are stale, and why?"
- "Go back to the version of the hero image from before I changed the style." Sloppy can't list versions, can't say what changed between them, and can't restore one.
- "How many takes have we done of this narration? Which prompt made the good one?"
- "You just edited 4 elements — what did that invalidate?" After `edit_script`, Sloppy has no way to check the damage it did.

The LIMITS block in `lib/agent/prompt.ts:55-62` papers over this with an unconditional "tell the user to press generate" line because Sloppy can't know whether there is anything to generate.

- "let sloppy see generation history for elements and whether or not an element is generated"

## Relevant code

- `lib/generation/snapshots.ts:21-32` — `ElementSnapshot`: the current state per element.
- `lib/generation/graph.ts:153-173` — `needsGeneration` / `isNodeStale`, the canonical tests. Reuse, do not reimplement.
- `lib/generation/staleReason.ts:46-60` — human-readable stale reason, already written for the badge tooltip; the same string is what the model should read.
- `lib/generation/history.ts` — `ElementHistory.get(id)` / `load(id)` / `status(id)`; note history is lazily hydrated per element, so a read may need to `await load(id)` first.
- `lib/generation/versions.ts:7-17` — `ElementVersion`, the per-take record.
- `lib/generation/restore.ts:19` — `restoreElementVersion(queue, history, version)`: restores owned dependencies (the still behind an animated image) alongside; a tool must call this, not `queue.restoreResult` directly.
- `lib/generation/ElementHistoryProvider.tsx` — `useElementHistoryStore`, the handle to hand into `useAgentTools`.
- `lib/agent/tools/context.ts:20-41` — where the accessor(s) belong. `elementImage` is the existing precedent for "project a snapshot into something the model can safely read."
- `lib/agent/tools/useAgentTools.ts:30,44-56` — `queue` is already in scope; `history` needs to be brought in the same way.
- `lib/agent/tools/readScript.ts` — the read path to extend for current state.
- `lib/agent/tools/registry.ts:103-111` — `SNAPSHOT_TOOLS` (`snapshot: true`) keeps turn-scoped output from lingering; both new reads belong there.
- `lib/agent/prompt.ts:55-62` — LIMITS boilerplate to retire once state is visible.
- `lib/generation/useNodeBuilder.ts` + `app/components/canvas/hooks/useGenerateScope.ts:73-96` — how a node is built from an element to run `needsGeneration` against; the agent layer needs the same `buildNode(forElement(element))` step.

## Suggested direction

Two capabilities, one projection shape. The current state is cheap and should ride along with every `read_script`; the history is per-element and on demand.

1. **Current state on `read_script`** (this is #610, restated against today's code). Add an `elementStates()` accessor to `AgentToolContext` that maps every content element to `{ id, state, reason?, durationSec?, error?, versions }` where `state` is one word — `ungenerated` / `queued` / `generating` / `generated` / `stale` / `failed` / `pinned` — collapsed at the boundary from `status` + `result` + `isNodeStale` + `pinned` + `error`. `reason` is `staleReason()` verbatim. `versions` is the count of recorded takes, so the model knows history exists before it asks for it. Render it as a compact block under the OSML in `read_script` output, never as OSML attributes.

2. **`read_element_history` tool** (new). Input: `{ id }`. Output: the element's versions, oldest to newest, each as `{ index, createdAt, prompt, attributes, pinned, durationSec?, current: boolean }` — the same fields the popover shows, plus a `current` flag matched by `resultIdentity(version.result)` against the live snapshot, and a diff-friendly `changed` list naming which inputs differ from the previous version (reuse the `changedInputs` logic behind `staleReason` rather than writing a second comparator). Awaits `history.load(id)` first; `status === "failed"` is reported as a tool error, an unhydrated-but-empty list reads as "no versions yet". Flag `snapshot: true`.

3. **`restore_element_version` tool** (new, small). Input: `{ id, index }`. Calls `restoreElementVersion(queue, history, version)` so owned dependencies come along. This is the write half of "go back to the one from before"; without it, the history read just lets Sloppy describe a button the user still has to press. Confirmation policy is the same as `edit_script`: Sloppy should say what it restored.

4. **Never expose internals.** No urls, no `resultInputs`, no dependency identity hashes, no connector config. Prompt and attributes are the user's own words and are fine; `dependencies` collapses to a `changed` entry like "character avatar changed".

5. **Retire the guesswork.** With state visible, LIMITS should say "you cannot *run* generation" (until #462 / #721 land) and drop the "tell the user to press generate" boilerplate — Sloppy can now say "3 elements are stale (scene 2 narration: voice changed; …), hit Generate All."

6. **Timeline durations.** Where a version's `result.durationSec` is real (audio/video), surface it; `measure_element_lengths` (`lib/agent/tools/measureElementLengths.ts`) is an estimate from word count and should defer to the generated duration when one exists.

## Acceptance criteria

- [ ] `read_script` output tells Sloppy, per element, whether it is ungenerated, queued, generating, generated, stale (with the same reason the badge shows), failed (with the error), or pinned.
- [ ] Staleness and its reason come from `isNodeStale` / `staleReason`, not a recomputation in the agent layer.
- [ ] Sloppy can list an element's generation history: when each take was made, the prompt/attributes that produced it, what changed from the previous take, and which one is current.
- [ ] Sloppy can restore a past version by index, and owned dependencies restore with it (via `restoreElementVersion`).
- [ ] History reads hydrate lazily via `ElementHistory.load` and handle the `loading` / `failed` states.
- [ ] Both reads are `snapshot: true` so they cannot linger as stale context after an edit.
- [ ] OSML serialization is unchanged; no state is smuggled in as attributes.
- [ ] No urls, dependency identities, `resultInputs`, or provider config reach the model.
- [ ] The "press generate" boilerplate in `prompt.ts` LIMITS is replaced by state-aware guidance.

## Related

- #610 — Let Sloppy read element generation state. This issue supersedes it against the now-landed `ElementHistory` and `staleReason`; the "current state" half is the same ask, and #610 can be closed as a duplicate of this or vice versa.
- #451 — Multiple generations per element + history. The model/persistence layers from that issue have landed (`lib/generation/history.ts`, `element_history` table); this is the agent-facing read of the same data.
- #598 — Sloppy views generated image results (closed). `elementImage` on `AgentToolContext` is the precedent for a safe snapshot projection.
- #537 — Show why an element is stale (closed). `staleReason` is the output this issue reuses.
- #667 — History panel offers no way back to the latest version; a `restore_element_version` tool has the same "which one is current" question and should share the answer.

umairnadeem:
## Summary

Sloppy could not tell whether an element was generated, stale, failed, queued or generating. This projects the state the app already keeps (`ElementSnapshot` plus `staleReason`) into `read_script`.

- **`read_script` gains a "Generation state" section** under the OSML: one line per element with a single state word (`ungenerated` / `queued` / `generating` / `generated` / `stale` / `failed` / `pinned`) and, for stale and failed, the badge's stale reason or the error verbatim. OSML itself is unchanged.
- **`elementStates()` on `AgentToolContext`**, built in `useAgentTools` from the queue snapshot and `staleReason` over the same `nodeBuilder` the canvas uses.
- **Prompt**: LIMITS drops the unconditional "press generate" line for state-aware guidance.
- `seconds` formatter shared between `fit_durations` and `measure_element_lengths`.

No urls, dependency identities, `resultInputs`, or provider config reach the model.

## Selected issue

Closes #720
Closes #610

The issue's history read and version restore are not taken: exposing generation state is what closes it.

## Validation

- `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run knip`: clean
- `npx vitest run`: 177 files, 1596 tests passing
- `npm run build`: succeeds

https://claude.ai/code/session_019BJCsSS3VjjPRAZckCLRgu


lib/agent/tools/useAgentTools.ts
```diff
@@ -54,6 +69,29 @@ export function useAgentTools(editor: Editor) {
...
+					const [element, path] = found;
+					await history.load(id);
+					const versions = history.get(id);
+					return {
+						versions: summarizeVersions(nodeFor(element), versions, queue),
+						restore: async (number) => {
+							const version = versions[number - 1];
+							if (!version) throw new Error(`${id} has no version ${number}`);
+							const restored = restoreElementVersion(queue, history, version);
+							applyNodeVersion(editor, path, version);
```
umairnadeem: `path` is captured before the `await history.load(id)` above, so if the user adds/removes an element above this one while that supabase read is in flight, the take's text + attrs get written onto whatever element now sits at the stale path. the history button dodges this by resolving `findPath` at click time; maybe re-run `findNodeById(editor, id)` inside `restore` and apply to that path instead.

umairnadeem (commented):
lgtm, thanks for the revisions! narrowing it to just the state read was the right call

PR #739, opened 2026-09-06
fix: ride out dropped poll requests instead of failing the generation

umairnadeem:
## Summary
- A single dropped poll request (browser `net::ERR_TIMED_OUT`, fetch rejecting with TypeError) no longer fails the whole generation. `apiFetch` classifies it as a typed `UnreachableError` at the boundary and both poll loops (asset jobs, export progress) retry on the next tick up to the existing deadline. Server-answered errors and aborts still throw immediately.
- Asset job poll interval raised from 1s to 5s; the render client shares the same constant.

## Test plan
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck`, `npx vitest run` (1593 passing)
- [x] New tests: typed error at the fetch boundary, abort passthrough, retry through an unreachable poll in both loops, deadline still fires when the server never answers
- [ ] Manual: start a generation, briefly toggle the network offline in DevTools, confirm the element completes once back online

https://claude.ai/code/session_01PyATdBRznz2dPH2Sg4QLZD

### DIFF
--- a/lib/clients/__tests__/http.test.ts
+++ b/lib/clients/__tests__/http.test.ts
@@ -1,5 +1,5 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import { apiFetch, apiJson } from "../http";
+import { apiFetch, apiJson, UnreachableError } from "../http";
 
 describe("apiFetch", () => {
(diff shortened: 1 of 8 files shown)


umairnadeem (commented):
lgtm

PR #742, opened 2026-09-07
feat: add Kling 3 Turbo with per-model resolution, image format and loop controls

umairnadeem:
## Summary
- Adds Kling 3 Turbo (`klingai:kling-video@3.0-turbo`) for clips and animated images, hosted as "Slop Video v1 Fast" and as "Kling 3 Turbo" on a Runware key.
- New per-element controls: resolution (options from the model's catalog entry), image output format (JPG default, since Kling only takes JPEG/PNG frames), a boolean loop toggle with Play/Repeat icon faces, and volume on animated images. The Seedance audio-off setting is removed.
- Runware video requests are sized by preset when a frame image is present and by pixels otherwise; a looping clip restarts while an unlooped one holds its last frame; the scene builder is untouched so `trimVisualsToDialogue` still governs visual length.
- `modelEntry` is generic over the catalog it receives, and Sloppy's "Attributes by type" contract is derived from the attribute schemas.

## Test plan
- [x] `npm run lint`, `format:check`, `typecheck`, `knip`, `build`, and the full Vitest suite pass (1604 tests).
- [ ] Generate an animated image on Slop Video v1 Fast at 720p and 1080p; confirm no `unsupportedParameter` or WebP frame error.
- [ ] Generate a clip on Kling and on Seedance; confirm Seedance now carries audio and the volume slider applies.
- [ ] Toggle loop on a clip whose dialogue outruns it; confirm it repeats, and holds its last frame when off.
- [ ] Switch a 1080p Kling element to Seedance; confirm resolution resets to 720p.

https://claude.ai/code/session_019BJCsSS3VjjPRAZckCLRgu


lib/providers/video/runware.ts
```diff
@@ -17,6 +23,20 @@ function toVideoJob(video: {
...
 }
 
+const DEFAULT_MODEL = RUNWARE_VIDEO_MODELS["Seedance 2 Fast"].id;
+
+const DEFAULT_SIZE =
+	ASPECT_RATIO_DIMENSIONS[DEFAULT_ASPECT_RATIO].video[DEFAULT_VIDEO_RESOLUTION];
+
+/** A frame-conditioned video takes its aspect from the frame, so it is sized by preset. */
+const sizeFor = (params: VideoGenerateParams) =>
+	params.frameImages && params.resolution
```
umairnadeem, line 33: does seedance actually take `resolution` with a frame image? this path isn't kling-only: Slop Video v1 (seedance) is still the default, so every existing animated_image now sends `resolution: "720p"` and no width/height where it used to send 1280x720. the test plan only covers frame-conditioned gen on kling, and the PR body already mentions `unsupportedParameter` as the failure mode. worth one probe on seedance before merge, or gate the preset on the model's catalog entry instead of on frameImages.

umairnadeem: Yes. Probed Runware's validator with the exact shapes the provider now sends, using an unreachable frame URL so validation runs but nothing bills:

| Request | Result |
|---|---|
| Seedance, `frameImages` + `resolution: "720p"`, no width/height, no `settings` | passes validation, fails only on downloading the fake frame (`invalidValueDownloadFailed`) |
| Seedance clip, `width`/`height` 720x1280 | passes validation, fails only on the deliberate bad duration (`invalidDuration`, 4-15s) |
| Kling, same frame-image shape | passes validation, same download failure |

Runware's Seedance doc states the same rule as Kling's: `resolution` requires `frameImages`, cannot be combined with width/height, and 720p is an allowed value. The aspect comes from the frame, so a 9:16 still yields the same 720x1280 clip the old pixel request produced.

Kept the gate on `frameImages` rather than the catalog entry because the rule is Runware's and both shipped models follow it; a per-model flag would state the same fact twice. If a future model breaks it, a capability on its entry is the place to add it then.

PR #737, opened 2026-09-05
fix: raise Remotion Lambda timeout to 900s for 4K exports

umairnadeem:
Closes #736

## What

- `TIMEOUT` in `lib/video/lambda-config.ts` goes from 240s to 900s (the AWS Lambda ceiling, and Remotion's `MAX_TIMEOUT`). The function name derives from the spec, so this is a new function: `remotion-render-4-0-481-mem3008mb-disk10240mb-900sec`.
- The progress route picks the fatal error with Remotion's own predicate (`isFatal && !willRetry`, the same one that drives `fatalErrorEncountered`), so a retried chunk flake ahead of the real failure is no longer reported as the cause. The Remotion message itself is passed through unchanged, CloudWatch links included.

## Deploy state

- **Already deployed**: the 900s function was created with `npm run deploy:remotion-function` and is listed by `remotion lambda functions ls`. Callers pick it up via `getFunctionName()` the moment this merges.
- **Not done, do after merge**: delete the old functions (`...-4-0-481-...-240sec`, plus the stale `4-0-459` and `4-0-477` ones still listed). Deleting before merge would break production exports.

## Not taken (from the issue)

- Rewording the timeout error for users. Deliberately kept raw: the project is open source and the CloudWatch links are useful.
- `framesPerLambda` as a function of `scale`, and raising `RAM` so the stitch finishes faster. The issue asks for measurement on a real 4K export before committing to numbers, and there is no measurement yet. These are the levers if 900s still is not enough.

## Checks

Lint, format, typecheck, knip, build, and the full vitest suite pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019BJCsSS3VjjPRAZckCLRgu

PR #745, opened 2026-09-07
refactor: nightly maintainability 2026-09-07

umairnadeem:
## Summary

Completes the server-side provider contract for video. The video provider was the last one whose `generate`/`poll` took an optional `model` and, in `RunwareVideo`, quietly fell back to Seedance when it was missing. ARCHITECTURE.md already states that a provider takes the request as its vendor's API does with the model id resolved by the route, and #732 moved the image, TTS, SFX, music and LLM providers to that contract. This PR does the same for video.

- `lib/providers/video/base.ts`: `VideoRequest = VendorParams<VideoGenerateParams>`; `VideoProvider`, `BaseVideoProvider` and `poll` are typed on it, so `model: string` is required.
- `lib/providers/video/runware.ts`: drops the `DEFAULT_MODEL` fallback and the import of the connector-layer Runware model table. The provider no longer knows any catalog name.
- `lib/providers/__tests__/runware-video.test.ts`: every request names its model, as the job handler already does in production.

No behavior change: the video job handler always passes `jobVendorParams(job)`, which carries the resolved id, so the fallback was unreachable on the real path.

## Skipped

- `HANDLERS` Partial + throw in `lib/api/job-handlers.ts`: needs `JobRow.connector_type` narrowed past `llm`/`animated_image`, an architecture call for the arch job.
- Runware image `width`/`height` defaults: dimension knowledge rather than model knowledge, previously reviewed and kept.
- Everything merged since the last pass (#733, #737, #739, #742, #744) was read and is clean.

## Checks

lint, format:check, typecheck, knip, `npm run build`, `vitest run` (178 files, 1612 tests) all pass.

## Merge-conflict guard

```
```

### DIFF
--- a/lib/providers/__tests__/runware-video.test.ts
+++ b/lib/providers/__tests__/runware-video.test.ts
@@ -20,6 +20,8 @@ vi.mock("@runware/sdk-js", () => ({
 
 import { RunwareVideo } from "../video/runware";
 
+const MODEL = "bytedance:seedance@2.0-fast";
+
 describe("RunwareVideo", () => {
 	beforeEach(() => {
 		vi.clearAllMocks();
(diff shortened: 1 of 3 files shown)


umairnadeem (commented):
lgtm, thanks!

PR #731, opened 2026-09-05
refactor: nightly conventions sweep 2026-09-05, dead guards in the LLM mock and serializer

umairnadeem:
## Summary

Nightly CONVENTIONS.md sweep for 2026-09-05. The delta since the 2026-09-04 sweep (PR #725) is the eight PRs that merged on 09-04 (#711, #712, #713, #714, #715, #722, #724, #727), about 60 source files once the 13 files held by open PR #730 are excluded. All of them were read against the seven rules; the repo-wide invariant greps were re-run as well.

All changes are no-behavior-change: lint, format, typecheck, the full test suite (1589 tests) and `npm run build` pass.

## Auto-fixed (2 files)

**Rule 1, define errors out of existence**
- `lib/providers/llm/mock.ts`: `mockCall` guarded `last.text ?? ""`, but `lastToolResult` already returns `text: string` (it goes through `outputText`, which never yields undefined).
- `lib/providers/llm/mock.ts`: `agentModel(model = "mock")` kept a default the `LLMProvider` interface made required in #725; the only caller passes `request.model: string`.
- `lib/canvas/osmlSerializer.ts`: `getElementText` mapped `child.text ?? ""` over `CanvasText[]`, whose `text` is a required `string`.

**Rule 5, one canonical way**
- `lib/providers/llm/mock.ts`: `MockLLM.stream` restated `{ text: string; done: boolean }` inline; it now names `LLMStreamChunk` like `LLMProvider`, `AnthropicLLM` and the HTTP gateway do.

## Audited clean

Every other delta file reads as intended: the agent tool registry is one table with derived views, the preview registry derives from `ELEMENT_TYPES`, `motionEffects` and `ResizeHandle` are lookup tables, `autosave` and `queue` route every failure to a typed handler, and the API routes stay thin over the shared handler factories.

Repo-wide greps unchanged: zero bare `catch {}`, zero catch-return-null, zero `as any` / non-null assertions, zero provider sniffing, no hand-rolled `clamp` or `sleep`, no deprecated `z.string().url()`; the only `switch` statements are the three discriminated-union dispatches (`applyOps:25`, `scene-builder:121`, `VideoComposition:59`).

## Flagged, not changed

- `lib/project/autosave.ts:79`: `persist` both `console.error`s and calls `onError`, and the caller's `onError` already toasts. The console line is a second report of the same failure. Dropping it is a logging change, so it is left for a human call. Same for the `not hydrated` line at :91, which is standing flag #32.
[...]

PR #746, opened 2026-09-07
refactor: nightly conventions sweep 2026-09-07, dead guards and one canonical read in the post-#731 delta

umairnadeem:
Nightly CONVENTIONS.md sweep. Delta-audited the 66 source files merged since the last sweep (#730, #732, #733, #737, #739, #742) and applied only no-behavior-change fixes. Repo-wide invariants unchanged: zero bare `catch {}`, zero catch-return-null, zero `as any` / non-null / provider sniffing, the same 3 discriminated-union switches.

## Fixed (no behavior change)

- `lib/connectors/plugins.ts`: `requireGateway` had zero callers. Deleted with its import.
- `lib/connectors/animated_image/plugins/still-frame.ts`: `STILL_ONLY_KEYS` read `formatDef.key` instead of re-spelling `"format"`; dropped a doc comment that restated the name.
- `lib/connectors/image/plugins/character-references.ts`: `ctx` is contextually typed by the `ConnectorPlugin<ParamsWithCharacters>` return type. Dropped the annotation and the now-unused import.
- `lib/providers/tts/cartesia.ts`: `WebsocketResponse.Chunk.audio` is `Buffer | null`, so after the `&& response.audio` narrowing the `Buffer.from` arm was unreachable. `PCM_WAV_PARAMS` is keyed by the SDK's `RawEncoding` so a typo in `ENCODING` fails at compile time instead of module load. `_search` takes the `VoiceSearchParams` instead of unpacking two optionals to repack them. `GenerationRequest` is a type-only export, so `import type`.
- `lib/providers/llm/anthropic.ts`: `response.response.modelId` is a `string` on the AI SDK type, so `?? params.model` was dead. `thinking(effort)` now takes `ThinkingLevel`, which is what both callers pass.
- `lib/generation/snapshots.ts`: `resetToIdle` rebuilt every `ElementSnapshot` field by hand; it now goes through `update` (which does not bump `resultVersion` since `result` is not in the patch, same as before). `isBusy` is `getActiveCount() > 0`.
- `lib/api/asset-bundle.ts`: inlined the single-use module `fetchJson` into `AssetBundle.fetchJson`.
- `app/components/canvas/elements/GenerationIndicator.tsx`: `className = ""` default is dead since `cn` drops `undefined`; the icon paints `currentColor` so its own `text-foreground` duplicated the wrapper's.
- `remotion/compositions/VideoComposition.tsx` + `lib/video/transitions.ts`: layered sequences premounted a bare `fps` (1s). Named as `LAYER_PREMOUNT_SEC = 1` beside `VIDEO_PREMOUNT_SEC`; `toFrames(1, fps) === fps`.
- `lib/video/elementAttributes.ts`: dropped a comment that restated the body of `layoutAttributeSignature`.

## Flagged, not fixed (design-level or behavior change)

- **`app/page.tsx:75-77` "Get Started" button is a no-op.** No handler, not inside a `<form>`, `AccessCodeInput` has no submit. Either wire it to submit the access code or remove it. Possibly a live bug.
- **`GenerationIndicator.tsx:50-62`** renders a status marker as a `disabled` `<button>`. `disabled` suppresses pointer events in Firefox/Safari, so the tooltip will not open there, and it uses one-off `disabled:cursor-not-allowed` instead of the `unavailable:` variant. A `<span role="status">` fits better.
- **`GenerationIndicator.tsx:25,28,54` and `PlayPauseFlash.tsx:15`** pair `bg-on-media/55` with `text-foreground`; DESIGN.md says scrims take `text-on-media-foreground`.
- **`lib/providers/tts/cartesia.ts:181-189`** catch-warn-continue: a failed embedding call returns unranked voices and the caller cannot tell. Rule 6 says propagate or surface. Dates from #108, so not new, but it is the only `.catch(` with a swallow left in `lib/`.
- **`lib/providers/tts/cartesia.ts:262-264`** the provider pads `durationSec` by `+1` for a downstream pause opinion; ElevenLabs reports raw duration. Policy in the mechanism. Tests encode the `+1`.
- **`lib/providers/tts/cartesia.ts:213`** `if (!params.voiceId) throw`: the route schema requires it but `TTSGenerateParams.voiceId` stays optional. Requiring it on the TTS request type defines the guard away (14 references to check).
- **`lib/api/handlers/video.ts:30-37`** sniffs `metadata.status === "failed"` inside the `pending` variant of `VideoPoll`. A `failed` variant produced in `BaseVideoProvider.poll` would let the handler switch on `kind` only. Touches `lib/providers/video/base.ts`, under open #745.
- **`lib/api/handlers/video.ts:18-23`** the no-`jobId` throw exists only because `VideoProviderResponse.metadata` is typed optional while every producer sets it. Same file as above.
- **`lib/api/providerKeys.ts:60-66`** hosted key row uses `""` sentinels for `last4`/`createdAt`; the settings UI special-cases `MANAGED_PROVIDER` in three places so they never render. A discriminated `hosted | stored` record would remove both.
- **`lib/connectors/types.ts:83`** `PluginContext.gateway` is written by `HttpLLMConnector.pluginContext()` and read by nobody now that `requireGateway` is gone. Removing the field is a contract change.
- **`lib/connectors/plugins.ts:15-31`** `requireState`/`requireModel` plus the hand-rolled guard in `voice-search.ts:23-27` are one keyed mechanism; a generic `requireContext(ctx, key, plugin)` collapses them without a cast. API change across 6 call sites.
- **`lib/connectors/tts/plugins/voice-search.ts:30`** `language || "en"` is an unnamed default inside the plugin.
- **`lib/connectors/video/openslop/models.ts`** hosted entries copy every field from the Runware table by hand (same for image). Aliasing changes dependency direction.
- **`lib/connectors/animated_image/plugins/still-frame.ts:29-39`** `VIDEO_ONLY_KEYS` names two of `clipPlaybackDefs`' five keys; deriving it from the schema changes what the still node carries.
- **`lib/connectors/types.ts:213-216`** `imageProvider`/`imageModel` are spelled here and in `STILL_MODEL`.
- **`lib/connectors/image/enums.ts:13-18`** `enum EffectType` where every sibling closed set is an `as const` array (two other enums exist, repo-wide call).
- **`lib/video/types.ts:57-64`** 1920x1080 lives in `DEFAULT_CONFIG`, `ASPECT_RATIO_DIMENSIONS["16:9"]` and `CAPTION_BASE_HEIGHT`; tying them changes what scale 1 means for portrait.
- **`lib/video/elementAttributes.ts:46-90`** `getVolume`/`getLoops`/`getDuration` read through `flatAttributes` while `getLoop`/`getMotion` read `layoutAttributes` directly.
- **`AttributeBadge.tsx:20-21` + `AttributeTrigger.tsx:21`** hand-roll the `Badge` pill tokens twice; wants a `Badge` size. (Flag #20 if-chain still standing.)
- **`lib/api/jobs.ts:93-94`** `omitBy` + empty-patch early return guard a case no caller produces; a transition union would define it away.
- **`lib/providers/music/elevenlabs.ts:32`** `params.model as "music_v1"` asserts what the catalog guarantees.
- Low: `lib/api/asset-bundle.ts:68` env default `""`, `snapshots.ts:134` `commit` returns its input for a test only, `app/page.tsx:14-17` module-scope readdir + logo strip in a route.

## Not taken on purpose

- `fitDurations.ts:86-88` empty-ops ternary: `registry.test.ts:465` throws if `editScript` is called with nothing to change, so the no-call contract is tested.
- `cartesia.ts:190` `limit || ranked.length` → `slice(0, limit)`: the route schema excludes 0 but the type does not.
- `components/ui/icon.tsx` export-list re-sort: churn only.
- `useAgentTools.ts:52-59` picture snapshot read: rewrite is no simpler.

## Checks

`npm run lint`, `npm run typecheck`, `npm run format:check`, `npm run test:run` (178 files, 1612 tests) and `npm run build` all pass.

## Merge-conflict guard

```
PR #745: clean
```

### DIFF
--- a/app/components/canvas/elements/GenerationIndicator.tsx
+++ b/app/components/canvas/elements/GenerationIndicator.tsx
@@ -34,7 +34,7 @@ export function GenerationIndicator({
 	status,
 	seconds = 0,
 	size = "md",
-	className = "",
+	className,
 }: {
 	status: ActiveGenerationStatus;
 	seconds?: number;
@@ -58,7 +58,7 @@ export function GenerationIndicator({
 				)}
 				disabled
 			>
-				<Icon className={cn(sizes.icon, "text-foreground", animation)} />
+				<Icon className={cn(sizes.icon, animation)} />
 			</button>
 		</SimpleTooltip>
 	);
--- a/lib/api/asset-bundle.ts
+++ b/lib/api/asset-bundle.ts
@@ -9,11 +9,6 @@ async function fetchOk(url: string, label: string): Promise<Response> {
 	return res;
 }
 
-async function fetchJson<T>(url: string, label: string): Promise<T> {
-	const res = await fetchOk(url, label);
-	return res.json() as Promise<T>;
-}
-
 type AssetManifest = {
 	version: number;
 	type: string;
@@ -80,7 +75,8 @@ export class AssetBundle {
 	}
 
 	async fetchJson<T>(key: string): Promise<T> {
-		return fetchJson<T>(this.resolve(key), `Failed to fetch "${key}"`);
+		const res = await fetchOk(this.resolve(key), `Failed to fetch "${key}"`);
+		return res.json() as Promise<T>;
 	}
 
 	static buildUrl(type: string, provider: string, id: string): string {
--- a/lib/connectors/animated_image/plugins/still-frame.ts
+++ b/lib/connectors/animated_image/plugins/still-frame.ts
@@ -13,6 +13,7 @@ import {
 import { buildImagePlugins } from "@/lib/connectors/image/plugins/imageChain";
 import { resolveModel } from "@/lib/connectors/models";
 import { ELEMENT_MODEL } from "@/lib/connectors/attributes/model";
+import { formatDef } from "@/lib/connectors/attributes/common";
 import { STILL_MODEL } from "../attributes";
 import {
 	derivedDependency,
@@ -36,7 +37,7 @@ const VIDEO_ONLY_KEYS = [
 const STILL_MODEL_KEYS = Object.values(STILL_MODEL);
 
 /** Attributes of the still, which the animation's video generation has no use for. */
-const STILL_ONLY_KEYS = ["format", ...STILL_MODEL_KEYS];
+const STILL_ONLY_KEYS = [formatDef.key, ...STILL_MODEL_KEYS];
 
 const STILL = "still";
 
@@ -97,9 +98,6 @@ export function pictureNode(node: GenerationNode): JobNode | null {
 	return makesPicture(target.job.elementType) ? target : null;
 }
 
-/**
- * The snapshot of the still a node depends on
- */
 export const stillSnapshot = (
 	node: GenerationNode,
 	queue: GenerationQueue,
--- a/lib/connectors/image/plugins/character-references.ts
+++ b/lib/connectors/image/plugins/character-references.ts
@@ -3,7 +3,7 @@ import {
 	CHARACTERS_ATTR,
 	parseCharacterNames,
 } from "@/lib/canvas/characterNames";
-import type { ConnectorPlugin, PluginContext } from "@/lib/connectors/types";
+import type { ConnectorPlugin } from "@/lib/connectors/types";
 import { characterAvatarElementId } from "@/lib/project/characterAvatar";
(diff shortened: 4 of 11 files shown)


umairnadeem (commented):
lgtm, thanks!

PR #747, opened 2026-09-07
refactor: rethink openslop architecture 2026-09-07

umairnadeem:
## App understanding

OpenSlop is a script-to-video editor. A project is a Slate script whose elements resolve into a generation graph; the client queues stale nodes and posts each to an asset route, which records a `jobs` row and enqueues it on the Vercel Queue. A worker picks the handler for the job's connector type, uploads the result, and the client polls. Video redelivers itself to poll the vendor. LLM turns stream over SSE instead of queueing, and an animated image reaches the worker as a video job.

## From-scratch architecture

Everything that can be a queued job says so in one place, and the row schema, the queue message, the asset routes and the handler map all read that one list. A job row that reaches the worker can only name a type the worker serves, so there is no "no handler registered" branch to write or test.

## Implemented

- `JOB_CONNECTOR_TYPES` / `JobConnectorType` in `lib/api/jobs.ts`: the five types an asset route queues (image, music, sfx, tts, video). `JobRowSchema.connector_type`, `AssetQueueMessage`, `createJob` and `enqueueJob` are typed on it, so a row with any other type is rejected at the parse boundary.
- `lib/api/job-handlers.ts`: `HANDLERS` is a total `Record<JobConnectorType, JobHandler>` and `getJobHandler` always returns a handler.
- `lib/api/process-job.ts`: the runtime "No job handler registered" throw is gone, along with its test.
- `lib/api/asset-routes.ts`: `AssetRoute.connectorType` is a `JobConnectorType`, so a route cannot queue an LLM or animated-image job.

No behavior change for any valid job. An invalid `connector_type` on a row now fails at `JobRowSchema.parse` instead of one line later in the worker.

## Validation

- `npm run lint` passed
- `npm run format:check` passed
- `npm run typecheck` passed
- `npm run knip` passed
- `npm run test:run`: 178 files, 1611 tests passed
- `npm run build` passed

## Merge-conflict guard

```
PR #746: clean
PR #745: clean
```

## Skipped

- Folding the hosted and BYOK vendor tables into one (`lib/api/providers/{openslop,byok}.ts`): trades a compile-time class reference for a runtime lookup for about ten lines; left as is.
- `AssetQueueMessage.connectorType` is never read by the callback, which re-reads the row on purpose. Kept as queue-payload context rather than dropped.
- Everything merged since #732 through #744 was audited clean by the maintainability pass this morning and was not re-read.

### DIFF
--- a/lib/api/__tests__/process-job.test.ts
+++ b/lib/api/__tests__/process-job.test.ts
@@ -64,16 +64,6 @@ describe("processQueuedJob", () => {
 		expect(mockUpdateJob).not.toHaveBeenCalled();
 	});
 
-	it("throws when no handler is registered for the connector type", async () => {
-		mockLoadJobForProcessing.mockResolvedValue(pendingJob());
-		mockGetJobHandler.mockReturnValue(undefined);
-
-		await expect(processQueuedJob(JOB_ID)).rejects.toThrow(
-			"No job handler registered for image",
-		);
-		expect(mockUpdateJob).not.toHaveBeenCalled();
-	});
-
 	it("marks the job processing then completed with the handler result", async () => {
 		const job = pendingJob();
 		mockLoadJobForProcessing.mockResolvedValue(job);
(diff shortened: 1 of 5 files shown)


umairnadeem (commented):
lgtm, thanks!

PR #748, opened 2026-09-07
fix: open the Export popover from the export toast's View control

### LINKED ISSUE #735: "View" button in the export toast does not open the Export popover
## Problem

The "View" button in the export toast does nothing. Both the in-progress toast ("Exporting video…") and the finished toast ("Export ready", where the title itself is the View control) call `onView`, which is wired to `setOpen(true)` on the render context so the Export popover should open. In practice clicking it leaves the popover closed. The user has to find and click the Export button in the toolbar to see progress or the finished video. Download still works.

- "the "view" button in the export toast doesnt work"

## Relevant code

- `app/components/video/ExportToast.tsx:31-33` — progress toast View button (`onClick={onView}`).
- `app/components/video/ExportToast.tsx:52-58` — done toast, the "Export ready" title is the View control.
- `app/components/video/RenderProvider.tsx:29-66` — effect that renders the toasts and passes `onView={() => setOpen(true)}`; also dismisses the toast whenever `open` is true.
- `app/components/video/ExportButton.tsx:39` — the Radix `Popover` controlled by that same `open`/`setOpen`; content is docked to a fixed `PopoverAnchor` at bottom-right (`:53`).
- `app/components/AppToaster.tsx` — the sonner `Toaster`. Note the export toast asks for `position: "bottom-right"` while the toaster is mounted `bottom-center`.

## Things to check

Root cause is not confirmed. The toast lives in sonner's portal and the popover in Radix's portal, and the two interact around the same click. Likely candidates, in order:

1. **Popover dismissed by the same interaction that opened it.** Radix `DismissableLayer` closes on pointer/focus outside. If `open` flips to true while the View click is still completing (pointerup, click, focus leaving the unmounting toast), `onOpenChange(false)` may fire straight after `setOpen(true)`. Check by logging `onOpenChange` calls in `ExportButton`, or by adding `onInteractOutside`/`onFocusOutside` handlers to the `PopoverContent` and seeing whether they fire on a View click.
2. **Stale closure.** `toast.custom` is re-issued on every state change with the same id. Confirm the rendered toast actually has the latest `onView` and that the click handler runs at all.
3. **Toast stacking or pointer capture.** sonner attaches pointer handlers for swipe-to-dismiss on the toast element; confirm the inner button click is not swallowed.

## Suggested direction

Whatever the mechanism, the fix should keep one owner for "is the export popover open" (the render context) and make the toast a plain caller of it. If the cause is (1), the clean options are to defer the open until after the click settles, or to tell the popover to ignore interactions originating inside the toast (for example via `onInteractOutside` checking the event target against the toaster). Avoid a second popover state or a duplicate viewer.

## Acceptance criteria

- [ ] Clicking View on the in-progress export toast opens the Export popover showing render progress.
- [ ] Clicking "Export ready" on the done toast opens the Export popover with the finished video.
- [ ] The toast is dismissed when the popover opens and returns if the popover is closed while a render is still active (existing behaviour in `RenderProvider`).
- [ ] Download from the done toast keeps working.
- [ ] Lint, format, typecheck, and tests pass.

---

umairnadeem:
## Summary

Clicking **View** on the in-progress export toast, or **Export ready** on the finished toast, did nothing: the popover opened and closed in the same interaction.

Root cause, confirmed by reading the sonner and Radix sources in `node_modules`:

1. Clicking View focuses the button inside sonner's toast list. Sonner's list `onFocus` records the previously focused element (typically the canvas editor).
2. `setOpen(true)` mounts the Radix popover, whose `FocusScope` auto-focuses its first control.
3. Focus leaving the toast list fires sonner's `onBlur`, which restores focus to the recorded element.
4. Radix's `DismissableLayer` sees a document `focusin` outside the popover, the non-modal popover's `onInteractOutside` does not exempt it (it is neither the trigger nor preceded by a pointer-down outside), and `onDismiss` sets `open` back to false.

The flow works only when the previously focused element was `body` or the Export trigger, which is why it looked intermittent.

## Fix

`RenderProvider` owns the popover's `open` state, so the fix lives there: `setOpen(true)` drops focus from the active element before flipping the state. Sonner's restore then runs while no popover is listening, and the popover's own auto-focus lands as normal. The toasts stay plain callers of `onView`, and any future caller of `setOpen(true)` gets the same behavior.

Alternatives considered and not taken:
- `onFocusOutside={(e) => e.preventDefault()}` on the popover content: one line, but focus then ends up back in the editor after the popover opens (sonner's restore still fires), and Tab-out no longer closes the popover.
- Making the popover modal: blocks the canvas while an export runs.

## Selected issue

#735, `size: S`, `priority: medium`, `bug`. Highest-priority non-newcomer S issue open tonight. Single-file change in a component with one owner of the state.

## Validation

- `npm run lint` — pass, 0 warnings
- `npm run typecheck` — pass
- `npm run knip` — pass
- `npm run build` — pass
- `npm run test:coverage` — 178 files, 1612 tests pass, thresholds met
- `npm run test:e2e` — skipped (unchanged smoke surface, needs a browser install and Supabase env)

Not verified in a browser (repo rule: CI checks are the verification). There is no DOM test environment in this repo (no jsdom/happy-dom), so no focused test was added; the acceptance criteria depend on real browser focus semantics that jsdom would not model faithfully anyway.

## Merge-conflict guard

```
PR #747: clean
PR #746: clean
PR #745: clean
```

Closes #735

### DIFF
--- a/app/components/video/RenderProvider.tsx
+++ b/app/components/video/RenderProvider.tsx
@@ -1,6 +1,12 @@
 "use client";
 
-import { useEffect, useMemo, useState, type ReactNode } from "react";
+import {
+	useCallback,
+	useEffect,
+	useMemo,
+	useState,
+	type ReactNode,
+} from "react";
 import { toast } from "sonner";
 import { createRequiredContext } from "@/lib/components/createRequiredContext";
 import { ExportDoneToast, ExportProgressToast } from "./ExportToast";
@@ -24,7 +30,19 @@ const TOAST_OPTIONS = {
 
 export function RenderProvider({ children }: { children: ReactNode }) {
 	const { state, render, reset } = useRendering();
-	const [open, setOpen] = useState(false);
+	const [open, setOpenState] = useState(false);
+
+	// The toasts open the popover from inside sonner's list. Sonner returns focus
+	// to whatever had it before the toast as soon as focus leaves the list, and
+	// Radix reads that restored focus as a focus-outside and dismisses the popover
+	// it just opened. Dropping focus first lets the restore run while nothing is
+	// listening.
+	const setOpen = useCallback((next: boolean) => {
+		if (next && document.activeElement instanceof HTMLElement) {
+			document.activeElement.blur();
+		}
+		setOpenState(next);
+	}, []);
 
 	useEffect(() => {
 		if (open || state.status === "idle") {
@@ -63,7 +81,7 @@ export function RenderProvider({ children }: { children: ReactNode }) {
 			id: TOAST_ID,
 			position: "bottom-right",
 		});
-	}, [state, open]);
+	}, [state, open, setOpen]);
 
 	useEffect(() => {
 		return () => {
@@ -73,7 +91,7 @@ export function RenderProvider({ children }: { children: ReactNode }) {
 
 	const value = useMemo(
 		() => ({ state, render, reset, open, setOpen }),
-		[state, render, reset, open],
+		[state, render, reset, open, setOpen],
 	);
 	return <RenderContext value={value}>{children}</RenderContext>;
 }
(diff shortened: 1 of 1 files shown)


umairnadeem (commented):
lgtm, thanks!

PR #754, opened 2026-09-08
refactor: nightly maintainability 2026-09-08

umairnadeem:
## Summary
- `lib/api/error-envelope.ts` declares the `{ error }` body every route answers failures with; `lib/api/response.ts` builds through it and `lib/clients/http.ts` parses it with `safeParse` instead of hand-rolled typeof checks.
- `lib/connectors/plugins.ts`: the three identical hook folds (`runBeforeGenerate`, `runAfterGenerate`, `runTransformPrompt`) share one private `fold` with a single boundary cast. Exported names and call sites unchanged.
- `AssetEditProvider`: four dialog flags become one `AssetEdit | null` union, so two asset dialogs can never be open at once by construction. The `AssetEditors` interface is unchanged.
- Character attribute writes move out of `CharactersPicker.tsx` into `app/components/canvas/utils/characterOps.ts` (lodash `xor`/`without`), and the comma-join for the `characters` attribute lands in `lib/canvas/characterNames.ts` as `formatCharacterNames`, the inverse of the existing parser, with a round-trip test.

No behavior change.

## Skipped (for a human call)
- `lib/providers/video/runware.ts` casts the vendor status to `VideoJobStatus` (`queued|processing|completed|failed`) but the Runware SDK compares against `"success"`, `"ready"` and `"error"`. The `status === "failed"` branch in `lib/api/handlers/video.ts` may never fire. Making the parse fail loudly would break live generation until the real vocabulary is confirmed, so it was left alone.
- `VOLUME_OPTIONS` (0-10) and `LOOPS_OPTIONS` (1-8) in `lib/connectors/attributes/common.ts` vs `VOLUME_MAX`/`LOOPS_MAX` (10 / 1000) in `lib/video/elementAttributes.ts`: loops disagree, so unifying is a behavior decision.
- Voice trait field lists are declared in four places (`lib/project/types.ts`, `lib/agent/tools/inputs.ts`, `lib/connectors/tts/plugins/voice-search.ts`, `lib/providers/tts/voiceSimilarity.ts`); deriving them from `voiceTraitsSchema` changes `.catch` semantics on the agent tool input.
- `AttributeBadge` type-switch to registry: TypeScript already narrows the enum branch, so the "silent fallthrough" cost is not real; left as is.
- `runScript` + `clearEditor` ordering in `useAgentTools.ts`, and a `useVoiceSearch` hook extraction from `VoicePicker`: real seams, deferred to keep this PR coherent.

## Merge-conflict guard
```
```

## Test plan
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run knip`
- [x] `npm run build`
- [x] `npm run test:run` (178 files, 1613 tests)

### DIFF
--- a/app/components/canvas/elements/CharactersPicker.tsx
+++ b/app/components/canvas/elements/CharactersPicker.tsx
@@ -1,7 +1,6 @@
 "use client";
 
 import { UserPlus } from "@/components/ui/icon";
-import { Editor } from "slate";
 import { useSlateStatic } from "slate-react";
 import {
 	DropdownMenu,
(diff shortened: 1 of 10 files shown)


umairnadeem (commented):
lgtm

PR #757, opened 2026-09-08
refactor: rethink openslop architecture 2026-09-08

umairnadeem:
## App understanding

OpenSlop is a script-first AI video maker. A project is a Slate canvas of scenes and content elements (image, animated image, video, TTS, music, SFX). A generation graph turns elements into queued jobs that run through connectors and gateways to either the hosted provider or a user's own vendor key (BYOK), results land in Blob, and Remotion renders the cut. Sloppy is the in-app agent that edits the script. Supabase holds auth, projects, jobs and the key vault. Settings expose the provider keys and per-type default models.

## From-scratch architecture

The current layering (routes → `lib/api` handlers → connectors → gateways → providers) already matches what a rebuild would look like, and it has been audited clean over the last weeks. What a rebuild would still do differently is refuse to let one record type serve two kinds of row, and refuse to let a write take a patch that can be empty. Both cases were handled at runtime with sentinels and guards; the type system can carry them instead.

- **Provider key records are a `hosted | stored` union.** `HostedProviderKey` carries only `provider` and `status`; `StoredProviderKey` carries the key metadata. The server no longer builds the hosted row with `""` for `last4` and `createdAt`, and the shape that reaches the client is what it says it is.
- **Settings render from the records.** `ModelsTab` maps the key rows straight to cards, narrowing on `provider`: the hosted card takes its `status` as a prop instead of looking itself up and falling back to `"invalid"`, and `ProviderCard` takes its stored row instead of re-finding it in the store. A linked provider with no row yet is the one card built without a record. The "No providers connected yet" branch was unreachable, since the hosted row is always present, and is gone.
- **`useProviderKey` is gone; `useProviderKeyLookup` stays** as the one existence check used by the model menu and the add-providers list.
- **Job updates take a `JobTransition` union.** `processing`, `completed` with its result, `failed` with its error, or a metadata write. Each variant is non-empty by construction, so the lodash `omitBy` and the empty-patch early return are deleted, and the statuses are anchored to `JobStatus` so they cannot drift from the row schema.

Behavior is unchanged: the same rows, cards, order and writes.

## Validation

- `npm run test:run`: 178 files, 1611 tests pass
- `npm run build`: pass
- `/simplify` ran with four review agents; applied: stored row passed into the card (dropping the generic lookup), transition statuses tied to `JobStatus`, duplicate comments removed.

## Merge-conflict guard

```
PR #756: clean
PR #755: clean
PR #754: clean
```

## Skipped items

- **`PluginContext.gateway` (ledger 746j).** Removing the unread field leaves `PluginContext<TParams, TResult>` with phantom type parameters across 8 files including `lib/connectors/plugins.ts`, which PR #754 changes. It should go together with the de-genericization, once #754 merges.
- **Cartesia TTS contract (746d/e/f) and the video poll `failed` variant (746g/h)** are the next two batches of the conventions follow-up job and were left to it.
- **`params.model as "music_v1"` (746u).** Typing the vendor params on the catalog needs a mapped id type over both model tables for one cast; sub-nightly on its own.
- `JobTransition` deliberately has no variant that writes a terminal status and metadata in one call; no caller does today.

### DIFF
--- a/app/components/settings/ModelsTab.tsx
+++ b/app/components/settings/ModelsTab.tsx
@@ -11,7 +11,10 @@ import {
 	differsFromRecommended,
 	modelEntry,
 } from "@/lib/connectors/models";
-import { MANAGED_PROVIDER } from "@/lib/connectors/providerCatalog";
+import {
+	isByokProvider,
+	MANAGED_PROVIDER,
+} from "@/lib/connectors/providerCatalog";
 import type { Provider } from "@/lib/connectors/types";
 import { toastError } from "@/lib/toastError";
 import { useSettings } from "@/lib/settings/useSettings";
@@ -33,11 +36,15 @@ export function ModelsTab({
 	const providerKeys = useAccount((state) => state.providerKeys);
 	const settings = useSettings();
 
-	const stored = providerKeys.map((row) => row.provider);
 	// A link that named a provider leads here: it goes first, so what the link
-	// was about is the first thing read.
-	const shown =
-		selected && !stored.includes(selected) ? [selected, ...stored] : stored;
+	// was about is the first thing read, even before it has a key.
+	const pending =
+		selected &&
+		isByokProvider(selected) &&
+		!providerKeys.some((row) => row.provider === selected)
+			? selected
+			: null;
+	const dismiss = () => settings.open("models");
 
 	return (
 		<div className="flex flex-col gap-6">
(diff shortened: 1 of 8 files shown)


app/components/settings/ModelsTab.tsx
```diff
@@ -52,27 +59,29 @@ export function ModelsTab({
...
-									// A link pointing at a row that is gone has nothing to open.
-									onDismissed={() => settings.open("models")}
-								/>
-							),
-						)}
-					</div>
-				)}
+				<div className="flex flex-col gap-2">
+					{pending && (
+						<ProviderCard
```
umairnadeem, line 64: this card lives outside the keyed `providerKeys.map`, so when `saveKey` lands the row it unmounts and a fresh `ProviderCard` mounts from the list with `selected` still true (the url param never clears) → `useState(selected || !key)` reopens the form right after saving, and a rejected key's error message is gone with the old form instance. the old `shown.map` kept one instance per `key={provider}` across that hop, so imo fold the pending provider back into the same keyed list.

umairnadeem:
useless

PR #758, opened 2026-09-08
fix: stop a cancelled generation's poll loop

### LINKED ISSUE #694: Cancelling a canvas generation doesn't stop its poll loop

umairnadeem:
## Summary

Cancelling, discarding, or unmounting a canvas generation left its poll loop running until the job finished or hit the 15-minute timeout. The per-job `AbortController` was only read as an "ignore the result" flag; its signal never reached the connector.

This threads the existing signal through the layers that already had a slot for it:

- `queue.runJob` passes `controller.signal` to `generateForElement`, which puts it on the `GenerationContext`.
- `PluginContext` gains `signal`, so `BaseConnector.generate` hands it to `_generate` (mirroring `LLMConnector.stream(params, signal?)`).
- `BaseAssetConnector._generate` forwards it to every `gateway.poll(jobId, signal)`.
- `HttpAssetGateway.poll` and `ApiClient.get` pass it to `fetch`. `ApiClient.get` now takes `{ params, signal }` so the poll can name its signal without a positional placeholder.

An aborted fetch rejects with a `DOMException`, which `awaitCompletion` already rethrows (it only swallows `UnreachableError`), so the loop ends with no further requests for that job. The existing `signal.aborted` guards in the queue's terminal handlers stay in place and swallow the resulting `AbortError` for a job that completes in the same tick as the cancel. No new cancellation machinery.

## Selected issue

#694, size M, priority medium, labelled `bug`. It names the exact seam where the signal is lost and lists acceptance criteria, so the outcome was unambiguous. Every file it needed was free of open-PR overlap tonight.

## Validation

- `npm run build`: pass
- `npm run test:coverage`: 178 files, 1617 tests pass
- New seam tests: queue `cancel` / `cancelAll` abort the signal handed to the job, `generateForElement` forwards it in the context, the asset connector polls with it, `HttpAssetGateway.poll` and `ApiClient.get` hand it to `fetch`.
- Skipped: `npm run test:e2e`. The Playwright smoke suite covers only the public onboarding routes, which this change does not touch. CI runs it.

## Merge-conflict guard

```
PR #757: clean
PR #756: clean
PR #755: clean
PR #754: clean
```

Closes #694

### DIFF
--- a/lib/agent/client.ts
+++ b/lib/agent/client.ts
@@ -16,7 +16,7 @@ export async function loadAgentTranscript(
 ): Promise<SloppyMessage[]> {
 	const { messages } = await client.get<{ messages: SloppyMessage[] }>(
 		AGENT_PATH,
-		{ projectId },
+		{ params: { projectId } },
 	);
 	return messages;
 }
--- a/lib/clients/__tests__/apiClient.test.ts
+++ b/lib/clients/__tests__/apiClient.test.ts
@@ -39,7 +39,9 @@ describe("ApiClient", () => {
 				json: () => Promise.resolve({ voices: [] }),
 			});
 
-			await client.get("/api/v1/tts/voices", { gender: "feminine" });
+			await client.get("/api/v1/tts/voices", {
+				params: { gender: "feminine" },
+			});
 
 			expect(fetchMock).toHaveBeenCalledWith(
 				"https://api.test.com/api/v1/tts/voices?gender=feminine",
@@ -61,6 +63,21 @@ describe("ApiClient", () => {
 			);
 		});
 
+		it("hands the abort signal to fetch", async () => {
+			fetchMock.mockResolvedValue({
+				ok: true,
+				json: () => Promise.resolve({}),
+			});
+			const { signal } = new AbortController();
+
+			await client.get("/api/v1/video/job-1", { signal });
+
+			expect(fetchMock).toHaveBeenCalledWith(
+				"https://api.test.com/api/v1/video/job-1",
+				expect.objectContaining({ method: "GET", signal }),
+			);
+		});
+
 		it("filters out undefined param values", async () => {
 			fetchMock.mockResolvedValue({
 				ok: true,
@@ -68,8 +85,7 @@ describe("ApiClient", () => {
 			});
 
 			await client.get("/api/v1/tts/voices", {
-				gender: "masculine",
-				age: undefined as unknown as string,
+				params: { gender: "masculine", age: undefined as unknown as string },
 			});
 
 			const url = fetchMock.mock.calls[0][0] as string;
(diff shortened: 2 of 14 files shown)


umairnadeem (commented):
lgtm, signal threads cleanly through the slots that were already there and the aborted-fetch rejection lands in the existing `signal.aborted` guards. nice small fix.

PR #767, opened 2026-09-09
refactor: nightly maintainability 2026-09-09

umairnadeem:
Nightly maintainability pass over `lib/connectors`. No behavior change except the one noted under VoicePicker. Four of the changes are user-validated flags from the conventions follow-up ledger (746j, 746k, 746l, 746m), unblocked now that #754 has merged.

## Plugin context contract

- `PluginContext` loses its dead `gateway` field. `HttpLLMConnector` was the only writer and no plugin ever read it (`requireGateway` went away earlier). With the field gone the `<TParams, TResult>` generics carried nothing, so `PluginContext` is now plain; `ConnectorPlugin` keeps its generics, which are still load-bearing on `beforeGenerate`/`afterGenerate`. Eight files stop spelling phantom type arguments.
- `requireContext(ctx, key, plugin)` replaces `requireState`, `requireModel`, and the hand-rolled `searchVoices` guard in `voice-search.ts`. One keyed mechanism, no cast; the return type narrows from the field. Ten call sites.
- `BaseConnector.pluginContext()` is gone. It was a virtual hook returning `{}` with one override left (TTS), so `HttpTTSConnector` now overrides `contextFor` directly and adds `searchVoices` to the base context.

## Hosted model tables

Every `openslop/models.ts` table restated its vendor entry by hand (`id`, `cost`, `speed`, `resolutions`). They now alias the vendor row they serve: `"Slop Video v1": RUNWARE_VIDEO_MODELS["Seedance 2 Fast"]`, and likewise for image, TTS, LLM, SFX, and music. The ids cannot drift from the vendor tables any more, and `typeof OPENSLOP_*_MODELS` still keys `OPENSLOP_PROVIDERS` in `lib/api/providers/openslop.ts` unchanged.

## Voice search

- `useVoiceSearch(filters, model)` in `lib/connectors/tts/useVoiceSearch.ts` owns the TTS connector, the 300ms debounce, the cancellation flag, and the result limit. It returns a `loading | ready | failed` union in the shape `usePeaks` already uses. `VoicePicker` is now presentational: it calls the hook and renders the three states.

**One visible difference:** when a voice search fails, the list from the previous successful search no longer stays on screen under the error. The failed state renders the error alone.

## Not taken

- Dropping the `plugin` name argument from `requireContext` by attributing failures in the plugin runner: needs a try/rethrow around every hook call and changes every plugin error's text. Left for a human call.
[...]

PR #774, opened 2026-09-10
refactor: nightly maintainability 2026-09-10

umairnadeem:
## Summary

Nightly maintainability pass. No behavior change. Net **+51 / -241** across 8 files.

**Delete the `ApiClient` pass-through.** `lib/clients/apiClient.ts` was a three-method class whose only job was to prefix a `baseUrl` onto a path before calling `apiJson`/`apiFetch`. That prefix is always empty: `ConnectorConfig.baseUrl` is never set anywhere in the app. The HTTP gateways now call the transport helpers directly and own their full route string (the `baseUrl` constructor parameter stays, since the connectors still pass it and `ConnectorConfig` is reserved by an open PR). The agent transcript loader does the same. `apiClient.test.ts` (160 lines) re-tested the transport through the wrapper; the two cases it alone covered (abort signal forwarding, status-line fallback for non-envelope JSON bodies) moved into `http.test.ts` as a table.

**`sendMagicLink` throws like its siblings.** It was the only function in `lib/auth/session.ts` returning `{ error?: string }` while `signInWithGoogle` and `signOut` throw. `AuthForm` was handling both conventions for one call. It now handles one.

## Merge-conflict guard

```
PR #772: clean
PR #771: clean
PR #770: clean
PR #769: clean
PR #768: clean
PR #767: clean
PR #766: clean
PR #765: clean
PR #764: clean
PR #763: clean
PR #762: clean
PR #761: clean
PR #760: clean
PR #759: clean
```

## Checks

- `npm test`: 180 files, 1616 tests pass
- `npm run build`: pass

## Not taken

- Removing `ConnectorConfig.baseUrl` and the gateway parameter entirely: `lib/connectors/types.ts` and two connectors are reserved by #767/#769/#771. One-line follow-up once they land.
- `PlayerControlContext` exposing commands instead of the nullable `PlayerRef`: two consumers read the nullness deliberately to disable seeking, so the leak only half closes.
- `ConfirmDeleteDialog` owning its target and dropping the `latched` state: the fourth call site (`ProviderCard`) is reserved by #765; half a migration reads as churn.


umairnadeem (commented):
lgtm

PR #780, opened 2026-09-11
refactor: nightly maintainability 2026-09-11

umairnadeem:
## Summary

Two deletion-shaped simplifications the last two maintainability passes named as follow-ups and that were blocked on PRs that have since merged.

**`ConnectorConfig.baseUrl` is gone.** No caller ever set it: the registry configs carry only plugins, and `createConnector` spreads the config as-is. The three HTTP gateways now build their route from the model's prefix alone, the seven connector classes stop threading a positional arg, and the interface has one field left. The gateway test parses app-relative routes against a fixed origin instead of prepending one.

**`ConfirmDeleteDialog` owns its target.** It takes `target: T | undefined` and is open while there is one. `title` and `onConfirm` receive the target, so the four call sites (`ProjectsList`, `Storyboard`, `ComposerAssets`, `ProviderCard`) drop their `onOpenChange` plumbing and the `if (deleting)` guards that could never fail. The dialog latches one value through the Radix exit animation instead of comparing three strings. Confirming still closes via `AlertDialogAction`, so the explicit reset after confirm went too.

No behaviour change. +92 / -107 across 15 files.

## Checks

`npm run lint && npm run typecheck && npm run format:check && npm run test:run && npm run build` all pass (185 test files, 1635 tests).

## Skipped

- `ElementHistory.record` toasting from `lib/generation/history.ts` (a lib module reaching for UI). Deliberate UX choice from an earlier PR; a product call, not a refactor.
- The rest of the surface merged since #758 (error boundaries, provider search, account store, element history storage, video poll contract) read clean.

## Merge-conflict guard

```
PR #779: clean
PR #778: clean
```

### DIFF
--- a/app/components/copilot/ComposerAssets.tsx
+++ b/app/components/copilot/ComposerAssets.tsx
@@ -23,11 +23,6 @@ export function ComposerAssets({ uploadingCount }: { uploadingCount: number }) {
 
 	const hasNarration = Object.keys(narration).length > 0;
 
-	const confirmDelete = () => {
-		if (deletingName) deleteCharacter(store, queue, deletingName);
-		setDeletingName(undefined);
-	};
-
 	return (
 		<div className="flex flex-wrap gap-2 pb-2">
 			{hasArtStyle && <ArtStyleAssetTile />}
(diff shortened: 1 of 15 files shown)


umairnadeem (commented):
lgtm, the latched target through the exit animation is a nice cleanup

PR #778, opened 2026-09-11
refactor: nightly readability 2026-09-11

umairnadeem:
## Summary

First run of the nightly readability job. One area: the video pipeline, `lib/video` and `app/components/video`. Names say what things are, flow is straightened, and no behavior changes. Every touched file ends the same size or smaller.

**Lines: +58 / -69 across 14 files.**

### Renames

| File | Old | New |
| --- | --- | --- |
| `lib/video/motionEffects.ts` | `ramp(frame, d, a, b)` | `ramp(frame, duration, from, to)` |
| `lib/video/motionEffects.ts` | `d`, `ar`, `c` in `motionTransform` | `duration`, `aspect`, destructured `{ tx, ty, rotation, extraScale }` |
| `lib/video/fadeRamp.ts` | `f` | `edge` |
| `lib/video/timestamps.ts` | `m`, `s` | `minutes`, inlined `total % 60` |
| `lib/video/sceneSegments.ts` | `fg` | `foreground` |
| `lib/video/resolve.ts` | `el` | `element` |
| `lib/video/layoutKey.ts` | `el` | `element` |
| `app/components/video/ScrubBar.tsx` | `p` | `hover` |
| `app/components/video/usePlayerState.ts`, `PlayerControls.tsx` | `p` | `player` |
| `app/components/video/useAssetPrefetch.ts` | `m`, `h`, `err` | `remotion`, `handle`, `error` |
| `app/components/video/useRendering.ts` | `err` | `error` |
| `app/components/video/VideoPreview.tsx` | `f` | `prev` |
| `app/components/video/BottomTransportBar.tsx` | `dir` | `direction` |

### Straightened flow

- `buildSequenceIndex`: `new Map(series.map(...))` instead of a mutable loop.
- `findSceneSequence`: one guarded expression instead of an early-return pair.
- `findSegmentIndexAtFrame`: `findIndex` instead of an indexed loop; the explicit empty-list `-1` guard stays.
- `trimSequencesAt` (scene-builder): a pop-while plus one clamp instead of a while/continue/return loop.

### Skipped

- `lib/components/Waveform.tsx` has the densest short-name cluster (`a` for the audio element, six times) but is outside the video folder pair, so it is a candidate for the next pass.
- `useActiveSegmentIndex.ts` keeps its `(p)` callback: renaming it wraps the line and grows the file.
- The `SPECS` lambdas in `motionEffects.ts` keep `(f, d)`; renaming all thirteen would grow the file, so the type signature keeps `d` too for consistency.
[...]

PR #768, opened 2026-09-09
refactor: nightly conventions follow-up 2026-09-09, TTS fails loudly

umairnadeem:
Nightly conventions follow-up. Works the design-level flags the sweep could not apply. Theme: TTS provider fails loudly.

## Behavior changes
- `lib/providers/tts/cartesia.ts`: `CartesiaTTS.search` logged a voice-ranking failure and returned the unranked list → the failure propagates to the caller (the voices route returns its error envelope, the `VoicePicker` shows the message, the voice-search plugin fails the generation). Why: CONVENTIONS.md "Fail loudly"; the only swallow-style `.catch` left in `lib/`; flag 746d from #746, user-validated. PR #108 added the fallback so a missing `OPENAI_API_KEY` still returned voices; that key is server-side on both the hosted and BYOK voice routes, so the degraded result only ever hid a misconfiguration. Tests: `cartesia-tts.test.ts` "propagates an embedding failure instead of returning unranked voices".

## Contract changes
- `lib/api/logger` import dropped from `cartesia.ts`; no other reference.

## Flags resolved
- 746d from #746: ranking failure propagates.

## Flags rejected
- 746e from #746 (move the one-second pause out of the provider into the layout): the pause is a Cartesia implementation detail and stays in the provider. An earlier revision of this PR moved it to a `DIALOGUE_GAP_SEC` in the scene builder; that leaked into the layout, the length estimate, and every layout test, and was reverted.

## Skipped tonight
- 746f (`voiceId` required on `TTSGenerateParams`): blocked by PR #767 (`lib/connectors/types.ts`).
- 746g/746h (video poll `failed` variant, required `metadata`): open; a union-only rewrite is the shape #757 was closed for, so they wait to ride with 756a's boundary change as one "video request and poll contract" batch.
- 746n, 746o: blocked by PR #767 (`still-frame.ts`, `lib/connectors/types.ts`).
- 746u (`params.model as "music_v1"`): rejected, type-only cast rewrite of the class closed in #756/#757.
- 746p, 746q, 746r, 746s, 746v, 756a, 756b: off theme tonight.

## Checks
lint, typecheck, test (cartesia, lib/video, timeline suites): all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BrKVnGmie2TV5W4Ftr2xLt

PR #781, opened 2026-09-11
refactor: nightly conventions follow-up 2026-09-11, parse at the boundary

umairnadeem:
Nightly conventions follow-up. Theme: parse at the boundary, delete the guard the type made unnecessary.

## Behavior changes
- `lib/project/serialize.ts`: `deserializeWithScenes` cast every parsed OSML node to a canvas element; now it keeps only nodes that pass `isParsedContentElement`, the same split `withOSMLClipboard` and `useScriptSync` already make. A stored script carrying a `metadata_*` tag no longer lands in Slate as a scene child no renderer handles. Why: CONVENTIONS.md "Parse, don't validate" and "Define errors out of existence"; flag 910i. Tests: `serialize.test.ts` "keeps metadata tags out of the scene".
- `lib/providers/cache.ts`: `audioBundleCache.fromMetadata` rehydrated a Pinecone row through hand-rolled `typeof`/`Number.isFinite` checks and `String(m.description)` (a missing description became the string `"undefined"`); now a zod object parses the row and any shape we did not write is a cache miss, as before. Why: "Parse, don't validate" at an external boundary; flag 910k. Tests: `cache.test.ts` legacy `audioUrl`, no-URL and no-duration misses kept, "forces a miss when the row has no description" added.

## Contract changes
- `lib/providers/cache.ts`: the local `Metadata` alias is gone; `CacheMatch`, `toMetadata` and `fromMetadata` use Pinecone's own `RecordMetadata`, which deletes the `m.metadata as Metadata` cast (1 reference).
- `components/ui/color-field.tsx`: `onChange` is `(value: string) => void`; a layer that can be switched off passes `empty: { label, onSelect }` instead of `emptyLabel`. The picker can no longer hand a caller `null` it did not ask for, so `CaptionTextStyleFields` drops the `fill ?? value.fill` guard and each cleared layer names its own null (2 call sites). Flag 910g.
- `app/components/canvas/elements/AssetTile.tsx`: `name` is required (all four `AssetTiles` callers pass one); the five `name ?? ...` / `name &&` fallbacks and the `!initial` icon fallback are deleted. Flag 910f.

## Flags resolved
- 910i: serialize cast replaced by the shared narrow, test added.
- 910k: cache row parsed with zod, hand guards deleted.
- 910g: ColorField cannot emit null unless asked; consumer guard deleted.
- 910f: AssetTile.name required; dead fallbacks deleted.

## Skipped tonight
- 910j (`emotion` parsed as `z.enum(TTSEmotion)` at the TTS route): rejected, needs product call. The hosted LLM falls back to `MockLLM` without an Anthropic key, and that mock script writes `cheerful`, `warm`, `gentle`, `delighted`, `kind`, `hearty`, `wonder`, none of which the catalog names; enforcing the catalog at the route would 400 those lines, and Cartesia currently takes the string as-is. Question for the owner: is the emotion catalog a hard contract (then the mock script and the OSML parse boundary should enforce it), or a prompt hint (then the route stays a string)?
- 746f (`voiceId` required on `TTSGenerateParams`): blocked by PR #780 (`lib/connectors/types.ts`).
- 746o: blocked by PR #780 (`lib/connectors/types.ts`).
- 746d, 746e: rejected by the owner on #768 ("this is actually desired, since ranking is best-effort only"); ledger updated.
- 746m: already shipped by #767 (the hosted video and image tables alias the Runware rows).
- 746l, 746n: each too small to carry a night on its own; wait for 746o to free so the connector group ships together.
- 746q: the "what scale 1 means for portrait" half lives in `useRendering.ts`, reserved by PR #778; tying three literals alone is trivial.
- 746r, 910b: off theme; batched next as "element attribute bounds". Note for that batch: the LLM prompt tells the writer `loops` is any integer while the picker offers 1..8, so clamping the render to the picker is a product call.
- 910a (`readSSE` swallow): off theme, and the owner's #768 comment makes a swallow removal a question to ask first.
- 910d, 910e: `confirm-delete-dialog.tsx` is reserved by PR #780.
- 910c: visual change, waits for its own batch with DESIGN.md read.

## Checks
lint, typecheck, format:check, test:run (185 files, 1637 tests), build, knip: all pass. `/simplify`, composition, web-design and react-best-practices reviews applied (flat cache schema, dead `!initial` branch, `empty` capability instead of a props union).

## Merge-conflict guard
```
PR #780: clean
PR #779: clean
PR #778: clean
```

### DIFF
--- a/app/components/canvas/elements/AssetTile.tsx
+++ b/app/components/canvas/elements/AssetTile.tsx
@@ -38,7 +38,7 @@ export function AssetTile({
 	removeAffordance = "overlay",
 	fallback = "initial",
 }: {
-	name?: string;
+	name: string;
 	previewUrl?: string;
 	Icon: IconComponent;
 	elementId?: string;
@@ -51,9 +51,12 @@ export function AssetTile({
 	const status = useQueueSelector(
 		(q) => q.getElementSnapshot(elementId).status,
 	);
-	const initial = name?.trim().charAt(0).toUpperCase();
 	const fallbackContent =
-		fallback === "icon" || !initial ? <Icon className="h-5 w-5" /> : initial;
+		fallback === "icon" ? (
+			<Icon className="h-5 w-5" />
+		) : (
+			name.trim().charAt(0).toUpperCase()
+		);
 	return (
 		<div className="group/tile relative flex w-16 flex-col gap-1 sm:w-20">
 			<div className="relative aspect-square overflow-hidden rounded-md border border-border bg-card">
@@ -61,7 +64,7 @@ export function AssetTile({
 					<ImageWithShimmer
 						key={previewUrl}
 						src={previewUrl}
-						alt={name ?? ""}
+						alt={name}
 						fill
 						unoptimized
 						className="object-cover"
@@ -92,7 +95,7 @@ export function AssetTile({
 				{onEdit && (
 					<OverlayButton
 						icon={Pencil}
-						label={`Edit ${name ?? "asset"}`}
+						label={`Edit ${name}`}
 						onClick={onEdit}
 					/>
 				)}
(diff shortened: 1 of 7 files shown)


umairnadeem (commented):
lgtm, verified all four AssetTile callers pass a name and ColorField has the one consumer

PR #782, opened 2026-09-11
refactor: rethink architecture 2026-09-11

umairnadeem:
## The app

OpenSlop is a script-to-video editor. A Slate canvas holds the script, a Zustand store holds project metadata, and a generation queue holds each element's snapshot (result, inputs, `pinned` for a result the user supplied). The three persist together as one `ProjectContent` and restore together through `ProjectDocument.write`.

## From scratch

Each fact has one home. The queue's snapshot already says whether an avatar was supplied (`pinned`), so the store never records it again. A store is built from the row it loads, so nothing has to ask whether it has been loaded yet.

## What changed

Two booleans mirrored facts that already lived elsewhere. Both are gone.

**`avatarUploaded` on a character duplicated the avatar snapshot's `pinned`.** Both were written at the same call sites (the upload button, the regenerate button, a version restore), and they had drifted: `applyTemplate` seeds avatars with `pinned: true` but never set `avatarUploaded`, so template avatars were left out of art-style references and reported to Sloppy as generated. Now `characterAvatarState(results, name)` reads the snapshot, `uploadedAvatarUrls` and the agent context derive from it, `characterFromAvatarInputs` restores only appearance and model, and the modal's `regenerateAvatar` wrapper is gone (a regenerated result commits unpinned already).

One visible effect, by design: avatars a template supplied now count as supplied. They join the art-style references and Sloppy hears "avatar uploaded by the user" for them.

**`hydrated` on the project store was true by construction.** `createProjectStore` had one caller, which applied the parsed snapshot in the same `useState` initializer, so the flag could never be observed false. The store now takes its initial data: `createProjectStore(parseStoreSnapshot(initialStore))`. Deleted with it: `applyStoreSnapshot`, the `ProjectPersisted` twin of `ProjectData`, `AutosaverOptions.store` (its only use was the hydrated check), the autosave abort branch and its test, and the `if (!hydrated) return null` gate in `AssetsSection`. Ten test fixtures lost a `hydrated: true` line.

28 files, +96 / -203.

## What I ran

`npm run lint && npm run typecheck && npm run format:check && npm run test:run && npm run build`, all green (185 files, 1633 tests). `/simplify` over the diff; the repo has no `simplify` skill of its own, so the built-in one ran. The three Vercel skills in `.claude/skills` had nothing to apply: every TSX change here is a deletion. `knip` and a clone scan are clean.

## Skipped

- Folding `useProjectRehydrate` (the script applied in an effect) into `ProjectDocument.write` so all three columns load through one path. It changes when `markSaved` sees the loaded document; separate change.
- Parsing the `generation` column on `app/projects/[id]/page.tsx` (ledger flag 769a). Rows saved before `pinned`/`connectorType` existed would throw without defaults; human call.
- The route files restating one table (`connectorType`, `fields`, `label`, `models`) twice per asset type across `app/api/v1` and `app/api/third-party`, and `AttributeSchema`'s `hideModel` option threaded through for one TTS schema. Both are next-night candidates.
- Relocating vendor-response interpretation out of the four bespoke routes (render progress, validate-code, upload). Relocation, not simplification.
- Renaming `storeSnapshot.ts` and its helpers now that the alias is gone. The `store` column is still a snapshot of the store, and the rename touches five importers for no structural gain.
- A shared `resultSource(snapshot)` for the "has a result, pinned decides" predicate that `characterAvatarState`, `UploadedBadge` and `elementState` each spell out. Real, third reader now, follow-up.

## Merge-conflict guard

```
PR #781: clean
PR #780: clean
PR #779: clean
PR #778: clean
```

### DIFF
--- a/app/components/canvas/elements/AssetsSection.tsx
+++ b/app/components/canvas/elements/AssetsSection.tsx
@@ -1,7 +1,6 @@
 "use client";
 
 import { memo, useState } from "react";
-import { useProject } from "@/lib/project/useProject";
 import {
 	AddCharacterTile,
 	ArtStyleAssetTile,
@@ -12,10 +11,7 @@ import { CollapsibleHeader } from "./CollapsibleHeader";
 import { ReferenceImages } from "./ReferenceImages";
 
 function AssetsSectionComponent() {
-	const hydrated = useProject((s) => s.hydrated);
 	const [collapsed, setCollapsed] = useState(false);
-
-	if (!hydrated) return null;
 
 	return (
 		<section className="group/collapsible mb-4 select-none" aria-label="Assets">
--- a/app/components/canvas/elements/character/CharacterEditModal.tsx
+++ b/app/components/canvas/elements/character/CharacterEditModal.tsx
@@ -89,11 +89,6 @@ function CharacterEditDialogBody({
 	const update = (partial: Partial<MetadataCharacter>) =>
 		updateCharacter(name, partial);
 
-	const regenerateAvatar = () => {
-		if (character.avatarUploaded) update({ avatarUploaded: false });
-		avatar.generate();
-	};
-
 	const avatarModel = resolveModel("image", character.avatarModel);
 	const isStale = avatar.staleReason !== null;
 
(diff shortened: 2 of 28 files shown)


umairnadeem (commented):
lgtm, traced pinned through commitResult, job completion and restoreVersion and the snapshot really is the single source now

PR #785, opened 2026-09-12
feat: clips open on a start frame; animated image is gone

umairnadeem:
## Summary
- Deletes the `animated_image` element, connector, derived still node and `deps` seeding; only `image` and `clip` visuals remain, with `<clip>` taught to Sloppy in their place.
- Clips gain `startFrame`, a three-tile picker that previews each choice: **none**, **previous scene** (default, resolved from document order at build time, so reordering stales the clip; a clip source has its last frame captured in the browser) or **uploaded picture** (remembered across switches). Pointing at an arbitrary visual by id is plumbed but not offered. Node building reads the canvas through a `BuildContext`.
- One shared text teaches write_script, adapt_script and Sloppy what `startFrame`, `trimToDialogue` and `loop` do and the three shapes they fall into: Slideshow, Film, Motion explainer.
- Per-clip `trimToDialogue` toggle on the card and timeline replaces the hardcoded project constant. The animate button retypes an image into a clip opening on its picture and asks Sloppy for the motion.

## Test plan
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck`, `npx vitest run` (184 files)
- [x] Browser, mock providers, visible Chrome: settings popover opens a three-tile start-frame picker showing the previous visual's picture and the uploaded picture; first clip generates with no frame; a later clip sends the previous visual's picture in `frameImages`; removing the visual before a clip marks it Stale; upload sets the frame and is sent; animate retypes to a clip on its picture and Sloppy writes the motion; timeline badge present
- [ ] Reviewer: open an existing project that used `<animated_image>` and confirm the expected non-migration is acceptable

https://claude.ai/code/session_01Qm4xBHRxbqu5zoN3Xynhrh


lib/agent/prompt.ts
```diff
@@ -28,11 +29,12 @@ const ROLE = dedent`
...
   # How long a visual is on screen
-  - A visual (image, animated_image, clip) is on screen for the dialogue after it, up to the
-    next visual. \`duration\` sets the generated video's length, not its time on screen.
+  - A visual (image, clip) trimmed to dialogue is on screen for the dialogue after it, up to
+    the next visual; an untrimmed clip plays its full length. \`duration\` sets the generated
+    video's length, not its time on screen.
   - Never guess a length. measure_element_lengths reads them off the canvas and says how to
     change one.
-  - fit_durations sets every animated_image and clip to a \`duration\` that covers the dialogue
+  - fit_durations sets every clip to a \`duration\` that covers the dialogue
```
umairnadeem: this flattens a Film: `durationFits` still fits every clip to `words + leeway`, so an untrimmed clip with no dialogue under it snaps down to 4s, and Sloppy is told to call fit_durations as the last call of every turn. imo skip `trimToDialogue: false` clips in the fit, their `duration` is the on-screen length so there's nothing to fit to.

app/components/canvas/utils/nodeOps.ts
```diff
@@ -17,6 +21,21 @@ export function updateElementAttrs(
...
+/** Turns an image into a video opening on its picture, by URL since the image is gone. Returns its scene. */
+export function animateElement(
+	editor: Editor,
+	element: CanvasContentElement,
+	picture: string,
+	defaultModels?: ConnectorModels,
+): number {
+	const path = ReactEditor.findPath(editor, element);
+	retypeNode(editor, path, element, "video", {
+		attrs: { [START_FRAME_ATTR]: picture },
```
umairnadeem: animate sets `startFrame` to the url but not `uploadedFrame`, so the picker says "Uploaded picture" with no tile for it, and picking None writes `uploadedFrame: null` and the image is gone for good (it was retyped away). set `[UPLOADED_FRAME_ATTR]: picture` here too so it stays selectable.

lib/canvas/types.ts
```diff
@@ -37,13 +37,7 @@ export const ELEMENT_TYPES = {
...
 		layer: "visual",
 	},
-	animated_image: {
-		connector: "animated_image",
-		outputKind: "video",
-		role: "foreground",
-		layer: "visual",
-	},
-	clip: {
+	video: {
```
umairnadeem, line 40: renaming `clip` to `video` breaks old projects too, not just animated_image. `deserializeWithScenes` filters out tags it doesn't know, so every saved `<clip>` gets dropped on load along with its prompt, and the next save writes the script without it. is that intended? if not, maybe parse `clip` as `video` on load (and at least call it out next to the animated_image checkbox).

umairnadeem: backwards compatibility is optional

umairnadeem (commented):

lib/agent/tools/measureTotalLength.ts
```diff
@@ -1,45 +1,45 @@
...
 	input: z.object({}),
 	output: z.string(),
 	icon: Hourglass,
-	label: "Measuring the video's length",
+	label: "Measuring the total length",
 	execute: async (_input, ctx) => {
 		const words = ctx.countSpokenWords();
+		const runtime = ctx
+			.measureElementLengths()
+			.reduce((total, length) => total + length.seconds, 0);
```
umairnadeem, line 28: this undercounts: `measureElementLengths` drops dialogue before the first visual, but the renderer plays it over a blank scene, so it's real runtime. a script that opens on narration (which the spec now allows, "Start with a visual" is gone) reads short, and a narration-only script measures 0s and gets told "under by 180s" with 700 words on it. maybe add `secondsForWords` of the leading speech to the total, or count it as a span in `measureElementLengths`.

app/components/canvas/elements/attributes/ContinuitySection.tsx
```diff
@@ -0,0 +1,105 @@
...
+	type ReferenceImagesPopoverProps,
+} from "./ReferenceImagesPopover";
+import {
+	usePreviousPictureNames,
+	usePreviousVisualPictures,
+	type PreviousPictures,
+} from "./usePreviousVisualPictures";
+
+/** Whether a video references the visual before it, and the pictures it would take, known before anything decodes. */
+function useContinuity(element: CanvasContentElement) {
```
umairnadeem: the implementation of useContinuity is poorly done. usePreviousVisualPictures and           
  usePreviousPictureNames separately reference HANDED_ON_FRAMES, creating an implicit tight   
  coupling that is redundant. instead, we should simply have a single module (function or     
  hook) that returns a list of screenshots of the previous element (if one exists) that are   
  labeled. also we ideally shouldn't need parseStartFrame at all - startFrame can either be a 
  special value like "previous" or a url. lets simplify and remove as much code and           
  complexity as possble here

lib/script/prompt/videoPrompt.ts
```diff
@@ -0,0 +1,30 @@
...
+  - Objects: important props and set pieces, where they sit and what state they are in.
+  - Depth: what fills the foreground and what stands in the background.
+  - Then any Speech, then Sound.
+
+  - Keep the geography steady from shot to shot: a character on the left of frame stays on the left unless the shot shows them moving, and eyelines match (when Kai looks up at Mia, Mia looks down at Kai).
+  - Sound: name what makes the noise and how near it is ("rain drumming on a tin roof, a car passing far off"), never a mood. NEVER include music or narration in the video prompt.
+  - Speech: Write it as Name (tone): "line", one short sentence per shot in simple everyday words. Ensure the speech is perfectly idiomatic, appropriate, and realistic for the shot.
+  - A video that continues from the one before (startFrame="previous") starts where that video's last shot ends: Shot 1 keeps its place, framing, characters and action, then moves on from there.
+  - Refer to characters by name in the video prompt, do not describe their appearance since this is already in the reference images passed to the model.
+	- If there's only one shot, do not pass in a label such as Shot 1
```
umairnadeem, line 26: these new lines are tab-indented while the rest of the block uses spaces, so `dedent` takes the tab as the common indent and only strips one char: every other line in the block now keeps a stray leading space, the heading doesn't, and the slideshow example keeps a literal tab (ran it, the output is lopsided). indent with two spaces like the lines above.

lib/agent/tools/editScript.ts
```diff
@@ -28,7 +29,10 @@ const PICTURE_ATTRIBUTES = [
 const SCRIPT_ATTRIBUTES: Partial<Record<CanvasElementType, string[]>> = {
 	character: ["name"],
 	image: PICTURE_ATTRIBUTES,
-	animated_image: PICTURE_ATTRIBUTES,
+	video: [
+		...PICTURE_ATTRIBUTES,
+		'startFrame (none | previous, or a picture URL: an image\'s URL from view_image, with continuity="false" so the look before it does not fight that picture; leave a URL already set alone)',
```
umairnadeem, line 34: same footgun as the old animateElement, just via sloppy now: a `set` that writes `startFrame` to a URL never writes `uploadedFrame`, so the picker reads "Uploaded picture" with no tile selected, and picking None drops the url for good (the image was retyped away). imo have applySet write `openingOn(url)` when startFrame is a URL, it's otherwise dead code now.

umairnadeem (commented):

umairnadeem (commented):

umairnadeem (commented):

lib/generation/resolveGraph.ts
```diff
@@ -40,47 +37,53 @@ const toNode = (
...
+	// edit, which is a new canvas, is the only thing that builds an element again.
+	const revisions = new WeakMap<CanvasContentElement[], Map<string, JobNode>>();
 	return (spec) => {
-		// Scoped to one call: it dedupes nodes shared within a single graph and
-		// detects cycles. Held across calls it would serve a stale node back once
-		// its element changed, since element content is not part of `state`.
-		const resolved = new Map<string, JobNode>();
+		// Read once per build, so every node in the graph sees the same canvas.
+		const ctx = context();
+		const resolved = revisions.get(ctx.canvas) ?? new Map<string, JobNode>();
```
umairnadeem, line 61: this keys on id, but slate-react fires `useLiveNode`'s listener with last render's `build` (old element object) before react re-renders with the new one, so the stale node lands in the new canvas's cache and the re-render's build hits it. net effect every card lags one edit behind: type into an empty prompt and `hasPrompt` stays false, hit generate and it runs the text minus your last keystroke since `build()` at the click hits the same cache (repro'd with a real `<Slate>`, and each `useNodeBuilder` has its own `revisions` so the toolbar building fresh doesn't cover for it). keying `resolved` on the element object instead of its id fixes it, slate hands back a new object per revision anyway.

lib/script/prompt/review.ts
```diff
@@ -0,0 +1,47 @@
...
+	  - Dialogue: everyday words, idiomatic to the format, true to who says it and to what just
+	    happened. It should be appropriate, simple, and make perfect sense in its context.
+	  - Pacing and beats: the video progresses at a natural speed, with no sudden awkward cuts or overly dragged out scenes.
+		- Prompts: all prompts are clear, specific, and give the model enough information to generate the intended visuals and audio.
+
+	  Work through every scene and give each element a single finding.
+
+	  Write each finding as one bullet, three fields in this order:
+
+	  - <element id> | <the rule it breaks> | <the smallest change that fixes it>
```
umairnadeem, line 33: the prompt tuning changed strings the tests pin: 4 in review.test.ts (the em-dash finding shape, "never report a metadata tag", "every scene to the last one") and 3 in osml.test.ts (the dialogue line, "starts where that video's last shot", "Film: the picture tells the story"). ci is red on the head sha, so the assertions need to move with the text.

lib/script/prompt/review.ts
```diff
@@ -0,0 +1,47 @@
...
+	  - Video prompts: every requirement of that section, in every video prompt.
+	  - Language: each part of the script in the language that section assigns it.
+
+	  Then judge what those rules cannot state, reading the script whole:
+	  - Continuity: characters, place, light and props hold from one visual to the next, and
+	    each video's startFrame and continuity match what its pictures actually do.
+	  - Dialogue: everyday words, idiomatic to the format, true to who says it and to what just
+	    happened. It should be appropriate, simple, and make perfect sense in its context.
+	  - Pacing and beats: the video progresses at a natural speed, with no sudden awkward cuts or overly dragged out scenes.
+		- Prompts: all prompts are clear, specific, and give the model enough information to generate the intended visuals and audio.
```
umairnadeem, line 27: same dedent footgun as videoPrompt: this line and the "Say what to change" one below are tab-indented while the rest is tab+2 spaces, so dedent strips 2 chars and every other line keeps a stray leading space while these two don't (ran it, output is lopsided). indent them like their neighbours.

PR #787, opened 2026-09-12
refactor: nightly maintainability 2026-09-12

umairnadeem:
## What changed

A route file now picks a route family and a connector type. The family reads that type's models from the catalog (`listModels` / `hasModel` in `lib/connectors/models.ts`), so nothing else about a route varies.

- `lib/api/route-families.ts`: `RouteFamily<TPicked>` drops the `TModels` parameter. `model(type)` replaces `model(models)`.
- `lib/api/generation-schema.ts`: `hostedModel(type)` and `byokModel(type)` take a connector type. BYOK membership is `hasModel`, the one catalog reader, instead of a hand-rolled `model in table` check.
- `lib/api/asset-routes.ts`: an `ASSET_ROUTES` table holds each job type's request fields and log label. `createAssetRouteHandler(family, type)` reads it.
- `lib/api/llm-routes.ts`, `lib/api/voice-routes.ts`: the four factories lose their `models` argument.
- 18 route files under `app/api/v1` and `app/api/third-party` each shrink to one call and lose a models import. The `v1` and `third-party` twins can no longer drift.
- `lib/connectors/<type>/models.ts` (6 files): the `BYOK_*_MODELS` projections are gone. Each full table lists its providers directly.
- Two tests build schemas from a type instead of a table.
- `ARCHITECTURE.md`: two owner-authored sentences reworded to match ("A route file picks a family and a connector type"; a new asset type also adds a row to `ASSET_ROUTES`).

No behaviour change. Same schemas, same error messages, same labels.

+99 / -201 over 32 files.

## Not taken

- Deriving the log label from the type. Four rows are `<Type> generation` but video is `Video submission`, so deriving would rename a log line.
- Inlining the single-use `*_FIELDS` exports into the table. Pure move churn; the shapes stay in the schema module.
- One registry per job type carrying fields, label and handler, so `JOB_CONNECTOR_TYPES`, `ASSET_ROUTES` and `HANDLERS` stop being three hand-kept lists. Needs `lib/api/job-handlers.ts`; a separate night.
- Dropping the `as z.ZodType<AssetBody>` cast (ledger 756b). `projectId` still comes through the raw field shape as `unknown`; needs the generic handler contract change.

## Checks

`npm run lint && npm run typecheck && npm run format:check && npm run test:run && npm run build` all pass (185 test files, 1635 tests).

## Merge-conflict guard

```
PR #786: clean
PR #785: clean
```

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -36,7 +36,7 @@ Users see models and providers; developers also see connectors. Each word means 
 
 A model is a `{ provider, model }` pair, never a bare name. `MODELS[type][provider][name]` in `lib/connectors/models.ts` is the whole table, and nothing anywhere derives a provider from a name. Every element stores its own pair as attributes, speech included. A voice in project metadata stores the pair it was picked on, and its narration and character elements speak with that pair, falling back to their own until the voice picks one. A plugin declares where such an inherited model comes from, the same way it declares what it reads. Defaults resolve element, then project, then account, then the recommendation.
 
-Everything that differs between the two route families is one object each in `lib/api/route-families.ts`. `HOSTED` is API-access gated, takes only a model name, and runs on our keys. `BYOK` is session gated, takes the pair, and runs on the account's key. A route file picks a family and the models it serves. A job stores the pair, and the worker builds the provider from it at the last moment, which is the one place the server branches on the family.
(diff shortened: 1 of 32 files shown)


umairnadeem (commented):
lgtm

PR #789, opened 2026-09-12
refactor: rethink architecture 2026-09-12

umairnadeem:
## The app

OpenSlop is a script-to-video editor. A project row holds a script, a metadata store and generation snapshots. The client builds a store from each column when the page opens, the Slate canvas edits the script, an autosaver debounces every change back to the row, and version history checkpoints those saves.

## The from-scratch design

Every client store is constructed from its column, once, and nothing has to run "after hydration". Before this PR the editor was the one exception: it mounted on an empty document, a ref-guarded effect loaded the saved script, and autosave had to take its baseline in a later effect, with a comment explaining the ordering. The previous rethink (#782) removed the `hydrated` flag from the project store. This is the editor-side twin.

## What changed

- `useEditorSetup(script)` builds the editor and loads the saved script inside the `useState` initializer, through the same `applyScriptToEditor` version restore uses. `useProjectRehydrate` and its `ranRef` are deleted. `<Slate>` takes `editor.children` as its initial value instead of an empty document. An empty script still opens an empty editor, for the same reason the old hook skipped it: a forced normalize on nothing would seed the layout's narration ahead of the first streamed script.
- `createAutosaver` takes its baseline at construction. `markSaved` is gone from the `Autosaver` contract, along with the effect in `useAutosave` and the "runs after the rehydration effect" ordering. The autosaver is created with a `useState` initializer like its siblings in `useCanvasVersions`, since the baseline is only right if it is built exactly once.
- Tests: the autosave suite drops every `markSaved()` call and two cases that the construction baseline made duplicates of each other; a "does not save the document it was built from" case replaces the effect-order test. The `CanvasHistory` fake autosaver loses the method.

+49 / -112 over 9 files.

## What I ran

`npm run lint && npm run typecheck && npm run format:check && npm run test:run && npm run build`, all green (185 files, 1633 tests). knip clean, jscpd 0 clones, madge cycles unchanged. `/simplify` ran with four review agents; applied: module-level `createCanvasEditor` (no shadowed `editor`), the `read` doc comment now says it runs at construction, duplicate tests folded. `/vercel-composition-patterns` and `/vercel-react-best-practices` read and applied (lazy state init). `/web-design-guidelines` not run: no visual code changed.

## Skipped

- Moving the empty-script guard into `applyScriptToEditor` as an early return: it would leave a version restore whose script is empty half applied (store and generation replaced, script kept).
- Dropping the narration seed from `withLayout` so the guard is unnecessary rather than relocated: behaviour-visible when a user deletes every element, a product call.
- The baseline serialize now runs once during the first render instead of after commit. Same work, one document, once per open; not measurable at script sizes here.
- `hasSavedScript` predicate shared with `ScriptProvider` and `EditorPanelContext`: rule of three is met but the helper is one comparison.
- The remaining free candidates from the 09-11 notes (`resultSource`, `ProjectData` as `z.infer`) are thin on their own; the `hideModel` fold and job-type registry wait on #785 and #787.

## Merge-conflict guard

```
PR #788: clean
PR #787: clean
PR #786: clean
PR #785: clean
```


app/components/canvas/hooks/useEditorSetup.ts
```diff
@@ -4,27 +4,35 @@ import { withReact } from "slate-react";
...
+		withHistory,
+		withReact,
+		withLayout(defaultModels),
+		withScenes,
+		withFlatPaste,
+		withNodeId,
+		withOSMLClipboard(defaultModels),
+	)(createEditor());
+	// Loading an empty script would seed the layout's narration ahead of the first one streamed in.
+	if (script) applyScriptToEditor(editor, script, defaultModels());
```
umairnadeem, line 30: this runs during SSR now too (ProjectEditor is a client component the page server-renders), and `deserializeWithScenes`/`parseOSML` mint fresh nanoids per call, so the html ships one set of `data-scene-id`s and the client hydrates with another. react 19 doesn't patch attribute mismatches, so `scrollToScene` (player controls / autoscroll) silently misses every scene loaded from the row, plus a hydration error in dev. before, the SSR'd editor was empty and the script only ever loaded client side. maybe keep the initial load client-only, or make the ids deterministic?

PR #794, opened 2026-09-13
fix(sloppy): stick-to-bottom only while user is at bottom (#791)

### LINKED ISSUE #791: Sloppy panel forces scroll to bottom while streaming; can't scroll up mid-turn
## Problem

While Sloppy is responding, the transcript jumps to the bottom on every streamed chunk. If you scroll up to reread an earlier message or tool call, it drags you straight back down on the next token. You can't read anything mid-turn until the turn finishes.

## Relevant code

- `app/components/sloppy/SloppyPanel.tsx:81-83`: the effect runs `scrollIntoContainer(endRef.current, "end", "auto")` on every `messages` change. It doesn't check where the user has scrolled, and streaming changes `messages` constantly.
- `lib/components/scrollIntoContainer.ts:28-47`: finds the scrollable ancestor (the panel's scroll container), so the "is the user at the bottom?" check can use the same element.

## Suggested direction

Standard stick-to-bottom behavior:

- Follow the stream only while the user is at, or within a few px of, the bottom of the scroll container.
- Once the user scrolls up, stop auto-scrolling until they scroll back to the bottom.
- Sending a new message always scrolls to the bottom and turns following back on.
- Keep the initial jump to the bottom when a restored transcript loads.

Track "at bottom" with a scroll listener on the container (or an `IntersectionObserver` on `endRef`) and check it in the existing effect. No new dependency needed.

## Acceptance criteria

- [ ] While a turn streams, scrolling up stays put; new tokens don't pull the view down
- [ ] Scrolling back to the bottom resumes following the stream
- [ ] Sending a message scrolls to the bottom
- [ ] Opening a project with an existing transcript still starts at the bottom

---

mohandamged70m:
Fixes #791.

While Sloppy streamed, the transcript called `scrollIntoContainer` on every `messages` change, so scrolling up mid-turn yanked you back down on the next token.

### Changes
- `lib/components/scrollIntoContainer.ts`: export `findScrollableAncestor` (was private `scrollableAncestor`) for reuse.
- `app/components/sloppy/SloppyPanel.tsx`:
  - Track `isAtBottomRef` (starts `true`, no re-renders).
  - Scroll listener on the container updates it via `scrollHeight - scrollTop - clientHeight <= 48px`.
  - Gate the existing `[messages]` effect on it.
  - User message at the end forces follow back on (covers send, since Composer is a sibling).

### Acceptance
- [x] Mid-stream scroll-up stays put
- [x] Scroll back to bottom resumes following
- [x] Send scrolls to bottom
- [x] Restored transcript starts at bottom

### Checks
- `prettier --check` on touched files: pass
- `tsc --noEmit`: no errors in touched files (repo has pre-existing errors elsewhere)
- Manual: `npm run dev` + long Sloppy turn


app/components/sloppy/SloppyPanel.tsx
```diff
@@ -73,13 +76,38 @@ function Transcript({
...
+		const end = endRef.current;
+		if (!end) return;
+		const target = findScrollableAncestor(end) ?? document.scrollingElement;
+		if (!target) return;
+		const onScroll = () => {
+			const distance =
+				target.scrollHeight - target.scrollTop - target.clientHeight;
+			isAtBottomRef.current = distance <= BOTTOM_THRESHOLD_PX;
+		};
+		onScroll();
```
umairnadeem: this breaks reopening the panel: switching sidebar tabs remounts SloppyPanel, so on a long transcript this `onScroll()` sees scrollTop 0, flips `isAtBottomRef` to false before the `[messages]` effect runs, and you land at the top instead of the bottom. imo just drop the initial call, the ref already starts `true`.

mohandamged70m: removed the initial onScroll() call. Ref starts true so the first [messages] run lands at the bottom, then the listener takes over.

umairnadeem (commented):

PR #802, opened 2026-09-16
refactor: nightly maintainability 2026-09-16

umairnadeem:
Maintainability pass: no behavior change. Two boundaries now parse outside data once, and the drag-over logic leaves the canvas hook for `lib/canvas`.

## A saved project is parsed once, on the way in

- `lib/project/projectContent.ts` (new): `parseProjectContent` types the `script`, `store` and `generation` columns of a saved row, composing the existing `parseStoreSnapshot` and `GenerationSnapshotSchema`. `PROJECT_CONTENT_COLUMNS` is derived from the schema, the same idiom as `PROJECT_ROW_COLUMNS`, so the two selects cannot drift from it.
- `app/projects/[id]/page.tsx` parses the row on the server and hands the editor one `initial: ProjectContent` instead of `initialScript` / `initialStore` (unknown) / `initialGeneration` (typed as parsed, but it was the raw column). The `script ?? ""` default is gone: the column is `not null default ''`.
- `lib/project/canvasVersionStorage.ts` `read` uses the same parser; its private `RowSchema` is deleted. Version restore and first load now agree on what a saved project is.
- The `ProjectContent` type stays in `projectDocument.ts` because `lib/project/autosave.ts` and `useAutosave.ts` import it from there and both sit in #785. Moving it next to its parser is a three-line follow-up once #785 merges.

## The access-code route parses the RPC outcome

- `app/api/validate-code/route.ts`: `Outcome = z.enum(["valid", "invalid", "inactive", "expired"])` replaces a `typeof data !== "string"` check plus a `Record<string, string>` with a `?? ERROR_MESSAGES.invalid` fallback. The rejection table is now total, and an answer the SQL function never gives is a thrown 500 rather than "Invalid access code". New test covers that case.
- The 401 goes through `unauthorized(message)` from `lib/api/response.ts` (which gained the optional message), so this route stops being the one place that hand-builds the error envelope.

## Drag-over logic has one home

- `lib/canvas/dragOps.ts`: `crossSceneTransfer(editor, active, over)` owns "is this a cross-scene drop and where does it land". `contentSlot` is the one rule for "append to a scene, or take the hovered content's slot", shared with `moveDraggedElement`, which used to restate it. `SortableData` names the payload `SortableItem` gives dnd-kit (`satisfies` on the author side), and the single cast that reads it back lives beside the type.
- `useDragAndDrop.ts` `handleDragOver` is one line. It no longer reads `active.data.current?.sceneId` as `any`, no longer walks the tree itself, and no longer fires a `set(null)` before every `set(transfer)` (two notify passes per pointer move on a cross-scene hover).
- `DragTransfer` shrank to `{ sceneId, atIndex }`: `itemId` and `fromSceneId` had no reader except a dead `fromSceneId === sceneId` branch in `dropIndexIn` (the hook never produced such a transfer). `dropIndexIn` is inlined into `useDropIndex`. The store now dedupes by value, so a sustained hover over one slot notifies the cards once instead of on every pointer move.
- Tests: `crossSceneTransfer` has five cases in `dragOps.test.ts`; the transfer store test covers the value dedupe.

## Not taken tonight (blocked or judged out of scope)

- `apiJson<T>` casting the server's JSON at the client boundary (agent transcript, upload, account store): the schema-taking variant also touches `lib/render/render-client.ts`, reserved by #785.
- Job-type registry (`jobs.ts` / `asset-routes.ts` / `job-handlers.ts`): #785 / #787.
- `SnapshotStore.get(id?)`: the only `undefined` caller is `AssetTile` for non-character tiles; making it required needs `lib/generation/graph.ts` (#785).

Checks: lint, typecheck, format, 1640 tests, build. madge cycle count unchanged (20). Ran /simplify, /vercel-composition-patterns, /vercel-react-best-practices; /web-design-guidelines has no markup change to review.

## Merge-conflict guard

```
PR #801: clean
PR #794: clean
PR #789: clean
PR #788: clean
PR #787: clean
PR #785: clean
```


app/projects/[id]/page.tsx
```diff
@@ -35,11 +39,9 @@ export default async function ProjectPage({
...
 	return (
 		<ProjectEditor
-			key={project.id}
-			projectId={project.id}
-			initialScript={project.script ?? ""}
-			initialStore={project.store}
-			initialGeneration={project.generation}
+			key={id}
+			projectId={id}
+			initial={parseProjectContent(project)}
```
Comment (umairnadeem) [inline app/projects/[id]/page.tsx:44]: this turns a stale row into a 500 on open: `ElementSnapshotSchema` requires `pinned` (only written since #480, aug 1) and afaict nothing ever backfilled the `generation` json, so a project last autosaved before then throws here where before the raw column went straight into `SnapshotStore` and a missing field was just falsy. was the column backfilled? if not, imo the schema needs defaults for the newer fields (or this falls back to `{}` on a parse failure) so old projects stay openable.

Issue #800, opened 2026-09-16
Request Anthropic prompt caching for the stable system prefix


umairnadeem:
## Problem

`lib/agent/prompt.ts:86` states the intent:

> The settings snapshot goes last so the stable half stays a cacheable prefix.

The layout is right — the stable half really is ordered first — but nothing ever asks for the cache. `grep -rn "cache_control\|cacheControl" lib app components` returns **zero matches**, so every request pays full price for a prefix that is byte-identical across calls.

## Where

Two call sites pass a plain string as `instructions`:

- `lib/api/agentTurn.ts:96` — `instructions: sloppyInstructions(request.context)`, Sloppy's system prompt (~6 KB of literal).
- `lib/providers/llm/anthropic.ts:102` — `instructions: params.systemPrompt || undefined`, which is how `review_script` passes `scriptRules` (`projectPreamble` + `osmlSpec`, ~11–12 KB).

## Why it's worth doing

Both prefixes are re-sent unchanged, repeatedly:

- A Sloppy turn re-sends the system prefix on **every tool round trip**, up to `MAX_TOOL_CALLS = 30` (`lib/api/agentTurn.ts:34`).
- `review_script` re-sends the same `scriptRules` prefix on **each review pass** (up to three per new script).

## It is already supported

- Installed: `ai@7.0.66`, `@ai-sdk/anthropic@4.0.39`.
- `ai`'s `Instructions` type accepts `string | SystemModelMessage | Array<SystemModelMessage>`.
- The Anthropic provider reads `cache_control` off a system message's `providerOptions`.

Sketch, at `sloppyInstructions`' call site — mark only the stable half:

```ts
instructions: [
  {
    role: "system",
    content: SLOPPY_SYSTEM_PROMPT,
    providerOptions: { anthropic: { cacheControl: { type: "ephemeral" } } },
  },
  { role: "system", content: renderAgentContext(context) },
]
```

## Notes / risks

- Mark **only** the stable half. The settings snapshot changes per request, which is exactly why `prompt.ts` already orders it last — marking it would defeat the cache.
- Anthropic enforces a minimum cacheable prefix length; confirm both prefixes clear it before assuming a hit.
- Only the direct Anthropic provider path benefits. The `openslop` gateway path (`lib/gateway/http.ts`) is separate and would need its own look.
- `sloppyInstructions` currently joins into one string, so it needs to return the two parts separately (or the caller needs to compose them).

## Acceptance

[...]

PR #807, opened 2026-09-18
feat: Sloppy's prompt caches; the settings move to read_settings

umairnadeem:
## Summary
- The Sloppy system prompt is one block, marked `cache_control: ephemeral` through the provider's `cachedPrefix`; the request also sets top-level automatic caching, so each round trip reads the transcript up to the previous one. `review_script`'s system prompt gets the same explicit mark. The vendor literal lives only in `lib/providers/llm/anthropic.ts`; `lib/agent` sees an opaque `SharedV3ProviderOptions` via `AgentModel`.
- The settings snapshot leaves the system prompt. It changed on every `set_*` call, which invalidated the whole transcript cache mid-turn. Sloppy now reads it with `read_settings`, a snapshot tool beside `read_script`; the agent request carries no `context` field and `agentContextSchema` is gone.
- Usage (`cacheReadTokens` / `cacheWriteTokens`) is logged after every model call.

Stacked on #785. Closes #800.

## Test plan
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run knip`, `npm run build`, `npx vitest run` (190 files)
- [ ] Live: one Sloppy turn with a tool call, then confirm the second round trip logs `cacheReadTokens > 0`
- [ ] Browser: Sloppy calls `read_settings` before its first change on a fresh project

## Notes
- Haiku 4.5 needs a 4096-token prefix; `review_script`'s system prompt alone is short of that, so on Haiku it will not cache. Opus 5 (512) and Sonnet 5 (1024) are fine.
- Behavioural change: mid-turn, the model no longer sees settings the user edits by hand while it is working; it learns them on the next turn or the next `read_settings`.

https://claude.ai/code/session_01DydWuScAb29MDuxetLFktw


umairnadeem (commented):

